PR #16, opened 2021-08-02
Restructured code

beardyinc:
- [x]  moved code to the according new packages
- [ ]  updated package names, replace `com.microsoft` with `org.eclipse` and adapt application and component names
- [ ]  replaced all license headers with Apache license


beardyinc:
will reopen another PR once the build pipeline is more stable.

PR #29, opened 2021-08-04
Updated copyright headers and LICENSE

beardyinc:
Updated all code files to include the correct license header. Also updated the `LICENSE` file to contain Apache Software License 2.0


MoritzKeppler:
cool!
just one question: what does "Contributors: 1" mean? Is "1" a counter?

MoritzKeppler:
thx for adjusting. I'm fine with the PR but as it's about changing the license header of a Microsoft contribution it would be better if @jimmarino approves it.

PR #30, opened 2021-08-06
Modularize common/utils

jimmarino:
The purpose of this PR is to clean up the common/utils module. I did this as a series of commits, so it is easier to see the changes:

1. The AWS text fixture was moved to its own module under the AWS extensions, and the common dependency on AWS libraries was removed.
2. The Azure text fixture was moved to its own module under the Azure extensions and the common dependency on Azure libraries was removed. 
3. The Loopback dispatcher was moved from commons to the sample extension as it is not intended for general use. This is also the first step in decoupling utils from the SPI and core modules.
4. Move the HTTP basic auth helper back to Atlas where it is used and remove the utils dependency on the OkHttp library.
5. Move the connector id function to ServiceExtensionContext and complete removal of utils depending on SPI and the boostrap modules
6.  Move the Junit runtime launcher to its own module. This allows downstream repositories to depend on the JUnit extension for testing without depending on core EDC modules and implementation dependencies.
7. Move two utility classes from SPI to utils
8. Centralize Jupiter version in build dependencies
9. Some license and import cleanup that was missed.
  
Can one of you two, in addition to Paul, please take a look?

Thanks


beardyinc (approved):

data-protocols/ids/ids-core/src/main/java/org/eclipse/dataspaceconnector/ids/core/IdsCoreServiceExtension.java
```diff
@@ -60,7 +58,7 @@ public void initialize(ServiceExtensionContext context) {
         context.registerService(IdsDescriptorService.class, descriptorService);
 
         var identityService = context.getService(IdentityService.class);
-        var connectorName = getConnectorId(context);
+        var connectorName = context.getConnectorId();
```
MoritzKeppler, line 61: better call var connectorId (there are a few more places where connectorName is used...)

launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/EdcExtension.java
```diff
@@ -25,13 +25,18 @@
...
 
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 
-import static org.eclipse.dataspaceconnector.common.Cast.cast;
+import static org.eclipse.dataspaceconnector.common.types.Cast.cast;
 
 /**
  * A JUnit extension for running an embedded DA-GX runtime as part of a test fixture.
```
MoritzKeppler, line 42: rename DA-GX

MoritzKeppler (approved):

launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/EdcExtension.java
jimmarino: Hi Moritz, let's do this change in a separate PR to catch all of those.

data-protocols/ids/ids-core/src/main/java/org/eclipse/dataspaceconnector/ids/core/IdsCoreServiceExtension.java
jimmarino: OK updated

PR #34, opened 2021-08-09
Change names

jimmarino:
Cleanup leftover DA-GX names


MoritzKeppler (changes requested):
found two more "Gaia-X Data Appliance" occurences in terraform scripts

beardyinc:
be careful with replacing stuff in terraform scripts, they sometimes point to existing resources (such as the state backend). Simply replacing them will break stuff.

jimmarino:
It looks like there are two instances, both related to the organization. One is commented out. @beardyinc can you take a look at those to instances to see if they are safe to change?

beardyinc:
Ok, i took a look, seems those two occurrences of `"Gaia-X Data Appliance"` can be changed safely since they're there just for informational purposes.

MoritzKeppler (approved):

PR #69, opened 2021-09-13
Remove Atlas and Nifi

beardyinc:
This PR removes all references to Apache Atlas and Nifi since they were merely used as proof-of-concepts and we don't plan on maintaining them at the moment. Also, they make up for a major chunk of the integration test runtime.

This was done in an attempt to make the code more readable and less cluttered.


kbData:
what is the planned alternative for those dependencies?

beardyinc:
> what is the planned alternative for those dependencies?

We're currently looking into different approaches to implement the catalogueing feature as well as options to make the out-of-band data transfer simpler. No definite answer at this time I'm afraid, as the project is still in its infancy and docs aren't always up-to-date.

kbData:
how is it possible to follow the discussions around this topic? I am already on the eclipse project mailing list, but this does not seem to be the right place.

beardyinc:
once we have something concrete we'll definitely create issues here. your best bet is to watch this repo and periodically check back.

kbData:
ok, thx!

jimmarino (approved):

jimmarino:
Does anyone object to me merging this? If not, I'll go ahead later today.

Jim

Issue #41, opened 2021-08-13
Implementent Basic IDS Self-Description API of the Connector


DominikPinsel:
This is the corresponding issue for the defined offer interfaces in the draft pull request #38 

**The connector should be able to answer IDS self-description requests for itself. It is not necessary to support description requests of artifacts and resources yet.**

I also assume for the implementation
* A new _DescriptionMultipartController_ should be added to a new _IDS Multipart API Extension_.
* Instead of using the official java library for the IDS Information Model its probably better if we implement the IDS classes, that are necessary for this implementation, ourselves in a new _IDS Information Model_ in the IDS Core Extension.
* A new _Core Extension_ for Contracts must be added. This will be the place for our business logic regarding all contract topics (negotiation, offering, ...)
* The AssetIndex will be implemented in a new _SampleExtension_. Its totally fine if it just works with some hard coded assets. Please check whether this extension needs to support the Catalog interfaces, too.
* The interfaces from pull request #38 will be added to the _SPI Extension_

We at Daimler will start with the implementation today. It would be nice if a committer could create a new branch in this repository (e.g. _feature/41-ids-self-description_), so that we push our changes upstream.

Feel free to comment our suggestions from above. We appreciate any comments.

<sub>Dominik Pinsel <dominik.pinsel@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

DominikPinsel:
Replaced by #144

Issue #79, opened 2021-09-15
Discuss and improve extension-dependency naming scheme


paullatzelsperger:
Lets see if we can come up with a consistent and intuitive way to express dependencies (i.e. the `requires()` and `provides()`) that is consistent over all modules.

There are a few suggestions:
1) fully qualified classname. Can be problematic when one extension provides several features and when several modules provide the same feature. 
2) a hierarchical namespace like `edc:<feature>[:<subfeature>]`, e.g. `edc:policy-registry` or `edc:communication:http-client`

We should also think about defining constants for that in the interface class, e.g. in `PolicyRegistry.java`

jimmarino:
I like the approach in 2. IMO the problem with 1 is that it ties dependencies to specific implementation classes. Instead, they should be a symbolic name. For example, one extension may provide 2 or more features that clients depend on.

The advantage of 2 the way you defined it is the symbolic name can also be a URN and we can reserve "edc" to avoid clashes.

DominikPinsel:
Maybe another, complete different, approach. Wouldn't a dependency injection framework solve this problem, too? Because the framework should be able to resolve dependencies on it's own, so it won't be necessary to track them manually in an array.
It also would simplify other things, like the primordial load phase. And using DI frameworks is somewhat state of the art, so every developer knows how they work without reading any documentation.

jimmarino:
There were a number of reasons why we chose not to use a DI framework such as Spring (version clashing, a gazillion dependencies, abstractions on abstractions, non-standard APIs, keeping the connector codebase small so its security footprint is manageable, etc.). I think these are fair trade-offs for an application but they become problematic for low-level code that has to be deployed in varied environments.
  
If we were to go down this route, it would entail a complete re-write of the codebase...I'd prefer not to do that.

PR #77, opened 2021-09-14
Create and apply checkstyle xml config

beardyinc:
This PR adds a checkstyle config file (located at [resources/edc-checkstyle.xml](resources/edc-checkstyle.xml) which can be used by plugins of differend Java IDEs (e.g. IntelliJ IDEA, Eclipse, etc.) or standalone using the reviewdog app.

Also, Github Actions now performs Checkstyle checks (see [this workflow](.github/workflows/verify.yaml)). 

So far a few modifications from the basic Google Style have been made. The following things need to be improved still:

- [x] naming schemes (variables, parameters, classes, ...)
- [x] SaveAction and a Java Code Style configs, so that valid code is generated in the first place.

PR #80, opened 2021-09-15
#41 Contract Offer Interfaces (Part 1)

DominikPinsel:
Hello everyone,

@denisneuling and I are now finished with the initial implementation of the Contract Offer Interfaces. We decided to split the issue #41 in two separate pull requests. The first one will contain the implementation of the EDC Contract Offer Interfaces. The second part will be the implementation for the IDS self-description, that contains contract offers.

This pull request contains several ToDo's. Our intent with them is, to add these ToDo's to the main branch (for now) and resolve them at a later time. We decided to do it this way, so that this pull request doesn't get even larger and because some of them need further discussions.
If you think some of these ToDo's should be resolved within this pull request, please let us now.

Also feel free to ask questions or provide us with feedback, when something should be improved. We discussed a lot ourself and we would be very surprised, if this code doesn't raise some questions :)


jimmarino (commented):
Hi Dominik and Denis,

Here are a few comments/suggestions:

1. Why are the types in the "types.domain.policy" package needed? Aren't these the same as the types in the "policy" module? As background, the "policy" module types are separated out so that other systems can potentially embed a policy engine with dragging in all runtime types.

2. Most of the domain builders already use the static "newInstance()" idiom instead of new ...builder(). I don't really care which one is used other than we should be consistent across the codebase. Should we stick with newInstance(), or is there a reason we should switch to new ...builder()?  

3. It may be useful for ContractOffer and other domain objects (OfferedAsset) to be JSON serializable
types.

4. It would be good practice if we document all types in SPI (I'm sure I probably violated this somewhere myself ;-) )

If you want, let's get on a chat to discuss, perhaps tomorrow, so we can get this merged.

samples/demo-asset-index/src/main/java/org/eclipse/dataspaceconnector/demo/assets/DemoAssetExtension.java
```diff
@@ -0,0 +1,58 @@
...
+import org.eclipse.dataspaceconnector.spi.monitor.Monitor;
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtension;
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtensionContext;
+
+import java.util.Set;
+
+public class DemoAssetExtension implements ServiceExtension {
+    private static final String NAME = "Demo Asset Index Extension";
+    private static final String[] PROVIDES = {
+            AssetIndex.class.getName()
```
beardyinc, line 27: also see issue #79. nothing to do at the moment, just something to keep in mind.

beardyinc:
Here are some general comments:
1) Instead of having the `AssetIndexLocator`, why not just maintain a collection/list of `AssetIndex`-instances in the `CompositeAssetIndex`? 
2) Same for `ContractOfferFrameworkLocator`
3) How is a `UsagePolicy` different in semantics from a `Policy`? 
4) I am not the biggest fan of having `final` local variables and method parameters. In fact in the code style (see PR #77) those would get removed upon reformatting.

_please find specific comments directly in the code._

denisneuling:
Jim and Paul, thank you for taking time looking into our work. I'll try to answer all questions asked.

@jimmarino:
1.  Although the policy module you're referring to is pretty extensive and work put into it is impressive we aimed for a simpler domain representation of policies associated to an offered asset, which then can be read out in other application layers in an ordinary manner (getter/setter). Maybe it's a good idea to have a chat about how to integrate this with the existing policy classes tomorrow.

2.  I have no strong feelings about what idiom to use here, `DomainObject.builder().build()` is shorter than `DomainObject.Builder.newInstance().build()` though. IMO the shorter feels more intuitive to me... But as said, no strong feelings on my side. Let's just jointly decide on one and apply it to all concerning types.

3.  It depends on the purpose of these objects. The `ContractOffer` and other objects are not meant to be serialized somewhere, so having Jackson annotations in place is not needed IMO. They rather act as vehicle to pass though information through the business layer. Once they leave the business layer e.g. by reaching a presentation layer a data mapper should act on those informations carried in these objects and build object for the purpose (e.g. mapping domain objects into its corresponsing IDS/GaiaX/etc representation)

4.  We promise improvement and will add documentation to all SPI types (- also in future) ;)

@beardyinc
1. Isn't maintaining an internal list of instances the whole purpose of the ServiceExtensionContext? I somehow do not see the value maintaining another list of object references - if there is already a dependency resolution mechanism why not simply using it? Also those `*Locator` classes are package-only visible thus do not form part of the packages interface and having just the task of getting the ContractOfferServiceImpl wired up. 

2. Pls see above.

3.  We tried to portray the IDS policy model. Spontaneously I'd say no, there aren't semantic differences between UsagePolicy and Policy. 

4.  I am indeed a big fan of final local variables wherever possible, since it prevents a developer from accidentially modifying a value. Also I "heard" it enables the compiler to optimize the class files, although I doubt the impact is huge...

beardyinc:
@denisneuling 
1. that's fine, but my suggestion would be that the `CompositeAssetIndex` maintain the list of `AssetIndex` objects directly instead of going throught the `Locator`. We have two levels of (unnecessary) indirection: 
  - `CompositeAssetIndex` 1->1 `AssetIndexLocator`
  - `AssetIndexLocator` 1->N `AssetIndex`
thus, semantically, the Locator and the composite index are doing similar/same things.
2. Policy: We should definitely avoid having two domain models for the same entity, especially since the domain is not different :)
3. `final`: I find it makes code very verbose and cluttered, but at the same time, I don't care that much. It needs to be consistent throughout the code base. And since we don't have final (variables/params) anywhere...

DominikPinsel:
Hello everyone :)

First of all, thanks for all this feedback! Let’s try to address all the issues

### Builder Idiom
_Jim mentioned, that we should try to create builder ether with ‘newInstance()’ or ‘new builder()’, but not both._

That’s correct and I didn’t see, that we used another idiom here. I will update our builders to ‘newInstance()’.

### Serialization of ContractOffer and other domain objects
_Jim mentioned, that it may be useful to have all the domain objects serializable._

That may be true, but I still think there are also reasons against having them serializable.
1. YAGNI. If it’s not needed, I would not do it. Unused Code adds complexity, makes it more difficult to refactor, and so on.
2. My personal opinion is, in general,  that we should try to avoid serializing our domain objects. I’d like it more if these objects are not persisted 1:1 or send, as they are, to another connector / layer.

**Suggested Solution**: I will update the new builders, so that they look like the existing ones.

### Document all types in the SPI
_Jim mentioned, that we should document all types in the SPI._

I fully agree and will document the types. I should have done this from the start :D

**Suggested Solution**: I will update the documentation in the SPI.

### Extension-Depedency Naming Scheme
_Paul mentioned the issue #79, where the naming scheme is discussed._

And again I fully agree :) It feels really strange to map these Strings in an array. And other extensions must have mapped the same Strings, or else the application won’t start. That’s also the reason we used the fully qualified class name here. But it still feels wrong. I suggested using a DI framework in the linked issue.

**Suggested Solution**: We will let it as it is and solve this problem within issue #79.

### Move AssetIndex collection from AssetIndexLocator to CompositeAssetIndex (and same for ContractOfferFrameworkLocator)
_Paul mentioned, that we could move the collection from the Locator to the Composite._

We also discussed this internally and decided, to keep it out for single responsibility reasons.

**Suggested Solution**: I would not change the code. As the Locators are only visible within their package, this should not add much complexity.

### Coding Style (e.g. final)
_Paul mentioned, that the finalization of method arguments is not part of the current coding style._

Yea, coding style is definitely a difficult topic :D 
I’m coming more from a coding style, that creates minimal noise. Denis is using a coding style that is more verbose and descriptive. My personal opinion is, that there are good reasons for both. For an open source project we came to the conclusion that a more verbose coding style is better. If most of the developers are not our opinion, we will simply adopt :)

### Policy Extension vs Policy SPI
_Jim and Paul mentioned, that there is already a policy extension and we created new policy classes instead reusing the existing code._

This is again a really difficult topic. We needed to discuss this several times to come to a good conclusion. Let me say first, that we really liked the existing policy code. It is well written and supports simple policy operations. We still decided, that this code is not able to support our solution.
I will try to summarize the discussion. 

1. YAGNI

The existing dependency extension is able to do a lot more than is actually necessary. For the MVP, the only policy the connector needs to be able to describe is: **The consumer is allowed to do everything with the data**.
I assume this requirement will stay as it is for several month from now. Assuming this is true, there is no reason to implement a full fledged policy model today. Requirements from our stakeholders could chance, IDS could change, there could come requirements from GAIA-X.

We also see how an extra policy extension could add benefits in the future, like Jim mentioned. But we (again) don’t see them today.

2. KISS

 Because many different extensions and developers may come in contact with this domain in the future, we think the policy domain should be easy to use. The new approach gave us benefits like
- express how the policy domain is work in progress, using a marker interface and comments. This adds documentation in code and helps future developers to understand our policy domain better.
- less code / simpler model.
- It doesn’t look like the connector is able to do something it cannot (for example enforce regions).

3. Better map-able to IDS

We needed a policy model that is map-able to IDS. We think the best way to do this is strongly typed. If we would have used the policy extension, we would have to do this with LiteralExpressions. The IDS extensions would need some “common policy naming schemes” for these literal expressions. As this is definitely still possible, we like a strongly types approach a lot more, as it clearly describes out domain (and doesn’t hide it).

---

Sorry for the wall of text.
I added suggestions how we could continue with each topic (except coding style and policy model/extension). Please let me know if this is fine for you and what you think about the two open topics :)

beardyinc:
**Ad serialization:** while I can see the merit in separating domain objects and transfer objects ("DTOs") I would not introduce that duality simply for the sake of having them. Unless we have an actual requirement to either limit the amount of information exposed via APIs, I'd say lets keep them joint. Also, in Java, every object can be serialized, even without explicitly annotating them. To my knowledge this is done using reflection. The trouble begins when attempting to deserialize, esp. when using builders. So without explicitly annotating, we're essentially creating a broken mechanism.

**Code style**: mostly, I care about a democratic decision and consistency. So whatever code style we adopt, it needs to be coherent throughout the code base. So at some point we'll have to stop the world, reformat _the entire code_ and resume the world :)

jimmarino:
Hi,

I had a bunch of comments...

> ### Serialization of ContractOffer and other domain objects
> _Jim mentioned, that it may be useful to have all the domain objects serializable._
> 
> That may be true, but I still think there are also reasons against having them serializable.

> 
> 1. YAGNI. If it’s not needed, I would not do it. Unused Code adds complexity, makes it more difficult to refactor, and so on.
> 2. My personal opinion is, in general,  that we should try to avoid serializing our domain objects. I’d like it more if these objects are not persisted 1:1 or send, as they are, to another connector / layer.
> 
> **Suggested Solution**: I would not make the domain objects serializable now. In case we need it later we can do it any time.
> 



I had some other stuff serialized so it would make it easier to persist types (and have them indexable) or write part of an extension in a different language (which is less important). For example, TransferProcess is serializable to JSON so it can be easily written to a store and indexed for implementations that support such features with JSON types. We will eventually need to make all of this stuff auditable. What do you think?  



> ### Document all types in the SPI
> _Jim mentioned, that we should document all types in the SPI._
> 
> I fully agree and will document the types. I should have done this from the start :D


I'm sure I am guilty of violating this

.....

> And again I fully agree :) It feels really strange to map these Strings in an array. And other extensions must have mapped the same Strings, or else the application won’t start. That’s also the reason we used the fully qualified class name here. But it still feels wrong. I suggested using a DI framework in the linked issue.


Please see my rant on DI frameworks in the other issue...I really like some of them, but I don't think one is needed.


....

> _Jim and Paul mentioned, that there is already a policy extension and we created new policy classes instead reusing the existing code._
> 
> This is again a really difficult topic. We needed to discuss this several times to come to a good conclusion. Let me say first, that we really liked the existing policy code. It is well written and supports simple policy operations. We still decided, that this code is not able to support our solution.
> I will try to summarize the discussion.
> 
> 1. YAGNI
> 
> The existing dependency extension is able to do a lot more than is actually necessary. For the MVP, the only policy the connector needs to be able to describe is: **The consumer is allowed to do everything with the data**.
> I assume this requirement will stay as it is for several month from now. Assuming this is true, there is no reason to implement a full fledged policy model today. Requirements from our stakeholders could chance, IDS could change, there could come requirements from GAIA-X.
> 
> We also see how an extra policy extension could add benefits in the future, like Jim mentioned. But we (again) don’t see them today.
> 
> 1. KISS
> 
> Because many different extensions and developers may come in contact with this domain in the future, we think the policy domain should be easy to use. The new approach gave us benefits like
> 
> * express how the policy domain is work in progress, using a marker interface and comments. This adds documentation in code and helps future developers to understand our policy domain better.
> * less code / simpler model.
> * It doesn’t look like the connector is able to do something it cannot (for example enforce regions).
> 
> 1. Better map-able to IDS
> 
> We needed a policy model that is map-able to IDS. We think the best way to do this is strongly typed. If we would have used the policy extension, we would have to do this with LiteralExpressions. The IDS extensions would need some “common policy naming schemes” for these literal expressions. As this is definitely still possible, we like a strongly types approach a lot more, as it clearly describes out domain (and doesn’t hide it).
> 
> Sorry for the wall of text.
> I added suggestions how we could continue with each topic (except coding style and policy model/extension). Please let me know if this is fine for you and what you think about the two open topics :)
> 


I have a different view on this. At a high level, I don't think there should be two implementations of policy in the runtime. 

Looking at the code, I don't see what the issue is with using the existing policy model types. Is there something specific that makes it unworkable (if so, this should be fixed)? It seems like the policy model can be integrated in a straightforward manner but I may be missing something.  

I believe the existing policy model can be typed. Additional Expression implementations can be added if needed, although I do believe IDS policy is not typed (i.e. it identifies some policy expressions but does not define a typed grammar).  

Related to this, IDS is based on ODRL, which we tried to capture in the policy model. Since we are going to need to support that, we will likely wind up with something that is similar to the model anyway.

I don't think Actions should be an enum due to the constraints of the Java type system, specifically the inability to extend them. It should be possible for extensions (e.g. IDS) to add custom Action types without the core runtime having direct knowledge of them. 
  

> Dominik Pinsel [dominik.pinsel@daimler.com](mailto:dominik.pinsel@daimler.com), Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)

DominikPinsel:
@jimmarino @beardyinc 
Hello everyone,

We updated the pull request:
- Squashed commits down to one commit 
- Added comments to the SPI classes & interfaces
- Builder pattern is now consistent
- Integrated existing policy extension

Please have another look :)

jimmarino (approved):
Hi Guys,

Thanks for turning around the changes. I'm fine to merge this in if you think it is ready.

One thing we should discuss is the question of one-vs-multiple AssetIndexes in the runtime but I think we should do that after this PR is merged in.

If you're ready for the merge, let me know and I'll commit it.

denisneuling:
Thanks Jim for your approval. Feel free to merge. 
Regarding the multiplicity of AssetIndices we gonna create a follow-up issue for discussing this.

Issue #96, opened 2021-09-23
Provide S3-to-S3 transfer demo


MikhailGordienk:
Would be helpful to have a complete demo where a consumer can request
a file from a provider and the transfer will be done completely on the AWS side.

This demo would show the case when resource provisioning is not necessary.

MikhailGordienk:
PR: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/97

bscholtes1A:
@jimmarino I guess this one can closed as well

Issue #95, opened 2021-09-23
Decouple status checking from provisioned resources


jimmarino:
Currently StatusChecker takes a type of Provisioned resource: 

`boolean isComplete(T provisionedResource);`

StatusCheckerRegistry also dispatches based on the ProvisionedResource type. This has the unintended side-effect of coupling transfer status checking to resource provisioning. For example, if an extender wants to implement a DataFlowController and supply a StatusChecker, they must also implement a ProvisionedResouce, a Provisioner and a ManifestGenerator.

For some cases, a StatusChecker will need access to a ProvisionedResource, for example, to check if data has arrived in a target bucket. However, it may be the case that only the TransferProcess id is necessary for A StatusChecker implementation. 

To decouple status checking, I plan on changing the registry to dispatch based on DataAddress type and then pass the transfer process id and a collection of ProvisionedResources to the StatusChecker. For cases when no provisioned resources are created, the collection will be empty.

In addition to making life easier for extension implementors, this will have the other benefit of treating status checking the same for managed and unmanaged resources.

PR #97, opened 2021-09-23
samples: add s3-to-s3 transfer demo sample (#96)

MikhailGordienk:
This PR should be merged after https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/94 since it depends on the serialization fix.

### DIFF
--- a/samples/other/file-transfer-s3-to-s3/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/api/ClientApiController.java
+++ b/samples/other/file-transfer-s3-to-s3/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/api/ClientApiController.java
@@ -47,14 +47,20 @@ public class ClientApiController {
 
     private final Monitor monitor;
 
+    private final String destinationRegion;
+
+    private final String destinationBucket;
+
     public ClientApiController(
             RemoteMessageDispatcherRegistry dispatcherRegistry,
             TransferProcessManager processManager,
-            Monitor monitor
-    ) {
+            Monitor monitor,
+            String destinationRegion, String destinationBucket) {
         this.dispatcherRegistry = dispatcherRegistry;
         this.processManager = processManager;
         this.monitor = monitor;
+        this.destinationRegion = destinationRegion;
+        this.destinationBucket = destinationBucket;
     }
 
     @GET
@@ -85,8 +91,6 @@ public class ClientApiController {
         var connector = (String) request.get("connector");
         var artifact = (String) request.get("artifact");
 
-        var destinationRegion = System.getProperty("destinationRegion");
-        var destinationBucket = System.getProperty("destinationBucket");
         var destinationName = extractDestinationArtifactName(artifact);
 
         var usRequest = createRequest(
--- a/samples/other/file-transfer-s3-to-s3/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/api/ClientApiController.java
+++ b/samples/other/file-transfer-s3-to-s3/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/api/ClientApiController.java
@@ -1,0 +1,137 @@
+/*
+ *  Copyright (c) 2021 Siemens AG
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Siemens AG - initial implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.transfer.demo.api;
+
+import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import org.eclipse.dataspaceconnector.spi.message.RemoteMessageDispatcherRegistry;
+import org.eclipse.dataspaceconnector.spi.monitor.Monitor;
+import org.eclipse.dataspaceconnector.spi.transfer.TransferProcessManager;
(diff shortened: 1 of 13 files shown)


samples/other/file-transfer-s3-to-s3/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/api/ClientApiController.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    @POST
+    @Path("data/request")
+    public Response initiateDataRequest(Map<String, String> request) {
+        var connector = (String) request.get("connector");
+        var artifact = (String) request.get("artifact");
+
+        var destinationRegion = System.getProperty("destinationRegion");
```
jimmarino, line 88: Config settings should come from the context so they can be sourced by a configuration extension. Here's an example of how to do this:

https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/e9f13a3a10aaa5a91adde99a2a988dc07cdcb571/extensions/iam/oauth2/src/main/java/org/eclipse/dataspaceconnector/iam/oauth2/Oauth2Extension.java#L80

We're still discussing the key formatting but it will likely be something similar to this: 

https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/e9f13a3a10aaa5a91adde99a2a988dc07cdcb571/extensions/iam/oauth2/src/main/java/org/eclipse/dataspaceconnector/iam/oauth2/Oauth2Extension.java#L39


We're also trying to keep settings resolution in the extension classes and passing them as config into controllers and other services.

One other thing that is not documented is we are trying to annotate the key constants with @EdcSetting so that later we can have tooling generate configuration documentation:

https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/e9f13a3a10aaa5a91adde99a2a988dc07cdcb571/extensions/iam/oauth2/src/main/java/org/eclipse/dataspaceconnector/iam/oauth2/Oauth2Extension.java#L38

samples/other/file-transfer-s3-to-s3/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/vault/SingletonVaultExtension.java
```diff
@@ -0,0 +1,57 @@
...
+    public PrivateKeyResolver getPrivateKeyResolver() {
+        return null;
+    }
+
+    @Override
+    public CertificateResolver getCertificateResolver() {
+        return null;
+    }
+
+    private final String SECRET_PROPERTY = "creds";
```
jimmarino, line 49: one stylistic nit: can you move this to the top of the class and make it static so it is consistent with other classes?

jimmarino (commented):
Thanks, Mikhail! 

I had a few minor comments inline.

samples/other/file-transfer-s3-to-s3/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/api/ClientApiController.java
MikhailGordienk: Hello! 

But _connector_ and _artifact_ are not config settings, these are request parameters. I believe you meant _destinationRegion_ and _destinationBucket_, right?

MikhailGordienk: I've updated the PR

samples/other/file-transfer-s3-to-s3/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/vault/SingletonVaultExtension.java
beardyinc: about code formatting: please refer to the [styleguide](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/styleguide.md) and associated editor configs to keep code style consistent.

MikhailGordienk: I used the codestyle for the IntelliJ IDEA ([here](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/tree/main/resources)). Code inspection with the CheckStyle plugin points to a deprecated API - Metadatastore, which is crucial for the demo at the moment.

samples/other/file-transfer-s3-to-s3/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/api/ClientApiController.java
jimmarino: Yes, sorry my mistake

samples/other/file-transfer-s3-to-s3/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/vault/SingletonVaultExtension.java
jimmarino: The MetadataStore will be replaced with AssetIndex. It's OK for the time being to use the deprecated class

The SECRET_PROPERY should be moved up to the top of the class and use a descriptive string with "." similar to the previous examples. Thanks!

MikhailGordienk: I believe that's been done already. It now looks like this (at the beginning of the class):

```
@EdcSetting
private static final String CREDS = "dataspaceconnector.transfer.demo.s3.destination.creds";
```

MikhailGordienk: @jimmarino Is it OK?

jimmarino: Looks good. I'll squash and merge. As a heads up, we had to make a change to StatusChecker (#95) that was just merged as well. This change should not impact you, but if your end-to-end test does not complete or throw an error, it could be related to that change. If so, ping either myself or @beardyinc.

PR #98, opened 2021-09-23
Decouple status checking from provisioned resources

jimmarino:
See #95 for details

### DIFF
--- a/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
+++ b/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
@@ -225,14 +225,14 @@ public class TransferProcessManagerImpl extends TransferProcessObservable implem
                 var checker = statusCheckerRegistry.resolve(process.getDataRequest().getDestinationType());
                 if (checker == null) {
                     monitor.info(format("No checker found for process %s. The process will not advance to the COMPLETED state.", process.getId()));
-                } else if (checker.isComplete(process.getId(), resources)) {
+                } else if (checker.isComplete(process, resources)) {
                     // checker passed, transition the process to the COMPLETED state
                     transitionToCompleted(process);
                 }
             } else {
                 var checker = statusCheckerRegistry.resolve(process.getDataRequest().getDestinationType());
                 if (checker != null) {
-                    if (checker.isComplete(process.getId(), emptyList())) {
+                    if (checker.isComplete(process, emptyList())) {
                         //checker passed, transition the process to the COMPLETED state automatically
                         transitionToCompleted(process);
                     }
--- a/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
+++ b/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
@@ -27,8 +27,6 @@ import org.eclipse.dataspaceconnector.spi.transfer.provision.ResourceManifestGen
 import org.eclipse.dataspaceconnector.spi.transfer.response.ResponseStatus;
 import org.eclipse.dataspaceconnector.spi.transfer.store.TransferProcessStore;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataRequest;
-import org.eclipse.dataspaceconnector.spi.types.domain.transfer.ProvisionedDataDestinationResource;
-import org.eclipse.dataspaceconnector.spi.types.domain.transfer.ProvisionedResource;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.ResourceManifest;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.StatusCheckerRegistry;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcess;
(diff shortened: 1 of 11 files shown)


core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
```diff
@@ -219,30 +217,41 @@ private int checkProvisioned() {
...
-                //no managed resources, just update
-                process.transitionCompleted();
-                monitor.debug("Process " + process.getId() + " is now " + TransferProcessStates.COMPLETED);
-                invokeForEach(listener -> listener.completed(process));
+                var checker = statusCheckerRegistry.resolve(process.getDataRequest().getDestinationType());
+                if (checker != null) {
+                    if (checker.isComplete(process.getId(), emptyList())) {
+                        //checker passed, transition the process to the COMPLETED state automatically
+                        transitionToCompleted(process);
+                    }
```
beardyinc, line 238: This will always throw an `EdcException` for unmanaged S3 buckets, i.e. when the `S3StatusChecker` is used: 
```java
        // in S3StatusChecker.java: 
        for (var resource : resources) {
              // do completion checking
        }
        throw new EdcException(format("No bucket resource was associated with...."));
```
so the exceptio is raised when an empyt list is passed.

One option would be to obtain the transfer process by its `id` and use 
```java
TransferProcess.getDataRequest().getDataDestination()
``` 
as a fallback.

The same is true for the `ObjectContainerStatusChecker`

extensions/aws/s3/provision/src/main/java/org/eclipse/dataspaceconnector/provision/aws/s3/S3StatusChecker.java
```diff
@@ -17,41 +17,54 @@
...
-
-            var rq = ListObjectsRequest.builder().bucket(bucketName).build();
-            var response = Failsafe.with(retryPolicy)
-                    .getStageAsync(() -> s3client.listObjects(rq))
-                    .join();
-            return response.contents().stream().anyMatch(s3object -> s3object.key().endsWith(".complete"));
-        } catch (CompletionException cpe) {
-            if (cpe.getCause() instanceof NoSuchBucketException) {
-                return false;
+    public boolean isComplete(String id, List<ProvisionedResource> resources) {
```
beardyinc, line 42: Can we rename `id` to `transferProcessId` to make it more intuitive?

extensions/aws/s3/provision/src/test/java/org/eclipse/dataspaceconnector/provision/aws/s3/S3StatusCheckerTest.java
```diff
@@ -36,9 +37,6 @@
 
 @IntegrationTest
 class S3StatusCheckerTest extends AbstractS3Test {
```
beardyinc, line 39: Since the `S3StatusChecker` changed its implementation by moving over to a for-loop, all tests here will *fail* because an `emptyList()` is passed. See my comment in the `S3StatusChecker.java` for reference.

The same is true for the Azure variant of the Checker and -test.

beardyinc (changes requested):
The way the S3- and ObjectStorage Status checkers are now implemented, they will deterministically raise an exception when there are unmanaged resources (`TransferProcessManagerImpl.java`, Line 235).

Also, integration tests will fail for that exact reason.

extensions/azure/blob/provision/src/test/java/org/eclipse/dataspaceconnector/provision/azure/blob/ObjectContainerStatusCheckerTest.java
```diff
@@ -57,28 +57,19 @@ void isComplete() {
         putBlob("hello.txt", helloTxt);
         putBlob(testRunId + ".complete", helloTxt);
 
-        assertThat(checker.isComplete(createResource(containerName))).isTrue();
+        assertThat(checker.isComplete("123", emptyList())).isTrue();
```
beardyinc, line 60: "123" is not the right name.
```suggestion
        assertThat(checker.isComplete(bucketName, emptyList())).isTrue();
```

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
jimmarino: OK thanks, Paul and I discussed and he supplied a patch that I just committed

extensions/aws/s3/provision/src/main/java/org/eclipse/dataspaceconnector/provision/aws/s3/S3StatusChecker.java
jimmarino: This should be resolved by the patch

extensions/aws/s3/provision/src/test/java/org/eclipse/dataspaceconnector/provision/aws/s3/S3StatusCheckerTest.java
jimmarino: Should be fixed in the patch

extensions/azure/blob/provision/src/test/java/org/eclipse/dataspaceconnector/provision/azure/blob/ObjectContainerStatusCheckerTest.java
jimmarino: Fixed in the patch

beardyinc:
Just a reminder, Checkstyle has found an issue in `TransferProcessManagerImplConsumerTest`, Line 268

Issue #84, opened 2021-09-20
Standardize EDC Setting Keys


jimmarino:
The EDC setting keys should be standardized and implement a namespace feature. Dot notation is used already, so we should probably reserve the first segment for that purpose. The issue is some settings use "dataspaceconnector" while others use "edc". For example:

public class CosmosTransferProcessStoreExtension implements ServiceExtension {

    @EdcSetting
    private static final String COSMOS_DBNAME_SETTING = "edc.cosmos.database.name";
    @EdcSetting
    private static final String COSMOS_PARTITION_KEY_SETTING = "dataspaceconnector.cosmos.partitionkey";
    private static final String DEFAULT_PARTITION_KEY = "dataspaceconnector";
    private static final String CONTAINER_NAME = "transferprocess";

----------------------
I would vote for brevity, and hence "edc".

jimmarino:
@bscholtes1A did your PR solve this and can we close this issue out?

bscholtes1A:
@jimmarino yeap, we can close it

PR #113, opened 2021-09-29
Standardize EDC setting keys (#84)

bscholtes1A:
Apply standard naming for all EDC settings: edc.*

### DIFF
--- a/core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/CoreServicesExtension.java
+++ b/core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/CoreServicesExtension.java
@@ -16,6 +16,7 @@ package org.eclipse.dataspaceconnector.system;
 
 import net.jodah.failsafe.RetryPolicy;
 import okhttp3.OkHttpClient;
+import org.eclipse.dataspaceconnector.spi.EdcSetting;
 import org.eclipse.dataspaceconnector.spi.monitor.Monitor;
 import org.eclipse.dataspaceconnector.spi.system.ServiceExtension;
 import org.eclipse.dataspaceconnector.spi.system.ServiceExtensionContext;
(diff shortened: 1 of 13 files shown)


core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/CoreServicesExtension.java
```diff
@@ -28,9 +29,15 @@
 
     public static final String FEATURE_HTTP_CLIENT = "dataspaceconnector:http-client";
     public static final String FEATURE_RETRY_POLICY = "dataspaceconnector:retry-policy";
-    private static final String MAX_RETRIES = "dataspaceconnector:core.retry.max-retries";
-    private static final String BACKOFF_MIN_MILLIS = "dataspaceconnector.core.retry.backoff.min";
-    private static final String BACKOFF_MAX_MILLIS = "dataspaceconnector.core.retry.backoff.max";
+
+    @EdcSetting
+    private static final String MAX_RETRIES = "edc.core.retry.max-retries";
```
jimmarino, line 34: Hi Benjamin, do you think we should also change max-retries to max.retries for consistency? I don't have any particular preference, I just wanted to raise it as a qyestion

jimmarino (commented):
Thanks, Benjamin, looks good to me. I had one minor question, otherwise, let's see what others think and we can merge if there are no issues.

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/CoreServicesExtension.java
bscholtes1A: Hi @jimmarino 
if we want to be consistent with the others settings just below (e.g. edc.core.retry.backoff.min...) it should even be edc.core.retry.retries.max :)
I dont have preference either, let me know which version you prefer and I'll made the change.

jimmarino: That seems reasonable

bscholtes1A: @jimmarino renamed as discussed

beardyinc:
looks good to me

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/CoreServicesExtension.java
jimmarino: @MoritzKeppler are you OK with this? If so, I'll merge

MoritzKeppler (approved):
fine for me, thx for cleaning up!

PR #114, opened 2021-09-30
Improve and simplify Launcher concept

beardyinc:
There is now a base class for all runtimes called `BaseRuntime.java` which has various methods that can be overridden if so desired.
This means that for a new launcher, only a `build.gradle(.kts)` is required.

An example of how to create a custom runtime is located in `samples/other/custom-runtime`


core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/runtime/BaseRuntime.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * Initializes the context. If {@link BaseRuntime#createContext(TypeManager, Monitor)} is overridden and the (custom) context
+     * needs to be initialized, this method should be overridden as well.
+     *
+     * @param context The context.
+     */
+    protected void initializeContext(ServiceExtensionContext context) {
+        if (context instanceof DefaultServiceExtensionContext) {
```
MoritzKeppler, line 81: why is initialize() not part of the interface?

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/runtime/BaseRuntime.java
```diff
@@ -0,0 +1,183 @@
...
+     * @param context A context to which all the service extensions should be registered
+     * @return a list of {@code ServiceExtension}s
+     */
+    @NotNull
+    protected List<ServiceExtension> createExtensions(ServiceExtensionContext context) {
+        return context.loadServiceExtensions();
+    }
+
+    /**
+     * Create a {@link ServiceExtensionContext} that will be used in this runtime. If e.g. a custom DI mechanism should be used,
```
MoritzKeppler, line 125: I wouldn't call it "DI" :-)

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/runtime/BaseRuntime.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+
+    /**
+     * Hook point to instantiate a {@link Monitor}. By default, the {@link ExtensionLoader#loadMonitor()} method is called.
+     * <p>
+     * <strong>Do not use this to provide a custom Monitor implementation! Use the extension mechanism for that! </strong>
+     */
+    @NotNull
+    private Monitor createMonitor() {
```
MoritzKeppler, line 179: load or create?

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/runtime/BaseRuntime.java
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * Hook point to initialize the vault. By default this calls {@link org.eclipse.dataspaceconnector.system.ExtensionLoader#loadVault(ServiceExtensionContext)}
+     * to instantiate and configure the vault.
+     * <p>
+     * <strong>Do not use this to provide a custom Vault implementation! Use the extension mechanism for that! </strong>
+     *
+     * @param context An {@code ServiceExtensionContext} to resolve the {@code Vault} from.
+     */
+    protected void initializeVault(ServiceExtensionContext context) {
```
MoritzKeppler, line 160: load or init?

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/runtime/BaseRuntime.java
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * Create a list of {@link ServiceExtension}s. By default this is done using the ServiceLoader mechanism. Override if
+     * e.g. a custom DI mechanism should be used.
+     *
+     * @param context A context to which all the service extensions should be registered
+     * @return a list of {@code ServiceExtension}s
+     */
+    @NotNull
+    protected List<ServiceExtension> createExtensions(ServiceExtensionContext context) {
```
MoritzKeppler, line 120: load or create?

MoritzKeppler (changes requested):
just some minor wording findings. Really appreciate to get rid of all that duplicated classes, thx!

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/runtime/BaseRuntime.java
beardyinc: I think `create` is fine, because the fact that the default runtime (i.e. `BaseRuntime`) actually loads a monitor based on all available `MonitorExtension` classes that are loaded using the `ServiceLoader` mechanism is an implementation detail. I'll update the comment.

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/runtime/BaseRuntime.java
beardyinc: again, since the method returns `void`, and since the fact that internally the vault is actually _loaded_ (as opposed to created) is an implementation detail, I think `initialize` is fine here. It implies that the vault is already there beforehand. 
I'll improve the comments however.

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/runtime/BaseRuntime.java
beardyinc: like before it is not necessary for an implementor to know that the default implementation actually _loads_ `ServiceExtensions`.

MoritzKeppler: You can also argue the other way round and state that it's an implementation detail if something needs to be created first or can be loaded from somewhere - at the end it's about loading.
But feels not important enough to discuss that :-) , I'm fine with it and will merge it as soon as you are ready.

MoritzKeppler (approved):

PR #115, opened 2021-10-01
ids: add self-description-api, closes #41

### LINKED ISSUE #41: Implementent Basic IDS Self-Description API of the Connector

DominikPinsel:
Hello everyone,

this pull-request contains the implementation for IDS self-description. @denisneuling and I (again) tried to keep it small by adding ToDo's and follow up issues for some of the required changes.

I think, besides the normal implementation, there are three major issues within this pull-request.

Issue # 1: Mapping EDC Model -> IDS Model
We were not able to map the EDC Asset to a triplet of IDS Resource+Representation+Artifact as planned. The reason is, that the IDS Resource contains offers. And the EDC Contract Offer is able to contain multiple assets. A mapping of an EDC contract offer, that contains multiple assets, would look very strange if we map all the assets as triplets.
Therefore we are now creating one resource per contract offer. And one representation+artifact pair per asset.

Issue # 2: Old IDS Extensions
There were already some IDS extension ins the code base. But nether of them implements IDS. So we removed them or the non-IDS code.

Issue # 3: Policy Model
The current EDC policy model is not map-able to the IDS contract model. You can see it best in the ContractOfferFactory.java.
We should create an extra discussion or issue where we can discuss a possible solution. For now its not that of a big problem, because we are able to map the "just use all my data like you want"-contract perfectly fine between the two models. But it doesn't work as soon as one of the models specifies any details.

If you have questions about these three topics please feel free to ask. We discussed them for days and the current implementation just shows our current state of discussion.
To decide how we should proceed with the policy model I would suggest to open a new GitHub discussion.


launchers/ids-connector/src/main/java/org/eclipse/dataspaceconnector/runtime/EdcRuntime.java
```diff
@@ -0,0 +1,68 @@
+/*
```
bscholtes1A, line 1: These EdcRuntime files are no longer needed. Instead in the build.gradle.kts file you should replace
`
application {
    @Suppress("DEPRECATION")
    mainClassName = "org.eclipse.dataspaceconnector.runtime.ConnectorRuntime"
}
`

by

`
application {
    @Suppress("DEPRECATION")
    mainClassName = "org.eclipse.dataspaceconnector.system.runtime.BaseRuntime"
}
`

PR #130, opened 2021-10-06
monitor: add warning level, closes #129

### LINKED ISSUE #129: Add warning and error level to the monitor
Hello everyone,

at the moment it's not possible to log warnings. I think it would be helpful if the EDC would be able to show warnings, in case some configuration is missing or sample extensions are used in production.
And, if the monitor is updated anyway, I would suggest adding the error level, too.

DominikPinsel:
Signed-off-by: Dominik Pinsel <email@dominik-pinsel.com>

### DIFF
--- a/core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/monitor/ConsoleMonitor.java
+++ b/core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/monitor/ConsoleMonitor.java
@@ -25,10 +25,16 @@ import java.util.function.Supplier;
  * Default monitor implementation. Outputs messages to the console.
  */
 public class ConsoleMonitor implements Monitor {
+
+    private static final String SEVERE = "SEVERE";
+    private static final String WARNING = "WARNING";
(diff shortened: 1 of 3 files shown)


core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/monitor/ConsoleMonitor.java
```diff
@@ -25,10 +25,17 @@
  * Default monitor implementation. Outputs messages to the console.
  */
 public class ConsoleMonitor implements Monitor {
+
+    private static final String SEVERE = "SEVERE";
+    private static final String ERROR = "ERROR";
```
beardyinc, line 30: What exactly would be the difference between `SEVERE` and `ERROR`? Do we really need both?

beardyinc (changes requested):

beardyinc:
looks like you deleted `Monitor#severe(Map<String, Object> data)`. Whats the reason? If you do delete something, you should state the reason.

PR #131, opened 2021-10-07
iam: Refactor token validation in Oauth2 service #147

bscholtes1A:
This PR is a first step in order to make the Oauth2 IAM service compatible with DAPS.


MoritzKeppler:
thx for the contribution!
Suggest to avoid "DynamicAttributeToken" in the naming of the classes etc. - or is there anything that makes it DAT specific? Guess not if it should replace the current OAuth implementation in future.

MoritzKeppler:
one more question: proper impl of public key retrieval would mean something similar to the PrivateKeyResolver concept? Guess there will be cases where the public key of a signer can be retrieved on the fly (e.g. by GET .well-known/jwks.json) but also cases where we a forced to trust a fixed and preconfigured list of public keys.

data-protocols/ids/ids-identity-provider/src/main/java/org/eclipse/dataspaceconnector/ids/daps/validation/ValidationRuleResult.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.Objects;
+
+/**
+ * Evaluation result of a custom rule, can be either success or failure.
+ */
+public class ValidationRuleResult {
+
+    private final boolean success;
+    private final String message;
```
paullatzelsperger, line 25: It might make sense to actually have a `List<String>` here, since there might be multiple violations, since we can have multiple rules, and each rule might validate multiple things.

PR #137, opened 2021-10-09
ids-spi: fixed typo in artifactId of `data-protocols.ids-spi`

oxisto:
Just a small typo that results in a weird package name when published to maven local.


jimmarino:
Thanks for the contribution!

PR #139, opened 2021-10-10
Feature/51 No more NiFi HTTP edition

jimmarino:
Transfer functions are a simple way to create data flow extensions for EDC: implement an HTTP endpoint that handles data transfer according to the documentation and configure the Transfer Functions extension.

The HTTP endpoint can be written in any language and hosted on cloud infrastructure (such as an Azure Function or AWS Lambda) for scale-out. In the future, we can support [Keda](https://keda.sh/) for on-premise or DIY scale-out.

### DIFF
--- a/extensions/transfer-functions/transfer-functions-core/README.md
+++ b/extensions/transfer-functions/transfer-functions-core/README.md
@@ -1,7 +1,7 @@
 # Transfer Functions Core
 
-This extensions provides required services for the Transfer Functions feature (.cf transfer-functions-spi).
+This extensions provides required services for the [Transfer Functions](../transfer-functions-spi/README.md) feature.
 
 # Status
 
-In development. 
+In development. This code has not yet been fully tested. 


extensions/transfer-functions/transfer-functions-core/src/main/java/org/eclipse/dataspaceconnector/transfer/functions/core/flow/http/HttpFunctionConfiguration.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.function.Supplier;
+
+/**
+ * Configuration for the HTTP data flow controller and status checker.
+ */
+public class HttpFunctionConfiguration {
+    private String transferEndpoint;
+    private String checkEndpoint;
+    private Set<String> protocols;
+    private Supplier<OkHttpClient> clientSupplier;
```
beardyinc, line 30: Why not directly use the `OkHttpClient`? It is resolvable via the `ServiceExtensionContext`

extensions/transfer-functions/transfer-functions-core/src/main/java/org/eclipse/dataspaceconnector/transfer/functions/core/TransferFunctionsCoreServiceExtension.java
```diff
@@ -0,0 +1,177 @@
...
+        var httpClient = createHttpClient(context);
+        context.registerService(TransferFunctionInterceptorRegistry.class, httpClient::addInterceptor);
+
+        var typeManager = context.getTypeManager();
+        var transferEndpoint = context.getSetting(TRANSFER_URL_KEY, DEFAULT_LOCAL_TRANSFER_URL);
+        var checkEndpoint = context.getSetting(CHECK_URL_KEY, DEFAULT_LOCAL_CHECK_URL);
+        var configuration = HttpFunctionConfiguration.Builder.newInstance()
+                .transferEndpoint(transferEndpoint)
+                .checkEndpoint(checkEndpoint)
+                .clientSupplier(httpClient::build)
```
beardyinc, line 136: the `httpClient` can be directly resolved using the context. In that case a `requires` dependecy to `"dataspaceconnector:http-client"` must be defined.

```suggestion
                .clientSupplier(context.getService(OkHttpClient.class)
```

Using one HttpClient globally has the upside of central configuration (timeouts, etc.) and error handling.

extensions/transfer-functions/transfer-functions-core/src/test/java/org/eclipse/dataspaceconnector/transfer/functions/core/TransferFunctionsCoreHttpTest.java
```diff
@@ -0,0 +1,95 @@
...
+                .destinationType("foo")
+                .dataEntry(dataEntry)
+                .managedResources(false)
+                .dataDestination(DataAddress.Builder.newInstance().type("test-protocol1").build())
+                .connectorId("test").build();
+
+        processManager.initiateProviderRequest(dataRequest);
+
+        latch.await(1, TimeUnit.MINUTES);
+        Assertions.assertTrue(invoked.get());
```
beardyinc, line 77: we can always `assertTrue(latch.await(...))`, which would save us the `assertTrue(invoked.get())`. 
```suggestion
       assertThat(latch.await(1, TimeUnit.MINUTES)).isTrue()
```
Also, I think we should try and stick with one assertion framework (here: AssertJ)

extensions/transfer-functions/transfer-functions-core/src/test/java/org/eclipse/dataspaceconnector/transfer/functions/core/TransferFunctionsCoreLocalTest.java
```diff
@@ -0,0 +1,89 @@
...
+                .destinationType("foo")
+                .dataEntry(dataEntry)
+                .managedResources(false)
+                .dataDestination(DataAddress.Builder.newInstance().type("test-protocol1").build())
+                .connectorId("test").build();
+
+        processManager.initiateProviderRequest(dataRequest);
+
+        latch.await(1, TimeUnit.MINUTES);
+        Assertions.assertTrue(invoked);
```
beardyinc, line 62: we can always assertTrue(latch.await(...)), which would save us the assertTrue(invoked.get()).
```suggestion
        assertThat(latch.await(1, TimeUnit.MINUTES)).isTrue();
```
Also, I think we should try and stick with one assertion framework (here: AssertJ)

launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/MockVault.java
```diff
@@ -0,0 +1,45 @@
...
+import org.eclipse.dataspaceconnector.spi.security.VaultResponse;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * An in-memory vault for testing.
+ */
+public class MockVault implements Vault {
```
beardyinc, line 26: This class seems not to be used anywhere. There is another `MockVault` in `DemoProtocolsTransferExtensionTest.java`, which looks pretty much the same.

beardyinc (changes requested):

launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/MockVault.java
jimmarino: Oh this looks like a rebase problem. I'll delete it

extensions/transfer-functions/transfer-functions-core/src/main/java/org/eclipse/dataspaceconnector/transfer/functions/core/flow/http/HttpFunctionConfiguration.java
jimmarino: I need to document this...The extension will allow the use of custom interceptors to set headers for things like authorization.

beardyinc:
good to go

extensions/transfer-functions/transfer-functions-core/src/main/java/org/eclipse/dataspaceconnector/transfer/functions/core/flow/http/HttpFunctionDataFlowController.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public boolean canHandle(DataRequest dataRequest) {
+        return protocols.contains(dataRequest.getDestinationType());
+    }
+
+    @Override
+    public @NotNull DataFlowInitiateResponse initiateFlow(DataRequest dataRequest) {
+        var requestBody = RequestBody.create(typeManager.writeValueAsString(dataRequest), JSON);
```
MoritzKeppler, line 64: I always get a little stomach ache when passing internal objects to processes outside without beeing in control what's actually inside - do we know that a DataRequest contains solely information the other party is allowed to see?
Would it make sense to document that usage in the DataRequest class? e.g. "just store information that can be passed to ..."

MoritzKeppler (approved):

Issue #127, opened 2021-10-06
Acquire IDS DAT token from DAPS


denisneuling:
In order to enable two IDS-speaking connectors to communicate with each other we have to make them able to obtain and use [DAT](https://github.com/International-Data-Spaces-Association/IDS-G/blob/main/Components/IdentityProvider/DAPS/README.md) tokens from [DAPS](https://github.com/International-Data-Spaces-Association/IDS-G/blob/main/Components/IdentityProvider/DAPS/README.md) instances. The IDS DAPS service implements the [JWT Bearer Token client authentication flow](https://datatracker.ietf.org/doc/html/rfc7523#section-2.2) of the [OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749).

<sub> Denis Neuling <denis.neuling@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)

jimmarino:
@bscholtes1A @paullatzelsperger Has this issue been handled?

bscholtes1A:
> @bscholtes1A @paullatzelsperger Has this issue been handled?

Hi Jim, 
Yeap acquiring the DAPS token is now handled through the Oauth2 extension + the DAPS extension which is a very small wrapper around the Oauth2 extension. So this one can be closed.

PR #141, opened 2021-10-11
ids-identity-provider: draft, contributes to #127

DominikPinsel:
Hello everyone,

like discussed here is the draft pull request for the DAPS acquisition from @denisneuling and me.
Contributes to issue #127 


MoritzKeppler:
we will provide a patch for the existing OAuth lib instead.

PR #142, opened 2021-10-11
Fixed the AssetIndex 

beardyinc:
This PR aims at fixing the `AssetIndex` by removing the composite pattern and simplifying the `AssetSelectorExpression`.
Also, the code has been cleaned up so as to conform with the rest of the code base.

There is now the `AssetIndex` interface and an in-memory implementation which is ready-to-use. 

The sample implementation has been cleaned up, it now contains an incomplete implementation to spark a discussion, whether we need it at all. 
Personally I think we don't since we already have the in-mem impl.


DominikPinsel:
> @DominikPinsel lets go through this, this PR can serve us as a "straw man"

Sure. Did you see that I already implemented the things from our review already on the pull request branch?
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/115

Also, it seems that the title of your pull request does not reflect its purpose. For my understanding "fixing" does not
mean introducing new functionality as you did among others by extending the AssetIndex interface. Additionally, I'm not
sure what is fixed on this branch, as the core-contract [is not even in a runnable state](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/142/files#diff-3b4641f08672c88287416ed598226cf278d530bb482754efc651980336722dc7R60) and some changes seem more like draft implementations to support the discussion.

According to contributing guideline, we should [not combine independent things in one pull request](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/CONTRIBUTING.md#submit-a-pull-request). So I suggest we split this up into multiple pull requests, to separately address

- adding new functionality
- discussing and modifying existing solutions

Maybe let's just start with the following two topics, until they are resolved or the PR is split, so that the discussion
stays well-arranged.

**In Memory Asset Index**

Comparing to the current sample `AssetIndex` implementation your approach seems not to be an enhancement over it. I'd
appreciate if you would explain this in comparison to the current sample implementation.

It looks like there are the following downsides in comparison to the sample implementation:

- It's not possible to add data outside a unit test. So it's a running connector instance would always have zero assets.
- It is not as good as a sample, as the data behind is missing (even if its fixture data and IDSA documents).
- No functionality "out of the box". If each of the EDC sample/demo extensions requires the user to invest some time, to
  understand and set up the extension, running a functional connector for the first time will be very time-consuming.

**Asset Selector Expression**

The current implementation uses labels as selectors. Your implementation changed it to predicates. While it's convenient
for a unit test to use predicates, it's not so good for other purposes.

It looks like there are the following downsides in comparison to the labels:

- If the storage behind the `AssetIndex` is a database, the EDC has to load the complete database in memory, which would
  lead to performance degradation.
- Labels are easy to implement, explain and document. They offer a natural way for implementing custom filtering logic
  tied to not yet known pluggable data retention systems.
- Currently, there are two kinds of extensions, that are working with the selector. The `ContractOfferFramework`, that
  creates the selector, and the `AssetIndex`, that uses the selector. If you use a predicate  the `ContractOfferFramework`
  might have to know a lot about the `AssetIndex` and its `Asset`s. Labels are able to work as some kind of abstraction
  layer between those two components.
- A company that organizes their `Asset`s and `ContractOffer`s within the data space connector can use the labels to do
  so (a little like with Kubernetes labels). While it's also possible to do this with predicates, it looks like an
  unnecessary layer of complexity, when labels might be sufficient.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/asset/Asset.java
```diff
@@ -48,91 +51,73 @@ public String getVersion() {
...
-    public String getFileName() {
-        return fileName;
-    }
 
     public Integer getByteSize() {
         return byteSize;
     }
 
-    // TODO in theory each asset may have multiple representations/media types, depending on the transfer capabilities of the EDC
-    public String getFileExtension() {
```
DominikPinsel, line 60: If we remove the file extension from the Asset, there is no way to describe the IDS Representation. In this cast it's just a class without any set properties.

extensions/in-memory/assetindex-memory/src/main/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexExtension.java
```diff
@@ -0,0 +1,56 @@
...
+    private Monitor monitor;
+
+    @Override
+    public Set<String> provides() {
+        return Set.of(AssetIndex.FEATURE);
+    }
+
+    @Override
+    public LoadPhase phase() {
+        return LoadPhase.PRIMORDIAL;
```
DominikPinsel, line 34: Why not use LoadPhase.DEFAULT like all the other ServiceExtensions?

extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexTest.java
```diff
@@ -0,0 +1,163 @@
...
+    void queryAssets_fieldNull() {
+        var testAsset = Asset.Builder.newInstance().id(UUID.randomUUID().toString()).name("foobar").version("1").build();
+        index.add(testAsset, createDataAddress(testAsset));
+        var assets = index.queryAssets(AssetSelectorExpression.Builder.newInstance().whenEquals("description", "barbaz").build());
+        assertThat(assets).isEmpty();
+    }
+
+    @Test
+    void queryAssets_multipleFound() {
+        var testAsset1 = Asset.Builder.newInstance().id(UUID.randomUUID().toString()).name("foobar").version("1").build();
```
DominikPinsel, line 55: Maybe remove the asset creation and the addition of the asset to the index to a dedicated method to reduce noise and amount of duplicated code.

extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexTest.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @Test
+    void findById_multiple_raisesException() {
+        String id = UUID.randomUUID().toString();
+        var testAsset = Asset.Builder.newInstance().id(id).name("foobar").version("1").build();
+        var testAsset2 = Asset.Builder.newInstance().id(id).name("foobar").version("1").build();
+
+        index.add(testAsset, createDataAddress(testAsset));
+        index.add(testAsset2, createDataAddress(testAsset2)); //in an actual DB this would likely cause problems already
```
DominikPinsel, line 110: This is an interesting comment. Maybe you can also explain why this leads to problems. So that future developers, that see these comment, understand the reasoning.

DominikPinsel:
Maybe you can move the removal of the SampleAssetIndex to a dedicated pull-request, so we don't have to discuss it here.

extensions/in-memory/assetindex-memory/src/main/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndex.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * An ephemeral asset index.
+ */
+public class InMemoryAssetIndex implements AssetIndex {
+    private final List<Asset> cache = new LinkedList<>();
+    private final Map<String, DataAddress> dataAddresses = new HashMap<>();
+    private final ReentrantReadWriteLock lock;
```
ndr-brt, line 39: Maybe we don't need a read-write lock on an "ephemeral" in-memory index

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndex.java
```diff
@@ -15,10 +15,45 @@
...
+     */
+    Asset findById(String assetId);
+
+    /**
+     * Resolves a {@link DataAddress} for a given {@code Asset}. A {@code DataAddress} can be understood as a pointer into
+     * a storage system like a database or a document store.
+     *
+     * @param assetId The {@code assetId} for which the data pointer should be fetched.
+     * @return A DataAddress
+     * @throws NullPointerException     if {@code asset} was null
```
DominikPinsel, line 55: I don't understand how the resolveForAsset can throw an NullPointerException. Shouldn't it just return null (or better Optional<DataAddress) in case the assetId is invalid (e.g. null)?

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndex.java
```diff
@@ -15,10 +15,45 @@
...
+     * @throws IllegalArgumentException if no corresponding {@code DataAddress} was found for a certain asset
+     */
+    DataAddress resolveForAsset(String assetId);
```
DominikPinsel, line 58: This should not be part of the AssetIndex, as the AssetIndex only cares about Assets. Not in regards to their future contract or data transfer.

extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexTest.java
ndr-brt: Maybe would be better to get the error on the `add` call directly?

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndex.java
```diff
@@ -15,10 +15,45 @@
...
+     * @throws NullPointerException if the {@code AssetSelectorExpression} is null
+     */
     Stream<Asset> queryAssets(AssetSelectorExpression expression);
+
+    /**
+     * Fetches the {@link Asset} with the given ID from the metadata backend.
+     *
+     * @param assetId A String that represents the Asset ID, in most cases this will be a UUID.
+     * @return The {@link Asset} if one was found, or null otherwise.
+     * @throws NullPointerException If {@code assetId} was null or empty.
```
DominikPinsel, line 45: Add `@org.jetbrains.annotations.NotNull` to the argument

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndex.java
```diff
@@ -15,10 +15,45 @@
...
+     */
+    Asset findById(String assetId);
```
DominikPinsel, line 47: Since we use Java 11 we should utilize its semantic capabilities. I suggest using Optional<Asset> for  obvious nullable return values.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/asset/Asset.java
beardyinc: Ok once again. If IDS requires a file extension or a byte size, that's a bug in the spec and it will be changed. The reason for this is because streams don't **have** file names or byte sizes and we can't send erroneous information down.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/asset/Asset.java
```diff
@@ -48,91 +51,73 @@ public String getVersion() {
         return version;
     }
 
-    public String getFileName() {
```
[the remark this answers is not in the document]
beardyinc: see the comment above.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetSelectorExpression.java
```diff
@@ -14,43 +14,104 @@
...
+
+    public AssetSelectorExpression(String l, String op, String r) {
+        criteria = new ArrayList<>();
+        criteria.add(new Criterion(l, op, r));
     }
 
-    public Map<String, String> getFilterLabels() {
-        return Collections.unmodifiableMap(filterLabels);
+    public List<Criterion> getCriteria() {
+        return criteria;
```
DominikPinsel, line 55: Please return an immutable list

core/contract/src/test/java/org/eclipse/dataspaceconnector/contract/ContractOfferServiceImplTest.java
```diff
@@ -60,87 +63,87 @@ void testConstructorNullParametersThrowingIllegalArgumentException() {
...
-        EasyMock.expect(contractOfferFramework.queryTemplates(EasyMock.isA(ContractOfferFrameworkQuery.class)))
+        expect(contractOfferFramework.queryTemplates(isA(ContractOfferFrameworkQuery.class)))
                 .andReturn(Stream.of(contractOfferTemplate));
-        EasyMock.expect(contractOfferTemplate.getSelectorExpression())
-                .andReturn(Optional.empty());
-        EasyMock.expect(contractOfferTemplate.getTemplatedOffers(EasyMock.isA(Stream.class)))
-                .andReturn(Stream.empty());
+        expect(contractOfferTemplate.getSelectorExpression())
+                .andReturn(null);
+        //        expect(contractOfferTemplate.getTemplatedOffers(isA(Stream.class)))
```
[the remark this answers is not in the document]
beardyinc: done.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetSelectorExpression.java
```diff
@@ -14,43 +14,104 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        AssetSelectorExpression that = (AssetSelectorExpression) o;
+        return criteria == that.criteria || criteria.equals(that.criteria);
```
DominikPinsel, line 72: Wouldn't it be better to use `Objects.hashCode(Object o)`

extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexTest.java
beardyinc: i changed the in mem store to use a `Map`, so it becomes a non-issue.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndex.java
```diff
@@ -15,10 +15,45 @@
...
+
+    /**
+     * Resolves a {@link DataAddress} for a given {@code Asset}. A {@code DataAddress} can be understood as a pointer into
+     * a storage system like a database or a document store.
+     *
+     * @param assetId The {@code assetId} for which the data pointer should be fetched.
+     * @return A DataAddress
+     * @throws NullPointerException     if {@code asset} was null
+     * @throws IllegalArgumentException if no corresponding {@code DataAddress} was found for a certain asset
```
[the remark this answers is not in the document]
beardyinc: Javadoc was still referring to the "old" variable name. If `assetId` is null, a NPE is the way to go because it's an error.

extensions/in-memory/assetindex-memory/src/main/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndex.java
beardyinc: unfortunately, we do.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetSelectorExpression.java
```diff
@@ -14,43 +14,104 @@
...
     }
 
+    @JsonPOJOBuilder(withPrefix = "")
     public static final class Builder {
-        private final Map<String, String> labels = new HashMap<>();
+        private final AssetSelectorExpression expression;
```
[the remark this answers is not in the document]
beardyinc: this has already been changed.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndex.java
jimmarino: Hey Dominik, we discussed this at length, the codebase doesn't use Optionals in this way, and we've standardized on using nulls and annotations where appropriate (an alternative approach). That works well, and it's semantically valid in Java to do so.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndex.java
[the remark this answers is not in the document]
jimmarino: Please see my comment above

extensions/in-memory/assetindex-memory/src/main/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndex.java
ndr-brt: I miss the reasons, sorry, could you point me where to find them?

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndex.java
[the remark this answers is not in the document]
jimmarino: Please see my comment above

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndex.java
ndr-brt: @jimmarino could worth defining a "code guideline" markdown to avoid misunderstandings?

spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/asset/Asset.java
DominikPinsel: So, if we don't add these metadata to the labels and still delete them. You suggest to not describe the IDS representation and artifact (except byte size) until they change their spec?

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/Criterion.java
```diff
@@ -0,0 +1,63 @@
...
+ * {@code
+ * "operandLeft" = "name",
+ * "operator" = "=",
+ * "operandRight" = "someone"
+ * }
+ * <p>
+ * can be translated to [select * where name = someone]
+ * }
+ */
+public class Criterion {
```
DominikPinsel, line 21: Just proposing:

- Left operand and right operand should implement a marker interface called `Operand`.
- Operator should implement a marker interface `Operator`. 

Enum would have the disadvantage that we are limited to a specific set of operators. Whereas interfaces are easily extendible, without touching the existing code base.

Using `String.Equals()` is not a good way to check the operator type. For example:
Some people could check for equals using the operator `=` or `==`. Which one is correct? Aren't these operators the same? 

In the policy model, we use marker interfaces, too.

beardyinc: I should have explained this but a `Criterion` does not care whether you write "=" or "==" since its only a container. The object that evaluates the criterion - in the in-memory-index case thats the `CriterionToPredicateConverter` - needs to be aware of that. There could be converters that convert into SQL, Gremlin, etc, and that will really depend on the implementation of the index.
In the case of the InMemoryAssetIndex we're just converting to Predicates, and our "DSL" only accepts "=" as an operand - much like it was before.

extensions/in-memory/assetindex-memory/src/main/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexSeeder.java
```diff
@@ -0,0 +1,17 @@
+package org.eclipse.dataspaceconnector.metadata.memory;
+
+import org.eclipse.dataspaceconnector.spi.types.domain.asset.Asset;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataAddress;
+
+public class InMemoryAssetIndexSeeder {
```
DominikPinsel, line 6: I'm more sure whether I understood the intention. The idea is that other extensions reference the in memory asset index extension, create a new seeder, and. use it to add data to the asset index?

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndex.java
DominikPinsel: I don't like the idea of merging a pull request to main with the intention to create follow up issues. Why not discuss it here, where the method was introduced?

extensions/in-memory/assetindex-memory/build.gradle.kts
```diff
@@ -0,0 +1,30 @@
...
+    `java-library`
+}
+
+
+dependencies {
+    api(project(":spi"))
+}
+publishing {
+    publications {
+        create<MavenPublication>("in-mem.assetindex") {
```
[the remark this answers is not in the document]
DominikPinsel: Thanks :) Feel free to mark this as resolved

extensions/in-memory/assetindex-memory/src/main/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexSeeder.java
DominikPinsel: Thanks for your answer. Maybe you can help me here. Is it ok for low level extensions to reference each other? I always thought dependencies between extensions must be handled using the SPI and ServiceExtensionContext.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/asset/Asset.java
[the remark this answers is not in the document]
jimmarino: This goes back to the previous discussions we've had on the architectural vision we have for the EDC. We'd like it to be a general data-sharing platform, supporting IDS and other protocols provided through extensions.

This requires (among other things) the platform to be architected so that it can convey extensible types. One analogy that may help clarify this statement is that we are building software akin to messaging middleware. Those systems, by definition, have a concept of a "message," which nearly always contains a header and payload; they don't know anything about the semantics of the data they carry. 

The *core* EDC modules should be similar to this. Extensions "above" and "below" the core may understand additional qualities about the extensible types. To make this a little more concrete, the issue with the fileName can be handled with this architectural principle in mind by treating Asset as an extensible type (which it should be). I would not expect the base Asset type to include the union of all properties mandated by all protocols EDC supports. Because IDS defines it (IMO a bug in the spec that we can easily change), it does not mean we should include it on the base Asset class. 

There are two ways of handling extensible types, and I think both are valid. One is to use extensible properties (e.g. a map). The IDS InfoModel does this as well as many other information models. That the IDS folks thought of adding those properties is a good thing. Another way of doing this is to have lower or higher-level services such as the IDS modules and AssetIndex implementations subclass Asset combined with Jackson's ability to handle polymorphic (de)/serialization. That's why we have the TypeManager in the EDC (which also handles this in the Jersey JAX-RS container.

I prefer the loosey-goosey extensible properties approach because it avoids the proliferation of modules containing only types and plays nicely with schemaless and Graph databases.  However, I think that discussion is separate from this PR as the latter is concerned with the core and not how extensible types are handled in other architectural layers.

Does that make sense?

denisneuling (approved):
Also from my side thanks for the discussion. Improvements will be introduced when were facing issues with the current implementation. Approving.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/asset/Asset.java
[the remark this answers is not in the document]
DominikPinsel: @jimmarino Yea, I think you are right. My thoughts came from a wrong assumption about the asset index and how it would be implemented :) Paul was able to convince me in a spontaneous meeting this afternoon.

PR #148, opened 2021-10-13
Feature/ids rest

jimmarino:
This PR defines the IDS module structure and key services required to implement REST-based IDS protocol support in the EDC.

### DIFF
--- a/data-protocols/ids/ids-api-rest-dispatcher-v1/build.gradle.kts
+++ b/data-protocols/ids/ids-api-rest-dispatcher-v1/build.gradle.kts
@@ -1,0 +1,40 @@
+/*
+ *  Copyright (c) 2020, 2021 Microsoft Corporation
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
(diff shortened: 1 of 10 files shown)


jimmarino:
@beardyinc It looks as if the server-side Checkstyle and Intellij definitions don't directly match. Can you take a look?

jimmarino:
> Hi @jimmarino, thanks for your proposal, I think this is a very good starting point.
> 
> When working with the new transformer API I'd perhaps introduce an mechanism to mitigate infinite transformation calls, which would occur when passing in cyclic object graphs (e.g. by extension of the `TransformerContextImpl` to memoize already transformed objects and then just returning those).
> 

Sounds good to me.

> How shall we proceed with this PR? Making it pass the checkstyle test and then just getting it into the main branch?
>

Yeah, I need to talk to Paul about that, will do.
 
> Denis Neuling [denis.neuling@daimler.com](mailto:denis.neuling@daimler.com), Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)

beardyinc:
> @beardyinc It looks as if the server-side Checkstyle and Intellij definitions don't directly match. Can you take a look?

@jimmarino Did you also configure the `SaveActions` plugin? It does a bit more additional formatting. For reference check [here](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/resources/save_actions_screenshot.png).

jimmarino:
So, should we merge this?

denisneuling (approved):

Issue #152, opened 2021-10-15
Distributed Identity Enhancements and Web DID Support


jimmarino:
## Overview
We want to support a distributed identity solution for the next Hackathon based on [Web DIDs](https://w3c-ccg.github.io/did-method-web/). Web DIDs provide an alternative to anchoring DIDs on a DLT such as a blockchain.

The basic flow is described [here](https://github.com/Metaform/didweb).

This issue will cover three areas:

1. Cleanup the distributed identity module dependencies 
2. Add support for multiple DID methods in a runtime
3. Implement Web DID resolution and verification

## Cleanup Module Dependencies

Currently, the distributed identity modules are tied to the ION modules. There are also a number of interdependencies that should not exist.  The following dependency tree will be introduced (note this shows only the dependencies between distributed identity modules, not core modules such as *spi*):

**did-spi** - Distributed identity extension points

**crypto** (did-spi) - A library that contains utility classes. Note this module is not an extension with runtime services.

**core** (did-spi, crypto) - The Identity Hub impl and eventually the Resolver Registry

**did-service** (did-spi, crypto) - The Distributed Identity Service

**did-web** (did-spi, likely crypto) - The Web DID implementation

--

**registration service-api** (did-spi) - Ingress API for the registration service
**registration-service** (did-spi) - Implementation, will need to remove azure events dependency

--

**ion-core** (did-spi) - ION-related classes
**ion-client** (did-spi, ion-core, crypto for testing) - ION client classes

## Support for multiple DID methods in a runtime

This will involve introducing a Resolver Registry which will delegate to DID Document Resolvers based on DID method.

## Web DID Resolution and Verification

This will be implemented in an extension

jimmarino:
I've expanded the scope of this PR as we will need to reorganize and generalize some of the distributed identity modules.

PR #153, opened 2021-10-15
 Feature/152-Web-DIDs, Part 1: Cleanup Module Dependencies

jimmarino:
This is the PR for Part 1 of #152, Cleanup Module Dependencies.


jimmarino:
@beardyinc Can you take a look at this? I'd like to get it in ASAP as I need to have this in place before beginning on the Web DID implementation. 

This commit has a boatload of changes and mostly deals with Step 1 of Issue #153 (a DidResolverRegistry interface is also added).  Unfortunately, with module restructuring, there is no way to avoid touching so many files.

extensions/ion/ion-client/src/main/java/org/eclipse/dataspaceconnector/ion/IonClientExtension.java
```diff
@@ -14,7 +15,7 @@
 
     @Override
     public Set<String> provides() {
-        return Set.of(IonClient.FEATURE, DidResolver.FEATURE);
+        return Set.of(IonClient.FEATURE, DidResolverRegistry.FEATURE);
```
paullatzelsperger, line 18: I suppose the `DidResolverRegistry` should be `required`, right? Since the `IonClientExtension` will likely register the `IonClient` with it. 
Then, this extension should `provide` both `IonClient.FEATURE` and `DidResolver.FEATURE`

jimmarino: yes, but let's discuss the concept or provides and requires as a separate issue.

I didn't see any other comments but this could be because of the number of changes. Is there anything we need to address in this PR?

PR #155, opened 2021-10-15
Create UML sequence diagram depicting data flow based on HTTP push

alexandrudanciu:
Hi everyone,

this PR introduces a new UML sequence diagram depicting data flow based on HTTP push.


MoritzKeppler:
thx, @alexandrudanciu !
Would be great to embed the diagram in a markdown file describing some of the details, its draft status and what's missing (e.g. we need to add a logging agent/service, the data app in the background which triggers the exchange etc.) 
But we can do this as a separate PR if you want. What do you think?

PR #156, opened 2021-10-15
spi: Allow Asset to contain custom properties

bscholtes1A:
Allow the customization of Asset for display in the federated catalogue.


beardyinc:
Is there any particular reason you chose to model the `Operator` as an `enum` rather than `interface`?  Same for the `PropertyType` and `DataType`. Also, why not just use common  Content Type strings for the `Asset#DataType`? Wouldn't that be more interoperable?

The problem with `enum` is that they're not extensible, so whenever we need to add something we need to re-compile (and likely create a release), which is not going to work.

paullatzelsperger (approved):

PR #157, opened 2021-10-15
Fix Metadatastore service extension

jimmarino:


paullatzelsperger (approved):
LGTM, but I think we should finally remove the `MetadataStore` altogether, since it was replaced with the AssetIndex.

Issue #54, opened 2021-09-03
Get all integration tests running again in Eclipse


jimmarino:

beardyinc:
Most i-tests need some form of credentials for external systems (such as Azure, AWS,...), which need to be stored as repo secrets. Only Admins can add repo secrets, so we must request (temporary) admin status from EF and then add the secrets.

mspiekermann:
related to #25

paullatzelsperger:
Repo secrets are now in (#25), the integration test is also in, but we still need to fine-tune when they actually run:
- it **cannot** run in a fork, as the fork will likely not have the secrets
- it **cannot** run during PRs originating from a fork, because Github does not pass secrets to those PRs due to security concerns.

I think they should run before we actually merge a PR into `main`, and they should run on every push within the EDC upstream repo.

mspiekermann:
so #215 and #216 failed again. can you check what might be the difference @ndr-brt?

ndr-brt:
Looking at the history:
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/actions/workflows/integrationtests.yaml

Seems like they fails for reasons that don't depend from the code modification itself, a better analysis should be done

Issue #118, opened 2021-10-04
Replace Metadatastore with AssetIndex


DominikPinsel:
Hello everyone,

I hope I understood correct, that the AssetIndex replaces the Metadata store. In case I understood this wrong, please feel free to close this issue.

If I'm correct this issue includes

- Delete the Metadatastore interface
- Adjust existing implementations, so that they use the AssetIndex instead
- Extend the AssetIndex Interface, so that it can be queried by Asset ID

At least some sample extensions depend on the possibility to add data to the Metadatastore/AssetIndex. This is a function I would not see in the AssetIndex, and I would suggest do remove this functionality instead. What do you think?

<sub>Dominik Pinsel <dominik.pinsel@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

jimmarino:
Sounds good, but we should only do this after the AssetIndex work is functional and the extensions using MetadataStore have been migrated. Otherwise, we will break those downstream extensions.

ndr-brt:
Maybe the issue could be divided in some tasks which would be solvable by different PRs, like:
- introduce an in-memory implementation of the `AssetIndex` (similar to the `DemoAssetIndex`)
- add the `findById` method to `AssetIndex`
- substitute the usages of `MetadataStore` with `AssetIndex` (and migrate the extension to use the in-memory `AssetIndex`)

We also need to establish a mapping between `Asset` and `DataEntry` (if that does not exist yet), right?

DominikPinsel:
You are right. Maybe a ``findById`` with an ``Optional<Asset>>`` as return value is the better solution.

Regarding the ``DataEntry``.  I assume this class should be deleted, too. But the current ``Asset`` implementation cannot describe any means of transportation,  like it is done with the ``DataAddress``. So maybe we must add this to the list as well.

PR #159, opened 2021-10-18
Replace MetadataStore with AssetIndex

ndr-brt:
Ref #118 
The PR is still draft because there are some points I need help with:
* `DataEntry` could be completely replaced by `Asset` and deleted from the codebase?
* a mapping between a `List<Policy>` and `AssetSelectorExpression` is still missing (see `QueryEngineImpl`), how can we convert them?


samples/04-file-transfer/api/src/main/java/org/eclipse/dataspaceconnector/extensions/api/ConsumerApiController.java
```diff
@@ -57,10 +58,7 @@ public Response initiateTransfer(@PathParam("filename") String filename, @QueryP
                 .connectorAddress(connectorAddress) //the address of the provider connector
                 .protocol("ids-rest") //must be ids-rest
                 .connectorId("consumer")
-                .dataEntry(DataEntry.Builder.newInstance() //the data entry is the source asset
-                        .id(filename)
-                        .policyId("use-eu")
-                        .build())
+                .asset(Asset.Builder.newInstance().id(filename).property("policyId", "use-eu").build())
```
bscholtes1A, line 61: could be cool to reuse the POLICY_ID constant here (see my previous comment)

samples/04-file-transfer/transfer-file/src/main/java/org/eclipse/dataspaceconnector/extensions/api/FileTransferExtension.java
```diff
@@ -50,15 +54,19 @@ private void savePolicies(ServiceExtensionContext context) {
...
+        DataAddress dataAddress = DataAddress.Builder.newInstance()
                 .property("type", "File")
                 .property("path", "/home/paul/Documents/")
                 .property("filename", "test-document.txt")
                 .build();
 
-        DataEntry entry1 = DataEntry.Builder.newInstance().id("test-document").policyId(USE_EU_POLICY).catalogEntry(file1).build();
-        metadataStore.save(entry1);
+        String assetId = "test-document";
+        Asset asset = Asset.Builder.newInstance().id(assetId).property("policyId", USE_EU_POLICY).build();
```
bscholtes1A, line 67: POLICY_ID here again

data-protocols/ids/ids-api-catalog/src/main/java/org/eclipse/dataspaceconnector/ids/api/catalog/CatalogQueryController.java
```diff
@@ -69,6 +69,6 @@ public Response query(QueryMessage message) {
         var consumerToken = verificationResult.token();
 
         var results = queryEngine.execute(correlationId, consumerToken, connectorId, language, query);
-        return Response.ok(results.stream().map(DataEntry::getId).collect(toList())).build();
+        return Response.ok(results.stream().map(Asset::getId).collect(toList())).build();
```
paullatzelsperger, line 72: can we return `Asset`s here instead of strings?

extensions/in-memory/assetindex-memory/src/main/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndex.java
```diff
@@ -68,7 +68,8 @@ public Asset findById(String assetId) {
         return assets.isEmpty() ? null : assets.get(0);
     }
 
-    void add(Asset asset, DataAddress address) {
+    // TODO: address is not used and it should be deleted
+    public void add(Asset asset, DataAddress address) {
```
paullatzelsperger, line 72: feel free to do so, but we need to touch that again in [this issue](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/172) anyway

samples/04-file-transfer/consumer/build.gradle.kts
```diff
@@ -26,8 +26,7 @@ dependencies {
 
     implementation(project(":extensions:in-memory:transfer-store-memory"))
     implementation(project(":extensions:in-memory:policy-registry-memory"))
-    implementation(project(":extensions:in-memory:metadata-memory"))
-    
+
```
paullatzelsperger, line 29: The dependency will probably need to be replaced with the in-memory `AssetIndex`, right?

spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/transfer/DataRequest.java
```diff
@@ -38,7 +38,7 @@
 
     private String connectorId;
 
-    private DataEntry dataEntry;
+    private Asset asset;
```
paullatzelsperger, line 41: how about using the Asset-ID (i.e. a String)? Not much use in passing the whole asset object here. The only usage I could find was in `DataRequestMessageSender` and that only uses the Asset ID

paullatzelsperger (changes requested):
Looks mostly fine to me, apart from a few minor nits.
However, we need to be really careful to replace all the `in-memory.metadata` references with `in-memory.asset-index` ones, otherwise stuff will break and we'll only notice that when e.g. trying to run the demos.

paullatzelsperger:
I checked the samples 01-05 and fixed them exemplarily, I have a patch which I can send over, unfortunately I cannot attach it to this conversation directly. 
We will need to do that for all the other samples as well.

mspiekermann:
@paullatzelsperger and @ndr-brt So in combination with fix #200 (if it passes the last check) we can merge this PR to complete switching from `MetadataStore` to `AssetIndex`? Since this is related to many other issues it would be great to have this merged. In the meantime I will open a ticket to resolve #54 to allow I-test for all samples.

paullatzelsperger:
Lets do the following: 
 - lets merge in #201 which will make sure all the samples still can boot. 
 - then please rebase this PR onto `main` and force-push (which should trigger the sanity checks on CI) 
 - double-check locally that everything works
 - then me merge it. 
makes sense?

ndr-brt:
> Lets do the following:
> 
> * lets merge in [Feature/200 fix broken samples #201](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/201) which will make sure all the samples still can boot.
> * then please rebase this PR onto `main` and force-push (which should trigger the sanity checks on CI)
> * double-check locally that everything works
> * then we merge it.
>   makes sense?

Absolutely :+1:

MoritzKeppler:
#201  is merged, so please go ahead with rebase, @ndr-brt 

ndr-brt:
Done, the check failed, but "it works on my machine" :)

ndr-brt:
I was able to replicate this `NullPointerException` locally: it happens when the process is killed before the service extensions are initialized. I just raise the sleep time to 10, not the best way but it should work.
What about catching exception on service extension's shutdown?
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/20acbd9c12dbe158286a3180bab8191a22632bb7/core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/runtime/BaseRuntime.java#L145

bscholtes1A:
> I was able to replicate this `NullPointerException` locally: it happens when the process is killed before the service extensions are initialized. I just raised the sleep time to 10, not the best way but it should work.

Hi @ndr-brt 
for me a better approach would have been to have State associated with each Extension (WAITING, INITIALIZED, SHUTDOWN), so that we call the shutdown method only for extensions that are in the INITIALIZED state. But I am pretty sure this would be a huge change that can be handle there.

ndr-brt:
@bscholtes1A thanks for the advice, I think that should be good, but I was wrong, it does not work on my machine and the problem is another, I will fix that and push asap :+1:

ndr-brt:
@paullatzelsperger ready for merge

PR #163, opened 2021-10-19
Implement DidResolverRegistry

jimmarino:
This PR relates to #152, part 2. It implements `DidResolverRegistry` and switches the codebase to use it instead of a `DidResolver` directly.

### DIFF
--- a/extensions/iam/distributed-identity/identity-did-core/src/main/java/org/eclipse/dataspaceconnector/iam/did/resolution/DidResolverRegistryImpl.java
+++ b/extensions/iam/distributed-identity/identity-did-core/src/main/java/org/eclipse/dataspaceconnector/iam/did/resolution/DidResolverRegistryImpl.java
@@ -24,8 +24,9 @@ import java.util.Objects;
  * Default implementation.
  */
 public class DidResolverRegistryImpl implements DidResolverRegistry {
-    public static final int DID_PREFIX = 0;
-    public static final int DID_METHOD_NAME = 1;
+    private static final String DID = "did";
+    private static final int DID_PREFIX = 0;
+    private static final int DID_METHOD_NAME = 1;
 
     private Map<String, DidResolver> resolvers = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


extensions/iam/distributed-identity/identity-did-spi/src/main/java/org/eclipse/dataspaceconnector/iam/did/spi/resolution/DidPublicKeyResolver.java
```diff
@@ -11,9 +10,35 @@
...
+     * Resolves the public key.
      */
-    @Nullable
-    PublicKeyWrapper resolvePublicKey(String did);
+    Result resolvePublicKey(String did);
 
+    /**
+     * The response of a resolve operation.
+     */
+    class Result {
```
ndr-brt, line 20: I think we should standardize the way to return something that could be either a Result or an Error, there's a lot of `*Result` and `*Response`, maybe we could find a common naming convention and, why not, a common abstraction

extensions/iam/distributed-identity/identity-did-core/src/main/java/org/eclipse/dataspaceconnector/iam/did/resolution/DefaultDidPublicKeyResolver.java
```diff
@@ -0,0 +1,56 @@
...
+        if (didDocument.getVerificationMethod() == null || didDocument.getVerificationMethod().isEmpty()) {
+            return new Result("DID does not contain a public key");
+        }
+
+        var verificationMethods = didDocument.getVerificationMethod().stream().filter(vm -> ALLOWED_VERIFICATION_TYPES.contains(vm.getType())).collect(Collectors.toList());
+        if (verificationMethods.size() > 1) {
+            return new Result("DID contains more than one allowed verification type");
+        }
+
+        var verificationMethod = didDocument.getVerificationMethod().get(0);
```
ndr-brt, line 47: should get the first element from the `verificationMethods` variable

paullatzelsperger (approved):

PR #179, opened 2021-10-21
Feature/152-Web-DIDs, Part 3: Implement Web DID Resolver

jimmarino:
This PR implements Part 3 of #152

### DIFF
--- a/extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/WebDidExtension.java
+++ b/extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/WebDidExtension.java
@@ -1,0 +1,46 @@
+/*
+ *  Copyright (c) 2021 Microsoft Corporation
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+package org.eclipse.dataspaceconnector.iam.did.web;
+
+import okhttp3.OkHttpClient;
+import org.eclipse.dataspaceconnector.iam.did.spi.resolution.DidResolverRegistry;
+import org.eclipse.dataspaceconnector.iam.did.web.resolution.WebDidResolver;
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtension;
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtensionContext;
+
+import java.util.Set;
+
+/**
+ * Initializes support for resolving Web DIDs.
+ */
+public class WebDidExtension implements ServiceExtension {
+
+    @Override
+    public Set<String> requires() {
+        return Set.of(DidResolverRegistry.FEATURE);
+    }
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        var httpClient = context.getService(OkHttpClient.class);
+        var mapper = context.getTypeManager().getMapper();
+        var monitor = context.getMonitor();
+        var resolver = new WebDidResolver(httpClient, mapper, monitor);
+
+        var resolverRegistry = context.getService(DidResolverRegistry.class);
+        resolverRegistry.register(resolver);
+
+        monitor.info("Initialized Web DID extension");
+    }
+}
--- a/extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/WebDidExtension.java
+++ b/extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/WebDidExtension.java
@@ -16,10 +16,17 @@ package org.eclipse.dataspaceconnector.iam.did.web;
 import okhttp3.OkHttpClient;
 import org.eclipse.dataspaceconnector.iam.did.spi.resolution.DidResolverRegistry;
 import org.eclipse.dataspaceconnector.iam.did.web.resolution.WebDidResolver;
+import org.eclipse.dataspaceconnector.spi.EdcException;
 import org.eclipse.dataspaceconnector.spi.system.ServiceExtension;
 import org.eclipse.dataspaceconnector.spi.system.ServiceExtensionContext;
+import org.jetbrains.annotations.Nullable;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.Set;
+
+import static java.lang.String.format;
+import static org.eclipse.dataspaceconnector.iam.did.web.ConfigurationKeys.DNS_OVER_HTTPS;
 
 /**
  * Initializes support for resolving Web DIDs.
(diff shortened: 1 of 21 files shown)


jimmarino:
@MoritzKeppler can you take a look? I'd like to get this merged in so we can continue on the Hackathon prep. Thx

extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/resolution/WebDidResolver.java
```diff
@@ -0,0 +1,133 @@
...
+
+    private OkHttpClient httpClient;
+    private ObjectMapper mapper;
+    private Monitor monitor;
+
+    /**
+     * Creates a resolver that executes DNS over HTTPs queries against the given DNS server.
+     */
+    public WebDidResolver(@Nullable URL dnsServer, OkHttpClient httpClient, ObjectMapper mapper, Monitor monitor) {
+        if (dnsServer != null) {
```
ndr-brt, line 55: I think this condition evaluation should be done in the extension class (it's already done in the `getDnsServerUrl` method), and this class should just receive an `OkHttpClient` already configured. This will also avoid the need for a constructor just for testing purpose (the second one, that would become the only one)

extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/resolution/WebDidResolver.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Override
+    public @NotNull String getMethod() {
+        return DID_METHOD;
+    }
+
+    @Override
+    @Nullable
+    public DidDocument resolve(String didKey) {
```
ndr-brt, line 81: I think we should define when a method have to return an object/null or a `Result` instance, because the `DidPublicKeyResolver` returns a `Result` object, and this could also be a case to use that pattern, to improve the error message returned by `DidResolverRegistryImpl`

extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/resolution/WebDidResolver.java
```diff
@@ -0,0 +1,133 @@
...
+            monitor.severe("Invalid DID key: " + didKey, e);
+            return null;
+        }
+    }
+
+    /**
+     * Converts a DID URN to a URL according to the Web DID specification, .cf https://w3c-ccg.github.io/did-method-web/#read-resolve.
+     */
+    String keyToUrl(String didKey) throws URISyntaxException, IllegalArgumentException, MalformedURLException {
+        var uri = new URI(didKey);
```
ndr-brt, line 111: This operation could be encapsulated using a `Did` class, that could be passed to the resolver simplifying the validation process, in fact, the scheme validation is already done by `DidResolverRegistryImpl`

extensions/iam/distributed-identity/identity-did-web/src/test/java/org/eclipse/dataspaceconnector/iam/did/web/resolution/WebDidResolverTest.java
```diff
@@ -0,0 +1,98 @@
...
+        var resolver = createResolver(interceptor);
+        var document = resolver.resolve("did:web:foo.com:edc:EiDfkaPHt8Yojnh15O7egrj5pA9tTefh_SYtbhF1-XyAeA");
+        assertThat(document).isNull();
+    }
+
+    @Test
+    void verifyConvertToUrl() throws Exception {
+        var resolver = createResolver();
+
+        assertThat(resolver.keyToUrl("did:web:w3c-ccg.github.io")).isEqualTo("https://w3c-ccg.github.io/.well-known/did.json");
```
ndr-brt, line 76: I think testing private methods (actually this is a default one, but I think it has this visibility just for testing purposes) is not a good practice, in this case we should use the interceptor way to test that or extract the transformation logic into another component. I think the latter is better

extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/resolution/WebDidResolver.java
```diff
@@ -0,0 +1,133 @@
...
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Resolves a Web DID according to the Web DID specification (https://w3c-ccg.github.io/did-method-web).
+ */
+public class WebDidResolver implements DidResolver {
+    private static final String DID_SCHEME = "did";
+    private static final String DID_METHOD = "web";
+    private static final String DID_METHOD_PREFIX = DID_METHOD + ":";
+    private static final String HTTPS_PREFIX = " https://";
```
ndr-brt, line 43: leading whitespace

extensions/iam/distributed-identity/identity-did-web/src/test/java/org/eclipse/dataspaceconnector/iam/did/web/resolution/WebDidResolverTest.java
```diff
@@ -0,0 +1,98 @@
...
+        assertThat(resolver.keyToUrl("did:web:w3c-ccg.github.io:user:alice")).isEqualTo("https://w3c-ccg.github.io/user/alice/did.json");
+
+        try {
```
ndr-brt, line 79: Try `try/catch` block assertion could be substituted by:
```
assertThatIllegalArgumentException().isThrownBy(() -> resolver.keyToUrl("did:web:w3c-ccg.github.io:user:alice:"));
```
I think it's more readable

jimmarino: Yep, I agree

extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/resolution/WebDidResolver.java
jimmarino: Yikes! I fixed it

extensions/iam/distributed-identity/identity-did-web/src/test/java/org/eclipse/dataspaceconnector/iam/did/web/resolution/WebDidResolverTest.java
jimmarino: I'm not too hung up on that, I think it is OK to test non-public methods but we can discuss that in another thread.

I agree it's much better, in this case, to separate it out, so I went ahead and did that.

extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/resolution/WebDidResolver.java
jimmarino: I don't think it should. DID keys are serializable as strings (they are URNs) per the DID spec and this is the general DidResolver interface, which is used in various code paths. Hence, we can't assume the DID is valid and wrapping it in a class doesn't really do anything.

extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/resolution/WebDidResolver.java
jimmarino: This is another really good idea and simplification, I added it in.

extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/resolution/WebDidResolver.java
jimmarino: Really good idea! That made things a lot simpler.

jimmarino:
@ndr-brt thanks for the feedback. I've implemented most of the suggestions.

ndr-brt (approved):

jimmarino:
@MoritzKeppler I think @ndr-brt did a thorough review of this. Do you want to take a quick look?

extensions/iam/distributed-identity/identity-did-web/src/main/java/org/eclipse/dataspaceconnector/iam/did/web/WebDidExtension.java
```diff
@@ -0,0 +1,66 @@
...
+
+        monitor.info("Initialized Web DID extension");
+    }
+
+    private OkHttpClient getOkHttpClient(ServiceExtensionContext context) {
+        var httpClient = context.getService(OkHttpClient.class);
+        var dnsServer = context.getSetting(DNS_OVER_HTTPS, null);
+
+        if (dnsServer != null) {
+            // use DNS over HTTPS for name lookups
```
MoritzKeppler, line 59: suggest to log if DnsOverHttps is used or not

MoritzKeppler (approved):

MoritzKeppler:
lgtm, can be merged after resolving the conflict. Thanks!

PR #184, opened 2021-10-22
Feature/183 remove reflective access from criterionconverter

### LINKED ISSUE #183: Remove reflection from CriterionToPredicateConverter
Remove the reflection code from `CriterionToPredicateConverter`, move all instance variables into the `properties` field and access through constants, e.g.:
```java
public String getVersion(){
    return properties.get(VERSION_PROP_KEY);
}

public Builder version(String version) {
      this.properties.put(VERSION_PROP_KEY, version);
      return this;
}
```

paullatzelsperger:
This removes the reflective access from the `CriterionToPredicateConverter`, closes #183 .

### DIFF
--- a/samples/demo-contract-framework/src/test/java/org/eclipse/dataspaceconnector/demo/contracts/DemoContractOfferFrameworkExtensionTest.java
+++ b/samples/demo-contract-framework/src/test/java/org/eclipse/dataspaceconnector/demo/contracts/DemoContractOfferFrameworkExtensionTest.java
@@ -23,6 +23,7 @@ import okhttp3.Response;
 import org.eclipse.dataspaceconnector.junit.launcher.EdcExtension;
 import org.eclipse.dataspaceconnector.spi.system.ConfigurationExtension;
 import org.eclipse.dataspaceconnector.spi.types.TypeManager;
+import org.eclipse.dataspaceconnector.spi.types.domain.asset.Asset;
 import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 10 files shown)


spi/src/test/resources/serialized_asset.json
```diff
@@ -0,0 +1,9 @@
+{
+  "properties": {
```
ndr-brt, line 2: Wouldn't be neater to remove this level of nesting? Just serializing the properties at the root level.

samples/demo-contract-framework/src/test/java/org/eclipse/dataspaceconnector/demo/contracts/DemoContractOfferFrameworkExtensionTest.java
```diff
@@ -48,6 +48,8 @@ void setUp(EdcExtension extension) {
     @Test
     void retrieveContractOffers(OkHttpClient httpClient, TypeManager typeManager) {
         ConnectorClient client = new ConnectorClient(httpClient, typeManager);
+        // for the 'id' we must actually use the constant from Asset, otherwise an exception gets raised during
```
ndr-brt, line 51: Is this comment needed? Otherwise we could add another test that shows the failure in the case of the absence of the correct id property

jimmarino (approved):
LGTM

samples/demo-contract-framework/src/test/java/org/eclipse/dataspaceconnector/demo/contracts/DemoContractOfferFrameworkExtensionTest.java
paullatzelsperger: good idea

Issue #160, opened 2021-10-18
ids: CatalogQueryController should return list of Assets instead of list of String


bscholtes1A:

mspiekermann:
@bscholtes1A is this one still relevant or fixed with the merge of #159 ?

bscholtes1A:
> @bscholtes1A is this one still relevant or fixed with the merge of #159 ?

@mspiekermann yeap this one should be closed

PR #188, opened 2021-10-25
data-protocols: make CatalogQueryController return a list of Assets

ndr-brt:
Ref #160


jimmarino:
@ndr-brt do we need this anymore now that we have the Control API in place?

ndr-brt:
Don't know, but we can close this anyway

PR #192, opened 2021-10-26
Implement persistent AssetIndex using CosmosDB

bscholtes1A:
Part of the development of the Federated Catalogue feature. This persistent implementation of the AssetIndex will especially be used within the next Hackathon to describe in a dynamic way the Assets exposed by a given data provider.


extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexExtension.java
```diff
@@ -0,0 +1,114 @@
...
+        Vault service = context.getService(Vault.class);
+
+        // create client
+        var cosmosClient = createClient(service, configuration.getAccountName(), Collections.singletonList(configuration.getPreferredRegion()));
+
+        // get database, throw exception if not exists
+        var database = createDatabase(configuration, cosmosClient);
+        String containerName = configuration.getContainerName();
+        if (database.readAllContainers().stream().noneMatch(sp -> sp.getId().equals(containerName))) {
+            //todo: maybe create database if not exist?
```
DominikPinsel, line 67: I assume in the most cases the EDC connects to an existing CosmosDB instance. Therefore throwing an exception is perfectly fine here. In case the database must be created together with the EDC application, using something like Terraform is probably better suited anyway.
Therefore removing this comment should be fine.

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/AssetIndexCosmosConfig.java
```diff
@@ -0,0 +1,53 @@
...
+    @EdcSetting
+    private static final String COSMOS_ACCOUNTNAME_SETTING = "edc.assetindex.cosmos.account-name";
+    @EdcSetting
+    private static final String COSMOS_DBNAME_SETTING = "edc.assetindex.cosmos.database-name";
+    @EdcSetting
+    private static final String COSMOS_PARTITION_KEY_SETTING = "edc.assetindex.cosmos.partition-key";
+    @EdcSetting
+    private static final String COSMOS_PREFERRED_REGION_SETTING = "edc.assetindex.cosmos.preferred-region";
+    @EdcSetting
+    private static final String COSMOS_CONTAINER_NAME_SETTING = "edc.assetindex.cosmos.container-name";
```
DominikPinsel, line 18: Maybe you can add a little README.md that lists the existing configuration parameters. So that somebody, that wants to use this extension, just needs to look through the README.md.

extensions/azure/assetindex-cosmos/src/test/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexIntegrationTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+        assertThat(asset).isNotNull();
+        assertThat(asset.getProperties()).isEqualTo(asset2.getProperties());
+    }
+
+    @AfterEach
+    void teardown() {
+        CosmosContainerResponse delete = container.delete();
+        assertThat(delete.getStatusCode()).isGreaterThanOrEqualTo(200).isLessThan(300);
+    }
```
DominikPinsel, line 156: I'd avoid deleting infrastructure within the integration test.

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetQueryBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+                .map(CosmosAssetQueryBuilder::toSql)
+                .forEach(joiner::add);
+        return " WHERE " + joiner.toString();
+    }
+
+    private static String toSql(Criterion criterion) {
+        return "( AssetDocument.wrappedInstance.properties." + criterion.getOperandLeft() +
+                " " + criterion.getOperator() + " " +
+                criterion.getOperandRight() + " )";
+    }
```
DominikPinsel, line 36: It's possible do an SQL injection here by manipulating the expression. E.g.
```java
 var expression = AssetSelectorExpression.Builder.newInstance()
                .constraint("name", "=", "hello\") OR (\"1\"=\"1")
                .build();
```

I'd suggest using the classes from the Azure Cosmos Library, something like this
```java
ArrayList<SqlParameter> paramList = new ArrayList<SqlParameter>();
        paramList.add(new SqlParameter("@id", "AndersenFamily"));
        SqlQuerySpec querySpec = new SqlQuerySpec(
                "SELECT * FROM Families f WHERE (f.id = @id)",
                paramList);

        executeQueryWithQuerySpecPrintSingleResult(querySpec);
```
Example Code from
https://github.com/Azure-Samples/azure-cosmos-java-sql-api-samples/blob/main/src/main/java/com/azure/cosmos/examples/queries/sync/QueriesQuickstart.java#L392-L421

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndex.java
```diff
@@ -0,0 +1,93 @@
...
+     *
+     * @param container   The CosmosDB container.
+     * @param typeManager The {@link TypeManager} that's used for serialization and deserialization
+     */
+    public CosmosAssetIndex(CosmosContainer container, String partitionKey, TypeManager typeManager, RetryPolicy<Object> retryPolicy) {
+        this.container = container;
+        this.typeManager = typeManager;
+        this.partitionKey = partitionKey;
+        tracingOptions = new CosmosQueryRequestOptions();
+        tracingOptions.setQueryMetricsEnabled(true);
```
DominikPinsel, line 56: If this is a configuration set from within the EDC it would be nice to enable/disable this via EDC configuration setting.

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexExtension.java
bscholtes1A: Hi Dominik, 
agreed, I removed this comment.

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/AssetIndexCosmosConfig.java
bscholtes1A: Done, you can find the file here: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/192/files#diff-d4164de6f93c8b9c67bd2c9239ba2437da6c7cd725690ca89f470542bc456e75

extensions/azure/assetindex-cosmos/src/test/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexIntegrationTest.java
bscholtes1A: If we don't then it means some temporary container will stay there after the integration test execution. 
I would rather clean up everything once test is properly executed, as it is done there.
@paullatzelsperger  @MoritzKeppler  what is your opinion?

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetQueryBuilder.java
bscholtes1A: To be honest I do not see any added value in using the cosmos library here, as it does not facilitate in any manner the construction of the SQL query.  
However, based on your comment, I searched for some existing libraries that would actually help in construction the SQL query, and I found out JOOQ, which appears to be the main one. 
I reimplemented the query builder by leveraging on this library which, in my opinion, makes the query construction much better.

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndex.java
bscholtes1A: Indeed you are right, I added this as a parameter in the AbstractCosmosConfig class

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetQueryBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+                .getSQL();
+    }
+
+    private List<Condition> createConditions(AssetSelectorExpression expression) {
+        return expression.getCriteria().stream().map(this::createCondition).collect(Collectors.toList());
+    }
+
+    private Condition createCondition(Criterion criterion) {
+        return field(PATH_TO_PROPERTIES + "." + criterion.getOperandLeft()).eq(field((String) criterion.getOperandRight()));
+    }
```
DominikPinsel, line 47: Sorry, it's me again :) Maybe you can validate the criterion operator here. So that it throws an exception if the criterion requires a non-equal operation.

DominikPinsel:
Thanks for the contribution :) I've finished my review. As I'm no maintainer I cannot merge you PR into the main branch. You need one of them as reviewer, too.

DominikPinsel (approved):

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndex.java
```diff
@@ -0,0 +1,96 @@
...
+     * @param container             The CosmosDB container.
+     * @param typeManager           The {@link TypeManager} that's used for serialization and deserialization
+     * @param isQueryMetricsEnabled Activate metrics for query execution
+     */
+    public CosmosAssetIndex(CosmosContainer container, String partitionKey, TypeManager typeManager, RetryPolicy<Object> retryPolicy, boolean isQueryMetricsEnabled) {
+        this.container = container;
+        this.typeManager = typeManager;
+        this.partitionKey = partitionKey;
+        tracingOptions = new CosmosQueryRequestOptions();
+        tracingOptions.setQueryMetricsEnabled(isQueryMetricsEnabled);
```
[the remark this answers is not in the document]
bscholtes1A: We have the class AssetIndexCosmosConfig as POJO to carry the config. But this object is built from the Context, meaning that if we want to pass this object into the constructor of CosmosAssetIndex, we would need to instantiate a Context in the unit test, or mock the AssetIndexCosmosConfig, which would be a bit wobbly. 
@paullatzelsperger  what do you think?

extensions/azure/assetindex-cosmos/README.md
```diff
@@ -0,0 +1,14 @@
...
+This extension provides a persistent implementation of an `AssetIndex` using a CosmosDB container.
+
+The setting parameters of this extensions are listed below:
+
+| Parameter name  | Description  | Mandatory | Default value |
+| :-------------- |:---------------| :---------| :-----------------|
+| `edc.assetindex.cosmos.account-name`  | Account name | true | null |
+| `edc.assetindex.cosmos.database-name` | Name of the DB | true | null |
+| `edc.cosmos.partition-key` | This is the partition key that CosmosDB uses for r/w distribution. Contrary to what CosmosDB suggests, this key should be the same for all local (=clustered) connectors, otherwise queries in stored procedures might produce incomplete results. |  false | dataspaceconnector | 
+| `edc.assetindex.cosmos.preferred-region` | Preferred region for Cosmos client instance | false | West US |
```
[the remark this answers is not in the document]
bscholtes1A: done

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexExtension.java
```diff
@@ -0,0 +1,119 @@
...
+                .key(accountKey)
+                .preferredRegions(preferredRegions)
+                .consistencyLevel(ConsistencyLevel.SESSION)
+                .buildClient();
+    }
+
+    private static CosmosDatabase getDatabase(CosmosClient client, String databaseName) {
+        CosmosDatabaseResponse databaseResponse = client.createDatabaseIfNotExists(databaseName);
+        return client.getDatabase(databaseResponse.getProperties().getId());
+    }
```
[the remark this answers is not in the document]
bscholtes1A: I checked at the documentation and it seems there is no method to check the existence of a DB. Each code snippet I found on google was using exactly the approach of this PR.

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetQueryBuilder.java
jimmarino: @bscholtes1A what's the license JOOQ is distributed under?

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndex.java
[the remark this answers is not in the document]
paullatzelsperger: there is no hard number above which a settings object is mandatory IMO. I agree with @bscholtes1A that instantiating a `ServiceExtensionContext` or an `AssetIndexCosmosConfig` in a unit test is a code smell.

extensions/azure/assetindex-cosmos/src/test/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexIntegrationTest.java
bscholtes1A: @paullatzelsperger I am not sure about your final answer :)
Do I keep it as it is?

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetQueryBuilder.java
bscholtes1A: Hi @jimmarino 
looking at the doc: https://www.jooq.org/doc/3.0/manual/copyright/ it seems it is dual-licensed:
- Apache software license 2.0 for use with Open source database
- commercial for use with commercial databases.
Not an expert but does not seem to be good then... 

Note that it was only Apache License v2 until version <= 3.1 o jOOQ. So maybe a downgrade would be enough?

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexExtension.java
```diff
@@ -0,0 +1,119 @@
...
+
+    private CosmosClient createClient(Vault vault, String accountName, List<String> preferredRegions) {
+        var accountKey = vault.resolveSecret(accountName);
+        if (StringUtils.isNullOrEmpty(accountKey)) {
+            throw new EdcException("No credentials found in vault for Cosmos DB '" + accountName + "'");
+        }
+
+        // create cosmos db api client
+        String host = "https://" + accountName + ".documents.azure.com:443/";
```
paullatzelsperger, line 95: we might want to convert this into a constant and format it here with `String.format`

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetQueryBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+    public String from(AssetSelectorExpression expression) {
+        List<Condition> conditions = expression == AssetSelectorExpression.SELECT_ALL ? Collections.emptyList() : createConditions(expression);
+        return DSL.using(SQLDialect.MYSQL)
+                .select(field("*"))
+                .from(AssetDocument.class.getSimpleName())
+                .where(conditions)
+                .getSQL();
+    }
+
+    private List<Condition> createConditions(AssetSelectorExpression expression) {
```
paullatzelsperger, line 41: This method can be inlined

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetQueryBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+        return expression.getCriteria().stream().map(this::createCondition).collect(Collectors.toList());
+    }
+
+    private Condition createCondition(Criterion criterion) {
+        if (criterion.getOperator().equals("=")) {
+            return field(PATH_TO_PROPERTIES + "." + criterion.getOperandLeft()).eq(field((String) criterion.getOperandRight()));
```
paullatzelsperger, line 47: Although JOOQ is licensed under Apache v2.0, introducing it for just this one line seems kind of excessive to me. There are ways to prevent SQL injection (thanks @dominikPinsel for pointing out): 
- https://cheatsheetseries.owasp.org/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.html
- https://stackoverflow.com/questions/48378681/sqlinjection-against-cosmosdb-in-an-azure-function
- https://docs.microsoft.com/en-us/azure/cosmos-db/sql/sql-query-parameterized-queries

There is even a sample of using SProcs in this code, checkout `CosmosTransferProcessStore.nextForState`

extensions/azure/assetindex-cosmos/src/test/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.eclipse.dataspaceconnector.common.configuration.ConfigurationFunctions.propOrEnv;
+
+@IntegrationTest
+class CosmosAssetIndexIntegrationTest {
+
+    public static final String REGION = "westeurope";
+    private static final String ACCOUNT_NAME = "cosmos-itest";
+    private static final String DATABASE_NAME = "connector-itest";
```
paullatzelsperger, line 46: We should append a random string like a UUID, so that two concurrent integration tests don't conflict with each other. Just to be sure, we should do the same for `CONTAINER_NAME`

extensions/azure/assetindex-cosmos/src/test/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexTest.java
```diff
@@ -0,0 +1,101 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.easymock.EasyMock.anyObject;
+import static org.easymock.EasyMock.eq;
+import static org.easymock.EasyMock.expect;
+import static org.easymock.EasyMock.niceMock;
+import static org.easymock.EasyMock.replay;
+import static org.easymock.EasyMock.strictMock;
+import static org.easymock.EasyMock.verify;
+
+class CosmosAssetIndexTest {
```
paullatzelsperger, line 36: Even though it might seem trivial, lets make sure that invalid params like `null` or `""` are tested as well. Also, there are no tests for `.queryAssets`?

paullatzelsperger (changes requested):

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetQueryBuilder.java
bscholtes1A: I removed the dependency on JOOQ and used the cosmos library instead

extensions/azure/assetindex-cosmos/src/test/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexIntegrationTest.java
bscholtes1A: Done

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetQueryBuilder.java
bscholtes1A: Dependency on JOOQ removed

extensions/azure/assetindex-cosmos/src/test/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexTest.java
bscholtes1A: UT added

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexExtension.java
bscholtes1A: done

extensions/azure/cosmos-common/src/main/java/org/eclipse/dataspaceconnector/cosmos/azure/CosmosDbApi.java
```diff
@@ -0,0 +1,17 @@
...
+
+public interface CosmosDbApi {
+
+    void createItem(Object item);
+
+    @Nullable Object queryItemById(String id);
+
+    List<Object> queryAllItems();
+
+    List<Object> queryItems(SqlQuerySpec query);
```
paullatzelsperger, line 16: Let's try and keep Cosmos-SDK classes out of this interface. Why not simply pass in a String?

extensions/azure/cosmos-common/src/main/java/org/eclipse/dataspaceconnector/cosmos/azure/CosmosDbApiImpl.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private static CosmosClient createClient(Vault vault, String accountName, List<String> preferredRegions) {
+        var accountKey = vault.resolveSecret(accountName);
+        if (StringUtils.isNullOrEmpty(accountKey)) {
+            throw new EdcException("No credentials found in vault for Cosmos DB '" + accountName + "'");
+        }
+
+        // create cosmos db api client
+        String host = String.format("%s://%s.%s", CLIENT_PROTOCOL, accountName, CLIENT_URL_SUFFIX);
```
paullatzelsperger, line 117: sorry if I misspoke, what I actually meant was a constant `private final static string HOST_TEMPLATE="https://%s.documents.azure.com:443/"` and then here you simply write `String host = String.format(HOST_TEMPLATE, accountName)`

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetQueryBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+    private static final String PROPERTIES_FIELD = "properties";
+    private static final String PATH_TO_PROPERTIES = String.join(".", AssetDocument.class.getSimpleName(), ASSET_FIELD, PROPERTIES_FIELD);
+
+    public CosmosAssetQueryBuilder() {
+        // pre-check fields of the AssetDocument class that will be used to build the queries
+        assertClassContainsField(CosmosDocument.class, ASSET_FIELD);
+        assertClassContainsField(Asset.class, PROPERTIES_FIELD);
+    }
+
+    public SqlQuerySpec from(AssetSelectorExpression expression) {
```
paullatzelsperger, line 28: could this just return a String? Then we keep surfacing the Cosmos DB SDK to a minimum

paullatzelsperger (changes requested):
I like the changes you made, makes everything cleaner and better testable. A annotated few things I would still change to make it even slicker.

paullatzelsperger (approved):

Issue #193, opened 2021-10-27
Avoid direct accesses to ObjectMapper


ndr-brt:
As discussed in #191 , the responsability of ser/des objects to/from json should be of the `TypeManager` class, but here and there in the codebase there are some `ObjectMapper` direct instantiations and usages. (see `CrawlerJob` or `IdentityProviderKeyResolver` class)
Those should be avoided.
This convention should also be described into one of the `docs` markdowns.

PR #197, opened 2021-10-28
extensions: prefer TypeManager over ObjectMapper

ndr-brt:
Ref #193 
Update the `architecture-principles.md` file and replaced the `ObjectMapper` instantiations in the production code.


jimmarino (commented):
Looks good!

PR #198, opened 2021-10-29
Asset Index demo

ndr-brt:
A sample on how to interact with the asset index through a REST API.


paullatzelsperger:
Do we really want to give direct access to the asset index? Shouldn't an `Asset` always be exposed through the ContractFramework, and "accompanied" by Policies etc?

ndr-brt:
Good point, maybe I could move the implementation of this example into the `demo-contract-framework` sample, that already contains an implementation of the `ContractFramework`, and avoid direct queries to the AssetIndex

ndr-brt:
In this PR I used a temporary `AssetIndexWriter` component that will be replaced by the `AssetIndexLoader` of the #199

jimmarino:
Can we integrate this with the existing demo/onboarding samples somehow, e.g. as part of the contract offer framework work? I think we should avoid the proliferation of many different, disjointed samples and consolidate around key use cases.

ndr-brt:
@jimmarino that's what I did after the @paullatzelsperger , I implemented that into the existing `demo-contract-framework` sample

jimmarino:
ok thx

MoritzKeppler:
> 
> 
> In this PR I used a temporary `AssetIndexWriter` component that will be replaced by the `AssetIndexLoader` of the #199

should we then merge #199 first?

samples/demo-contract-framework/README.md
```diff
@@ -2,4 +2,20 @@
 
 This framework provides an implementation of the ContractOfferFramework interface.
 
-The implemented PublicContractOfferFramework creates free use contract offers for all assets.
\ No newline at end of file
+The implemented PublicContractOfferFramework creates free use contract offers for all assets.
```
MoritzKeppler, line 5:
```suggestion
A simple REST API is offered to add assets to an in memory catalog.
The implemented PublicContractOfferFramework creates free use contract offers for all assets.
```

samples/demo-contract-framework/src/main/java/org/eclipse/dataspaceconnector/demo/contracts/AssetIndexController.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Stream;
+
+import static java.util.stream.Collectors.toList;
+
+@Consumes({ MediaType.APPLICATION_JSON })
+@Produces({ MediaType.APPLICATION_JSON })
+@Path("/assets")
```
MoritzKeppler, line 37: do we already have an alignment on how to name pathes? the risk of clashes is quite high...
@paullatzelsperger  @jimmarino

samples/demo-contract-framework/src/main/java/org/eclipse/dataspaceconnector/demo/contracts/DemoContractOfferFrameworkExtension.java
```diff
@@ -34,14 +39,26 @@
...
         monitor = context.getMonitor();
 
-        final PublicContractOfferFramework contractOfferFramework = new PublicContractOfferFramework();
-
+        var contractOfferFramework = new PublicContractOfferFramework();
         context.registerService(ContractOfferFramework.class, contractOfferFramework);
 
+        var assetIndex = context.getService(AssetIndex.class);
+        var webService = context.getService(WebService.class);
+        var assetIndexWriter = new InMemoryAssetIndexWriter((InMemoryAssetIndex) assetIndex);
```
MoritzKeppler, line 56: would expext an AssetIndexWriter to be provided as a service - but guess that will be solved by #199

ndr-brt: Yes, I think we should merge #199 before if possible

samples/demo-contract-framework/src/main/java/org/eclipse/dataspaceconnector/demo/contracts/AssetIndexController.java
ndr-brt: I used a simple REST approach, AFAIK there's no alignment on api paths, in the other samples they are pretty diversified. I think we need to find a convention, maybe add a check to avoid clashes at runtime, maybe could we address this to another issue?

paullatzelsperger (approved):

ndr-brt:
Can be merged?

Issue #172, opened 2021-10-20
Create an `AssetLoader` and an in-memory impl


paullatzelsperger:
For those cases where the `AssetIndex` is based on memory, we need a way to load data into it and into the `DataAddressResolver`. 
`Assets` and `DataAddresses` **must always be strongly consistent**, that means there every asset must always be resolvable, otherwise the system is inconsistent.

Thus, there needs to be an `AssetLoader` interface, which has several methods to insert `Asset` and `DataAddress` objects. The in-memory loader implementation would also implement the `AssetIndex` and `DataAddressResolver` interface, so that it can  perform CRUD operations atomically.

PR #199, opened 2021-11-01
Feature/172 create assetloader

paullatzelsperger:
This PR implements an `AssetIndexLoader` plus an in-memory implementation for that. 
Since the loader must guarantee consistency between an `Asset` and the corresponding `DataAddress`, the in-mem implementation also implements the `AssetIndex` and `DataAddressResolver` interfaces. 

Fixes issue #172


extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexLoaderTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.Map;
+import java.util.Random;
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.easymock.EasyMock.niceMock;
+
+// this test aims at testing the thread-safety of the asset index
```
ndr-brt, line 21: Don't seems to be true, actually there's no test that verifies thread-safety

extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexLoaderTest.java
```diff
@@ -0,0 +1,88 @@
...
+public class InMemoryAssetIndexLoaderTest {
+
+    private final Random random = new Random();
```
ndr-brt, line 24: Unused

extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexLoaderTest.java
```diff
@@ -0,0 +1,88 @@
...
+    private AssetIndexLoader assetIndexLoader;
+
+    @Test
+    void insert() {
```
ndr-brt, line 28: This behavior is already covered by the `insert_exists` test, so it could be deleted

extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexLoaderTest.java
```diff
@@ -0,0 +1,88 @@
...
+        var dataAddress = createDataAddress(asset);
+        assetIndexLoader.insert(asset, dataAddress);
+
+        assertThat(((InMemoryAssetIndex) assetIndexLoader).getAssets()).hasSize(1);
+        assertThat(((InMemoryAssetIndex) assetIndexLoader).getDataAddresses()).hasSize(1);
+    }
+
+    @Test
+    void insert_illegalParams() {
+        assertThatThrownBy(() -> assetIndexLoader.insert(null, niceMock(DataAddress.class))).isInstanceOf(NullPointerException.class);
```
ndr-brt, line 39: Maybe should better avoid mocking data classes in favour of using dummy objects

extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexLoaderTest.java
paullatzelsperger: not quite - the `exists` one verifies that the DataAddress has in fact been replaced.

extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexLoaderTest.java
paullatzelsperger: is there any particular reason for this?

ndr-brt: I think that mocks should be used to verify behavior and not state. this test is quite simple and here's not that big issue, but generally mocking data classes could bring to unmaintainable and valueless tests

paullatzelsperger: ok, I don't have a strong opinion about that. I'll change it.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndexLoader.java
```diff
@@ -0,0 +1,14 @@
...
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataAddress;
+
+import java.util.Map;
+
+public interface AssetIndexLoader {
+    String FEATURE = "edc:asset:assetindex:loader";
+
+    void insert(Asset asset, DataAddress address);
+
+    void insertAll(Map<Asset, DataAddress> entries);
```
ndr-brt, line 13: Since `Asset` lacks the `equals` and the `hashCode` overrides, it's quite strange to see a `Map` with `Asset` as key class. I understand that, used in this way, `entries` becomes just a collection of pairs of `Asset` and `DataAddress`, and that's ok, but it's... peculiar. Maybe a `List` of `Map.Entry` could fit better (the implementation should be the same)

extensions/in-memory/assetindex-memory/src/main/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndex.java
```diff
@@ -15,28 +15,35 @@
...
 /**
- * An ephemeral asset index.
+ * An ephemeral asset index, that is also a DataAddressResolver and an AssetIndexLoader
  */
-public class InMemoryAssetIndex implements AssetIndex {
+public class InMemoryAssetIndex implements AssetIndex, DataAddressResolver, AssetIndexLoader {
     private final Map<String, Asset> cache = new ConcurrentHashMap<>();
+    private final Map<String, DataAddress> dataAddresses = new ConcurrentHashMap<>();
     private final CriterionToPredicateConverter predicateFactory;
+    private final ReentrantLock lock;
```
MoritzKeppler, line 41: why do we need this lock in addition to ConcurrentHashMaps? Just to ensure both Maps stay consistent?
If so maybe a ReentrantReadWriteLock would be better... but just a minor-minor improvement for an extension that shouldn't be used in production...

MoritzKeppler (approved):

extensions/in-memory/assetindex-memory/src/main/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndex.java
paullatzelsperger: Are you after the supposed added performance since we're expecting much more `read` operations than `write` operations?

MoritzKeppler: yes. but more than that I wanted to make ensure I understood why we need the lock at all? Is it for consistency over those two maps?

MoritzKeppler: If so my next question would be: why is DataAddress not just an attribute at Asset? From a DDD perspective: does a DataAddress has it's own identity? Don't think so - but maybe I'm wrong...

spi/src/main/java/org/eclipse/dataspaceconnector/spi/asset/AssetIndexLoader.java
paullatzelsperger: I agree, a list feels more natural.

extensions/in-memory/assetindex-memory/src/main/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndex.java
ndr-brt: Actually looking from a DDD perspective the `Asset` is an entity, it has an identity, it has an history and, why not, their attributes could be modified without changing its identity. 
Seems like `DataAddress` in this context is more of a value object, because it is just related to the `Asset`, and in fact it does not own an identity so I agree with your reasoning @MoritzKeppler .
Don't remember why we extracted `DataAddress` from `Asset`...

paullatzelsperger: yes exactly. My intention was to mimic the behavior of a database where inserting values into two tables does not happen at the exact same time, but rather one-after-the-other. A transaction takes care of consistency.

paullatzelsperger: I would not model the `DataAddress` as a property of the `Asset`, because the DataAddress will probably private information - we wouldn't want to expose where our data is stored. Thus, if the DataAddress is a property on the Asset, we would have to make prevent exposing it during serialization, which I find dangerous.

MoritzKeppler: What we offer to outside is e.g. an IDS Resource or something similar. In my eyes it would be an antipattern to provide the internal model 1:1 to the world outside.

MoritzKeppler: But a DataAdress can also be used in the information passed to others: it that case it would be the address under which the other participant can retrieve the data.
As soon as a consumer retrieved an asset it will be added to an internal catalog (as it now needs to be managed) and will get a DataAddress pointing to the local copy of the asset.

paullatzelsperger: >  But a DataAdress can also be used in the information passed to others: it that case it would be the address under which the other participant can retrieve the data.

that would imply giving consumers direct access to, say, a database, right? Because I think the DataAddress would contain database name, etc...

> As soon as a consumer retrieved an ...

sure, but then there will also be a different instance of the Asset within the consumer's realm, even if that may contain the same metadata.

ndr-brt (approved):

extensions/in-memory/assetindex-memory/src/main/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndex.java
MoritzKeppler: started a discussion, maybe we can copy over the different views.
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/discussions/214

PR #201, opened 2021-11-01
Feature/200 fix broken samples

paullatzelsperger:
This PR fixes the broken samples (01-05) and adds steps to the pipeline which compile and run the samples and verify their readiness state. 
This is basically done in three steps: 
1) run the sample, pipe the command line output into a file
2) wait for some time, make sure the log file contains a line `"INFO.*edc-.*ready"`, which is logged whenever a runtime has finished booting.
3) kill the java process

_Note: at this time we can only do this for samples that don't rely on external systems such as cloud storage etc. That's why sample 05 is **not** verified._

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -50,3 +50,7 @@ jobs:
           echo "Checking Sample 02"
           java -jar samples/02-health-endpoint/build/libs/connector-health.jar > /dev/null $
           kill -INT $!
+
+          echo "Checking Sample 03"
+          java -jar samples/03-configuration/build/libs/filsystem-config-connector.jar > /dev/null $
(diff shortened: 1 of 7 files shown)


MoritzKeppler (approved):
Like that we include those checks as part of the pipeline!
Guess we need a more reliable solution for waiting until the service is loaded and check if it's healthy in future but for now it feels sufficient for me.

paullatzelsperger:
> Like that we include those checks as part of the pipeline! Guess we need a more reliable solution for waiting until the service is loaded and check if it's healthy in future but for now it feels sufficient for me.

having a dedicated HTTP health endpoint would be helpful here. however that brings with it a few aspects to take into account.

PR #217, opened 2021-11-05
Connector to be able to expose data contract offers (closes #216)

### LINKED ISSUE #216: Connector to be able to expose data contract offers
The connector needs to be capable of providing IDS contract offers via its multipart API.

denisneuling:
PR for making the connector capable of exposing IDS contract offers via its multipart api.

This branch is based and dependent on [PR208](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/208).


spi/src/main/java/org/eclipse/dataspaceconnector/spi/contract/ContractOfferFrameworkQuery.java
```diff
@@ -14,29 +14,28 @@
...
+import org.eclipse.dataspaceconnector.spi.iam.VerificationResult;
 
 /**
  * The {@link ContractOfferFrameworkQuery} narrows down the number of queried {@link ContractOfferTemplate}.
  */
 public class ContractOfferFrameworkQuery {
 
-    // TODO define who the principal is and rename it (promisee, assigne, recipient, daps token, etc.?)
-    private Principal principal;
+    private VerificationResult verificationResult;
```
MoritzKeppler, line 24: feels strange to have a VerificationResult as a field. I think it should just be used as a return value.

DominikPinsel: Yea. We intent here to enable the ContractOfferFramework to be enable to create custom offers for each request. But I think you are right, maybe it's not a good idea to pass the whole `VerificationResult` object here.

Would it resolve the issue if we extract the claims from the `VerificationResult` and wrap it in another object? Maybe a principal?

jimmarino (approved):
LGTM!

PR #218, opened 2021-11-05
Asset Index Pagination (closes #178)

### LINKED ISSUE #178: Extend AssetIndex for Pagination/Cursoring
Hello, everyone,

I assume the content AssetIndex might be too big to be used in one piece. If we make it accessible via IDS or another protocol it's necessary to limit the amount of processed assets per request.

I would therefore like to suggest:

- Create an **AssetIndexQuery** class with a **expression**, **limit** and a **cursor** property.
- Create an **AssetIndexResult**. That returns an **Cursor**, to continue the query, additionally to the returned **Assets**.

DominikPinsel:

### DIFF
--- a/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractOfferServiceImpl.java
+++ b/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractOfferServiceImpl.java
@@ -15,6 +15,8 @@
 package org.eclipse.dataspaceconnector.contract;
 
 import org.eclipse.dataspaceconnector.spi.asset.AssetIndex;
+import org.eclipse.dataspaceconnector.spi.asset.AssetIndexQuery;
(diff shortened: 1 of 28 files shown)


ndr-brt:
I think you need to rebase against main because `InMemoryAssetIndex` changed a bit.
For the rest, introducing `AssetStorage` seems like a bit of unneeded indirection here.
I'm still dubious about the need for the `Cursor` thing, wouldn't the `offset`/`limit` pattern just enough?

DominikPinsel:
Offset would only work with storage systems where all items are in some kind of "list". E.g. if the items are stored in a tree an offset cannot be used. There may also be a problem with the offset when new items are not happened at the end of the list.

Regarding the `AssetStorage`. I assume it's cleaner if the `Asset` items are not stored in the `AssetIndex` itself. Like this one objects does the storing of the assets, and the other one creates the index and makes them queryable by the EDC. I assume in a productive implementation this both cases would be put in separated classes, too.

PR #223, opened 2021-11-09
Enable Sync and Async TransferProcesses

paullatzelsperger:
This PR enables synchronous and asynchronous TransferProcesses. Whether one or the other is used is determined by the (new) flag `DataRequest.isSyncRequest`.

In order to achieve this, a few key changes were necessary:
- a `DelegatingTransferProcessManager`, which multiplexes out to the Sync-/Async manager 
- synchronous transfer processes are created in the `COMPLETED` state, since they immediately return. For that, a few minor tweaks were made to the `TransferProcess` were made.
- providers can return a data object called a `ProxyEntry` which encapsulates any data, e.g. a REST endpoint, tokens, etc.
- IDS `ArtifactResponseMessages` now contain the aforementioned `ProxyEntry` which is then parsed by the IDS `DataRequestMessageSender`. _Currently this is REST only!_

One scenario, where this feature could be useful, is when we want to make assets available through a private backend (here another REST API). Then the connector would then act as a proxy or an API gateway.

**In order to make this available for IDS Multipart, a subsequent issue will be filed.**

_Note that the actual HTTP Proxy implementation is not part of this PR, it was removed to keep this PR comparably small and it can be found in the [Microsoft EDC Showcase Repository](https://github.com/microsoft/edc-showcase)_


paullatzelsperger:
@bscholtes1A I cannot add you as a reviewer, but would still appreciate your comments :)

extensions/catalog/federated-catalog-cache/build.gradle.kts
```diff
@@ -29,6 +29,10 @@ dependencies {
     // generates random names
     implementation("info.schnatterer.moby-names-generator:moby-names-generator:20.10.0-r0")
 
+    // generates random names
+    // fixme: remove before PR submit
+    implementation("info.schnatterer.moby-names-generator:moby-names-generator:20.10.0-r0")
```
ronjaquensel, line 34: Should this be removed before the merge?

data-protocols/ids/ids-core/src/main/java/org/eclipse/dataspaceconnector/ids/core/message/DataRequestMessageSender.java
```diff
@@ -138,12 +155,32 @@ public DataRequestMessageSender(String connectorId,
...
+        try {
+            var body = httpResponse.body().string();
+            var artifactResponse = mapper.readValue(body, ArtifactResponseMessage.class);
+            return artifactResponse.getProperties().get("dataspaceconnector-data-object");
+        } catch (IOException e) {
+            monitor.severe("Could not read body of response", e);
+            return null;
+        }
+    }
+
```
ronjaquensel, line 177: In case of IDS Rest there is no IDS message in the response body; the information will be in the headers. Nothing that has to be fixed now, but we should keep this in mind. Can you add a TODO/fixme?

ronjaquensel (approved):

extensions/catalog/federated-catalog-cache/build.gradle.kts
paullatzelsperger: I kind of have grown to love those funny names :)
yeah, I'll remove it... :(

data-protocols/ids/ids-core/src/main/java/org/eclipse/dataspaceconnector/ids/core/message/DataRequestMessageSender.java
paullatzelsperger: will do.

paullatzelsperger: When we abandon IDS enirely, we'll likely have to re-implement this feature in IDS multipart anyway. But that'll be done in a subsequent PR.

bscholtes1A (approved):

spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/transfer/DataRequest.java
```diff
@@ -47,6 +47,8 @@
 
     private TransferType transferType;
 
+    private boolean isSyncRequest = false;
```
ndr-brt, line 50: Would remove the `is` prefix on the field name, and also the `Request` suffix is not needed, since the class is already called `DataRequest`, I would call it just `sync`.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/transfer/DataRequest.java
```diff
@@ -147,6 +150,10 @@ public TransferType getTransferType() {
         return transferType;
     }
 
+    public boolean isSyncRequest() {
```
ndr-brt, line 153: Can be called just `isSync`

spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/transfer/DataRequest.java
```diff
@@ -195,6 +202,11 @@ public Builder contractId(String contractId) {
             return this;
         }
 
+        public Builder isSyncRequest(boolean isSync) {
```
ndr-brt, line 205: Can be called just `sync`

extensions/in-memory/transfer-store-memory/src/main/java/org/eclipse/dataspaceconnector/transfer/store/memory/InMemoryTransferProcessStore.java
```diff
@@ -72,7 +73,9 @@ public String processIdForTransferId(String id) {
     @Override
     public void create(TransferProcess process) {
         writeLock(() -> {
-            process.transitionInitial();
+            if (process.getState() == TransferProcessStates.UNSAVED.code()) {
```
ndr-brt, line 76: I think this operation should be done in the `TransferProcessManager`, the store should just store.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/SyncTransferProcessManager.java
```diff
@@ -0,0 +1,96 @@
...
+import java.net.ConnectException;
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+
+import static java.util.UUID.randomUUID;
+import static org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcess.Type.CONSUMER;
+import static org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcess.Type.PROVIDER;
+
+public class SyncTransferProcessManager implements TransferProcessManager {
```
ndr-brt, line 25: I think understood the reason behind the sync/async naming convention, but looking at the implementation, this seems not a sync transfer process manager, more like a ProxyTransferProcessManager.

ndr-brt (commented):
The only concern I have is the one regarding the naming of the two TransferProcessManagers as I specified in the comment, I think it would be better push on the `proxy` concept than on the sync/async one.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/SyncTransferProcessManager.java
paullatzelsperger: yes and no :)
It is one that allows for proxying, but in that it is synchronous. I'm fine with renaming it though.

extensions/catalog/federated-catalog-cache/build.gradle.kts
```diff
@@ -52,3 +52,8 @@ publishing {
         }
     }
 }
+//tasks.withType<Test> {
```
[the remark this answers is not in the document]
paullatzelsperger: removed it. probably a debugging relic :)

extensions/in-memory/transfer-store-memory/src/main/java/org/eclipse/dataspaceconnector/transfer/store/memory/InMemoryTransferProcessStore.java
paullatzelsperger: good catch

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/SyncTransferProcessManager.java
paullatzelsperger: also, now we use it solely for proxying, but in the future we might have additional use cases.

ndr-brt: We will find a better name then :)

paullatzelsperger: since this PR has been open for quite some time, can we either
- agree on a new name 
- leave this one there for now and find a better one later
and merge it?

ndr-brt: Also leaving it like it is now and opening an issue with a discussion about renaming is fine for me.

paullatzelsperger: ok resolving this now.

core/protocol/web/src/main/java/org/eclipse/dataspaceconnector/web/rest/CorsFilter.java
```diff
@@ -0,0 +1,19 @@
...
+import jakarta.ws.rs.container.ContainerResponseFilter;
+
+import java.io.IOException;
+
+public class CorsFilter implements ContainerResponseFilter {
+
+    @Override
+    public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext)
+            throws IOException {
+        responseContext.getHeaders().add("Access-Control-Allow-Origin", "*");
```
jimmarino, line 14: Can you make origins configurable and off by default?

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/synchronous/ProxyEntry.java
```diff
@@ -0,0 +1,56 @@
+package org.eclipse.dataspaceconnector.spi.transfer.synchronous;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+import java.util.Map;
+
+//@JsonTypeName("dataspaceconnector:proxyentry")
+public class ProxyEntry {
```
jimmarino, line 9: Can you please document this class?

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/SyncTransferProcessManager.java
```diff
@@ -0,0 +1,102 @@
...
```
jimmarino, line 25: Please document

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/DelegatingTransferProcessManager.java
```diff
@@ -0,0 +1,34 @@
+package org.eclipse.dataspaceconnector.transfer.core.transfer;
+
+import org.eclipse.dataspaceconnector.spi.transfer.TransferProcessManager;
+import org.eclipse.dataspaceconnector.spi.transfer.TransferResponse;
+import org.eclipse.dataspaceconnector.spi.transfer.store.TransferProcessStore;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataRequest;
+
+public class DelegatingTransferProcessManager implements TransferProcessManager {
```
jimmarino, line 8: Plese document

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/synchronous/ProxyEntryHandlerRegistry.java
```diff
@@ -0,0 +1,9 @@
+package org.eclipse.dataspaceconnector.spi.transfer.synchronous;
+
+import java.util.Map;
+
+public interface ProxyEntryHandlerRegistry extends Map<String, ProxyEntryHandler> {
```
jimmarino, line 5: Please document

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/synchronous/DataProxyManager.java
```diff
@@ -0,0 +1,11 @@
+package org.eclipse.dataspaceconnector.spi.transfer.synchronous;
+
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataRequest;
+
+public interface DataProxyManager {
```
jimmarino, line 5: Please document

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/synchronous/ProxyEntryHandler.java
```diff
@@ -0,0 +1,8 @@
+package org.eclipse.dataspaceconnector.spi.transfer.synchronous;
+
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataRequest;
+
+@FunctionalInterface
+public interface ProxyEntryHandler {
```
jimmarino, line 6: Please document

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/TransferProcessManager.java
```diff
@@ -35,11 +35,11 @@
     /**
```
jimmarino, line 35: Please document on this class that async and sync transfers are supported

jimmarino (commented):
I'm fine with this once the documentation comments are resolved

MoritzKeppler:
mmh, still need to understand more of it: does sync requests mean we do not have a separation of control and data plane anymore? My expectation was that if access to an existing REST API is necessary, we would have something like an API gateway in the data plane. And this gateway gets configured according to the contract signed before as soon as the consumer sends and ArtifactRequestMessage.  All actual 'data-calls' can then be made synchronously on the data plane directly to that API gateway.
Thanks for enlightening me!

paullatzelsperger:
@MoritzKeppler _synchronous_ refers to the fact that the TPM returns immediately and the TP does not trickle through the state machine. 
The transfer can still happen out-of-band, the TPM would simply return to the consumer an object containing the URL (and possibly a token) that points to the proxy. The consumer can then access that API. 

It's important to mention, that the proxy is not necessarily located "inside" the connector, it could be an external system like an Azure API Gateway. The asset here is the API access rather than a file or a stream.

jimmarino:
> @MoritzKeppler _synchronous_ refers to the fact that the TPM returns immediately with information and the TP does not trickle through the state machine. The transfer can still happen out-of-band, the TPM would simply return to the consumer an object containing the URL (and possibly a token) that points to the proxy. The consumer can then access that API.
> 
> It's important to mention, that the proxy is not necessarily located "inside" the connector, it could be an external system like an Azure API Gateway. The asset here is the API access rather than a file or a stream.
> 
> Also, the implementation of the HTTP Transfer Proxy was moved out of this repository, as it was tailored specifically to one particular use case.

FYI, this is used for IONA-X. I'd like to get this PR (as well as #384) in before we begin the larger refactoring tasks this week (extension loading and module reorg).

PR #226, opened 2021-11-11
fixed broken launcher config and added a sanity check

paullatzelsperger:
fixed the build file of `launchers/basic` and added a sanity check for it.


DominikPinsel (approved):
approved :) thanks for fixing

Issue #227, opened 2021-11-11
ContractService and Policy Integration


jimmarino:
# Overview

We need to support policy-based data contracts in the runtime. This issue will outline the work required to achieve this functionality in the EDC.

Policies are used for two purposes:

1. To define access rights for 1..N `Assets`
2. To define usage requirements for 1..N `Assets`

For example, to restrict an asset to "gold partners" and require users to store the asset data in a geographic region, two policy `Permission`s may be used:

- Permission: Partner=GOLD 
- Permission: Region=EU 

A policy associated with an asset is evaluated against the `claims` (e.g. verifiable credentials) and `attributes` of a `ParticipantAgent`. If the claims and attributes
satisfy the policy, a `ParticipantAgent` will be granted access to the asset.

# Participant Agent

A `ParticipantAgent` is a connector system running on behalf of a dataspace `Participant`. A `ParticipantAgent` is not the same as a dataspace `Participant` since the former may have claims and attributes that are distinct from the latter. For example, the Acme organization may have connector systems deployed in different geographical regions. 
While Acme may possess the claim/attribute Partner=GOLD, only the connector running in the EU will have the claim Region=EU. 

A `ParticipantAgent` is constructed by the `ParticipantAgentService` based on the `ClaimsToken` associated with the request.

### Claims and Attributes

When a `ParticipantAgent` is constructed, its claims derive from verifiable credentials; attributes are key/value pairs that are decorated by extension(s) in the provider runtime. For example, Partner=GOLD may be an attribute added to the `ParticipantAgent` by an extension that performs a lookup in a back-end system using the identity associated with the 
`ClaimsToken`.
 
# Contract Offers

A `ContractOffer` defines the policy that is _attached_ to a set of assets. The relationship between a `ContractOffer` and an asset is a multiplicity: a `ContractOffer` may be attached to multiple assets and an asset may have multiple `ContractOffers` attached to it. Policy attachment is the process of associating a policy with 1..N assets; this is performed using an `AssetSelectorExpression`. The latter are serializable representations of an asset grouping.

### ContractOfferFramework and ContractDefinition

A `ContractOfferFramework` is an extensible service that returns the set of `ContractDefinitions`s a `ParticipantAgent` qualifies for. In business terms, a `ContractDefinition` 
is the terms of use associated with one or more assets a `ParticipantAgent` has access to. It may be based on an existing business relationship between the data provider and
[...]

PR #237, opened 2021-11-13
ContractService and Policy integration, part 2: Policy Engine implementation and working ContractDescriptors

jimmarino:
This PR implements Part 2 of #227.

Note that this PR reorganizes classes contained in the contract package into sub-packages as their number is and will continue to increase. Some classes in the IDS extensions are therefore modified due to this package change.

Also, LiteralExpression has been generalized to support complex types in addition to Strings to accommodate the use of data formats such as JSON-LD.


jimmarino:
@DominikPinsel can you take a look? I'd like to get this merged tomorrow if possible so the three of us can work with @paullatzelsperger on part 3 and get the end-to-end setup in place.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/policy/PolicyEngineImpl.java
```diff
@@ -0,0 +1,135 @@
...
+        var context = new PolicyContextImpl(agent);
+
+        for (BiFunction<Policy, PolicyContext, Boolean> validator : preValidators) {
+            if (!validator.apply(policy, context)) {
+                return new PolicyResult(context.hasProblems() ? context.getProblems() : List.of("Pre-validator failed: " + validator.getClass().getName()));
+            }
+        }
+
+        var evalBuilder = PolicyEvaluator.Builder.newInstance();
+        var evaluator = evalBuilder.build();
```
MoritzKeppler, line 56: suggest to move call to ```build()``` down to line 77. It works as it returns a reference but looks confusing.

DominikPinsel (approved):

MoritzKeppler:
we should rename ```ContractOfferFramework``` to ```ContractDefinitionFramework```

MoritzKeppler:
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/docs/Terminology.md needs to get adjusted or dropped. I like the Javadocs in the corresponding classes - so I'd prefer to just link and remove the description in Terminology.md - or autocreate it from the javadocs (in future).

We also have to adjust the diagrams in https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/tree/main/docs/diagrams

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/policy/PolicyEngineImpl.java
```diff
@@ -0,0 +1,135 @@
...
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+package org.eclipse.dataspaceconnector.contract.policy;
```
MoritzKeppler, line 14: ```PoliceEngine``` now lives in ```spi```, ```PolicyEngineImpl``` in ```contract```. What will happen with the ```policy-engine``` module?

samples/demo-contract-framework/src/main/java/org/eclipse/dataspaceconnector/demo/contract/offer/ContractDescriptor.java
```diff
@@ -0,0 +1,94 @@
...
+ * governed by the contract. These requirements are therefore not advertised to the agent. For example, access control policy may require an agent to be in a business partner tier.
+ * Usage policy defines the requirements governing use an agent must follow when accessing the data. This policy is advertised to agents as part of a contract.
+ *
+ * A participant agent may access a contract (and its associated assets) if the union of the access policy and usage policy is satisfied by the agent.
+ */
+public class ContractDescriptor {
+    private String id;
+    private AssetSelectorExpression selectorExpression;
+    private Policy accessControlPolicy;
+    private Policy usagePolicy;
```
MoritzKeppler, line 34: I like that those types are modeled separately.
General observation: as accessControl is not known to the caller the visibility of a ContractOffer may change between calls (e.g. a caller gets downgraded form PARTNER = Gold to Silver). Not sure if that would also affect actual Contracts? Or are access policies just taken into account during contract offer + negotiation phase, but later just the existence of contract counts?

MoritzKeppler:
The demo controllers got dropped in favor of #238 - are we sure that we can provide that fast in enough or should we better keep a simple, insecure and stupid rest interface for the next two or three weeks?

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractServiceExtension.java
```diff
@@ -59,26 +63,49 @@ public void shutdown() {
...
-            contractOfferFramework = new NullContractOfferFramework();
-        }
+        var agentService = new ParticipantAgentServiceImpl();
+        context.registerService(ParticipantAgentService.class, agentService);
 
-        // Contract offer service calculates contract offers using a variety of contract offer frameworks and the given asset index.
-        ContractOfferService contractOfferService = new ContractOfferServiceImpl(contractOfferFramework, assetIndex);
+        // Lazily load the contract offer framework into this core module since the implementation is provided by an extension that will require other core services
+        var cofSupplier = new CofSupplier(context, monitor);
+        var contractOfferService = new ContractOfferServiceImpl(agentService, cofSupplier, assetIndex);
```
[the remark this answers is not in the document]
jimmarino: Let's talk about this. That is probably a better alternative. We can do this in another PR.

core/policy/policy-model/src/main/java/org/eclipse/dataspaceconnector/policy/model/LiteralExpression.java
```diff
@@ -14,23 +14,30 @@
 
 package org.eclipse.dataspaceconnector.policy.model;
 
+import org.jetbrains.annotations.NotNull;
+
 /**
  * A literal value used as an expression.
  */
 public class LiteralExpression extends Expression {
```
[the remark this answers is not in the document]
jimmarino: Let's discuss. It's still a literal but now a complex type.

jimmarino:
> we should rename `ContractOfferFramework` to `ContractDefinitionFramework`

How about ContractDefinitionService and in another PR?

> https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/docs/Terminology.md needs to get adjusted or dropped. I like the Javadocs in the corresponding classes - so I'd prefer to just link and remove the description in Terminology.md - or autocreate it from the javadocs (in future).
> 
> We also have to adjust the diagrams in https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/tree/main/docs/diagrams

Can you create separate issues for this and link them to here?

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/policy/PolicyEngineImpl.java
jimmarino: Oh shoot, I forgot about that name. I think it should be changed to policy-evaluator. We also need to adjust the package name. I think this should be done in a separate PR. Do you like the name policy-evaluator?

samples/demo-contract-framework/src/main/java/org/eclipse/dataspaceconnector/demo/contract/offer/ContractDescriptor.java
jimmarino: Good question and catch. We need to take into account the case where a participant has access to an asset and then the accessControlPolicy is changed, causing them to no longer have access. If there is an outstanding contract and in force, the connector will likely need to honor it. Let's discuss this and then based on that, file an issue as this will affect multiple work streams.

jimmarino:
> The demo controllers got dropped in favor of #238 - are we sure that we can provide that fast in enough or should we better keep a simple, insecure and stupid rest interface for the next two or three weeks?

We'll get something in. I already created a PR for that and assigned @paullatzelsperger  ;-)

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/policy/PolicyEngineImpl.java
jimmarino: That was a cut-and-paste error...fixing it.

samples/demo-contract-framework/src/main/java/org/eclipse/dataspaceconnector/demo/contract/offer/ContractDescriptor.java
DominikPinsel: As contracts needs to be immutable I'd store actual contracts in a separate store with a unique Id. If we don't do it like this the number of assets or policies may change afterwards, which I think would be a problem. Also a stored contract doesn't need an access policy.

If we do it like this maybe rename the `ContractDiscriptor` to `ContractOfferDescriptor`.

MoritzKeppler:
> 
> 
> > we should rename `ContractOfferFramework` to `ContractDefinitionFramework`
> 
> How about ContractDefinitionService and in another PR?
> 
> > https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/docs/Terminology.md needs to get adjusted or dropped. I like the Javadocs in the corresponding classes - so I'd prefer to just link and remove the description in Terminology.md - or autocreate it from the javadocs (in future).
> > We also have to adjust the diagrams in https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/tree/main/docs/diagrams
> 
> Can you create separate issues for this and link them to here?

done: #245

samples/demo-contract-framework/src/main/java/org/eclipse/dataspaceconnector/demo/contract/offer/ContractDescriptor.java
MoritzKeppler: rename usagePolicy to contractPolicy

jimmarino: Created issue for separate PR

PR #249, opened 2021-11-17
ids: Updates IDS info model version to v4.2.7 (closes #247)

### LINKED ISSUE #247: Update IDS Infomodel dependency
The edc currently uses the Infomodel library v4.0.0. The latest release is v4.2.7 (see [here](https://maven.iais.fraunhofer.de/ui/native/eis-ids-public/de/fraunhofer/iais/eis/ids/infomodel/java/)). As this not only coveres some ontology extensions but especially technical/implementation changes and improvements, we should update the dependency. Maybe this will also fix some issues when using builder classes and mapping objects etc.

denisneuling:
Updates the IDS infomodel version to [4.2.7.](https://maven.iais.fraunhofer.de/ui/native/eis-ids-public/de/fraunhofer/iais/eis/ids/infomodel/java/4.2.7). 


denisneuling:
@juliapampus Since you opened the corresponding [issue](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/247) would you like to review this?

juliapampus:
Sure, is it correct that the AWS Integration Test fails?

juliapampus:
What about the version in the `data-protocols/ids/ids-api-multipart-endpoint-v1/gradle.properties` file? This is still 4.0.0. Is this used anywhere?

data-protocols/ids/ids-transform-v1/src/main/java/org/eclipse/dataspaceconnector/ids/transform/IdsProtocol.java
```diff
@@ -15,5 +15,5 @@
 package org.eclipse.dataspaceconnector.ids.transform;
 
 public final class IdsProtocol {
-    public static final String INFORMATION_MODEL_VERSION = "4.0.0";
+    public static final String INFORMATION_MODEL_VERSION = "4.2.7";
```
juliapampus, line 18: Does this need to be a hardcoded string? Can't we just read the value from the project's `gradle.properties` file? I think this also depends on whether we want to support one single Infomodel version or allow the support of multiple versions in different extensions. Have we already agreed on something? Nevertheless, then we should use a properties/config file instead of defining this in the class, don't we?

denisneuling:
The file [data-protocols/ids/ids-api-multipart-endpoint-v1/gradle.properties](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/data-protocols/ids/ids-api-multipart-endpoint-v1/gradle.properties) was a leftover from an older PR and mistakenly wasn't deleted at that time - thanks for pointing out. The file has been removed, since the infomodel version is inherited by the [parent project gradle.properties](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/gradle.properties) file.

Unfortunately the Azure and AWS integration/system tests are somewhat subject to certain flakiness. We already have filed an [issue](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/221) for that; and afaik paul is working on it.

data-protocols/ids/ids-transform-v1/src/main/java/org/eclipse/dataspaceconnector/ids/transform/IdsProtocol.java
denisneuling: Absolutely agree with you, having it hardcoded might not be the best way here.
Anyway, we decided to keep it super simple and did it this way. When any upcoming requirement needs a more sophisticated solution we can introduce it then.
As far as I understood from past meetings a connector will only support one IDS info model version at a time. If multiple versions need to be supported, then another connector has to be deployed beside supporting the other version.

juliapampus: Sounds good!

juliapampus (approved):

PR #259, opened 2021-11-18
Control API for Data Transfer

DominikPinsel:


jimmarino:
Here is what I am thinking as a start for the design

# Control API

The purpose of the control API is to provide a mechanism for a trusted client to initiate data sharing operations in a connector. Data sharing operations include:

1. Initiating a contract offer query to a third-party connector and marshalling the results back
2. Performing a contract negotiation sequence with a third-party connector
3. Initiating a data sharing request with a third-party connector

The Control API will be offered as a RESTful interface over HTTPS.                                                                 

# Security

The Control API must be secured against potential external and internal threats. Initially, this will be done in the following ways:

### API Keys

The control API will be made available under a sub-path `control`. All endpoints under this sub-path will be secured by a JAX-RS `ContainerRequestFilter` that enforces a 
header-based API key check. The filter can be created by implementing a JAX-RS Provider and registering it with the EDC `WebService`:

```
    @Provider
    private static class FooController implements ContainerRequestFilter {

        @Override
        public void filter(ContainerRequestContext requestContext) throws IOException {
            System.out.println("Filtered");
        }
    }
```
The filter implementation will secure the `control` sub-path by comparing it to a set of cached API keys. The cached API keys will be obtained from the `Vault` at runtime startup.
Note that this version will not support API key refresh or expiration from the `Vault` This can be done in a subsequent release.
            
In a subsequent release, this filter needs to be made extensible to support the ability to plug in alternative authorization mechanisms and correlate requests with client `Subjects`.

### Write Access and Read Access

The API will not provide ***ANY*** write access to the connector. This includes create/update/delete operations for Assets, ContractDefinitions, API keys, etc. These capabilities 
will be provided by separate mechanisms such as the *Data Loader*.

Read Access will also be restricted to providing status and other data on the three supported data sharing operation types. The API will not provide ***ANY** access to listing
assets or other data items. 

## Future Considerations

In a subsequent release, we will need the ability to integrate RBAC frameworks for permission-based access to status/metadata reads. In addition, we may consider adding the ability to
segment the Control API on a different HTTPS port so access may be secured at the networking layer. This could be done by adding the ability to specify a port binding when 
registering a JAX-RS endpoint on the `WebService`. 

# Clustered deployment

Clustered deployments will be supported through HTTP-based loadbalancers. In these setups, a zero-trust deployment should be recommended where traffic from the loadbalancers to the 
EDC instances is secured.

# Documentation

It would be nice to include Swagger annotations on all Control API endpoints.

DominikPinsel:
@denisneuling  I added all the stuff from the review (except security). If you are interested feel free to implement this next week :)

jimmarino:
Was @denisneuling going to implement the JAX-RS security filter (provider) in a separate PR. If so, I'll merge this in.

DominikPinsel:
@jimmarino thanks for your comment. I assume Denis can also add this to this PR, as I will need to make here some additions next week anyway. Some things regarding the data transfer are not clear yet, so it should be ok to leave this PR open for now :)

jimmarino:
This can also be checked in after @denisneuling adds the filter. Let's keep the PRs iterative. We can add the contract negotiation integration in a subsequent PR once that is available.

DominikPinsel:
> This can also be checked in after @denisneuling adds the filter. Let's keep the PRs iterative. We can add the contract negotiation integration in a subsequent PR once that is available.

Ok., then I assume it can be merged :) There transfer itself will not work until we touch some other components, but I would not change them in this PR anyway, as this PR is only for the control API. You will see what I mean once I open a new PR for the data transfer and we can discuss this there.

denisneuling:
Just go ahead and merge. The security filter can be introduced in another PR since it's pretty disconnected from what was happening here.

jimmarino:
@denisneuling should we merge this and then introduce the filter in another PR?

denisneuling:
@jimmarino yes, please merge. The security lib plus it's integration will be introduced in subsequent PRs.

PR #263, opened 2021-11-19
Extensible dataloader, closes #238

### LINKED ISSUE #238: Extensible Data Loader
# Overview

There are requirements to load different data types in the runtime. This list is likely to expand in the near future:

1. `Asset` types to the `AssetIndexLoader`
2. `ContractDescriptor` types to the `DemoContractOfferFrameworkExtension`

Users may want to load data from a file, using a REST API or via an external system such as a messaging sink. This landscape is made more complex by the fact that extensible storage systems may manage loaded types. For example, the `AssetIndexLoader` may be backed by database or in-memory storage depending on the EDC deployment. 
                                                                                                                          
The best way to satisfy these requirements is through a generalized framework for loading data into a connector system with the following features:

- Support multiple data types through extensibility 
- Handle high-volume streaming and batch processing 
- Support multiple ingress methods
- Provide out-of-the-box support for a command-line client that loads data from files 
- Provide facilities for data processing and validation
- Be embeddable in another process, e.g. a messaging sink

# Extensibility Points
                
The data loader should provide the following extensibility points:

1. **Data Types** - the type of data processed and loaded through the framework 
2. **Ingress Points** - the protocols and APIs used to load data through the framework
3. **Validation Functions** - facilities for adding end-user validation functions into the processing pipeline
4. **Data Sinks** - the backing stores data is loaded into
             
The data loader can be built using the core EDC extensibility system. 
 
# Batch and Stream Processing

All ingress methods should support batch processing. This will allow the client to specify the data batch size and commit data to backing stores according to the specified batch sizing. For ingress points capable of streaming (e.g. the file system), all data must be loaded and streamed in batches through the processing
pipeline.

# Security Considerations

A REST ingress point can be provided, but it should not be run as part of a connector process due to security risks, given that the latter runtime is often publicly accessible. 

It is recommended that a loader process be run only in a secure environment and not be publicly accessible due to the risk of data leakage.

paullatzelsperger:
### Description
This PR implements a command-line tool that reads JSON files and stores the contents in a backing storage. 
In the current implementation only `Assets` (actually `AssetEntry` objects) can be parsed and stored in an `AssetIndex`.

### Synopsis:
```bash
java -jar <path-to-jar> --assets <path-to-file.json>
```

### A few things to notice
- by default no `AssetIndex` implementation is configured! Please adapt the `build.gradle.kts` file to suit your particular needs! The app _will_ log an error if you don't do this.
- If a database-backed `AssetIndex` is used, you'll likely also need a configuration extension plus a vault extension to handle credentials.
- The commandline tool is intended to run as standalone program
- Currently only `Asset/DataEntry` objects are supported, more commands will follow. This might change the synopsis of the tool.


core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/runtime/BaseRuntime.java
```diff
@@ -48,7 +48,7 @@ protected Monitor getMonitor() {
     /**
      * Main entry point to runtime initialization. Calls all methods.
      */
-    protected void boot() {
+    protected ServiceExtensionContext boot() {
```
jimmarino, line 51: Just curious, is this a convenience or was it required to return the context?

jimmarino (approved):
LGTM, I'd like to get this merged ASAP, so please let me know if you are OK if I do so.

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/runtime/BaseRuntime.java
paullatzelsperger: pure convenience. I could also intercept one of the initialization hooks to get the context, which I did in my latest commit.

ndr-brt:
> ### A few things to notice
> * by default no `AssetIndex` implementation is configured! Please adapt the `build.gradle.kts` file to suit your particular needs! The app _will_ log an error if you don't do this.
> * If a database-backed `AssetIndex` is used, you'll likely also need a configuration extension plus a vault extension to handle credentials.
> * The commandline tool is intended to run as standalone program
> * Currently only `Asset/DataEntry` objects are supported, more commands will follow. This might change the synopsis of the tool.
These things could be reported in a README file

PR #265, opened 2021-11-21
Feature/242 rename and move in memory cache query adapter registry

paullatzelsperger:
Moved the Catalog- and QueryAdapterRegistry from the `in-memory` module back to the main FCC module and renamed them to appropriately reflect their purpose.


bscholtes1A (approved):

extensions/catalog/federated-catalog-cache/src/test/java/org/eclipse/dataspaceconnector/catalog/cache/TestProtocolAdapterRegistry.java
```diff
@@ -11,30 +11,32 @@
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
-public class TestProtocolAdapterRegistry implements CatalogQueryAdapterRegistry {
-    private final Map<String, List<CatalogQueryAdapter>> map;
+public class TestProtocolAdapterRegistry implements NodeQueryAdapterRegistry {
```
ndr-brt, line 14: It's the same as `NodeQueryAdapterRegistryImpl` except for line 19, that should executed externally the constructor by the test. So my proposal is to remove this class.

ndr-brt (approved):

extensions/catalog/federated-catalog-cache/src/test/java/org/eclipse/dataspaceconnector/catalog/cache/TestProtocolAdapterRegistry.java
paullatzelsperger: removed it.

PR #271, opened 2021-11-22
Feature/cosmos cli loader

paullatzelsperger:
This PR adds some fixes to the CosmosDB backend and enables it for the DataLoader-CLI. Now we can use the DataLoader-CLI tool to insert `AssetEntry` objects into the `AssetIndex`.


jimmarino (approved):
LGTM

PR #272, opened 2021-11-23
Bugfix/221 fix flaky itests

paullatzelsperger:
This should fix the brittle integration tests for Azure and S3. I added defensive measures like timeouts and wait times to the tests as well as additional catch blocks to the code.

I also renamed the Integration tests to `*IntegrationTest`.

We'll know for sure only _after_ the merge though.


ndr-brt:
I think this should be merged. 
It's still failing sometimes because the `pull_request_target` workflow configuration, the integration tests are launched from the checkout of the target of the pull request (main) and not from the code of the PR.
This is safer because no unsafe workflow code can be executed from a PR, but it makes the execution of the IT useless on the PR because the code tested will be always the one on the main branch.

Take a look at the commit hash logged in the checkout phase for confirmation:
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/runs/4311160262?check_suite_focus=true

paullatzelsperger:
I agree, it makes modifying the itests themselves a bit of a trial-and-error endeavour. 
Let's merge and adapt if necessary

jimmarino:
@paullatzelsperger The AWS test continually fails. Cal you take a look?

paullatzelsperger:
as @ndr-brt pointed out, due to the fact how i-tests are set up (`pull_request_target`) the fixed tests will only run after the merge.

jimmarino (approved):

jimmarino:
merging as the the merge will fix the broken tests

Issue #239, opened 2021-11-15
Policy negotiation flow


juliapampus:
As discussed last week, @ronjaquensel and I want to share our thoughts about a possible contract negotiation flow containing states - similar to our current transfer process. 

`.puml` files: [edc-negotiation.zip](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/files/7538410/edc-negotiation.zip)

We tried to keep the number of states as small as possible. States on both sides express the same behaviour and, therefore, trigger the same action.
* `REQUESTED` (prov, cons): A contract request or counter offer has been initiated. The connector sends an appropriate message to the recipient. The state is changed to `REQUEST_ACK` as soon as the request or a response to a request has been received.
* `REQUEST_ACK` (prov, cons): The connector has sent a contract request/offer. The connector (on e.g. consumer side) that initiated the request/offer is waiting for a response. The connector that approved the receipt of a request/offer is switching to the next state.
* `VALIDATING` (prov, cons): Depending on the input (contract request/offer/agreement), the connector validates the contract by comparing it with the contract of the last state. E.g. request - offer, offer - offer, agreement - offer, agreement - request, agreement - agreement. Optionally, the user or another system is being involved in the negotiation process.
* `APPROVED` (prov, cons): As soon, as the contract's content is approved by the connector, a user, or an external system, the connector builds an appropriate contract agreement. This is stored in a dedicated location. As soon as the initial sender of an agreement has not received a matching contract agreement, the stored one is marked as `not confirmed`. Next, a message containing the agreement is sent. The status changes to `PENDING`.
* `DECLINED` (prov, cons): If a contract is malformed or invalid, or a user wants to abort the negotiation process, the state changes to `DECLINED`. Thereupon, a contract rejection is sent (and the process gets removed). If necessary, the rejection is logged on both sides and propagated involved users or systems.
* `PENDING` (prov, cons): This state implies that an agreement has been sent and the connector is waiting for confirmation. This can happen both on provider and consumer side, depending on the one that first agreed.
* `CONFIRMED` (prov, cons): If the validation of an agreement is successful, the state changes to `CONFIRMED`. As a result, the stored agreement is marked as confirmed and persisted permanently. Thus, it can be used for later access and usage control.
* `ENDED` (prov, cons): This is the final state that marks the negotiation process as ended. No further action is required.

[...]

Issue #194, opened 2021-10-27
Remove extension phases and standardize usage of provides/requires


jimmarino:
This is related to #121. I will work on a PR ASAP.

ndr-brt:
This was done by #475 right? Can be closed?

paullatzelsperger:
yes, can be closed.

PR #274, opened 2021-11-23
Feature/part1a contract negotiation: prepare to integrate data loader

### LINKED ISSUE #254: Rename usagePolicy to contractPolicy on ContractDescriptor

jimmarino:
Related to PR #239

Closes #254

This PR:

1.  Simplifies the ContractDefinitionService by combining ContractDescriptors and ContractDefinitions
2. Moves the demo ContractDefinitionService in the contract core (we can make it replaceable later, but it is generic).
3. Introduces a  ContractDefinitionStore with a default in-memory implementation (a Cosmos implementation will be added by @paullatzelsperger). This store will be integrated with the Data Loader.
4. Makes the policy model serializable so it can be sent through the Data Loader. A few cases (as noted in the code) will be left for later, such as serialization of parent references for Duty types. These cases are not typically encountered.

Once @paullatzelsperger implements the Data Loader integration, we will be able to load general policies into the runtime.


core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractServiceExtension.java
```diff
@@ -74,38 +81,18 @@ private void registerServices(ServiceExtensionContext context) {
         var agentService = new ParticipantAgentServiceImpl();
         context.registerService(ParticipantAgentService.class, agentService);
 
-        // Lazily load the contract offer framework into this core module since the implementation is provided by an extension that will require other core services
-        var cofSupplier = new CofSupplier(context, monitor);
-        var contractOfferService = new ContractOfferServiceImpl(agentService, cofSupplier, assetIndex);
+        var policyEngine = new PolicyEngineImpl();
```
paullatzelsperger, line 84: There should be on entry in the Set returned by `provides()` per registered service. Also, those Strings should be constants on the interface (`FEATURE`).

spi/src/main/java/org/eclipse/dataspaceconnector/spi/contract/offer/ContractDefinitionResult.java
```diff
@@ -13,6 +13,8 @@
  */
 package org.eclipse.dataspaceconnector.spi.contract.offer;
 
+import org.eclipse.dataspaceconnector.spi.types.domain.contract.offer.ContractDefinition;
+
 /**
  * The result of searching for a specific contract definition.
  */
```
paullatzelsperger, line 20: The entire class seems to be just an elaborate wrapper around a `null` check, so I think it can be removed.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractServiceExtension.java
```diff
@@ -74,38 +81,18 @@ private void registerServices(ServiceExtensionContext context) {
         var agentService = new ParticipantAgentServiceImpl();
```
paullatzelsperger, line 81: does it really make sense to have a null asset index? In case the `AssetIndex` is null the extension should throw an exception IMO.

paullatzelsperger (changes requested):
I think this is a good starting point. Added my comments directly in the code.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractServiceExtension.java
jimmarino: Provides and Requires are per feature, not necessarily per service. It will become unmanageable if each extension has to declare every service it relies on. The intent of provides/requires was to declare this at the feature level where 1..N services provide a feature. Let's discuss that as part #194

spi/src/main/java/org/eclipse/dataspaceconnector/spi/contract/offer/ContractDefinitionResult.java
jimmarino: That's a good point, I'll remove it

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractServiceExtension.java
jimmarino: I agree but not for this PR. That was stuff that was already there.

paullatzelsperger (approved):
I'm ok with merging this - I can take care of minor stuff in my related work then.

PR #277, opened 2021-11-24
initialize MonitorProvider correctly

ramunter:
When pulling in dependencies that rely on certain logging functionality [you can hit this error ](http://www.slf4j.org/codes.html#null_MDCA). It looks like this is caused by the returned nulls in the MonitorProvider, which [in this code ](http://www.slf4j.org/api/src-html/org/slf4j/MarkerFactory.html#line.44) result in the MarkerFactory/MDCAdapter to be null. 

Updated the provider to match the [NOPProvider that can be found here](http://www.slf4j.org/api/src-html/org/slf4j/helpers/NOPServiceProvider.html#line.8) for the MarkerFactory and MDCAdapter.


jimmarino:
@ramunter thanks for the contribution! Can you please sign the CLA so we can merge?

mspiekermann:
You can get information and start the process here: https://www.eclipse.org/legal/ECA.php

ramunter:
Have a mismatch between commit email and eclipse account email, reopening here https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/281

PR #278, opened 2021-11-24
extensions: add ids policy extension

ndr-brt:
Add an extension that defines two functions for ids policy validation:
* `AbsSpatialPositionConstraintFunction` that checks the region
* `PartnerLevelConstraintFunction` that checks the partner level


jimmarino:
@ndr-brt LGTM, can I merge this when the tests are successful?

ndr-brt:
I think we are ok now

PR #276, opened 2021-11-23
Draft PR for IDS Data Transfer 

DominikPinsel:
Signed-off-by: Dominik Pinsel <dominik.pinsel@daimler.com>

PR #285, opened 2021-11-24
control-api: Add controller to fetch resource catalog

denisneuling:
Adding the control-api to retrieve catalogs containing contract offers from a data provider.


denisneuling:
@DominikPinsel Do you want to merge that with your work done in PR #276 - or shall we keep it separated?

jimmarino:
Hi @denisneuling There are types added to SPI that are used by the control-api module. Can you please remove them from SPI and either put them in a separate "model" module that the control-api depends on?

DominikPinsel:
FYI
I resolved the merge conflicts with main and added some minor changes/fixes so that the fetching of the resource catalog works with sample5.

denisneuling:
@jimmarino The dispatcher and the control api modules need some common container object to make known and use of the contract information. I think there is just the way - as is - of having that `ResourceCatalog` container object to be contained in the main `spi` module. Otherwise by introducing another intermediary module exclusively used by dispatcher and control api, we'd couple them to tight together - which would lead to the question why we have separated those both into distinct modules initially. - And I suppose that pretty much clear due to them handling different concerns.

denisneuling:
@jimmarino As discussed `DataCatalog` has been renamed to `Catalog` and `CatalogRequest` has been moved into the same package. The large amount of changed files is due to the import reordering and variable renaming.

jimmarino (approved):
LGTM, thanks Denis

PR #289, opened 2021-11-25
ci: execute ITs also on every push on main branch

### LINKED ISSUE #288: Integration tests are not executed after a merge
I noticed that integration tests arent't running after a merge, so, currently, they are executed only on pull request push but against the main branch.
Is this the expected behaviour? I would make them running also after a PR merge.

ndr-brt:
Fix #288


jimmarino:
@ndr-brt can this be merged?

ndr-brt:
Yes

PR #292, opened 2021-11-26
extensions: fix integration tests

ndr-brt:
This should fix the problem with the AWS Integration Tests (was a problem with the mock definition).

I have a question... why there are two different aws sdks?
`software.amazon.awssdk` and `com.amazonaws:aws-java-sdk`?
Seems like the latter is the version 1.x and that has been substituted by the `software.amazon.awssdw` as far as I can read here: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration.html


paullatzelsperger:
the "old" SDK is used in the test environment. feel free to update to the `software.amazon.awssdk` in a subsequent PR :)

paullatzelsperger (approved):
step by step we get closer to stable i-tests :)

paullatzelsperger:
can we merge this?

ndr-brt:
Yes

ndr-brt:
Is the same as #311

PR #300, opened 2021-11-26
spi: extend DataRequest with custom properties field

### LINKED ISSUE #299: Need for extensible properties that can be forwarded from a consumer to a provider
I our scenario the consumer connector passes query parameters to the provider which makes a call to a remote API.
Until now we were passing the query parameters in DataDestination field in DataRequest object. But when provisioning a blob container as managed resource (using Azure provisioning extension) the DataDestination properties get overwritten.

We’d need a way to pass the custom properties from the consumer to the provider. For this we suggest using a dedicated data container to decouple it from other functionalities.

_Implementation_
A PR with a proposed solution that we used as a workaround will be published soon.

Izzzu:
This PR contains a proposal to solve issue #299  by extending the DataRequest object with a Map of custom properties.

- add properties Map field to DataRequest
- add the properties to the artifactMessage in DataRequestMessageSender so that they are passed to the provider
- read the properties in ArtifactRequestController
- add unit tests for verifying that the properties were passed from DataRequest to ArtifactRequestMessage in consumer and from ArtifactRequestMessage to DataRequest in provider

Closes #299

### DIFF
--- a/data-protocols/ids/ids-api-transfer/build.gradle.kts
+++ b/data-protocols/ids/ids-api-transfer/build.gradle.kts
@@ -30,8 +30,6 @@ dependencies {
     implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
 
     testImplementation("org.glassfish.jersey.core:jersey-common:${jerseyVersion}")
-    testImplementation("com.github.javafaker:javafaker:1.0.2")
-
 }
 
 publishing {
--- a/data-protocols/ids/ids-api-transfer/build.gradle.kts
+++ b/data-protocols/ids/ids-api-transfer/build.gradle.kts
@@ -14,6 +14,7 @@
 
 val infoModelVersion: String by project
 val rsApi: String by project
+val jerseyVersion: String by project
 
 plugins {
     `java-library`
@@ -27,6 +28,9 @@ dependencies {
     api("de.fraunhofer.iais.eis.ids.infomodel:java:${infoModelVersion}")
 
     implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
+
+    testImplementation("org.glassfish.jersey.core:jersey-common:${jerseyVersion}")
+    testImplementation("com.github.javafaker:javafaker:1.0.2")
 
 }
 
--- a/data-protocols/ids/ids-core/src/test/java/org/eclipse/dataspaceconnector/ids/core/message/DataRequestMessageSenderTest.java
+++ b/data-protocols/ids/ids-core/src/test/java/org/eclipse/dataspaceconnector/ids/core/message/DataRequestMessageSenderTest.java
@@ -1,0 +1,123 @@
+/*
+ *  Copyright (c) 2021 Copyright Holder (Catena-X Consortium)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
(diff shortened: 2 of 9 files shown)


spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/transfer/DataRequest.java
```diff
@@ -45,6 +48,8 @@
 
     private boolean managedResources = true;
 
+    private Map<String, String> additionalProperties = new HashMap<>();
```
algattik, line 51:
```suggestion
    private Map<String, String> additionalProperties = Map.of();
```

immutable

data-protocols/ids/ids-api-transfer/build.gradle.kts
```diff
@@ -28,6 +29,9 @@ dependencies {
 
     implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
 
+    testImplementation("org.glassfish.jersey.media:jersey-media-multipart:${jerseyVersion}")
```
algattik, line 32:
```suggestion
    testImplementation("org.glassfish.jersey.media:jersey-media-multipart:${jerseyVersion}")
```
```suggestion
        testImplementation("org.glassfish.jersey.core:jersey-common:${jerseyVersion}")
```

data-protocols/ids/ids-api-transfer/build.gradle.kts
```diff
@@ -28,6 +29,9 @@ dependencies {
 
     implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
 
+    testImplementation("org.glassfish.jersey.core:jersey-common:${jerseyVersion}")
+    testImplementation("com.github.javafaker:javafaker:1.0.2")
```
ndr-brt, line 33: Since it's a test dependency that could be used by any module it could be moved to the root `build.gradle.kts`

ndr-brt (approved):

data-protocols/ids/ids-core/src/test/java/org/eclipse/dataspaceconnector/ids/core/message/DataRequestMessageSenderTest.java
```diff
@@ -0,0 +1,123 @@
...
+        replay(identityService, tokenResult);
+        sender = new DataRequestMessageSender(connectorId, identityService, niceMock(TransferProcessStore.class), EasyMock.mock(Vault.class), httpClient, mapper, niceMock(Monitor.class));
+    }
+
+    @Test
+    public void initiateIdsMessage() throws IOException {
+        DataRequest dataRequest = createDataRequest();
+        DataAddress dataDestination = dataRequest.getDataDestination();
+
+        // record
```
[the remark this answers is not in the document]
Izzzu: I put these comments because it seemed to me that it’s a common approach for tests in ids module. And that they can help to separate the sections in the test, which improves the readability.
E.g. CatalogServiceImplTest, DescriptionHandlerTest, AssetToIdsRepresentationTransformerTest and other tests especially in ids-transform-v1 

Is there an agreement across the EDC repo, in which tests is it advisable to have the comments or it’s a matter of personal preference?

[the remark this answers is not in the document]
ndr-brt: I think they came from a single test and then some copy-and-paste work. I think they are pretty obvious so they can be deleted.

data-protocols/ids/ids-api-transfer/build.gradle.kts
Izzzu: Thanks for the suggestion. I moved the java faker dependency to the root

PR #313, opened 2021-11-29
contract-validation: add validation of agreement start/end time (closes #295)

### LINKED ISSUE #295: `ContractValidationService` Add functionality
This issue is about adding the following functionality to the `ContractValidationService`:

- The contract validation service should be able to tell when a contract is expired
- The contract validation service should also check whether the corresponding asset is part of the contract
- The contract validation service should return a result object with found errors, so that the caller is able to log them with some context information.

Additionally the `ContractValidationService` checks currently in two places whether a `contract` id is valid. Therefore it would be nice if
- The `contract id` would be a value objects, that cannot be formatted incorrectly. So that we don't need this checks in several places. 

Update:
> The contract validation service should also check whether the corresponding asset is part of the contract

This requirement will become invalid as future `ContractAgreement` will only contain one `Asset` -> Ignore it.

DominikPinsel:
Signed-off-by: Denis Neuling <denis.neuling@daimler.com>
Co-authored-by: Denis Neuing <denis.neuling@daimler.com>

### DIFF
--- a/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/validation/ContractValidationServiceImpl.java
+++ b/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/validation/ContractValidationServiceImpl.java
@@ -26,6 +26,7 @@ import org.eclipse.dataspaceconnector.spi.types.domain.contract.offer.ContractDe
 import org.eclipse.dataspaceconnector.spi.types.domain.contract.offer.ContractOffer;
 import org.jetbrains.annotations.NotNull;
 
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Objects;
 import java.util.function.Supplier;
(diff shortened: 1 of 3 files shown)


DominikPinsel:
@jimmarino introduces all the changes we talked about. Please take another look :)

jimmarino:
@DominikPinsel this PR is doing too many things. As discussed previously, the validation should only return a boolean and not give any indication to the client why the validation failed. Otherwise, this may create a security vulnerability. Details of why the validation failed can be passed to the monitor and correlated by an operations team. Please see how `IdentityService` works for an analogous example.

Also, the PR is changing code formatting in unrelated classes (which already passed Checkstyle) and reactors method signatures that don't need to be modified. That makes it difficult to understand what the intent of the PR is.

I think this PR should be a very small addition to check timestamps.

DominikPinsel:
> @DominikPinsel this PR is doing too many things. As discussed previously, the validation should only return a boolean and not give any indication to the client why the validation failed. Otherwise, this may create a security vulnerability. Details of why the validation failed can be passed to the monitor and correlated by an operations team. Please see how `IdentityService` works for an analogous example.
> 
> Also, the PR is changing code formatting in unrelated classes (which already passed Checkstyle) and reactors method signatures that don't need to be modified. That makes it difficult to understand what the intent of the PR is.
> 
> I think this PR should be a very small addition to check timestamps.

finished

jimmarino (approved):
@DominikPinsel thanks!

PR #314, opened 2021-11-29
Feature/rename maven artifacts

### LINKED ISSUE #20: [Eclipse] Include LICENSE and NOTICE file in every jar
Suggest to include those file in every jar, even if we just distribute those as part of the fat jar inside the docker container.
see https://www.eclipse.org/projects/handbook/#legaldoc-distribution
 
something like that inside build.gradle allprojects section should solve it:

```
tasks.jar {
            metaInf {
                from("${rootProject.projectDir.path}/LICENSE")
            }
        }
```

<sub> Dominik Pinsel <dominik.pinsel@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

paullatzelsperger:
Renames all Maven Publications and adds the NOTICE and LICENSE file to the resulting artifacts' META-INF directory.

Left in draft so that people can take a look and give their opinion

contributes to #267 
closes #20

### DIFF
--- a/core/bootstrap/build.gradle.kts
+++ b/core/bootstrap/build.gradle.kts
@@ -26,8 +26,8 @@ dependencies {
 
 publishing {
     publications {
-        create<MavenPublication>("core.bootstrap") {
-            artifactId = "core.bootstrap"
+        create<MavenPublication>("core-bootstrap") {
+            artifactId = "core-bootstrap"
             from(components["java"])
         }
     }
--- a/core/bootstrap/build.gradle.kts
+++ b/core/bootstrap/build.gradle.kts
@@ -26,8 +26,8 @@ dependencies {
 
 publishing {
     publications {
-        create<MavenPublication>("core-bootstrap") {
-            artifactId = "core-bootstrap"
+        create<MavenPublication>("bootstrap") {
+            artifactId = "bootstrap"
             from(components["java"])
         }
     }
--- a/extensions/azure/blob/build.gradle.kts
+++ b/extensions/azure/blob/build.gradle.kts
@@ -26,8 +26,8 @@ dependencies {
 
 publishing {
     publications {
-        create<MavenPublication>("blob") {
-            artifactId = "blob"
+        create<MavenPublication>("blobstorage") {
+            artifactId = "blobstorage"
             from(components["java"])
         }
     }
--- a/extensions/azure/blob/build.gradle.kts
+++ b/extensions/azure/blob/build.gradle.kts
@@ -26,8 +26,8 @@ dependencies {
 
 publishing {
     publications {
-        create<MavenPublication>("azure.blob") {
-            artifactId = "azure.blob"
+        create<MavenPublication>("blob") {
+            artifactId = "blob"
             from(components["java"])
         }
     }
--- a/extensions/transfer-functions/transfer-functions-core/build.gradle.kts
+++ b/extensions/transfer-functions/transfer-functions-core/build.gradle.kts
@@ -30,7 +30,7 @@ dependencies {
 publishing {
     publications {
         create<MavenPublication>("transfer-functions-core") {
-            artifactId = "transfer-functions-ore"
+            artifactId = "transfer-functions-core"
             from(components["java"])
         }
     }
--- a/extensions/transfer-functions/transfer-functions-core/build.gradle.kts
+++ b/extensions/transfer-functions/transfer-functions-core/build.gradle.kts
@@ -29,8 +29,8 @@ dependencies {
 
 publishing {
     publications {
-        create<MavenPublication>("transfer-functions.core") {
-            artifactId = "transfer-functions.core"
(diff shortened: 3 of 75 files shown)


launchers/junit/build.gradle.kts
```diff
@@ -34,8 +34,8 @@ dependencies {
 
 publishing {
```
ndr-brt, line 35: Seems like this is the only launcher with needs to publishing. I'm not sure that we should consider this as a launcher because I think it fit better as a test utils package. The name should be `junit` or `junit-extension`.
I know that we already had this discussion.

core/bootstrap/build.gradle.kts
```diff
@@ -26,8 +26,8 @@ dependencies {
 
 publishing {
     publications {
-        create<MavenPublication>("core.bootstrap") {
-            artifactId = "core.bootstrap"
+        create<MavenPublication>("core-bootstrap") {
```
ndr-brt, line 29: Why `bootstrap` and `contract` mantained the `core-`prefix and the other core modules did not?

extensions/azure/blob/build.gradle.kts
```diff
@@ -26,8 +26,8 @@ dependencies {
 
 publishing {
     publications {
-        create<MavenPublication>("azure.blob") {
-            artifactId = "azure.blob"
+        create<MavenPublication>("blob") {
```
ndr-brt, line 29: I know that `blob` is a specific azure service, but it's also a generic term widely used to describe a chunk of binary data. Maybe we should add the specification for this.

launchers/junit/build.gradle.kts
jimmarino: @ndr-brt This should be separate from test utils which are internal classes the EDC uses for testing, whereas this is a JUnit extension accessed by users. I'm happy to call it junit-extension, but it should still be a launcher since it bootstraps a runtime as part of a JUnit test fixture.

ndr-brt: Ok, but what about `BaseRuntime`? It's a runtime but it's contained in the `core:bootstrap` module.
For the release process we should find a clear separation between what should be published and what should be not.
We could separate the code (`EdcExtension` and eventually other test utils classes) from the launcher itself (the gradle file) as we did for the `BaseRuntime` that is separated from `basic` launcher. 
That will permit us to state that "no launcher has to be published"

core/bootstrap/build.gradle.kts
paullatzelsperger: we can remove it from there.

extensions/azure/blob/build.gradle.kts
paullatzelsperger: fair point. maybe `blobstore` is better?

launchers/junit/build.gradle.kts
paullatzelsperger: will change it to `junit-extension`. we can figure out whether we want to move this to a different subfolder later.

extensions/azure/blob/build.gradle.kts
ndr-brt: maybe `blobstorage` because the official name is `Azure Blob Storage`

jimmarino:
@paullatzelsperger is this ready for review?

paullatzelsperger:
I think so

extensions/transfer-functions/transfer-functions-core/build.gradle.kts
```diff
@@ -29,8 +29,8 @@ dependencies {
 
 publishing {
     publications {
-        create<MavenPublication>("transfer-functions.core") {
-            artifactId = "transfer-functions.core"
+        create<MavenPublication>("transfer-functions-core") {
+            artifactId = "transfer-functions-ore"
```
alexandrudanciu, line 33: Missing c (core)

alexandrudanciu (commented):
Minor comment: Parameterising the plugin creation with the artifact's id might suggest some relevant semantic implication:

e.g., create<MavenPublication>("ids-spi")

As far as I can tell, this parameter only results in the name of the plugin instance and has no influence on the resulting artifact.

extensions/transfer-functions/transfer-functions-core/build.gradle.kts
paullatzelsperger: > Minor comment: Parameterising the plugin creation with the artifact's id might suggest some relevant semantic implication:
> 
> e.g., create("ids-spi")
> 
> As far as I can tell, this parameter only results in the name of the plugin instance and has no influence on the resulting artifact.

so what are you saying exactly :)?

alexandrudanciu: No need for action, just a comment.

PR #284, opened 2021-11-24
feat: implement contract negotiation part 2

juliapampus:
- [x] Implement `ConsumerContractNegotiationManagerImpl`
- [x] Implement `ProviderContractNegotiationManagerImpl`
- [x] Implement `InMemoryContractNegotiationStore` and `InMemoryContractNegotiationStoreExtension`
- [x] Extend `ContractServiceExtension`
- [x] Implement state transitions in `ContractNegotiation`
- [x] Add tests

PR #318, opened 2021-11-29
control-api: add endpoint to initiate a contract negotiation

juliapampus:
Depends on PR #284


jimmarino:
@juliapampus if you guys are backlogged with the core contract negotiation stuff, is this something @denisneuling could help with?

jimmarino (approved):

PR #322, opened 2021-11-30
Added "IN" operator to all AssetIndex imlementations

### LINKED ISSUE #307: Verify that asset Index supports IN criterion

paullatzelsperger:
Both the in-memory and the CosmosDB implementation of the `AssetIndex` can now handle the "IN" operator. Tests were added to verify that.

_Please be aware that the concrete syntax depends on the implementation! E.g. CosmosDB requires the list to be in brackets and - depending on the datatype - surrounded by single quotation marks, whereas the in-memory variant doesn't!_
 
Querying for a String property named `"someProperty"` would result in the following valid Criterion arguments:
|Impl | operandLeft | operator | operandRight |
| --------| :------------- |-------------:| :-----|
|In-mem | someProperty | "IN" | "(val1, val2, val3)" |
|cosmos | someProperty | "IN" | "('val1', 'val2', 'val3')"|

closes #307 

_Smaller changes to the `CosmosAssetIndex` and `CosmosAssetQueryBuilder` were necessary: the automatic addition of single-quotation marks was removed._

### DIFF
--- a/extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndex.java
+++ b/extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndex.java
@@ -63,7 +63,7 @@ public class CosmosAssetIndex implements AssetIndex, DataAddressResolver, AssetL
 
         SqlQuerySpec query = queryBuilder.from(expression);
 
-        var response = with(retryPolicy).get(() -> assetDb.queryItems(query.getQueryText()));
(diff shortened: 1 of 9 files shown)


jimmarino:
@paullatzelsperger Can we have the Cosmos impl quote the values or is there an issue with that?

paullatzelsperger:
> @paullatzelsperger Can we have the Cosmos impl quote the values or is there an issue with that?

I would not do that because those quotes are only needed for String properties. Thus we would have to do schema introspection, which might not even be possible.

jimmarino:
@paullatzelsperger and I discussed that we will create an issue covering how to enhance the Criteria framework to deal with typed properties and he will come up with a proposal for addressing it.

PR #325, opened 2021-11-30
Transfer Process Manager: consistent update of transfer processes

### LINKED ISSUE #324: Inconsistent transfer process updates by TransferProcessManager
The current implementation of the `TransferProcessManager` updates changes to processes in an inconsistent way:
* Sometimes processes are updated although no changes are available. This causes unnecessary write operations and gives issues when updating the state of a process from a separate thread as the TransferProcessManager overwrites the state back constantly.
* Sometimes `TransferProcessListeners` are notified before updating the process, sometimes after.

My suggestion would be to only update processes whenever changes are available and always before notifying `TransferProcessListeners`.

marcgs:
Consistent update of TransferProcesses in `TransferProcessManagerImpl`:

* Only update processes whenever changes are done
* Call `TransferProcessListeners` always after process has been updated

Closes #324

### DIFF
--- a/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
+++ b/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
@@ -198,12 +198,12 @@ public class TransferProcessManagerImpl extends TransferProcessObservable implem
                 } else {
                     process.transitionStreaming();
                 }
+                transferProcessStore.update(process);
                 invokeForEach(l -> l.inProgress(process));
                 monitor.debug("Process " + process.getId() + " is now " + TransferProcessStates.from(process.getState()));
             } else {
                 monitor.debug("Process " + process.getId() + " does not yet have provisioned resources, will stay in " + TransferProcessStates.REQUESTED_ACK);
             }
-            transferProcessStore.update(process);
         }
 
         return requestAcked.size();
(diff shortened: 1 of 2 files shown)


jimmarino:
Hi @marcgs Can you also sign the ECA? Thx

paullatzelsperger (approved):
LGTM, but what problem does this solve? Semantically there it's almost identical.

paullatzelsperger (changes requested):
one of the unit tests failed. PRs should only be marked "ready for review" once all the tests are green.

marcgs:
> LGTM, but what problem does this solve? Semantically there it's almost identical.

The TransferProcessManager was constantly updating TransferProcess in a loop, even when this was not necessary. The problem we faced is that is that we did not manage to move the state of a TransferProcess to ERROR after a timeout from a separate watchdog thread, as this was being overwritten almost always. With this change the chance of this happening is way smaller as the TPM will only do so if the process advances to the next state, in which case is it ok as the process was completed roughly at the same time the timeout expired.

Generally I think we should prevent unnecessary write operations as well.

marcgs:
> Hi @marcgs Can you also sign the ECA? Thx

@jimmarino will do ASAP 👍🏻

paullatzelsperger:
@marcgs ok, fair point :)

marcgs:
@paullatzelsperger @jimmarino I signed the ECA. You can go ahead and review the PR.

paullatzelsperger (approved):

PR #328, opened 2021-12-01
Create developer instruction for versioning and publishing

alexandrudanciu:
Hi everyone,

this PR introduces an instruction for developers describing the versioning scheme, legal requirement towards distribution artifacts, and the process of publishing maven artifacts.

Additionally, the NOTICE file now states the fact that the list of 3rd party libraries covers all submodules of the project (which is relevant when the file is embedded in an individual module jar).

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -26,7 +26,7 @@ The project maintains the following source code repositoriy:
 
 * https://github.com/eclipse-dataspaceconnector/DataSpaceConnector
 
-## Third-party Content
+## Third-party Content (Overarching All Modules)
 
 AWS SDK for Java - BOM
 * Project: https://aws.amazon.com/sdkforjava
--- a/docs/developer/releases.md
+++ b/docs/developer/releases.md
@@ -1,0 +1,117 @@
+Release Approach
+================
+
+## Table of Contents
+
+* [Versioning](#versioning)
+  * [API Compatibility](#api-compatibility)
+  * [Modules Providing API](#modules-providing-api)
+  * [Towards a First Release](#towards-a-first-release)
+* [Legal Documentation Requirements](#legal-documentation-requirements)
+  * [License and Notice Files](#license-and-notice-files)
+  * [Creating the Notice File](#creating-the-notice-file)
+  * [Background Information](#background-information)
+* [Publishing Maven Artifacts](#publishing-maven-artifacts)
+  * [Naming Convention](#naming-convention)
+
+### Versioning
+
+The Eclipse Dataspace Connector will employ [SemVer](https://semver.org) for versioning and distinguish between the following releases as defined by the [Eclipse Handbook](https://www.eclipse.org/projects/handbook/#release):
+
+- Major releases, which introduce API changes,
+- minor releases, which add new functionality, but are API compatible with previous versions, and
+- service releases, which include bug fixes only and add no significant new functionality.
+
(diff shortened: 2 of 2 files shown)


docs/developer/Releases.md
```diff
@@ -0,0 +1,116 @@
...
+* [Legal Documentation Requirements](#legal-documentation-requirements)
+  * [License and Notice Files](#license-and-notice-files)
+  * [Creating the Notice File](#creating-the-notice-file)
+  * [Background Information](#background-information)
+* [Publishing Maven Artifacts](#publishing-maven-artifacts)
+  * [Naming Convention](#naming-convention)
+
+### Versioning
+
+The Eclipse Dataspace Connector employs [SemVer](https://semver.org) for versioning and distinguishes between the following releases as defined by the [Eclipse Handbook](https://www.eclipse.org/projects/handbook/#release):
```
jimmarino, line 19: I would say "will employ" so we don't give anyone the false impression this will happen before 1.0 (even though it is mentioned below).

docs/developer/Releases.md
```diff
@@ -0,0 +1,116 @@
...
+Between releases, snapshot versions reflecting the current state of modules can be packaged to distribution artifacts on a regular basis. 
+Snapshots, however, do not actually represent released versions.
+
+#### API Compatibility
+
+The concept of API compatibility is defined in terms of binary compatibility according to the [Java SE 11 Language Specification](https://docs.oracle.com/javase/specs/jls/se11/html/jls-13.html) 
+
+#### Modules Providing API
+
+The following modules specify a Java Service Provider Interface (SPI) and contribute public classes, interfaces and public members which are considered public APIs and are covered by the proposed versioning approach:
```
jimmarino, line 34: To be precise, I would say "The following modules define official extension points for the EDC and contribute..."

docs/developer/Releases.md
```diff
@@ -0,0 +1,116 @@
...
+- spi
+- data-protocols/ids-spi
+
+In addition to the SPI-based extension points, other public APIs (e.g., based on REST) are also contributed.
+So that changes to these APIs can be implemented without impacting the connector's release cycle, such modules can be located within an own repository.
```
jimmarino, line 39: This is a little unclear to me: 

So that changes to these APIs can be implemented without impacting the connector's release cycle, such modules can be located within an own repository.

Which APIs, the SPI modules or the REST interfaces?

docs/developer/Releases.md
```diff
@@ -0,0 +1,116 @@
...
+- extensions/*
+
+Extensions can in turn specify their own SPI-based extension points.
+Theses are, however, regarded as **internal SPI** and not as a public API.
+Therefore, changing internal SPI doesn't necessarily imply a version increment for the module.
+
+#### Towards a First Release
+
+Until its first major release, the Eclipse Dataspace Connector will be developed under the version 0.0.1 without complying to semantic versioning.
+According snapshot version might break the binary compatibility.
```
jimmarino, line 52: I would word it as, "Snapshot versions may break binary compatibility and are not guaranteed to be stable"

jimmarino:
Thanks, Alex, I left a few small comments inline.

docs/developer/Releases.md
```diff
@@ -0,0 +1,117 @@
+Release Approach
```
ndr-brt, line 1: I would use the naming convention that we used with the other documentation markdownsand call this `releases.md`, with the first letter in lowercase

ndr-brt (approved):

Issue #338, opened 2021-12-02
LeftOperand class is an enum which is not compliant with IDS or its use of ODRL


jimmarino:
The info model models the LeftOperand of a constraint as an enum. This means that the infomodel cannot express constraints other than those defined in the enum class. For example, it is impossible to express policies that include usage constraints other than the 17 defined in the enum class.

My understanding of IDS is it supports ODRL and extensible policy constraints so this is a bug in the info model. This is currently a severe blocker for the EDC and need to reach out to the infomodel folks for help.

juliapampus:
Tried to note down some observations, past discussion outcomes, and next steps @jimmarino.

## Policy Language in IDS

- IDS extends ODRL: the IDS Usage Control Language was initially defined as a [profile](https://www.w3.org/TR/odrl-model/#profile-mechanism) of the ODRL language
- Referring to [this comment](https://github.com/International-Data-Spaces-Association/InformationModel/issues/523#issuecomment-967040234), IDS introduces pre- and post-duties that could, nevertheless, be mapped to ODRL
- In IDS, it was tried to map use case requirements to policy classes (see [here](https://github.com/International-Data-Spaces-Association/InformationModel/tree/develop/examples/contracts-and-usage-policy/templates) for templates) and thus, attributes as the left operand or actions were restricted to a number of enums that seemed reasonable.

### Action: IDS vs ODRL

Taking a look at [this documentation](https://w3id.org/idsa/core), this is the current status:
- `Action`: A thing one might be permitted to do or prohibited from doing to something.
   - has super-classes: action
   - is in domain of: action refinement, is included in
   - is in range of: action, is included in
   - has members: _aggregate by consumer_, _aggregate by provider_, anonymize, compensate, delete, distribute, _ecrypt_, grant use, _log_, modify, next policy, _notify_, read, _track provenance_, use, write (16)

ODRL defines top-level actions (use, transfer). According to [the documentation](https://www.w3.org/TR/odrl-model/#action), Action terms must be defined using the includedIn property referring to an encompassing Action and either use or transfer as the top-level parent term by transitive means. Details on the includedIn and implies properties can be defined by ODRL profiles. 

The [ODRL vocabulary](https://www.w3.org/TR/odrl-vocab/#term-Action) defines 72 action types. Some of them appear in IDS, a lot of them not. On top of that, IDS defines other actions (marked cursive).

### LeftOperand: IDS vs ODRL

[...]

PR #342, opened 2021-12-02
Support IDS logical constraint transformations

jimmarino:
IDS logical constraints need to be included for the milestone. This implements transforming from both atomic and multiplicty constraints to IDS counterparts.


data-protocols/ids/ids-transform-v1/build.gradle.kts
```diff
@@ -23,6 +23,7 @@ plugins {
 dependencies {
     api(project(":spi"))
     api(project(":data-protocols:ids:ids-spi"))
+    implementation(project(":common:util"))
```
denisneuling, line 26: Just as a note: Seems this dependency has been introduced for making use of `CollectionUtil#isNotEmpty` in `IdsLogicalConstraintToConstraintTransformer`: IMO this could have also been placed directly there since it's just  `collection == null || collection.isEmpty()`.

data-protocols/ids/ids-transform-v1/src/main/java/org/eclipse/dataspaceconnector/ids/transform/IdsLogicalConstraintToConstraintTransformer.java
```diff
@@ -0,0 +1,113 @@
...
+            constraintTypes++;
+        }
+        if (constraintTypes == 0) {
+            context.reportProblem("Logical constraint must specify at least AND, OR, or XONE constraints");
+            return null;
+        } else if (constraintTypes == 1) {
+            return transformSingleConstraintType(constraint, context);
+        } else {
+            return transformMultipleConstraintTypes(constraint, context);
+        }
```
denisneuling, line 66: Code here seems a little bit complex, but don't have a idea on how to simplify right now. Let's improve this within our refactoring week.

denisneuling (approved):
@jimmarino Thanks for bringing this in! Except the things mentioned, everything looks good to me. But we can improve on this later. No objections to merge from my sinde.

data-protocols/ids/ids-transform-v1/build.gradle.kts
jimmarino: Yes, but it's an internal utility library so I think it is ok and avoids code duplication (albeit one line ;-) )

data-protocols/ids/ids-transform-v1/src/main/java/org/eclipse/dataspaceconnector/ids/transform/IdsLogicalConstraintToConstraintTransformer.java
jimmarino: Sure. Unfortunately, however, I think this complexity is a result of the InfoModel not using an inheritance hierarchy for Constraints. That choice also leads to other issues, including the difficulty of supporting nested constraints through recursion in IDS. My initial inclination is the InforModel itself will need to change for this to get simpler.

There is also #338 which IMO is a critical blocker.

PR #362, opened 2021-12-06
spi: add name method on ServiceExtension

### LINKED ISSUE #358: Give name to ServiceExtension
As discussed for the refactoring week

#### Description
At the moment every `ServiceExtension` has to log the `Initializated`/`Started`/`Shutdown` events, but that could be done by the `ServiceExtensionContext` and by the `ExtensionsLoader`.
The `ServiceExtension` could have a `name()` method that returns by default the class name but it could also be overwritten.

#### Scope
the `ServiceExtension` class and then all the implementations

#### Benefit
Less code, the service init/start/shutdown logging phase will be handled in a single place, no need to remember to add logs on the service extensions.

#### Risk
None

ndr-brt:
and avoid service lifecycle log duplication
Fix #358

### DIFF
--- a/spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/ServiceExtension.java
+++ b/spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/ServiceExtension.java
@@ -24,6 +24,13 @@ import java.util.Set;
 public interface ServiceExtension extends SystemExtension {
 
     /**
+     * Returns the name of the service, by default is its class name
+     */
+    default String name() {
+        return this.getClass().getSimpleName();
+    }
+
+    /**
      * Returns the list of features provided by this extension.
      */
     default Set<String> provides() {
--- a/spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/ServiceExtension.java
+++ b/spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/ServiceExtension.java
@@ -27,7 +27,7 @@ public interface ServiceExtension extends SystemExtension {
      * Returns the name of the service, by default is its class name
      */
     default String name() {
-        return this.getClass().getSimpleName();
+        return this.getClass().getName();
     }
 
     /**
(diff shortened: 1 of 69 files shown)


spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/ServiceExtension.java
```diff
@@ -23,6 +23,13 @@
  */
 public interface ServiceExtension extends SystemExtension {
 
+    /**
+     * Returns the name of the service, by default is its class name
+     */
+    default String name() {
+        return this.getClass().getSimpleName();
```
jimmarino, line 30: I think it should be the FQN to avoid name clashes

jimmarino (commented):
Looks good except for the one comment on using the FQN. Also, I didn't realize you were proposing removing the "user friendly" extension naming and just using the class name. I think it would provide more information to have the default method return the FQN but for extensions this repo provides, always implement name() and return a more descriptive string.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/ServiceExtension.java
ndr-brt: like `serviceExtensionName`? which name clashes could happen?

ndr-brt: oh, I realized, you talked about the result, not the method name. wouldn't be that too much verbose and confusing on the logs?

jimmarino: Per my other comment, for our extensions, I think we should always implement a name and return a user/operator friendly extension name like we currently do. The class name doesn't tell much but makes sense as a fallback default. However, this should be an FQN (package name included) since it is possible for two extension names to clash otherwise. I don't think it will be too verbose.

ndr-brt:
@jimmarino done, I also noticed that other extensions like Configuration and Vault one missed the name property, so I pulled up to `SystemExtension` and moved out the logs on those extensions

jimmarino:
@ndr-brt This is a great simplification!

PR #365, opened 2021-12-06
spi: introduce Result abstraction

### LINKED ISSUE #363: Abstract Result concept
#### Description
There's a lot of classes in the codebase that represent the concept of a `Result`, something returned by a component that could be a valid content or an error (with details)
#### Scope
All the codebase
#### Benefit
Fo3r EDC developers it will avoid to have to reinvent the wheel everytime they need to return something that could be a valid content or an error
#### Risk
Can't see any

ndr-brt:
Fix #363


spi/src/main/java/org/eclipse/dataspaceconnector/spi/Result.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public static <T> Result<T> success(@NotNull T content) {
+        return new Result<>(content, emptyList());
+    }
+
+    public static <T> Result<T> failure(String failure) {
+        return new Result<>(null, List.of(failure));
+    }
+
+    public static <T> Result<T> failure(List<String> failures) {
```
jimmarino, line 33: A nit: this should probably be "failures(...)"

spi/src/main/java/org/eclipse/dataspaceconnector/spi/Result.java
```diff
@@ -0,0 +1,64 @@
...
+    public boolean succeeded() {
+        return failures.isEmpty();
+    }
+
+    public boolean failed() {
+        return !failures.isEmpty();
+    }
+
+    public String getFailure() {
+        return failures.stream().findFirst().orElseThrow(() -> new EdcException("This result is successful"));
```
jimmarino, line 58: So far this looks really good. I don't think it should throw an exception here though. Would it be better to just offer a failures() method? For example, how would a client know if just one or many reasons were returned?

spi/src/main/java/org/eclipse/dataspaceconnector/spi/Result.java
```diff
@@ -0,0 +1,64 @@
...
+
+package org.eclipse.dataspaceconnector.spi;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+import static java.util.Collections.emptyList;
+
+public class Result<T> {
```
jimmarino, line 23: One thing to think about: there are cases when we need to distinguish the type of failure, for example, is it fatal or can it be retried.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/Result.java
ndr-brt: I thought it more in the way of: "This result represent a failure, and here's the detail" in opposition of "this result represent a success"

spi/src/main/java/org/eclipse/dataspaceconnector/spi/Result.java
ndr-brt: You're right, I think at the end there will be only the `getFailures` method

spi/src/main/java/org/eclipse/dataspaceconnector/spi/Result.java
jimmarino: Maybe it should be "problem" or "error detail" then?

spi/src/main/java/org/eclipse/dataspaceconnector/spi/Result.java
ndr-brt: I see, like it's handled in `HttpFunctionDataFlowController`. 
I think we can choose between two naming conventions:
```
Result.success
Result.failure
Result.fatalFailure
```

or
```
Result.ok
Result.error
Result.fatalError
```

jimmarino: Looks good, I don't have strong preference either way, but maybe "success"?

ndr-brt:
Ok, I think I reached a stable point in defining the design.
The proposal contained in this PR is to introduce an `AbstractResult`, that represent a result object that can be a success or a failure.
The success case will bring an object that represents the content of the computation, 
while the failure case bring an object that represents the failure of the computation.
The simpler failure is the one that report a list of messages, it's used in the `Result` extension class that covers the majority of the cases.
If some other details are needed (e.g. a status that can differentiate a simple error from a fatal one, like `DataFlowInitiateResult`) it's enough to extend `AbstractResult` setting as failure an extension of the `Failure` class.
In this way things are extensible and pretty clear (at least to me :) ).
I put the result core classes here: https://github.com/BMW-InnovationLab/DataSpaceConnector/tree/abstract-result/spi/src/main/java/org/eclipse/dataspaceconnector/spi/result

Attention points:
* This PR become quite big, so I'm not sure if this PR deserves to be merged eventually or instead we should introduce the `AbstractResult/Result/Failure` classes and then fix the duplication in some other PRs therefore
* I found a case that can't fit well with this implementation, is the `QueryResponse`, that provides a successful state with errors attached. This is a different pattern and should be left as it is.

jimmarino:
> Ok, I think I reached a stable point in defining the design. The proposal contained in this PR is to introduce an `AbstractResult`, that represent a result object that can be a success or a failure. The success case will bring an object that represents the content of the computation, while the failure case bring an object that represents the failure of the computation. The simpler failure is the one that report a list of messages, it's used in the `Result` extension class that covers the majority of the cases. If some other details are needed (e.g. a status that can differentiate a simple error from a fatal one, like `DataFlowInitiateResult`) it's enough to extend `AbstractResult` setting as failure an extension of the `Failure` class. In this way things are extensible and pretty clear (at least to me :) ). I put the result core classes here: https://github.com/BMW-InnovationLab/DataSpaceConnector/tree/abstract-result/spi/src/main/java/org/eclipse/dataspaceconnector/spi/result
> 
Hi @ndr-brt this looks really promising. I took a look at DataFlowInitiateResult and the case of distinguishing fatal and non-fatal errors. Given that this PR is quite large, and it may benefit others to see this, can you paste in a quick example of how client code would use DataFlowInitiateResult and distinguish between the three result case? I this it would also be worth documenting the fatal vs. non-fatal extension case this on the AbstractResult class itself in general terms.

ndr-brt:
Sure, the example is shown in the code of [`TransferProcessManagerImpl`](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/7d96af5d8aec928cffc1ae6b6d399625b776a28c/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java#L297):

```java
if (ResponseStatus.ERROR_RETRY == response.getFailure().status()) {
    monitor.severe("Error processing transfer request. Setting to retry: " + process.getId());
    process.transitionProvisioned();
    transferProcessStore.update(process);
    invokeForEach(l -> l.provisioned(process));
} else if (ResponseStatus.FATAL_ERROR == response.getFailure().status()) {
    monitor.severe(format("Fatal error processing transfer request: %s. Error details: %s", process.getId(), String.join(", ", response.getFailureMessages())));
    process.transitionError(response.getFailureMessages().stream().findFirst().orElse(""));
    transferProcessStore.update(process);
    invokeForEach(l -> l.error(process));
} else {
    if (process.getDataRequest().getTransferType().isFinite()) {
        process.transitionInProgress();
    } else {
        process.transitionStreaming();
    }
    transferProcessStore.update(process);
    invokeForEach(l -> l.inProgress(process));
}
```

I think that, with this design, will be up to those who'll need a particular failure to be handled in different ways to extend the `AbstractResult` and `Failure` classes to fit their needs. Or, if this case (the fatal/not fatal one) is generic enough we can provide an generic implementation too in the `spi` (I just noticed that `DataFlowInitiateResult` and `TransferInitiateResult` are pretty similar and also `NegotiationResult` is not that different)

jimmarino:
Hi, @ndr-brt if you think this is ready to merge, do you want to let us know? My preference would be to get this in ASAP and we can always make subsequent changes in separate PRs

PR #384, opened 2021-12-09
Replace easymock with mockito

ndr-brt:


paullatzelsperger (changes requested):
I had a few general comments, that apply to the entire changed codebase. 

1. Mockito requires explicit verification whereas easymock has explicit setups. That means, that EasyMock-mocks need to be _set up_ in a particular way, whereas Mockito requires a particular _verification_ sequence. 

1. mockito does not have the concept of a `strict` mock, thus the exact invocations must be 
made sure through verification sequences and calls to `verifyNoMoreInvocations`

1. declaring mock variables explicitly seems cleaner, for example 
  `SomeObject obj = mock(SomeObject.class)` 
   instead of 
  `var obj = (SomeObject)mock(SomeObject.class)`

Personally, I would not trust the new tests unless points 1 and 2 are adoped.

Unrelated: mocking POJOs seems weird, e.g. mocking `Assets` in `AssetToIdsArtifactTransformer`. I know that this is not part of this PR, just wanted to point out.

ndr-brt:
> 1. Mockito requires explicit verification whereas easymock has explicit setups. That means, that EasyMock-mocks need to be _set up_ in a particular way, whereas Mockito requires a particular _verification_ sequence.

Right, but we need to distinguish two cases:
* _assert the result of a call:_ in the majority of the cases this means that the mocks are there just to serve us some results that will compose the object returned by the SUT. In that case, drive the mocks with the `when` statement is just enough, we don't have to verify that every call to the mocks has happened, because we rely only on the result of the call
* verify that a method on a collaborator (mocked) is called in a certain way: in these cases, like where we test void methods or so, we can't assert the result of the call but we have to verify that the collaborator has been called in a certain way.

Doing this replacement I removed useless verifications where they were overkill, e.g. [`VaultPrivateKeyResolverTest` with EasyMock](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/3d477ac9256bb027141bb05ceb792b68a5910b7c/core/bootstrap/src/test/java/org/eclipse/dataspaceconnector/spi/security/VaultPrivateKeyResolverTest.java) -> [`VaultPrivateKeyResolverTest` with Mockito](https://github.com/BMW-InnovationLab/DataSpaceConnector/blob/42903196982e9297584f57082982f622e86043a1/core/bootstrap/src/test/java/org/eclipse/dataspaceconnector/security/VaultPrivateKeyResolverTest.java)

Instead, where the verification was crucial for the result of the test I reported it also on Mockito e.g. [`CoreServicesExtensionTest` with EasyMock](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/3d477ac9256bb027141bb05ceb792b68a5910b7c/core/bootstrap/src/test/java/org/eclipse/dataspaceconnector/system/CoreServicesExtensionTest.java) -> [`CoreServicesExtensionTest` with Mockito](https://github.com/BMW-InnovationLab/DataSpaceConnector/blob/42903196982e9297584f57082982f622e86043a1/core/bootstrap/src/test/java/org/eclipse/dataspaceconnector/system/CoreServicesExtensionTest.java)

I carefully controlled these occurrences but could be that I missed a few, in that case the test will be a little looser but not in a dangerous way. 

> 2. mockito does not have the concept of a `strict` mock, thus the exact invocations must be
>    made sure through verification sequences and calls to `verifyNoMoreInvocations`

I haven't found cases when the `verifyNoMoreInvocations` were needed, regarding the exact invocations, same as before, I tried to make the verifications as consistent as possibile but I could have missed a few. Could you point me to an example of that?

> 3. declaring mock variables explicitly seems cleaner, for example
>    `SomeObject obj = mock(SomeObject.class)`
>    instead of
>    `var obj = (SomeObject)mock(SomeObject.class)`

Yes, they are typos, will fix them asap

> Unrelated: mocking POJOs seems weird, e.g. mocking `Assets` in `AssetToIdsArtifactTransformer`. I know that this is not part of this PR, just wanted to point out.

Yes, totally agree, I replaced them with real instantiation where I noticed them, it's a thing that I've already pointed out before, it should become one of our best practices in writing tests.

ndr-brt:
> > Right, but we need to distinguish two cases:
> 
> Sure, but unless you can guarantee 100% that the tests are doing absolutely the same thing as they did before, I say we need to have those verifications in place, even if some of them might be superfluous. It's just a matter of safety. That being said, I totally agree that going forward we can employ the pattern of distinguishing between those two cases.

You mean, going through all the tests and replicate every `when` with a `verify` after the assertions?

paullatzelsperger:
> You mean, going through all the tests and replicate every when with a verify after the assertions?

I'm afraid that's the only way we can assert that the tests are 100% the same as they were before, which was a precondition to doing this task.

ndr-brt:
Ok, the task is huge and I think it will lead to bad-written tests but, if there are no alternatives I will do it, I need this to be merged, the rebase work is becoming insane :)

paullatzelsperger:
> Ok, the task is huge and I think it will lead to bad-written tests but, if there are no alternatives I will do it, I need this to be merged, the rebase work is becoming insane :)

We can always open another issue to improve all the tests (did you just volunteer for it :) ?), but we absolutely should do this in another step to ensure that we don't mix up things here.

ndr-brt:
Done. If it's ok we can merge this.

extensions/azure/assetindex-cosmos/src/test/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexTest.java
```diff
@@ -87,114 +80,87 @@ void inputValidation() {
...
-
-        CosmosAssetIndex assetIndex = new CosmosAssetIndex(api, TEST_PARTITION_KEY, typeManager, retryPolicy);
+        when(api.queryItemById(eq(id))).thenReturn(document);
 
         Asset actualAsset = assetIndex.findById(id);
-        assertThat(actualAsset.getProperties()).isEqualTo(document.getWrappedAsset().getProperties());
 
-        verify(api);
+        assertThat(actualAsset.getProperties()).isEqualTo(document.getWrappedAsset().getProperties());
+        verify(api).queryItemById(eq(id));
```
paullatzelsperger, line 88: this would be a situation where `verifyNoMoreInteractions` would actually contribute to the test strictness, since `api` was a strictmock before. same for all other testmethods in this class.

extensions/azure/contract-negotiation-store-cosmos/src/test/java/org/eclipse/dataspaceconnector/contract/negotiation/store/CosmosContractNegotiationStoreTest.java
```diff
@@ -12,23 +12,21 @@
...
+import static org.mockito.Mockito.when;
 
 class CosmosContractNegotiationStoreTest {
     private CosmosContractNegotiationStore store;
     private CosmosDbApi cosmosDbApiMock;
 
     @BeforeEach
     void setup() {
-        cosmosDbApiMock = strictMock(CosmosDbApi.class);
+        cosmosDbApiMock = mock(CosmosDbApi.class);
```
paullatzelsperger, line 29: since this is was a `strictMock` before, `verifyNoMoreInteractions` would actually make sense here.

extensions/catalog/federated-catalog-cache/src/test/java/org/eclipse/dataspaceconnector/catalog/cache/loader/LoaderManagerImplTest.java
```diff
@@ -8,88 +8,69 @@
...
     }
 
-
     @Test
     @DisplayName("Verify that the LoaderManager does not sleep when a complete batch was processed")
     void batchSizeReachedWithinTimeframe() throws InterruptedException {
-        for (var i = 0; i < loaderManager.getBatchSize(); i++) {
-            queue.offer(new UpdateResponse());
-        }
+        range(0, 3).forEach(i -> queue.offer(new UpdateResponse()));
```
paullatzelsperger, line 60: love it :)

extensions/dataloading/dataloading-spi/src/test/java/org/eclipse/dataspaceconnector/dataloading/DataLoaderTest.java
```diff
@@ -48,17 +48,15 @@ void setUp() {
...
     void insert() {
         var te = new TestEntity("Test Desc", 3);
-        sinkMock.accept(te);
-        expectLastCall();
-        replay(sinkMock);
+
         dataLoader.insert(te);
-        verify(sinkMock);
+
+        verify(sinkMock).accept(te);
```
paullatzelsperger, line 54: adding `verifyNoMoreInteractions` would resemble more closely the behaviour of a `strictMock`

paullatzelsperger (changes requested):
Looks much better now. However, I think that all mocks that were `strictMocks` before, should be verified with `verifyNoMoreInteractions`, as it is more semantically equivalent to what it was before and thus increases test strictness. I marked a few locations in the code. 

However, in order to get it merged quickly and to not cause an "in-between" state much longer, I'm OK with merging this as-is and handling that in another PR.

paullatzelsperger (approved):

PR #392, opened 2021-12-10
Introduce unified implementation of `DataFlowController` for managing inline data transfers (closes #364)

### LINKED ISSUE #364: Samples: improve the implementation of the `DataFlowControllers`
Currently there is only two `DataReader` and `DataWriter` implementations: one of each for S3 and Azure BlobStore. This should be expanded additional readers and writers can be supplied easily, e.g. through a registry.

The emerging `DataFlowController` could then even be adopted into the EDC "core" codebase (as opposed to it being a sample).

Linked to #246 

will be done together with @bscholtes1A

bscholtes1A:


spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/flow/DataOperatorRegistryImpl.java
```diff
@@ -0,0 +1,30 @@
+package org.eclipse.dataspaceconnector.spi.transfer.flow;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public class DataOperatorRegistryImpl implements DataOperatorRegistry {
```
ndr-brt, line 6: The implementation should go into the core:transfer module

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/flow/DataReader.java
```diff
@@ -0,0 +1,9 @@
+package org.eclipse.dataspaceconnector.spi.transfer.flow;
+
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataAddress;
+
+public interface DataReader {
+    boolean canHandle(String type);
+
+    String read(DataAddress source);
```
ndr-brt, line 8: Maybe the return type should be enclosed in a Result that could report also errors? 
Maybe byte[] or InputStream would be better to handle a binary data type

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/flow/DataWriter.java
```diff
@@ -0,0 +1,9 @@
+package org.eclipse.dataspaceconnector.spi.transfer.flow;
+
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataAddress;
+
+public interface DataWriter {
+    boolean canHandle(String type);
+
+    void write(DataAddress destination, String name, String data, String secretToken);
```
ndr-brt, line 8: Like above, a Result would help handling errors

ndr-brt (commented):
This PR should also replace some `DataFlowController` implementations with the same behaviour in the samples.

jimmarino:
Should this go into core, i.e. is it a "core" concept? For example, do these interfaces represent a capability that all or nearly all transfer extensions would implement? I think most production async transfer implementations would perform read and write operations out-of-band from the connector.

Another alternative is to create a higher-level SPI for extensions that implement this behavior.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/flow/DataOperatorRegistryImpl.java
bscholtes1A: Moved accordingly

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/flow/DataReader.java
bscholtes1A: Done

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/flow/DataWriter.java
bscholtes1A: Done

bscholtes1A:
> Should this go into core, i.e. is it a "core" concept? For example, do these interfaces represent a capability that all or nearly all transfer extensions would implement? I think most production async transfer implementations would perform read and write operations out-of-band from the connector.
> 
> Another alternative is to create a higher-level SPI for extensions that implement this behavior.

Hi @jimmarino 
I moved the inline-data-transfer part into the `extensions` folder next to the transfer-functions. Discussing with @paullatzelsperger , it seems the `InlineDataFlowController` introduced in this PR might become the cornerstone of the data-plane module that will be introduced soon.

extensions/aws/s3/writer/src/main/java/org/eclipse/dataspaceconnector/writer/s3/S3BucketWriter.java
```diff
@@ -0,0 +1,80 @@
...
+                .region(Region.of(region))
+                .build()) {
+            return transferData(s3, bucketName, data, name);
+        } catch (S3Exception ex) {
+            monitor.severe("Data request: transfer failed!");
+            return Result.failure("Failed to transfer data");
+        }
+    }
+
+    private Result<Void> transferData(S3Client client, String bucketName, byte[] data, String name) {
```
ndr-brt, line 54: this method can be inlined, since it does not a lot and the exception is already caught from the `write` method

extensions/aws/s3/writer/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+dependencies {
+    api(project(":spi"))
+    api(project(":core:schema"))
+    api(project(":extensions:aws:s3:provision"))
+    api(project(":extensions:inline-data-transfer:inline-data-transfer-spi"))
+}
+
+publishing {
+    publications {
+        create<MavenPublication>("s3-schema") {
```
ndr-brt, line 30: the artifact name should be `s3-writer`

samples/04-file-transfer/transfer-file/src/main/java/org/eclipse/dataspaceconnector/extensions/api/FileTransferDataStreamPublisher.java
```diff
@@ -2,41 +2,44 @@
...
+import org.eclipse.dataspaceconnector.transfer.inline.spi.DataStreamPublisher;
+import org.eclipse.dataspaceconnector.transfer.inline.spi.StreamContext;
 
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 
-public class FileTransferFlowController implements DataFlowController {
+public class FileTransferDataStreamPublisher implements DataStreamPublisher {
```
ndr-brt, line 15: The README of sample 04 needs to be updated

samples/05-file-transfer-cloud/transfer-file/src/main/java/org/eclipse/dataspaceconnector/extensions/transfer/CloudTransferExtension.java
```diff
@@ -1,12 +1,14 @@
 package org.eclipse.dataspaceconnector.extensions.transfer;
```
ndr-brt, line 1: The README of the sample 05 needs to be updated

ndr-brt (approved):
Neat implementation, i like it, just left some simple comments.
Would have been nice to see some tests especially on the `InlineDataFlowController`.
However, LGTM.

extensions/inline-data-transfer/inline-data-transfer-core/src/main/java/org/eclipse/dataspaceconnector/transfer/inline/core/DataOperatorRegistryImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import org.eclipse.dataspaceconnector.transfer.inline.spi.DataOperatorRegistry;
+import org.eclipse.dataspaceconnector.transfer.inline.spi.DataReader;
+import org.eclipse.dataspaceconnector.transfer.inline.spi.DataStreamPublisher;
+import org.eclipse.dataspaceconnector.transfer.inline.spi.DataWriter;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public class DataOperatorRegistryImpl implements DataOperatorRegistry {
```
paullatzelsperger, line 13: This could be registered by default in the CoreTransferExtension as it does not have any dependencies. This registry is an essential part and should always be there, even if the `InlineDataFLowController` is not used.

extensions/inline-data-transfer/inline-data-transfer-spi/src/main/java/org/eclipse/dataspaceconnector/transfer/inline/spi/DataWriter.java
```diff
@@ -0,0 +1,27 @@
...
+import org.eclipse.dataspaceconnector.spi.result.Result;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataAddress;
+
+/**
+ * A system responsible for writer data from to a destination (e.g. Azure blob storage,...)
+ */
+public interface DataWriter {
+    boolean canHandle(String type);
+
+    Result<Void> write(DataAddress destination, String name, byte[] data, String secretToken);
```
paullatzelsperger, line 26: Could this take an `InputStream` instead of the byte array?

extensions/azure/blob/reader/src/main/java/org/eclipse/dataspaceconnector/reader/azure/blob/BlobStoreReader.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    @Override
+    public boolean canHandle(String type) {
+        return "AzureStorage".equals(type);
+    }
+
+    @Override
+    public Result<ByteArrayInputStream> read(DataAddress source) {
+        var account = source.getProperty("account");
```
paullatzelsperger, line 25: This was not introduced by you, but could we use `AzureBlobStoreSchema` to access those constants?

paullatzelsperger (changes requested):
Other than a few minor nits I think we can merge this so as to not block upcoming tasks

PR #396, opened 2021-12-10
Move duplicated classes to new module core:base

### LINKED ISSUE #357: Introduce a foundation module
When implementing the contract negotiation subsystem, multiple services needed to be copied from the transfer process module, like the WaitStrategy. These services should be consolidated into a third foundation module that the others depend on.

ronjaquensel:
Closes #357


core/base/src/main/java/org/eclipse/dataspaceconnector/core/base/ExponentialWaitStrategy.java
```diff
@@ -9,17 +9,17 @@
...
+package org.eclipse.dataspaceconnector.core.base;
 
+import org.eclipse.dataspaceconnector.spi.contract.negotiation.NegotiationWaitStrategy;
 import org.eclipse.dataspaceconnector.spi.transfer.TransferWaitStrategy;
 
-/**
- * Implements an exponential backoff strategy for failed iterations.
- */
-public class ExponentialWaitStrategy implements TransferWaitStrategy {
+public class ExponentialWaitStrategy implements TransferWaitStrategy, NegotiationWaitStrategy {
```
ndr-brt, line 21: Do we still need three interfaces for this concept? Would `WaitStrategy` just be enough? (Eventually it can be moved from the `transfer` package in `spi`

ronjaquensel: Not sure. As the same wait strategy is currently used everywhere, we could probably work with just the `WaitStrategy` interface. But then we'd remove the possibility for extensions to provide a custom wait strategy for either contract negotiation or data transfer, as all new implementations would just be registered under the `WaitStrategy` interface.

jimmarino: Could we have a single interface but the ability to define different WaitStrategy configurations via a service?

ronjaquensel: That could work. Should this be part of this PR or a subsequent one?

ndr-brt: In another one for me would work too.

jimmarino: Either way, whichever you think is more manageable.

ronjaquensel: Then let's merge this one first and I'll open an issue for providing a service for the wait strategies.

jimmarino (approved):

core/base/src/main/java/org/eclipse/dataspaceconnector/core/base/ExponentialWaitStrategy.java
jimmarino: @ndr-brt Are you OK with this, if so, I'll merge it

ndr-brt: @jimmarino yes, ok!

ndr-brt (approved):

Issue #381, opened 2021-12-08
Sample 05-file-transfer-cloud broken (assetId)


algattik:
When running sample 05 according to its instructions, the consumer log contains:
```
SEVERE 2021-12-08T10:02:01.741348 Error caught in transfer process manager
java.lang.NullPointerException
	at java.base/java.net.URI$Parser.parse(URI.java:3104)
	at java.base/java.net.URI.<init>(URI.java:600)
	at java.base/java.net.URI.create(URI.java:881)
	at org.eclipse.dataspaceconnector.ids.core.message.DataRequestMessageSender.send(DataRequestMessageSender.java:92)
	at org.eclipse.dataspaceconnector.ids.core.message.DataRequestMessageSender.send(DataRequestMessageSender.java:44)
	at org.eclipse.dataspaceconnector.ids.core.message.IdsRemoteMessageDispatcher.send(IdsRemoteMessageDispatcher.java:46)
	at org.eclipse.dataspaceconnector.transfer.core.protocol.provider.RemoteMessageDispatcherRegistryImpl.send(RemoteMessageDispatcherRegistryImpl.java:47)
	at org.eclipse.dataspaceconnector.transfer.core.transfer.TransferProcessManagerImpl.sendOrProcessProvisionedRequests(TransferProcessManagerImpl.java:294)
	at org.eclipse.dataspaceconnector.transfer.core.transfer.TransferProcessManagerImpl.run(TransferProcessManagerImpl.java:116)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

Presumably the cause is a missing assetId in the query

paullatzelsperger:
quite a bit of work was done in that area and in the samples over the last couple of days. would you mind checking if this still persists?

algattik:
Now this fails with

`java -Dedc.fs.config=samples/05-file-transfer-cloud/provider/config.properties -jar samples/05-file-transfer-cloud/provider/build/libs/provider.jar`

```
[...]

PR #401, opened 2021-12-13
sample 5: generate service principal certificate with Terraform

algattik:
Generate X.509 certificate in Terraform, rather than on local machine.

Benefits:
- not requiring OpenSSL tools to be installed locally to deploy.
- not requiring the user to manually enter certificate information.
- useful Terraform code that can be adapted by end users for development environments.

Deployment fails against upstream main because of #381, but I tested it against an older head with minor modifications.

Other improvements:
- Locked terraform provider versions for maximum stability.
- Applied `terraform fmt` to format terraform files.


paullatzelsperger:
Cool stuff, thanks!

paullatzelsperger (approved):

ndr-brt (approved):
thanks

PR #412, opened 2021-12-13
Basic Connector Observability

### LINKED ISSUE #206: Implement basic connector observability
We should provide a module that implements basic connector observability by offering the following features: 
### A `health` endpoint
returns information about the system status. This is primarily intended to be used in [`HEALTHCHECK` instructions in Docker files](https://docs.docker.com/engine/reference/builder/#healthcheck). 

### A `liveness` endpoint
From the [Kubernetes docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes): 
> indicates whether the container is running. 

We can assume, that in the event that the connector crashes, all REST endpoints become unavailable. 

### A `readiness` endpoint
From the [Kubernetes docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes): 
> Indicates whether the container is ready to respond to requests.

Thus, the readiness endpoint must return a value indicating success _only after_ all extensions that contribute an API have started successfully. Other subsystems like the catalog crawlers might still not be completed at that point, but technically the connector is "ready to respond to requests", even if they might not produce the desired - or even a sensible - outcome. Catalog queries may produce an empty or incomplete result until crawlers have completed.

### A `startup` endpoint
From the [Kubernetes docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes): 
> Indicates whether the application within the container is started

This is very similar to the `readiness` endpoint, but a connector may be able to respond to requests as soon as all REST controllers have been registered (thus being `ready`), whereas the startup is only completed after all extensions have started. This can only be determined by the runtime. Again, parallel systems like crawlers will **not affect** system readiness.

### Implementation considerations
Some samples that can be found in the [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) uses a file (or rather: its absence or existence) as indication whether e.g. `liveness` is achieved. 
Another way is to expose REST Endpoints for each of the aforementioned probes, which makes it necessary to have `cURL` or similar tools available inside the docker Container. 

### Future extensibility
Other information and metrics would be cool too eg. transfer process throughput, average dwell-time of a TP  in a particular state, number of loaded extensions, statistics about catalog crawling, caught and/or uncaught exceptions etc. but lets focus on the first two in this issue and lets create and `advanced` module later.

paullatzelsperger:
This adds the base infrastructure and an API for a Docker `HEALTHCHECK` and Kubernetes readiness probes. Closes #206 

The functionality is split in two parts:
1. a `HealthCheckService` and associated classes in `spi` plus an impl for it in the `core:bootstrap` module.
2. an API to access the service plus documentation is located in `extensions:api:observability`. 

**Open questions:**
- So far, all cosmos-db-based datastores have been added as readiness providers, so they'll contribute to the overall readiness state. I used the `Container.read()` method as "heardbeat". Is that acceptable?
- What should we use as heartbeat signal for Azure Blobstore and AWS S3?
- should we provide a sample, like a Dockerfile and a K8S yaml file?
- if so, should that go into the `other` samples directory, or become ono of the numbered samples?

_Note: as #384 is on the verge of being merged, I already used Mockito in tests, so the local dependencies in `extensions:api:observability` and `core:bootstrap` need to be removed prior to merging this._


paullatzelsperger:
@algattik I cannot add you as a reviewer directly, but your thoughts would be appreciated :)

core/bootstrap/src/test/java/org/eclipse/dataspaceconnector/system/CoreServicesExtensionTest.java
```diff
@@ -71,6 +72,8 @@ void initialize() {
         expect(context.getService(Vault.class)).andReturn(niceMock(Vault.class)).anyTimes();
         expect(context.getService(eq(PrivateKeyResolver.class))).andReturn(niceMock(PrivateKeyResolver.class));
 
+        context.registerService(eq(ReadinessService.class), isA(ReadinessServiceImpl.class));
+        expectLastCall();
```
algattik, line 76: nit: use mockito? (#384)

launchers/basic/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR /app
+COPY ./build/libs/dataspaceconnector-basic.jar /app
+
+EXPOSE 8181
+
+# health status is determined by the availability of the /health endpoint
+HEALTHCHECK --interval=5s --timeout=5s --retries=10 CMD curl --fail -X GET http://localhost:8181/api/health || exit 1
+
+# by default curl is not available, so install it
+RUN apt update && apt install curl -y
```
algattik, line 13: move up to leverage cache

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/readiness/ReadinessServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.eclipse.dataspaceconnector.spi.system.readiness.LivenessProvider;
+import org.eclipse.dataspaceconnector.spi.system.readiness.ReadinessProvider;
+import org.eclipse.dataspaceconnector.spi.system.readiness.ReadinessService;
+import org.eclipse.dataspaceconnector.spi.system.readiness.StartupStatusProvider;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public class ReadinessServiceImpl implements ReadinessService {
```
algattik, line 11: rename to HealthCheckService? (also interface, package name)

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/readiness/ReadinessServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public void addStartupStatusProvider(StartupStatusProvider provider) {
+        startupStatusProviders.add(provider);
+    }
+
+    @Override
+    public boolean isLive() {
+        return livenessProviders.stream().allMatch(LivenessProvider::get);
```
algattik, line 39: Logging would be useful for failing providers. In consequence consider using a loop without short-circuit to ensure all failing providers are logged, not just the first one.

extensions/api/observability/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+From the [Kubernetes docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes):
+> indicates whether the container is running.
+
+Basically the same semantics as the `/health` endpoint. We can assume, that in the event that the connector crashes, all
+REST endpoints become unavailable.
+
+### `/check/readiness` endpoint
+
+From the [Kubernetes docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes):
```
algattik, line 19: consider adding a link to https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ (perhaps replacing the link provided)

extensions/api/observability/README.md
```diff
@@ -0,0 +1,35 @@
...
+> Indicates whether the container is ready to respond to requests.
+
+Thus, the readiness endpoint must return a value indicating success _only after_ all extensions that contribute an API
+have started successfully. Note, that other subsystems like the catalog crawlers might not yet be completed at that
+point, but technically the connector is "ready to respond to requests", even if they might not produce the desired - or
+even a sensible - outcome. Catalog queries may produce an empty or incomplete result until crawlers have completed.
+
+### `/check/startup` endpoint
```
algattik, line 27: I don't think a separate endpoint (and separate registration mechanism) is needed for startup. We can just use the standard health endpoint for startup. See example at https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

For example, in our PRS application we use this because of the time it takes for our Spring Boot app to start and Flyway to seed the database. See https://github.com/catenax/tractusx/blob/3565aeee25b34ebc23febb431f6cceddc41955fc/coreservices/partsrelationshipservice/cd/helm/prs/templates/deployment.yaml#L67

Flyway is just another health check in Spring Boot.

algattik (commented):
This is a great approach and we've used something similar in industrial dotnet projects with good results.

I would expect each extension to plug their own health checks. For example, Cosmos DB persistence would issue a test Cosmos DB query, Azure storage extension would check access to the storage account, etc.

The approach proposed here has scale limitations because those would be all done synchronously in sequence when a health check comes in, quickly leading to bottlenecks, timeout and instability. Nonetheless this is a great first step to establish the interfaces, and background threads can be established later.

Because extensions should provide health checks, perhaps this could be part of the base rather than an extension, as it might in practice always be a dependency.

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/readiness/ReadinessServiceImpl.java
algattik, line 39: Perhaps parallelism can be used here

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/readiness/ReadinessServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    private final List<StartupStatusProvider> startupStatusProviders;
+
+    public ReadinessServiceImpl() {
+        readinessProviders = new CopyOnWriteArrayList<>();
+        livenessProviders = new CopyOnWriteArrayList<>();
+        startupStatusProviders = new CopyOnWriteArrayList<>();
+    }
+
+    @Override
+    public void addLivenessProvider(LivenessProvider provider) {
```
algattik, line 23: consider reversing the pattern and have each extension implement a MetricsProvider interface which can provide liveness and readiness check, later custom metrics etc.

extensions/api/observability/src/main/java/org/eclipse/dataspaceconnector/api/observability/ObservabilityController.java
```diff
@@ -0,0 +1,70 @@
...
+    public Response checkHealth() {
+        return getLiveness();
+    }
+
+    @GET
+    @Path("liveness")
+    public Response getLiveness() {
+        return readinessService.isLive() ?
+                Response.ok().build() :
+                Response.status(404).build();
```
algattik, line 51: consider returning 503, as spring actuator does. 5xx also clearly indicates that the request can be retried.

core/bootstrap/src/test/java/org/eclipse/dataspaceconnector/system/CoreServicesExtensionTest.java
paullatzelsperger: there is #384, which should introduce mockito. not sure if its better to use the "old" tech here, or have two testing frameworks at a time

extensions/api/observability/README.md
paullatzelsperger: The site I linked has a good definition of the three K8S probes, plus it linkes to the link you provided anyways.

extensions/api/observability/src/main/java/org/eclipse/dataspaceconnector/api/observability/ObservabilityController.java
```diff
@@ -0,0 +1,70 @@
...
+    @Path("health")
+    public Response checkHealth() {
+        return getLiveness();
+    }
+
+    @GET
+    @Path("liveness")
+    public Response getLiveness() {
+        return readinessService.isLive() ?
+                Response.ok().build() :
```
algattik, line 50: consider returning a body, e.g. from Spring Actuator: `{"status":"UP","components":{"db":{"status":"UP","details":{"database":"PostgreSQL","validationQuery":"isValid()"}},"diskSpace":{"status":"UP","details":{"total":62725623808,"free":48520187904,"threshold":10485760,"exists":true}},"readinessState":{"status":"UP"}}}`

extensions/api/observability/README.md
paullatzelsperger: There may be a subtle difference between the health endpoint and the startup endpoint as they are defined now: while the health endpoint becomes available as soon as the `observability`-extension gets loaded, the system startup may not yet have completed. The startup endpoint should only return `success` once _all_ extensions have completed loading. 

Although not everyone may use the `startup` endpoint, there are certain use cases where it comes in handy, e.g. during testing: the sanity checks could leverage it

paullatzelsperger:
> I would expect each extension to plug their own health checks. For example, Cosmos DB persistence would issue a test Cosmos DB query, Azure storage extension would check access to the storage account, etc.

That's a good idea.

> The approach proposed here has scale limitations because those would be all done synchronously in sequence when a health check comes in, quickly leading to bottlenecks, timeout and instability. Nonetheless this is a great first step to establish the interfaces, and background threads can be established later.

Extensions can provide health information, which does not necessarily mean the _execute_ the aforementioned db-queries etc. synchronously on demand - an extension could monitor its health constantly, and simply return the status when their `get()` method is invoked.

> Because extensions should provide health checks, perhaps this could be part of the base rather than an extension, as it might in practice always be a dependency.

The readiness service itself is in the base (`:core:bootstrap`), but the REST API is not - some runtimes may prefer to use other APIs or mechanisms

extensions/api/observability/src/main/java/org/eclipse/dataspaceconnector/api/observability/ObservabilityController.java
paullatzelsperger: wouldn't that leak internal information?

extensions/api/observability/README.md
algattik: Yes we agree on this. My point is that, in every application I've seen, the health and the startup probe use the same endpoint. They're simply configured differently in Kubernetes, e.g. to kill the pod if unavailable for 1 minute but give it 5 minutes to be available when it starts up.

So while these may be two different *probes* in the pod configuration, it is actually one single *endpoint* in the application itself, i.e. are you up, ready and capable of serving requests.

algattik:
> Extensions can provide health information, which does not necessarily mean the _execute_ the aforementioned db-queries etc. synchronously on demand - an extension could monitor its health constantly, and simply return the status when their `get()` method is invoked.

That would work, but introduces overhead to add such a mechanism in various extensions. In the dotnet projects we delivered, we centralized the background thread that polls those metrics providers regularly. This makes extension development much simpler, and also allow to react (logging and alerting) immediately to health conditions, in a uniform manner, rather than waiting for an HTTP call to wake this.

Besides, you will surely want to extend the mechanism to provide detailed *metrics* in a pluggable manner, whereby such a periodic call makes sense too.

> The readiness service itself is in the base (`:core:bootstrap`), but the REST API is not - some runtimes may prefer to use other APIs or mechanisms

Looks good then :)

extensions/api/observability/src/main/java/org/eclipse/dataspaceconnector/api/observability/ObservabilityController.java
algattik: Health information should be posted on a separate port that is not accessible externally. With Jetty we can do this with `SelectChannelConnector`s.

paullatzelsperger:
> That would work, but introduces overhead to add such a mechanism in various extensions. In the dotnet projects we delivered, we centralized the background thread that polls those metrics providers regularly. This makes extension development much simpler, and also allow to react (logging and alerting) immediately to health conditions, in a uniform manner, rather than waiting for an HTTP call to wake this.

We may be able to kill two birds with one stone then. how about this: There could be a background worker periodically polling all `*provider`s and collecting their health status info into some temporary cache. The HTTP calls then simply query that cache to obtain readiness/liveness etc. instead of querying the `*provider`s upon request. This means the `ReadinessService` could simply implement the background worker thread.
So the mechanism is centralized at one place, easily changeable and we can still query it synchronously through HTTP. Plus we could proactively react to problems e.g. by logging and/or posting an event somewhere (although a observable/listener concept feels better suited for that). 

> Besides, you will surely want to extend the mechanism to provide detailed _metrics_ in a pluggable manner, whereby such a periodic call makes sense too.

Yup, that's planned but still a rather distant fantasy ATM :)

jimmarino:
> This is a great approach and we've used something similar in industrial dotnet projects with good results.
> 
> I would expect each extension to plug their own health checks. For example, Cosmos DB persistence would issue a test Cosmos DB query, Azure storage extension would check access to the storage account, etc.
> 
> The approach proposed here has scale limitations because those would be all done synchronously in sequence when a health check comes in, quickly leading to bottlenecks, timeout and instability. Nonetheless this is a great first step to establish the interfaces, and background threads can be established later.
> 
> Because extensions should provide health checks, perhaps this could be part of the base rather than an extension, as it might in practice always be a dependency.

+1 on base. This should go on that module after we merge #396

extensions/api/observability/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+### `/check/readiness` endpoint
+
+From the [Kubernetes docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes):
+> Indicates whether the container is ready to respond to requests.
+
+Thus, the readiness endpoint must return a value indicating success _only after_ all extensions that contribute an API
+have started successfully. Note, that other subsystems like the catalog crawlers might not yet be completed at that
+point, but technically the connector is "ready to respond to requests", even if they might not produce the desired - or
+even a sensible - outcome. Catalog queries may produce an empty or incomplete result until crawlers have completed.
```
marcgs, line 25: I do not fully understand why we would not wait for the container to return complete and correct results to mark it as ready? IMO the readiness endpoint should return success only when the container is ready to accept and properly process traffic (including catalog queries, etc...). Readiness probes are polled during the whole lifetime of containers, leading to Load Balancers not forwarding any traffic to them if they fail. Why would we forward traffic to containers that return incomplete results if there are others that do return correct results?

paullatzelsperger: I guess it depends what we regard as `ready to accept requests`. Why would we not accept other requests before the catalog cache is fully built? The catalog endpoint could theoretically still return a 503 until the cache is fully built.

I'd expect to be able to reach, say, my Plex server, even before its library is fully built.
Also, would triggering a new catalog crawl mean that the connector has become "unready", because its catalog may be out of date? 

I'd prefer reacting to an ongoing catalog crawl through a different mechanism. Haven't thought about it in detail, but to me, it does not contribute to readiness from an infrastructure/ops perspective

marcgs: Ok I see, assuming the catalog takes a very long time to load it might be better not to consider for infrastructure/ops readiness. But do we agree that `ready to accept requests` would include all components required for the asynchronous processing of TPs? For instance for sample 5 it would include both the Azure Storage Container as well as the S3 bucket?

paullatzelsperger: > But do we agree that ready to accept requests would include all components required for the asynchronous processing of TPs? For instance for sample 5 it would include both the Azure Storage Container as well as the S3 bucket?

yes i think so.

algattik: @marcgs do you mean the Azure Storage Account? On the Consumer side, the connector health provider could indeed check it is accessible, but does not know about any Storage Containers (which are managed in the Transfer lifecycle). On the Provider side, the health provider doesn't know about any Storage Accounts. (Correct me if I'm wrong)

paullatzelsperger: Assuming that the readiness of a connector is determined by it's own readiness plus the readiness of all its backend services, then every extension that deals with such a backend, should implement its own checks.

For Azure Storage, the readiness/availability could be determined by listing all Containers or getting the account info, same for S3 buckets. 

There may be backend systems where such a periodic heartbeat might not be applicable, e.g. serverless functions, which may only get provisioned on demand. Such backends should be assumed "live" until proven otherwise.

extensions/api/observability/README.md
paullatzelsperger: I was actually going to make `liveness` and `health` the same, but sticking to best practice also makes sense, so I will change it to `startup == health`.

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/readiness/ReadinessServiceImpl.java
paullatzelsperger: that is what has been done, although there are now `LivenessProviders`, `ReadinessProviders` and `StartupStatusProviders`. So every extension, like a CosmosDB-extension can contribute to the summary `readiness` state.

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/health/HealthCheckServiceImpl.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+
+    private void start(HealthCheckServiceConfiguration configuration) {
+        //todo: maybe providers should provide their desired timeout instead of a global config?
+        executor.scheduleAtFixedRate(this::queryReadiness, 0, configuration.getReadinessPeriod().toMillis(), TimeUnit.MILLISECONDS);
+        executor.scheduleAtFixedRate(this::queryLiveness, 0, configuration.getLivenessPeriod().toMillis(), TimeUnit.MILLISECONDS);
+        executor.scheduleAtFixedRate(this::queryStartupStatus, 0, configuration.getStartupStatusPeriod().toMillis(), TimeUnit.MILLISECONDS);
+    }
```
denisneuling, line 84: @paullatzelsperger Aren't those checks triggered from outside e.g. via a request handling thread / resp. why are these 3 extra threads required? Wouldn't caching results for a certain amount of time be sufficient?

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexExtension.java
```diff
@@ -54,6 +55,8 @@ public void initialize(ServiceExtensionContext context) {
         context.registerService(AssetLoader.class, assetIndex);
         context.registerService(DataAddressResolver.class, assetIndex);
 
+        context.getService(HealthCheckService.class).addReadinessProvider(() -> cosmosDbApi.get().forComponent(name()));
```
ndr-brt, line 58: I'm not that sure that a connection to an external data source should drive the readiness provider, it can lead to service fault and "connection refused" or "timeout" to the clients also for a temporary network failure, in my opinion the connector should handle those cases returning 500 and notifying the problem through the monitoring system.

ndr-brt (commented):
I have a big doubt about the readiness probe on datbase connections, for the rest all good for me

core/bootstrap/build.gradle.kts
```diff
@@ -22,6 +22,10 @@ plugins {
 dependencies {
     api(project(":spi"))
     api("org.slf4j:slf4j-api:${slf4jVersion}")
+    //todo: will be removed once PR #384 is merged
+    testImplementation("org.mockito:mockito-core:4.1.0")
+    testImplementation("org.awaitility:awaitility-kotlin:4.1.1")
```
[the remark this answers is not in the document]
paullatzelsperger: no reason, I'm sure that got in through IntelliJ somehow

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/health/HealthCheckServiceImpl.java
paullatzelsperger: So the reasoning for this was that the probe requests coming in from K8S or docker will only hit the cached results and ttherefore be very fast, at the cost of potentially returning "stale" status. 

The service then periodically performs it's health checks separately, therefore decoupling the two mechanisms and avoiding system clogging: imagine we have a lot of backend systems, and every `readiness` probe triggers a heart beat to every one of them - that would certainly cause CPU load spikes or let the probes run into a timeout.

The reason why it's 3 threads as opposed to one is, that we might want to schedule them at different intervals.

core/bootstrap/build.gradle.kts
```diff
@@ -22,6 +22,10 @@ plugins {
 dependencies {
     api(project(":spi"))
     api("org.slf4j:slf4j-api:${slf4jVersion}")
+    //todo: will be removed once PR #384 is merged
+    testImplementation("org.mockito:mockito-core:4.1.0")
```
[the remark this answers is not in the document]
paullatzelsperger: > I have a big doubt about the readiness probe on datbase connections, for the rest all good for me

I think they aren't a problem, as the heartbeat runs in a background thread, so the probe should be very quick.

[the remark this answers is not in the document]
ndr-brt: The problem for me is not about the speed of the probe, but what can bring this in an environment like k8s.
I explained my concerns in detail in the comment on the `CosmosAssetIntexExtension` ffile

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/health/HealthCheckServiceImpl.java
denisneuling: Okay.

What happens if a exception is thrown by any of the providers? - The spawning thread then dies and never gets restarted, right? Means either stale information is served fromt that point on or - if never run - the check never gets taken into consideration at all. Also `Stream#parallelStream` spawns another thread from the default `ForkJoinPool` for each element in the stream - seems a little bit wasteful. If you still want to keep the parallel stream execution please make it configurable which `ForkJoinPool` to use.
Btw, when latency of fulfilling a certain check exceeds the configured fixed delay there might be overlapping runs of that specific check.

paullatzelsperger: > What happens if a exception is thrown by any of the providers? - The spawning thread then dies and never gets restarted, right? Means either stale information is served fromt that point on - if ever run - or the check never gets taken into consideration at all.

good point. added exception handling.

> Also Stream#parallelStream spawns another thread from the default ForkJoinPool for each element in the stream - seems a little bit wasteful. If you still want to keep the parallel stream execution please make it configurable which ForkJoinPool to use.

The `threadPoolSize` of the `ScheduledExecutorService` is configurable, and the `parallelStreams` run within it's context, so I think we're good there. 

> Btw, when latency of fulfilling a certain check exceeds the configured fixed delay there might be overlapping runs of that specific check.

Yup, if heartbeats take longer than the interval configured in code, that can potentially create an ever-growing backlog of tasks. The alternative would be to keep track of running checks and only re-submit the check, if it's not already running. That will add more complexity and will still not prevent configuration errors, it would make them loggable though. Let me see if I can come up with a usable solution for that.

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexExtension.java
paullatzelsperger: In a K8S environment, a failing `readiness` probe caused by transient errors will make the loadbalancer stop routing traffic to that pod, so this is indeed the intended behaviour, since the pod is not "ready to respond to requests" anymore.

However, if the application fails to start, e.g. due to a config error, that would be caught by the `startup` probe, which disables all other probes until it succeeds.

Monitoring and K8s probes have very different use cases.

core/bootstrap/src/main/java/org/eclipse/dataspaceconnector/system/health/HealthCheckServiceImpl.java
paullatzelsperger: Just to amend the parallelism topic: I did some testing, and from a runtime perspective it does not make a difference whether `new ForkJoinPool(5)` or a `threadPoolSize(5)` is used. Also, ForkJoinPools attempt to re-use internal worker threads, so it's not wasteful. 
Since health providers potentially do stuff that may take some time, having them checked in multiple threads in parallel is advisable here.

extensions/azure/assetindex-cosmos/src/main/java/org/eclipse/dataspaceconnector/assetindex/azure/CosmosAssetIndexExtension.java
algattik: Indeed, and that is critical for resiliency (backpressure)

algattik:
Sorry, I lost track of our thread about the need for a startup probe. My opinion is that separate code is not useful for startup vs health checks. A kubernetes startup probe is just a configuration for probing the health endpoint with a longer timeout. By definition, application startup is the time it takes for it to be healthy.

paullatzelsperger:
@algattik as of recently, the health check and the startup probe execute the exact same code. I've made them available on two different endpoints simply for convenience reasons.

algattik:
> @algattik as of recently, the health check and the startup probe execute the exact same code. I've made them available on two different endpoints simply for convenience reasons.

Thank you. Why is that a convenience? As a developer I find it more convenient to have a single endpoint for both, so that my pod configurations are clear to that regard.

paullatzelsperger:
> Thank you. Why is that a convenience?
The `health` endpoint was intended with Docker in mind, rather than K8s, where the wording is "health" rather than "startup" etc.

jimmarino:
@paullatzelsperger where are we on this? Are there outstanding issues still?

paullatzelsperger:
> @paullatzelsperger where are we on this? Are there outstanding issues still?

There is a Pull Request (#412) which can be merged IMO.

ndr-brt (approved):

PR #413, opened 2021-12-14
Proposal: Directory Restructuring

DominikPinsel:
Hello everyone,

this is a proposal for the directory restructuring. I did most of it last week, so it's not 100% aligned with the main branch. Also there are probably still some minor mistakes in it, so don't expect you will be able to run everything after executing the script. But the general idea, where all the folders would be moved to, should be clear.

Please have a look at the proposed directory structure and tell me what you think. You can also always copy the script into your project root directory and run it yourself to see the changes.

### DIFF
--- a/reordering.sh
+++ b/reordering.sh
@@ -1,0 +1,367 @@
+#!/bin/bash
+./gradlew clean
+
+function move() {
+  local SOURCE="${1}"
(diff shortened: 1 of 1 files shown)


ndr-brt:
Is this still a thing? It could be closed now I suppose.

Issue #330, opened 2021-12-01
Add CommandQueue to TransferProcessManager


paullatzelsperger:
## Problem statement
The `TransferProcessManagerImpl` (TPM) uses several methods to check and advance the state of transfer processes. Those "check-methods" are not transactional, so another process/thread could intercede and perform another modification, which would get overwritten by the TPM, e.g.:
```java
private int checkCompleted(){
  var processesInProgress = transferProcessStore.nextForState(TransferProcessStates.IN_PROGRESS.code(), batchSize);
  for(var process : processesInProgress){
    // modify processes in a loop, advance state, etc
    // Problem when "process" gets modified by another thread before the call to update()!
    transferProcessStore.update(process);
  }
```
The fundamental problem is that the read and write calls (here: `nextForState()` and `update()`) are not performed within the same transaction, which was not a problem until now, since the TPM is single-threaded. However, as soon as a second thread comes into play, e.g. a watchdog, or a cleanup job, we're running into concurrency problems.

## Proposed solution
I propose the implementation of a command queue, which gets drained at specific points in the code. Any modifications to the TP's must be submitted through the command queue.
A command would basically be a `Function<TransferProcessStore, TransferProcess>`, that receives the TPS as parameter, loads the TP and does whatever  modification it needs to do, and returns the TP back out again. 

There is a few important considerations about this:
- The TPM must pass the TPS to the command so that it can load the TP
- The command cannot maintain a reference to the TP, because that might become "stale"
- The command should not persist the TP, rather it should return the TP and let the TPM manage persisting.

All put together this could look something like this:
```java
//TransferProcessManagerImpl.java, line 111
 try {
   drainCommandQueue();
   int provisioning = provisionInitialProcesses();
   drainCommandQueue();
   int sent = sendOrProcessProvisionedRequests();
   drainCommandQueue();
   //...
   if (provisioning + provisioned + sent + finished + deprovisioning + deprovisioned == 0) {
     Thread.sleep(waitStrategy.waitForMillis());
   }
   waitStrategy.success();
  } //catch...
```

### Modifying the `TransferProcessManager`
An additional method will be added to the `TransferProcessManager` allowing the submission of the aforementioned commands.

## Resilience considerations
[...]

PR #421, opened 2021-12-14
Isolate TransferProcessStore usage into TransferProcessManager

ndr-brt:
Part of #330

This PR bring all the code that calls methods that change a transfer process (`create`, `update`, `delete`) inside `TransferProcessManager`, that will help introducing the command queue.

I have two questions, both for @jimmarino and @paullatzelsperger 
1. The process of handling post-provision/deprovision actions using `ProvisionContext` is pretty hard to understand (it took me a while), I would suggest to change that implementation, I have two proposal:
  * Make `Provisoner`'s `provision` and `deprovision` and `ProvisionManager`'s `provision` and `deprovision` return a `CompletableFuture` containing the result, doing so, the definition of what should be done after the provisioning/deprovisioning will be set right after the method call, more easy to read. (preferred solution for me)
  * Make `Provisoner`'s `provision` and `deprovision` and `ProvisionManager`'s `provision` and `deprovision` accept the callback  as parameter.
  
2. On `ProvisionContext` there are 3 methods unused: `create`, `update` and `find`, can they be removed or will be needed (I would remove them anyway since they are not used at the moment)

### DIFF
--- a/spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/TransferProcessManager.java
+++ b/spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/TransferProcessManager.java
@@ -14,7 +14,9 @@
 
 package org.eclipse.dataspaceconnector.spi.transfer;
 
+import org.eclipse.dataspaceconnector.spi.result.Result;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataRequest;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcessStates;
 
 /**
  * Manages data transfer processes. Currently synchronous and asynchronous data transfers are supported.
@@ -38,4 +40,6 @@ public interface TransferProcessManager {
     void transitionProvisioned(String processId);
 
     void transitionError(String processId, String detail);
+
+    Result<TransferProcessStates> deprovision(String processId);
 }
--- a/spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/TransferProcessManager.java
+++ b/spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/TransferProcessManager.java
@@ -35,11 +35,15 @@ public interface TransferProcessManager {
      */
     TransferInitiateResult initiateProviderRequest(DataRequest dataRequest);
 
+    // TODO: will be substituted by the upcoming command queue introduction
     void transitionRequestAck(String processId);
 
+    // TODO: will be substituted by the upcoming command queue introduction
     void transitionProvisioned(String processId);
 
+    // TODO: will be substituted by the upcoming command queue introduction
     void transitionError(String processId, String detail);
 
+    // TODO: will be substituted by the upcoming command queue introduction
     Result<TransferProcessStates> deprovision(String processId);
 }
--- a/spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/TransferProcessManager.java
+++ b/spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/TransferProcessManager.java
@@ -33,4 +33,9 @@ public interface TransferProcessManager {
      */
     TransferInitiateResult initiateProviderRequest(DataRequest dataRequest);
 
+    void transitionRequestAck(String processId);
+
+    void transitionProvisioned(String processId);
+
+    void transitionError(String processId, String detail);
 }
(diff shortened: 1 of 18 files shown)


ndr-brt:
> Regarding item 1, we used a traditional callback mechanism (some actor models follow this pattern) since an asynchronous system, you don't want to always require the client to hold a reference (in this case the future) to get a result. For example, it could be 10, 20, 30 minutes, two days or more before a result comes back. The result could also be after a system crash and restart. Does that make sense?

I think that on the functional side the approaches are pretty similar.
With a Future, you don't have to hold a reference of it if you don't need it, you will just postpone the evaluation of the result of the computation in the future.
[In the `S3ProvisionPipeline` is already implemented this way](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/e9f13a3a10aaa5a91adde99a2a988dc07cdcb571/extensions/aws/s3/provision/src/main/java/org/eclipse/dataspaceconnector/provision/aws/s3/S3ProvisionPipeline.java#L98)
At the end of the `createAndSendToken` method the callback is evaluated, but it would be the same to return that Future to the caller of the provisioner and let it add a `.whenComplete(doTheAction)`.
Doing like that the consumer will not have to wait for the async computation, but it will be more readable.

I think we should always prefer Futures to Callbacks as a good practice.

jimmarino:
> > Regarding item 1, we used a traditional callback mechanism (some actor models follow this pattern) since an asynchronous system, you don't want to always require the client to hold a reference (in this case the future) to get a result. For example, it could be 10, 20, 30 minutes, two days or more before a result comes back. The result could also be after a system crash and restart. Does that make sense?
> 
> I think that on the functional side the approaches are pretty similar. With a Future, you don't have to hold a reference of it if you don't need it, you will just postpone the evaluation of the result of the computation in the future. [In the `S3ProvisionPipeline` is already implemented this way](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/e9f13a3a10aaa5a91adde99a2a988dc07cdcb571/extensions/aws/s3/provision/src/main/java/org/eclipse/dataspaceconnector/provision/aws/s3/S3ProvisionPipeline.java#L98) At the end of the `createAndSendToken` method the callback is evaluated, but it would be the same to return that Future to the caller of the provisioner and let it add a `.whenComplete(doTheAction)`. Doing like that the consumer will not have to wait for the async computation, but it will be more readable.
> 
> I think we should always prefer Futures to Callbacks as a good practice.
> 
> EDIT: I forgot to say that Futures are also easier to test, and the exception handling is also better

Ah, ok so you want to use .whenComplete. Now I understand. The return type will then need to be something that models the two different ProvisionContext.provision() parameters since those two cases need to be distinguished. We could try and see what that looks like. Consider that there may be multiple provisioners engaged (fanout) so the completion method needs to operate over a collection of futures. Let's see if that is simpler.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/TransferProcessManager.java
```diff
@@ -33,4 +35,11 @@
      */
     TransferInitiateResult initiateProviderRequest(DataRequest dataRequest);
 
+    void transitionRequestAck(String processId);
```
paullatzelsperger, line 38: We will likely substitute this with commands from the CommandQueue later, so can we add a comment indicating that here?

paullatzelsperger (approved):
LGTM. In preparation for the CommandQueue feature we may also think about internalising the TransferProcessStore, so that select objects like the TPM have access to it.

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/TransferProcessManager.java
ndr-brt: Sure. 
I was thinking about the api that the `TransferProcessStore` interface will have in future, do you think a single `execute(command)` method will be enough or better to have a method for each state change require?
Choosing the first option will mean that every implementation of `TransferProcessStore` would need to know about the commands, for me is ok, but is something we have to keep in mind

jimmarino:
@ndr-brt I am good with this, can you look into the integration test issue (likely unrelated)? Also, should we look to merge the Mockito PR first?

jimmarino (approved):

ndr-brt:
@jimmarino , yes, maybe the mockito one is more important (especially for the rebasing effort). The IT will be fixed by #441

ndr-brt:
@jimmarino ready for merge

PR #439, opened 2021-12-16
Provide library to support postgresql as data retention site (closes #438)

### LINKED ISSUE #438: Lib - Postgresql - Provide library to support postgresql as data retention site
Postgresql is one of the most widely used relational database management systems, so it's a natural need to provide support using it within the EDC.

denisneuling:
This PR ships a postgresql client library able to be used for creating e.g. a postgresql backed asset-index.

### DIFF
--- a/common/libraries/clients/postgresql/README.md
+++ b/common/libraries/clients/postgresql/README.md
@@ -1,0 +1,23 @@
+# PostgreSQL Clients
+
+This library provides a connection pooling client to perform SQL queries towards PostgreSQL databases. The
+implementation prevents from SQL injection due to exclusive usage of `java.sql.PreparedStatement`s. It yields an
+interface for transactional SQL queries and automatic transaction rollback while treating exceptions.
+
+## Usage
+
+```java
+    ConnectionFactoryConfig connectionFactoryConfig = ConnectionFactoryConfig.Builder.newInstance()
+        .uri(URI.create("jdbc://localhost:5432/mydb"))
+        .userName("username")
+        .password("password")
+        .build();
+    ConnectionFactory connectionFactory = new ConnectionFactoryImpl(connectionFactoryConfig);
+    CommonsConnectionPoolConfig commonsConnectionPoolConfig = CommonsConnectionPoolConfig.Builder.newInstance().build();
+    ConnectionPool connectionPool = new CommonsConnectionPool(connectionFactory, commonsConnectionPoolConfig);
+    PostgresqlClient postgresqlClient = new PostgresqlClientImpl(connectionPool);
+
+    List<Integer> i = postgresqlClient.execute((r) -> r.getInt(1), "SELECT 1;");
+
+    postgresqlClient.close(); // closes all connections managed by the underlying pool
+```
--- a/common/libraries/clients/postgresql/build.gradle.kts
+++ b/common/libraries/clients/postgresql/build.gradle.kts
@@ -1,0 +1,42 @@
+/*
+ *  Copyright (c) 2021 Daimler TSS GmbH
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Daimler TSS GmbH - initial build file
+ *
+ */
+
+plugins {
+    `java-library`
+    `java-test-fixtures`
+    `maven-publish`
+}
+
+val jupiterVersion: String by project
+val postgresqlVersion: String by project
+val apacheCommonsPool2Version: String by project
+val testContainersVersion: String by project
+
+dependencies {
+    implementation("org.postgresql:postgresql:${postgresqlVersion}")
+    implementation("org.apache.commons:commons-pool2:${apacheCommonsPool2Version}")
+
+    testImplementation("ch.qos.logback:logback-classic:1.2.6")
+    testImplementation("org.testcontainers:postgresql:${testContainersVersion}")
+    testImplementation("org.testcontainers:junit-jupiter:${testContainersVersion}")
+}
+
+publishing {
+    publications {
+        create<MavenPublication>("common-clients-postgresql") {
+            artifactId = "common-clients-postgresql"
+            from(components["java"])
+        }
+    }
+}
--- a/common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/PostgresqlClient.java
+++ b/common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/PostgresqlClient.java
@@ -1,0 +1,79 @@
+/*
+ *  Copyright (c) 2021 Daimler TSS GmbH
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Daimler TSS GmbH - Initial API and Implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql;
+
+import org.eclipse.dataspaceconnector.clients.postgresql.row.RowMapper;
+
+import java.sql.SQLException;
+import java.util.List;
+
+/**
+ * Client capable of issuing SQL queries against postgresql databases
+ */
+public interface PostgresqlClient extends AutoCloseable {
+
+    /**
+     * Intended for mutating queries.
+     *
+     * @param sql       the parametrized sql query
+     * @param arguments the parameteres to interpolate with the parametrized sql query
+     * @return rowsChanged
+     * @throws SQLException if execution of the query was failing
+     */
+    int execute(String sql, Object... arguments) throws SQLException;
+
+    /**
+     * Intended for reading queries.
+     *
+     * @param rowMapper able to map a row to an object e.g. pojo.
+     * @param sql       the parametrized sql query
+     * @param arguments the parameteres to interpolate with the parametrized sql query
+     * @param <T>       generic type returned after mapping from the executed query
+     * @return results
+     * @throws SQLException if execution of the query or mapping was failing
+     */
+    <T> List<T> execute(RowMapper<T> rowMapper, String sql, Object... arguments) throws SQLException;
+
+    /**
+     * Spans a transaction around the passed callback.
+     *
+     * @param callback transaction scoped callback
+     * @throws SQLException if the transaction was failing
+     */
+    void doInTransaction(TransactionCallback callback) throws SQLException;
+
+    /**
+     * Closes the postgres client and frees all underlying resources.
+     *
+     * @throws Exception in case an error was encountered while closing the postgres client
+     */
+    default void close() throws Exception {
+    }
+
+    /**
+     * A transaction scoped callback
+     */
+    @FunctionalInterface
+    interface TransactionCallback {
+
+        /**
+         * Callback.
+         *
+         * @param postgresqlClient the transaction scoped postgres client
+         * @throws SQLException if execution of any operations encountered an error
+         */
+        void accept(PostgresqlClient postgresqlClient) throws SQLException;
+    }
+}
--- a/common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/connection/ConnectionFactoryImpl.java
+++ b/common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/connection/ConnectionFactoryImpl.java
@@ -1,0 +1,85 @@
+/*
+ *  Copyright (c) 2021 Daimler TSS GmbH
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Daimler TSS GmbH - Initial API and Implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.connection;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.Objects;
+import java.util.Properties;
+
+/**
+ * A ConnectionFactory implementation that utilizes the ConnectionFactoryConfig for connection creation.
+ */
+public class ConnectionFactoryImpl implements ConnectionFactory {
+    private final String jdbcUrl;
+    private final Properties properties;
+    private final boolean autoCommit;
+
+    public ConnectionFactoryImpl(@NotNull ConnectionFactoryConfig connectionFactoryConfig) {
+        Objects.requireNonNull(connectionFactoryConfig, "connectionFactoryConfig");
+
+        this.jdbcUrl = Objects.requireNonNull(connectionFactoryConfig.getUri()).toString();
+        this.properties = getProperties(connectionFactoryConfig);
+        this.autoCommit = connectionFactoryConfig.getAutoCommit();
+    }
+
+    /**
+     * Constructs a map containing postgreSQL jdbc connection parameters to obtain a connection.
+     *
+     * @param connectionFactoryConfig container object containing set of connection configuration parameters
+     * @return properties to be used by the {@link java.sql.DriverManager}
+     * @see <a href="https://jdbc.postgresql.org/documentation/head/connect.html">PostgreSQL JDBC Driver - Connecting to the Database</a>
+     */
+    private Properties getProperties(ConnectionFactoryConfig connectionFactoryConfig) {
+        Properties properties = new Properties();
+        addProperty(properties, "user", connectionFactoryConfig.getUserName());
+        addProperty(properties, "password", connectionFactoryConfig.getPassword());
+        addProperty(properties, "ssl", connectionFactoryConfig.getSsl());
+        addProperty(properties, "sslmode", connectionFactoryConfig.getSslMode());
+        addProperty(properties, "sslcert", connectionFactoryConfig.getSslCert());
+        addProperty(properties, "sslkey", connectionFactoryConfig.getSslKey());
+        addProperty(properties, "sslrootcert", connectionFactoryConfig.getSslRootCert());
+        addProperty(properties, "sslhostnameverifier", connectionFactoryConfig.getSslHostNameVerifier());
+        addProperty(properties, "loggerLevel", connectionFactoryConfig.getLoggerLevel());
+        addProperty(properties, "loggerFile", connectionFactoryConfig.getLoggerFile());
+        addProperty(properties, "logUnclosedConnections", connectionFactoryConfig.getLogUnclosedConnection());
+        addProperty(properties, "connectTimeout", connectionFactoryConfig.getConnectTimeout());
+        addProperty(properties, "socketTimeout", connectionFactoryConfig.getSocketTimeout());
+        addProperty(properties, "loginTimeout", connectionFactoryConfig.getLoginTimeout());
+        addProperty(properties, "readOnly", connectionFactoryConfig.getReadOnly());
+        addProperty(properties, "ApplicationName", connectionFactoryConfig.getApplicationName());
+        return properties;
+    }
+
+    private void addProperty(Properties properties, String key, Object value) {
+        if (properties != null && key != null && value != null) {
+            properties.put(key, String.valueOf(value));
+        }
+    }
+
+    @Override
+    public Connection create() throws SQLException {
+        Connection connection = DriverManager.getConnection(jdbcUrl, properties);
+
+        connection.setAutoCommit(autoCommit);
+        // TODO transaction isolation made configurable
+        connection.setTransactionIsolation(Connection.TRANSACTION_READ_UNCOMMITTED);
+
+        return connection;
+    }
+}
--- a/common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/PostgresqlClientImpl.java
+++ b/common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/PostgresqlClientImpl.java
@@ -1,0 +1,212 @@
+/*
+ *  Copyright (c) 2021 Daimler TSS GmbH
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Daimler TSS GmbH - Initial API and Implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql;
+
+import org.eclipse.dataspaceconnector.clients.postgresql.connection.pool.ConnectionPool;
+import org.eclipse.dataspaceconnector.clients.postgresql.row.RowMapper;
+import org.eclipse.dataspaceconnector.clients.postgresql.statement.ArgumentHandler;
+import org.eclipse.dataspaceconnector.clients.postgresql.statement.ArgumentHandlers;
+import org.jetbrains.annotations.NotNull;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * An implementation of a {@link org.eclipse.dataspaceconnector.clients.postgresql.PostgresqlClient}
+ * backed by an {@link org.eclipse.dataspaceconnector.clients.postgresql.connection.pool.ConnectionPool}
+ */
+public class PostgresqlClientImpl implements PostgresqlClient {
+    private final ConnectionPool connectionPool;
+
+    /**
+     * Constructor for instantiating the PostgresqlClientImpl
+     *
+     * @param connectionPool mandatory for obtaining connections
+     */
+    public PostgresqlClientImpl(ConnectionPool connectionPool) {
+        Objects.requireNonNull(connectionPool, "connectionFactory");
+
+        this.connectionPool = connectionPool;
+    }
+
+    @Override
+    public int execute(String sql, Object... arguments) throws SQLException {
+        Objects.requireNonNull(sql, "sql");
+        Objects.requireNonNull(arguments, "arguments");
+
+        return connectionPool.doWithConnection(
+                connection -> new SingleConnectionScopedPostgresqlClient(connection).execute(sql, arguments));
+    }
+
+    @Override
+    public <T> List<T> execute(RowMapper<T> rowMapper, String sql, Object... arguments) throws SQLException {
+        Objects.requireNonNull(sql, "rowMapper");
+        Objects.requireNonNull(sql, "sql");
+        Objects.requireNonNull(arguments, "arguments");
+
+        return connectionPool.doWithConnection(
+                connection -> new SingleConnectionScopedPostgresqlClient(connection).execute(rowMapper, sql, arguments));
+    }
+
+    @Override
+    public void doInTransaction(TransactionCallback callback) throws SQLException {
+        Objects.requireNonNull(callback, "callback");
+
+        connectionPool.doWithConnection(connection -> {
+            doInTransaction(connection, callback);
+            return null;
+        });
+    }
+
+    public void close() throws Exception {
+        connectionPool.close();
+    }
+
+    private static void doInTransaction(Connection connection, TransactionCallback transactionCallback) throws SQLException {
+        new Transaction(connection, transactionCallback).execute();
+    }
+
+    private static class SingleConnectionScopedPostgresqlClient implements PostgresqlClient {
+        private final Connection connection;
+
+        public SingleConnectionScopedPostgresqlClient(@NotNull Connection connection) {
+            this.connection = Objects.requireNonNull(connection, "connection");
+        }
+
+        @Override
+        public int execute(String sql, Object... arguments) throws SQLException {
+            Objects.requireNonNull(sql, "sql");
+            Objects.requireNonNull(arguments, "arguments");
+
+            try (PreparedStatement statement = connection.prepareStatement(sql, PreparedStatement.RETURN_GENERATED_KEYS)) {
+
+                setArguments(statement, arguments);
+
+                return statement.execute() ? 0 : statement.getUpdateCount();
+            }
+        }
+
+        @Override
+        public <T> List<T> execute(RowMapper<T> rowMapper, String sql, Object... arguments) throws SQLException {
+            Objects.requireNonNull(rowMapper, "rowMapper");
+            Objects.requireNonNull(sql, "sql");
+            Objects.requireNonNull(arguments, "arguments");
+
+            try (PreparedStatement statement = connection.prepareStatement(sql, PreparedStatement.RETURN_GENERATED_KEYS)) {
+
+                setArguments(statement, arguments);
+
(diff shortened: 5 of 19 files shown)


common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/PostgresqlClient.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Intended for mutating queries.
+     *
+     * @param sql       the parametrized sql query
+     * @param arguments the parameteres to interpolate with the parametrized sql query
+     * @return rowsChanged
+     * @throws SQLException if execution of the query was failing
+     */
+    int execute(String sql, Object... arguments) throws SQLException;
```
ndr-brt, line 32: Personally I'm not much into checked exception handling, I would prefer returning an object that encapsulate the result or error.

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/PostgresqlClient.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    <T> List<T> execute(RowMapper<T> rowMapper, String sql, Object... arguments) throws SQLException;
+
+    /**
+     * Spans a transaction around the passed callback.
+     *
+     * @param callback transaction scoped callback
+     * @throws SQLException if the transaction was failing
+     */
+    void doInTransaction(TransactionCallback callback) throws SQLException;
```
ndr-brt, line 52: I'm not into callbacks too, I think they made the code harder to read/test.
The alternative could be to have a method (like `transaction`, or `begin`) that returns a TransactionContext that would take care of handling the transaction.

common/libraries/clients/postgresql/src/test/java/org/eclipse/dataspaceconnector/clients/postgresql/PooledPostgresqlClientTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.eclipse.dataspaceconnector.clients.postgresql.connection.ConnectionFactoryConfig;
+import org.eclipse.dataspaceconnector.clients.postgresql.connection.ConnectionFactoryImpl;
+import org.eclipse.dataspaceconnector.clients.postgresql.connection.pool.ConnectionPool;
+import org.eclipse.dataspaceconnector.clients.postgresql.connection.pool.commons.CommonsConnectionPool;
+import org.eclipse.dataspaceconnector.clients.postgresql.connection.pool.commons.CommonsConnectionPoolConfig;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+
+class PooledPostgresqlClientTest extends AbstractPostgresqlClientTest {
+    private PostgresqlClient postgresqlClient;
```
ndr-brt, line 27: the client could stay in the abstraction, being instantiated by an abstract method (that would do the same things of the `setUp`, this will avoid the duplication of the teardown and the need for that `getPostgresqlClient`.

data-protocols/ids/README.md
```diff
@@ -34,4 +34,4 @@ These extensions bridge between EDC core and IDS.
 #### Description Request Message
 
 - As not-negotiated ContractOffers are not persisted, it is not possible to send IDS DescriptionRequestMessages for the
-  corresponding IDS Resources, where they are mapped to.
+  corresponding IDS org.eclipse.dataspaceconnector.clients.postgresql.asset.Resources, where they are mapped to.
```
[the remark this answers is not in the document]
denisneuling: Already seen and resolved :)

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/SimplePostgresqlClient.java
```diff
@@ -0,0 +1,173 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Stream;
+
+public class SimplePostgresqlClient implements PostgresqlClient {
```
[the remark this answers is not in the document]
denisneuling: I partially agree. Since the `pooled` implementation relies on the `simple` one (due to scoping a transaction to a specific connection), we cannot remove it completely. But the `SimplePostgresqlClient` could be made package private and the `PooledPostgresqlClient` be ranmed to e.g. `PostgresqlClientImpl`.

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/PostgresqlClient.java
denisneuling: Nah, won't introduce an error wrapping return object, since `SQLException` are an integral part when talking to SQL database. Also IMO it would make less readable.

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/PostgresqlClient.java
denisneuling: The problem with your proposal is, that one could easily leak a connection if not properly returned to the pool, which then means the connection is blocked for all other accessors - So think about it as it would be a security net. 
One isn't forced to use the PostgresqlClient implementation - it would be able to obtain a `Connection` from a `ConnectionFactory`/`ConnectionPool` on its own and roll his own transaction handling.

common/libraries/clients/postgresql/src/test/java/org/eclipse/dataspaceconnector/clients/postgresql/PooledPostgresqlClientTest.java
denisneuling: Once the `SimplePostgresClient` is made package private this will be refactored - so the abstract test class will go away anyway :)

common/libraries/clients/postgresql/README.md
```diff
@@ -0,0 +1,43 @@
+# commons-clients-postgresql
```
alexandrudanciu, line 1:
```suggestion
# PostgreSQL Clients
```

common/libraries/clients/postgresql/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+This library provides two Postgresql clients `SimplePostgresqlClient` and `PooledPostgresqlClient`.
```
alexandrudanciu, line 3:
```suggestion
This library provides the two Postgresql clients `SimplePostgresqlClient` and `PooledPostgresqlClient`. 
```

common/libraries/clients/postgresql/README.md
```diff
@@ -0,0 +1,43 @@
+# commons-clients-postgresql
+
+This library provides two Postgresql clients `SimplePostgresqlClient` and `PooledPostgresqlClient`. 
+Both implementation provide a unified interface to issue SQL queries towards Postgresql databases.
```
alexandrudanciu, line 4:
```suggestion
Both implementations provide a unified interface for issuing SQL queries towards PostgreSQL databases.
```

common/libraries/clients/postgresql/README.md
```diff
@@ -0,0 +1,43 @@
...
+Both prevent from SQL injection due to exclusive usage of `java.sql.PreparedStatement`s and yield an interface for transactional SQL queries and automatic transaction rollback while treating exceptions.
+Additionally, the `PooledPostgresqlClient` performs connection pooling, which yields to better performance when the system is under load.
+Connection validation can be configured to identify ill connections.
```
alexandrudanciu, line 7:
```suggestion
Connection validation can be configured to identify broken connections. 
```

common/libraries/clients/postgresql/build.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+    `maven-publish`
+}
+
+val jupiterVersion: String by project
+val postgresqlVersion: String by project
+val apacheCommonsPool2Version: String by project
+val testContainersVersion: String by project
+
+dependencies {
+    implementation("org.postgresql:postgresql:${postgresqlVersion}")
```
alexandrudanciu, line 27: This library is considered as "approved" by the Eclipse Dash Licensing tool, cf. the output:

maven/mavencentral/org.postgresql/postgresql/42.3.1, BSD-2-Clause, approved, clearlydefined

common/libraries/clients/postgresql/build.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+    implementation("org.apache.commons:commons-pool2:${apacheCommonsPool2Version}")
+
+    testImplementation("ch.qos.logback:logback-classic:1.2.6")
+    testImplementation("org.testcontainers:postgresql:${testContainersVersion}")
```
alexandrudanciu, line 31: This library is considered as "approved" by the Eclipse Dash Licensing tool, cf. the output:

maven/mavencentral/org.testcontainers/postgresql/1.16.2, MIT, approved, clearlydefined

common/libraries/clients/postgresql/build.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+    testImplementation("org.testcontainers:junit-jupiter:${testContainersVersion}")
+}
+
+publishing {
+    publications {
+        create<MavenPublication>("common-clients-postgresql") {
+            artifactId = "common-clients-postgresql"
```
alexandrudanciu, line 38: Our current guideline for naming Maven artifacts is:
Artifact id describing the module name (disregarding the directory structure) separating terms by a dash

Therefore, I would propose to remove the terms "common" and and maybe also "clients".
My suggestion is: postgresql-clients

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/PostgresqlClient.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql;
+
+import org.eclipse.dataspaceconnector.clients.postgresql.row.RowMapper;
+
+import java.sql.SQLException;
+import java.util.List;
+
+public interface PostgresqlClient extends AutoCloseable {
```
alexandrudanciu, line 22: As this interface might be reused across multiple modules, a Java Doc on class level could be useful.

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/SimplePostgresqlClient.java
```diff
@@ -0,0 +1,177 @@
...
```
alexandrudanciu, line 47: As this client might be reused across multiple modules, a Java Doc on class level could be useful.

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/connection/ConnectionFactoryImpl.java
```diff
@@ -0,0 +1,78 @@
...
+        Objects.requireNonNull(connectionFactoryConfig, "connectionFactoryConfig");
+
+        this.jdbcUrl = Objects.requireNonNull(connectionFactoryConfig.getUri()).toString();
+        this.properties = getProperties(connectionFactoryConfig);
+        this.autoCommit = connectionFactoryConfig.getAutoCommit();
+    }
+
+    private Properties getProperties(ConnectionFactoryConfig connectionFactoryConfig) {
+        Properties properties = new Properties();
+        addProperty(properties, "user", connectionFactoryConfig.getUserName());
```
alexandrudanciu, line 43: Are these random string values or is there a rule/expectation from another module? Are the values maybe specified by another library (java.sql)?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/BigDecimalArgumentHandler.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.math.BigDecimal;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum BigDecimalArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 22: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/BooleanArgumentHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum BooleanArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 21: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/ByteArgumentHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum ByteArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 21: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/BytesArgumentHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum BytesArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 21: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/DateArgumentHandler.java
```diff
@@ -0,0 +1,34 @@
...
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Timestamp;
+import java.util.Date;
+
+public enum DateArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 23: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/DoubleArgumentHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum DoubleArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 21: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/FloatArgumentHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum FloatArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 21: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/InputStreamArgumentHandler.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.io.InputStream;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum InputStreamArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 22: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/IntArgumentHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum IntArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 21: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/LongArgumentHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum LongArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 21: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/NullArgumentHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum NullArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 21: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/ObjectArgumentHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum ObjectArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 21: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/ShortArgumentHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum ShortArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 21: Is an enum containing one element necessary?

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/StringArgumentHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+public enum StringArgumentHandler implements ArgumentHandler {
+    INSTANCE;
```
alexandrudanciu, line 21: Is an enum containing one element necessary?

alexandrudanciu (commented):
This review proposes some improvements related to naming, design patterns. Also some feedback on licensing is provided.

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/BigDecimalArgumentHandler.java
denisneuling: Those `ArgumentHandler`s are stateless so predestined for being singletons - and this is the shortest way to define a singleton in Java known to me. So no need for 
```java
    public static final Singleton INSTANCE = new Singleton();
    private Singleton() {}
    public static Singleton getInstance(){ return INSTANCE; }
```
nor any of its derivatives - just defining an enum :)
But, if you like, I can rewrite this into something else.

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/connection/ConnectionFactoryImpl.java
denisneuling: Those property keys are coming from the [postgres jdbc documentation](https://jdbc.postgresql.org/documentation/head/connect.html) - will note that down as javadoc.

common/libraries/clients/postgresql/build.gradle.kts
denisneuling: Many thanks for investigating - that's great!

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/LongArgumentHandler.java
ndr-brt: It's considered the best way in java to obtain a singleton instance.

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/BigDecimalArgumentHandler.java
ndr-brt: Maybe, since the implementation methods are all one-liner, we could implement a single enum with all the implementations inside, doing that we won't need to list every implementation in `SimplePostgresqlClient` because we will need only to call the static method `values()` on the enum class.

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/statement/DefaultArgumentHandler.java
```diff
@@ -0,0 +1,207 @@
...
+package org.eclipse.dataspaceconnector.clients.postgresql.statement;
+
+import java.io.InputStream;
+import java.math.BigDecimal;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Timestamp;
+import java.util.Date;
+
+public enum DefaultArgumentHandler implements ArgumentHandler {
```
ndr-brt, line 24: This could be called `ArgumentHandlers`

common/libraries/clients/postgresql/src/main/java/org/eclipse/dataspaceconnector/clients/postgresql/SimplePostgresqlClient.java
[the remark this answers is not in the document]
denisneuling: `SimplePostgresqlClient` has been removed and `PooledPostgresqlClient` was renamed to `PostgresqlClientImpl`.

paullatzelsperger (changes requested):
# PR Remarks

## Connection Pooling

- The connection pool should be decoupled from the client since the latter should not be tied to one particular pool imnplementation. For example, someone may want to use another connection pool or use a pool provided by a JTA transaction manager or Spring. This is not possible with the current implementation.  

## Transaction Management

- Transaction management is different when using resource-local connections and JTA-bound connections. This needs to be accounted for.  

## Connection Pool Configuration

- Configuration needs to account for the fact that JDBC driver properties vary by type. 

## Generalize the Implementation

- Nothing appears to tie the implementation to Postgres in particular, it seems to be a generic SQL client implementation. We should just generalize it by removing Postgres dependencies.

## Testing
- Testcontainers should be removed. Docker containers _must not_ be started from within code, as this will require docker to be installed
on every machine and will cause permission problems if the Java process does not have the appropriate rights.
- Docker containers are spun up in the workflow file.
- Those are integration tests and should be annotated as such
- There are no unit tests, they need to be added 


# Refactoring Proposal

## 1. Introduce a `TransactionContext` interface 

Create an interface `TransactionContext`, which is implemented by a second extension that abstracts the connection pool and transaction management. 

```java
public interface TransactionContext{
    void execute(Consumer<Connection> action);
}
```
The interface receives a `Connection` object from the pool implementation, on which all SQL/DML operations are to be executed, within the bounds of that lambda. Implementations are responsible for acquiring a connection and properly demarcating transactions. We can supply an Apache implementation of `TransactionContext` as default. There could be other implementations that work with C3P0, Spring, Atomikos, a JTA TM bound to JNDI, etc. 

The interface would not go into the core SPI, but exist in its own SPI module. This would also allow several extensions to share a pool.

## 2. Isolate Postgres specific code if needed

The client and trx context modules should not reference Postgres. It may make sense to create a specific module to handle the Postgres driver, although generic configuration should handle that.

## 3. Generalize configuration

Create a general JDBC configuration loader so it is possible to configure any JDBC driver with the database pool. 

### 4. General Comments
- The Stream of `*Handler`s is overengineered. We can just have a single Handler class with overloads for each datatype, since we do not expect the Java primitive types to change anytime soon. Also, that would not waste memory on allocating an entire Stream of handlers per query parameter. 
- the `RowMapper` looks a lot like the one provided by Spring. We may need to attribute accordingly.
- Having enums implement an interface is confusing, as enums semantically are just named integers. Of course we can abuse them as singletons (I know that some claim that to be good), but there are clearer and better ways to do this.

paullatzelsperger:
also, the Implementation of the [CommandQueue](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/470) will depend on a working transaction implementation

DominikPinsel:
> # PR Remarks
> ## Connection Pooling
> * The connection pool should be decoupled from the client since the latter should not be tied to one particular pool imnplementation. For example, someone may want to use another connection pool or use a pool provided by a JTA transaction manager or Spring. This is not possible with the current implementation.

I don't see why we should make a library choice configurable by moving it out of the extension. We don't do something similar for other extensions, too.

Additionally, if this becomes a requirement at some point (which I doubt) we can do this any time. I don't see a good enough reason to increase the complexity and total number of extensions in the EDC.

> ## Transaction Management
> * Transaction management is different when using resource-local connections and JTA-bound connections. This needs to be accounted for.

Could you elaborate this? Other storage extensions/libraries prefer a solution with a lower complexity, too. Why is this case different and a **necessary** feature for this PR, which implements a simple library to send queries to PostgreSQL?

> ## Connection Pool Configuration
> * Configuration needs to account for the fact that JDBC driver properties vary by type.

This library is only for PostgreSQL and in this case the configuration types are perfectly fine.

> ## Generalize the Implementation
> * Nothing appears to tie the implementation to Postgres in particular, it seems to be a generic SQL client implementation. We should just generalize it by removing Postgres dependencies.

Like I already said, the library is for PostgreSQL only (and the issue too). I would not increase the scope, and I also
don't see why this should be necessary.

> ## Testing
> * Testcontainers should be removed. Docker containers _must not_ be started from within code, as this will require docker to be installed
    >   on every machine and will cause permission problems if the Java process does not have the appropriate rights.
> * Docker containers are spun up in the workflow file.
> * Those are integration tests and should be annotated as such
> * There are no unit tests, they need to be added

To integration test a PostgreSQL library having a running PostgreSQL instance somewhere is necessary. I don't see how you would test it without it.

Also, I'd assume that a docker installation on integration test machines is pretty common. GitHub supports it, too.

We will add further unit tests and the integration test annotation.

> # Refactoring Proposal
> ## 1. Introduce a `TransactionContext` interface
> Create an interface `TransactionContext`, which is implemented by a second extension that abstracts the connection pool and transaction management.
>
> ```java
> public interface TransactionContext{
>     void execute(Consumer<Connection> action);
> }
> ```
>
> The interface receives a `Connection` object from the pool implementation, on which all SQL/DML operations are to be executed, within the bounds of that lambda. Implementations are responsible for acquiring a connection and properly demarcating transactions. We can supply an Apache implementation of `TransactionContext` as default. There could be other implementations that work with C3P0, Spring, Atomikos, a JTA TM bound to JNDI, etc.
>
> The interface would not go into the core SPI, but exist in its own SPI module. This would also allow several extensions to share a pool.

I would not move a library to its own extension and make it configurable without a good requirement. In your refactoring proposal it's already pretty good visible how this would increase the complexity by increasing the overall size of the interface and the total number of extensions and SPIs.

> ## 2. Isolate Postgres specific code if needed
> The client and trx context modules should not reference Postgres. It may make sense to create a specific module to handle the Postgres driver, although generic configuration should handle that.

> ## 3. Generalize configuration
> Create a general JDBC configuration loader so it is possible to configure any JDBC driver with the database pool.

Regarding 2 and 3. The PR and the issue are about supporting PostgreSQL. To increase the scope please create a follow-up issue.

> ### 4. General Comments
> * The Stream of `*Handler`s is overengineered. We can just have a single Handler class with overloads for each datatype, since we do not expect the Java primitive types to change anytime soon. Also, that would not waste memory on allocating an entire Stream of handlers per query parameter.
> * the `RowMapper` looks a lot like the one provided by Spring. We may need to attribute accordingly.

I don't see how this `RowMapper` has anything to do with Spring. The interface is completely different. There are
classes in other libraries that are called `RowMapper`, too. Anyway, we will rename it.

> * Having enums implement an interface is confusing, as enums semantically are just named integers. Of course we can abuse them as singletons (I know that some claim that to be good), but there are clearer and better ways to do this.

For other languages you may be right. But in Java this is perfectly fine at most a question of code style. This was also already reviewed and discussed with Andrea.

MoritzKeppler:
What would be the drawbacks of refactoring it to a generic jdbc solution + a postgres specific one?
Experience shows the it's hard to optimize for every database even if it's just a set of sql databases and postgres is widely used, so most likely we cover most needs by supporting it. But I like the idea to keep it open for other implementations in future.
Would also come with the benefit to test it against an in memory sql db (e.g. derby) - agree that we should not start containers from junit. @DominikPinsel  @denisneuling 

Same consideration could also apply to the connection pooling question, as @paullatzelsperger described - but I understand the reasons for keeping it simple. Assume it can be refactored in future if needed.

Definitely great to have the support of (opensource) sql databases!

denisneuling:
An alternative approach incorporating Paul's remarks has been created via [PR 499](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/499).

denisneuling:
Closing this PR in preference to the alternative approach introduced by [PR 499](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/499)

Issue #440, opened 2021-12-16
can not clone project, file name too long - please; some people use windows


a-trzewik:
unable to clone the repository please remote the test file

$ git clone https://github.com/eclipse-dataspaceconnector/DataSpaceConnector.git
Cloning into 'DataSpaceConnector'...
remote: Enumerating objects: 35169, done.
remote: Counting objects: 100% (2/2), done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 35169 (delta 0), reused 0 (delta 0), pack-reused 35167
Receiving objects: 100% (35169/35169), 8.29 MiB | 1.66 MiB/s, done.
Resolving deltas: 100% (10706/10706), done.
error: unable to create file extensions/iam/daps/src/test/resources/keys/OEI6REU6RUU6NEM6N0M6MkQ6N0I6OTM6NTE6MkY6RkQ6QTA6RDM6NTM6NEM6NTg6Q0E6OUQ6NzA6QzI6a2V5aWQ6Q0I6OEM6Qzc6QjY6ODU6Nzk6QTg6MjM6QTY6Q0I6MTU6QUI6MTc6NTA6MkY6RTY6NjU6NDM6NUQ6RTg=.cert: Filename too long

ndr-brt:
Do you know what's the limit for the file name on windows?
If I'm not wrong the filename is used by the omejdn to identify the name of the client, but we can find an alternative anyway

jimmarino:
@ndr-brt has this been resolved?

ndr-brt:
In my opinion, yes. We could wait for a feedback from @a-trzewik , but in the meantime I will close this.

PR #441, opened 2021-12-16
integration-tests: set absolute path on docker volumes

ndr-brt:
Seems like the docker service didn't like the use of environment variables in the volume path.
Set these as absolute path.
Ref #440

### DIFF
--- a/.github/workflows/integrationtests.yaml
+++ b/.github/workflows/integrationtests.yaml
@@ -99,8 +99,8 @@ jobs:
         ports:
           - 4567:4567
         volumes:
-          - $PWD/extensions/iam/daps/src/test/resources/config:/opt/config
-          - $PWD/extensions/iam/daps/src/test/resources/keys:/opt/keys
+          - /home/runner/work/DataSpaceConnector/DataSpaceConnector/extensions/iam/daps/src/test/resources/config:/opt/config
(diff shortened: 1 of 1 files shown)


bscholtes1A (approved):

.github/workflows/integrationtests.yaml
```diff
@@ -99,8 +99,8 @@ jobs:
         ports:
           - 4567:4567
         volumes:
-          - $PWD/extensions/iam/daps/src/test/resources/config:/opt/config
-          - $PWD/extensions/iam/daps/src/test/resources/keys:/opt/keys
+          - /home/runner/work/DataSpaceConnector/DataSpaceConnector/extensions/iam/daps/src/test/resources/config:/opt/config
+          - /home/runner/work/DataSpaceConnector/DataSpaceConnector/extensions/iam/daps/src/test/resources/keys:/opt/keys
```
paullatzelsperger, line 103: maybe you could use `${{ github.workspace }}`

ndr-brt: I don't know when and how the binding is done, i prefer to set something that will work now, since I can't try it (on bmw fork actions are disabled...)

paullatzelsperger: if it helps, I can try on my fork

jimmarino: I'll merge this now but let's fix the use of hardcoded paths in a subsequent PR

PR #448, opened 2021-12-17
Fix daps it definitely

ndr-brt:
This time it's working, I needed to run omejdn from a job step.
Proof: https://github.com/ndr-brt/DataSpaceConnector/runs/4561749563


bscholtes1A (approved):

ndr-brt:
The unit test failure is not my fault, I modified only the `integrationtests.yaml` file.

paullatzelsperger:
the failing test is somewhat strange, but could be caused by a slow runner. #449 should take care of that

ndr-brt:
This will fix also #440

PR #451, opened 2021-12-18
Make Provisioner async

ndr-brt:
As stated in #421 a refactor to make `Provisioner` and `ProvisionManager` asynchronous, returning `CompletableFuture`s instead of accepting callbacks, this makes the code easier to read and to test.

Part of #330.


extensions/aws/s3/provision/src/main/java/org/eclipse/dataspaceconnector/provision/aws/s3/S3BucketProvisioner.java
```diff
@@ -45,12 +106,9 @@ public S3BucketProvisioner(ClientProvider clientProvider, int sessionDuration, M
...
         this.sessionDuration = sessionDuration;
         this.monitor = monitor;
-        this.retryPolicy = retryPolicy;
-    }
-
-    @Override
-    public void initialize(ProvisionContext context) {
-        this.context = context;
+        this.retryPolicy = retryPolicy.copy()
+                .withMaxRetries(10)
```
jimmarino, line 110: Can you make this be configurable with a default that is represented as a constant?

jimmarino:
@ndr-brt Overall I like this. I'd prefer not to remove the S3 pipelines because the provisioner classes are becoming bloated and are more difficult to follow (the documentation on the pipelines is also dropped). I think we should keep the pipelines, which can be further decomposed into segments if needed.

ndr-brt:
@jimmarino I inlined them because was kinda confusing to have a test suite for `S3BucketProvisioner` but in fact it tested the pipeline class. To deflate the provisioner class maybe we could extract some "wrappers" one for each SdkClient, so we can reach a division by responsability and a better testability

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/CoreTransferExtension.java
```diff
@@ -100,7 +96,7 @@ public void initialize(ServiceExtensionContext context) {
 
         var vault = context.getService(Vault.class);
 
-        provisionManager = new ProvisionManagerImpl();
+        ProvisionManagerImpl provisionManager = new ProvisionManagerImpl();
```
jimmarino, line 99: A bit of a nit: since the existing code uses var, can you stick to that so it is consistent w/in the class?

extensions/aws/s3/provision/src/main/java/org/eclipse/dataspaceconnector/provision/aws/s3/S3BucketProvisionerConfiguration.java
```diff
@@ -0,0 +1,55 @@
+package org.eclipse.dataspaceconnector.provision.aws.s3;
```
jimmarino, line 1: Why does this class need a builder, it only has two parameters?

Also, I don't think the service extension context should be passed to it. That can be handled in the extension initialize (I would also trap the Integer parse and report the problem if an exception is thrown)

ndr-brt: Done. 
Regarding the Builder, there's not a clear rule about the minimum parameter size, but I agree that this was pretty overkill.

ndr-brt:
@jimmarino @paullatzelsperger did the modifications we agreed on. 
The CI is broken on dependency retrieval that returns 502, I cannot re-launch it.

paullatzelsperger:
> The CI is broken on dependency retrieval that returns 502, I cannot re-launch it.

looks like the mulesoft repo is in a wonky state. lets wait a bit and restart later.

ndr-brt:
@jimmarino this can be merged

ndr-brt:
Moved the async logic inside `ProvisionManagerImpl` as discussed in #468 .
I assumed that, if a single resource provision/deprovision fails, the whole provision/deprovision process fails and the transfer process goes to the `transitionError` state. Is this assumption correct?

In addition, I noticed that there's some unclarity about the error notification: it's present into the `ProvisionResponse` and `DeprovisionResponse`, but in the latter there's also a `ResponseStatus`, this is not clear at all...
There could be different solutions to this, I would use a `Result` and avoid to put the error into the `ProvisionResponse`/`DeprovisionResponse`. Thoughts?

ndr-brt:
Ok, I moved the error handling responsibility to the futures, now the process is more linear. 
Ready for review/merge

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/CoreTransferExtension.java
```diff
@@ -92,16 +90,16 @@ public void initialize(ServiceExtensionContext context) {
...
         context.registerService(ProvisionManager.class, provisionManager);
 
         var waitStrategy = context.hasService(TransferWaitStrategy.class) ? context.getService(TransferWaitStrategy.class) : new ExponentialWaitStrategy(DEFAULT_ITERATION_WAIT);
 
         var dataProxyRegistry = new DataProxyManagerImpl();
         context.registerService(DataProxyManager.class, dataProxyRegistry);
 
-
-        asyncMgr = AsyncTransferProcessManager.Builder.newInstance()
+        transferProcessStore = context.getService(TransferProcessStore.class);
```
paullatzelsperger, line 101: This will likely have to go away since it's going to be set using @Inject

core/transfer/src/test/java/org/eclipse/dataspaceconnector/transfer/core/provision/ProvisionManagerImplTest.java
```diff
@@ -14,57 +18,118 @@
...
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.isA;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 class ProvisionManagerImplTest {
```
paullatzelsperger, line 28: we might wanna add a few tests, feel free to use this: https://gist.github.com/44cf444fda5c705823530fff6af5b779

paullatzelsperger (commented):
some minor things, see comments in code

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/CoreTransferExtension.java
ndr-brt: Done

core/transfer/src/test/java/org/eclipse/dataspaceconnector/transfer/core/provision/ProvisionManagerImplTest.java
ndr-brt: In async components testing the case where a future fails and then everything fails is kinda of implicit (at least, in this case).
However I added the two tests and improved assertions and names

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/AsyncTransferProcessManager.java
```diff
@@ -163,59 +161,54 @@ public void transitionError(String processId, String detail) {
...
+                        }
+
+                        transferProcess.getDataRequest().updateDestination(dataDestination);
+                    }
+
+                    return destinationResource;
+                })
+                .forEach(transferProcess::addProvisionedResource);
+
+        if (TransferProcessStates.ERROR.code() != transferProcess.getState() && transferProcess.provisioningComplete()) {
```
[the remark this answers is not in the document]
ndr-brt: If transfer process is in the error state what should not be done? What could bring the TP to ERROR before the provisioning? 
However I added a guard before the loop.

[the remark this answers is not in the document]
ndr-brt: Added the same guard on deprovisioning too

extensions/aws/s3/provision/src/main/java/org/eclipse/dataspaceconnector/provision/aws/s3/S3DeprovisionPipeline.java
```diff
@@ -1,123 +1,124 @@
-/*
```
jimmarino, line 1: Looks like the license header was inadvertently removed

extensions/aws/s3/provision/src/main/java/org/eclipse/dataspaceconnector/provision/aws/s3/S3ProvisionPipeline.java
```diff
@@ -1,48 +1,31 @@
-/*
```
jimmarino, line 1: License header removal

extensions/aws/s3/provision/src/main/java/org/eclipse/dataspaceconnector/provision/aws/AwsProvisionExtension.java
```diff
@@ -37,16 +39,24 @@
...
+import static java.lang.String.format;
+
 /**
  * Provides data transfer {@link org.eclipse.dataspaceconnector.spi.transfer.provision.Provisioner}s backed by AWS services.
  */
 @Provides(ClientProvider.class)
 public class AwsProvisionExtension implements ServiceExtension {
     @EdcSetting
-    private static final String AWS_ACCESS_KEY = "edc.aws.access.key";
+    public static final String AWS_ACCESS_KEY = "edc.aws.access.key";
```
jimmarino, line 50: Is there a reason these are made public?

jimmarino (approved):
This PR is a really big improvement. I had two minor license fixes that need to be put in and another minor thing about constants. Otherwise, I think it's ready to be merged.

ndr-brt:
@jimmarino resolved your comments, if checks pass can be merged

PR #456, opened 2021-12-21
Moves over the transfer subsystem to IDS Multipart

paullatzelsperger:
also fixes a few smaller bugs that were introduced in one of the previous refactorings.

The `DataRequest.isSync` flag is now also contained in the `.properties` map, which had to made mutable.


core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/validation/ContractValidationServiceImpl.java
```diff
@@ -131,7 +132,7 @@ public boolean validate(ClaimToken token, ContractAgreement agreement, ContractO
     @NotNull
     private ArrayList<Criterion> createCriteria(ContractOffer offer, ContractDefinition contractDefinition) {
         var criteria = new ArrayList<>(contractDefinition.getSelectorExpression().getCriteria());
-        var criterion = new Criterion(Asset.PROPERTY_ID, "=", offer.getAsset().getId());
+        var criterion = new Criterion(Asset.PROPERTY_ID, "=", "'" + offer.getAsset().getId() + "'");
```
DominikPinsel, line 135: I assume this would break several existing implementations that expect the asset id in the right operand. Is this a workaround that is necessary to send the asset id somewhere else? If yes maybe there is a better place for this workaround.

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/MultipartArtifactRequestSender.java
```diff
@@ -104,6 +104,8 @@ protected Message buildMessageHeader(DataRequest request, DynamicAttributeToken
                 ._requestedArtifact_(artifactId)
                 ._transferContract_(contractId)
                 .build();
+        request.getProperties().forEach((k, v) -> message.setProperty(k, v));
```
DominikPinsel, line 107: The properties are attributed with `@JsonIgnore`. We should not use them, otherwise this functionally will be broken as soon as we serialize the Information Model class correctly.
Source:
https://github.com/International-Data-Spaces-Association/Java-Representation-of-IDS-Information-Model/blob/main/de/fraunhofer/iais/eis/ArtifactRequestMessageImpl.java

data-protocols/ids/ids-api-multipart-endpoint-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/handler/ArtifactRequestHandler.java
```diff
@@ -121,25 +122,34 @@ public boolean canHandle(@NotNull MultipartRequest multipartRequest) {
 
         DataAddress dataAddress = artifactRequestMessagePayload.getDataDestination();
 
+
+        var props = new HashMap<String, String>();
+        if (artifactRequestMessage.getProperties() != null) {
+            artifactRequestMessage.getProperties().forEach((k, v) -> props.put(k, v.toString()));
+        }
```
DominikPinsel, line 129: The properties are attributed with `@JsonIgnore`. We should not use them, otherwise this functionally will be broken as soon as we serialize the Information Model class correctly.
Source:
https://github.com/International-Data-Spaces-Association/Java-Representation-of-IDS-Information-Model/blob/main/de/fraunhofer/iais/eis/ArtifactRequestMessageImpl.java

extensions/azure/contract-negotiation-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/store/CosmosContractNegotiationStore.java
```diff
@@ -78,7 +86,7 @@ public void delete(String negotiationId) {
     @Override
     public @NotNull List<ContractNegotiation> nextForState(int state, int max) {
 
-        var partitionKey = String.valueOf(state);
+        var partitionKey = "negotiationPartition1"; //TODO: make this configurable!!!
```
DominikPinsel, line 89: Good input. Please do this before merge the PR. Just commenting so it isn't forgotten.

DominikPinsel (changes requested):

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/MultipartArtifactRequestSender.java
paullatzelsperger: since the INfomodel we be donated to EDC we're fine.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/validation/ContractValidationServiceImpl.java
paullatzelsperger: I agree. As far as I recall this has to do with Cosmos (or SQL in general) expecting string comparisons to be put in quotes. It might prove difficult to decide when to add the ticks and when not, because in order to do that reliably we'd need schema introspection.

data-protocols/ids/ids-api-multipart-endpoint-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/handler/ArtifactRequestHandler.java
paullatzelsperger: The Infomodel will get donated to EDC, so we're in control of that anyways. So we're good.

ronjaquensel (approved):

PR #462, opened 2021-12-21
Transfer Process Manager Command Queue

ndr-brt:
Ref #330 . Depends on #451 so it will remain in "draft" state until the other gets merged, 
but in fact this is complete and ready for review.

### Description
This PR introduce a Command Queue in the `TransferProcessManager`.
The motivations for this are listed in the issue.

#### Implementation details
Commands are simple POJOs that extend `TransferProcessCommand`.
For every command there's an handler, who execute the business logic and do the state change.
Note that "every state change should be done inside a command handler".
At the moment is not really like this, I leaved some TODO here and there, I noticed some unclear situations that are not tested and not documented.
The commands are executed by the loop inside the `TransferProcessManager`, every command brings with itself a future so it will be possible for a client to know when the command is handled (and also if some error occurred).
Every handler returns a `TransferProcessCommandResult`, that contains:
* `nextCommand`, a reference to the command that has to be executed at the end of the current one (note: this is a workaround needed because at the moment there are no events generated, otherwise an "event-to-command" translator would work better)
* `postAction`, a function that tells what `TransferProcessListener` method should be called when the command is executed correctly.

#### Tests
I introduced an `AsyncTransferProcessManagerIntegrationTest` that tests some "start-to-end" transfer processes using the in-memory store. This was necessary because the mocking would have been just crazy.
I moved all the tests from the `AsyncTransferProcessManagerConsumerTest` in the respective handler tests.

#### Next steps
There are a lot of "nice to do" things that could be implemented, but the most important in my opinion are:
* **define a better interface**: at the moment `TransferProcessStore` exposes 6 methods, but they are not consistent in the return types (in my opinion everything should be async so every one should return a `CompletableFuture`, one of them (`transitionProvisioning`) has no clear intents (I added a TODO on its only call)
* **avoid useless commands/states**: with this refactor is pretty clear that some states are not needed anymore (all those ending with `ING`) and the handlers/commands starting with `Prepare*` could be avoided.
* **better handling of async operations**: in two handlers (provision and deprovision) the async calls and their return are handled in different manners. Since there are no events the only way to handle async state changes is to pass the commandQueue to the command handler but that's not a good way to do things. (e.g. `ProvisionHandler`)
* **graceful shutdown**: at the moment there are no handling of the shutdown, I forced the `stop()` method to wait for all the commands to be executed but this concept needs to be improved
* **unit test handlers**: I added some tests on the handlers, but some of them are only covered by integration tests.


ndr-brt:
Closing as it will be replaced by another PR by @paullatzelsperger

PR #466, opened 2021-12-22
Remove unused mulesoft repository

ndr-brt:
The comment reported that this dependency was "used for the multihash lib", but we don't use that dependency (there are some copy-and-pasted files from it).
This is currently breaking CI (e.g. https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/runs/4607127286?check_suite_focus=true) because the mulsesoft repository is responding 502 but, since it's not needed, we can get rid of it.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -46,9 +46,6 @@ subprojects {
         maven {
             url = uri("https://maven.iais.fraunhofer.de/artifactory/eis-ids-public/")
         }
-        maven {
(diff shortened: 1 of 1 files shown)


ndr-brt:
(Integration tests are broken because they run on the current main, that still has the mulesoft repository)

paullatzelsperger:
as far as I recall it was used for the rate limiting API on ION.

jimmarino (approved):

PR #473, opened 2021-12-23
fix: Repair transformers related to Asset and fix NullPointerException in MultipartCatalogDescriptionRequestSender

bscholtes1A:


data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/MultipartCatalogDescriptionRequestSender.java
```diff
@@ -132,4 +112,34 @@ protected Catalog getResponseContent(IdsMultipartParts parts) throws Exception {
 
         return transformResult.getContent();
     }
+
+    private static void createOfferResourcesFromProperties(ResourceCatalog catalog, ObjectMapper mapper) {
+        if (catalog.getProperties() != null) {
```
bscholtes1A, line 117: This test is the fix for the NullPointerException

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/MultipartCatalogDescriptionRequestSender.java
```diff
@@ -84,44 +85,23 @@ protected Message buildMessageHeader(CatalogRequest request, DynamicAttributeTok
...
             throw new EdcException("Payload was null but connector self-description was expected");
         }
 
-        BaseConnector baseConnector;
-        try {
-            baseConnector = objectMapper.readValue(parts.getPayload().readAllBytes(), BaseConnector.class);
-        } catch (IOException exception) {
-            throw new EdcException(String.format("Could not deserialize connector self-description: %s", exception.getMessage()));
+        BaseConnector baseConnector = getBaseConnector(getObjectMapper(), parts);
+        if (baseConnector.getResourceCatalog() == null || baseConnector.getResourceCatalog().isEmpty()) {
```
jimmarino, line 94: Wouldn't it be valid to have no resources, e.g. if the client does not have access to any resources or the current connector does not have any configured?

data-protocols/ids/ids-api-multipart-endpoint-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/handler/DescriptionHandler.java
```diff
@@ -47,15 +47,14 @@
...
-    public DescriptionHandler(
-            @NotNull Monitor monitor,
-            @NotNull String connectorId,
-            @NotNull TransformerRegistry transformerRegistry,
-            @NotNull ArtifactDescriptionRequestHandler artifactDescriptionRequestHandler,
-            @NotNull DataCatalogDescriptionRequestHandler dataCatalogDescriptionRequestHandler,
-            @NotNull RepresentationDescriptionRequestHandler representationDescriptionRequestHandler,
-            @NotNull ResourceDescriptionRequestHandler resourceDescriptionRequestHandler,
-            @NotNull ConnectorDescriptionRequestHandler connectorDescriptionRequestHandler) {
+    public DescriptionHandler(@NotNull Monitor monitor,
```
jimmarino, line 50: Minor nit: If this is just formatting, is it possible to avoid this since it makes it difficult to see what the actual changes are?

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/MultipartCatalogDescriptionRequestSender.java
bscholtes1A: Yes but what do we do in this case?
If you look at the current main code, if `baseConnector.getResourceCatalog().isEmpty()`, then it would lead `resourceCatalog` to be null and the transformer would be called with a null value while this is not allowed (input argument is annoted with @NotNull).
My concern was then to cover this case which does not seem cover in the present implem and could lead to NullPointerException.

jimmarino: @bscholtes1A it should be initialized to an empty list, which will be passed to the transformer

jimmarino:
@bscholtes1A LGTM pending the one comment about passing an empty list.

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/MultipartCatalogDescriptionRequestSender.java
bscholtes1A: I updated so that now `resourceCatalog` gets an empty catalog if none is present in the base connector:
`ResourceCatalog resourceCatalog = baseConnector.getResourceCatalog().stream()
                .findFirst()
                .orElse(new ResourceCatalogBuilder().build());`

@jimmarino  could you please confirm this is what you expected?
If yes I guess we can merge this one

PR #474, opened 2021-12-30
fix: Repair InMemoryFederatedCacheStore as non-unique ContractOffer is used as key in the store

bscholtes1A:
Current implementation uses the id of the inserted `ContractOffer` as key in the underlying map of the in-memory store. 

However this id is not contains a random string (see `ContractOfferServiceImpl`), thus leading the store to contain duplicate versions of the same asset and to grow indefinitely.


paullatzelsperger (approved):
LGTM, thanks for adding tests

PR #475, opened 2021-12-30
Refactor module loading

### LINKED ISSUE #469: Refactor the module loading
## Restructuring the `core` module

This will be refactored to contain only three sub-modules, those are: 
- `core:base`: will contain all modules currently located in `core`, except `transfer` and `contract`
- `core:contract`: contains all contract-management-related implementations
- `core:transfer`: contains all transfer-management-related implementations

## Dependency loading - requirements and preconditions

_Disclaimer: the term "dependency" here refers to the runtime-dependency expressed by `requires()` and `provides()` rather than Gradle dependencies._

The three `core` modules are added automatically according to the following rules, so we do not need to `require` them or any of their implementations explicitly. Should a module contain several submodules and/or extensions then the rules apply to all of them.
1. `core:base`: implicitly required by all other modules, that it is automatically loaded first during extension loading. One way of doing is would be to simply add a `requires` entry to all extensions on the classpath. 
1. `core:transfer`: implicitly required by all extensions (except other `core` extensions) as soon as it's on the classpath. That means that the `core:transfer` module is always loaded _after_ `core:base` and _before_ all other modules. Technically this could be achieved by adding a `required` entry to all modules (except core extensions) that have the `core:transfer` module on the classpath. So putting the `core:transfer` module on the classpath is enough, it never needs to be `"required"` explicitly.  
1. `core:contract`: implicitly required by all extensions that require the `core:transfer`, so the same mechanism can be applied there.

## Dependency loading - implementation considerations

- Establish a pattern where feature names must be accessible through constants, e.g. requiring the AssetIndex is done by adding `@Requires(AssetIndex.FEATURE)` to the extension.
- Introduce annotations: `requires()` and `provides()` methods will be substituted with annotations, `@Provides` and `@Requires` which take a list of Strings. This would give us the following advantages: 
  - can then be used to inspect dependencies without instantiating the extension class, 
  - perform some sort of dryrun or preflight check to detect dependency problems faster and without actually instantiating and initializing the extension
  - group dependencies together in another annotation, which is useful if the same set of dependencies are used together often, e.g. an extension for a specific AssetIndex implementation could be annotated with `@AssetExtension`, which be an alias for `@Provides(AssetIndex.FEATURE, AssetLoader.FEATURE, DataAddressResolver.FEATURE)`
  - dependency graphs could be generated as a means of documentation

- Introduce hierarchical features: feature names should be namespaced such that they express hierarchy. For example features within the `extension:foo` module have to be named `extension:foo:bar:stuff`, `extension:foo:baz:stuff`. Simply adding a dependency onto the parent feature, here `extension:foo`, would implicitly add depencies onto all "child" features "underneath" the `extension:foo` namespace. The parent feature would be synthetic, meaning we would never explicitly `@Provide` this, only the child features. This is a convenience functionality, as it reduces the amount of dependencies we need to express.

- Modules should expose their features (including parent features) through constants in an interface class in an SPI module, e.g. via `AssetIndex.FEATURE`. 

- remove the `LoadPhase` altogether

paullatzelsperger:
closes #469


data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/IdsMultipartDispatcherServiceExtension.java
```diff
@@ -51,32 +51,27 @@
...
     private Monitor monitor;
+    @Inject
+    private OkHttpClient httpClient;
+    @Inject
+    private IdentityService identityService;
+    @Inject
+    private TransformerRegistry transformerRegistry;
+
+    public IdsMultipartDispatcherServiceExtension() {
+    }
```
bscholtes1A, line 62: Do we need this constructor?

core/base/src/main/java/org/eclipse/dataspaceconnector/core/BaseExtension.java
```diff
@@ -0,0 +1,14 @@
+package org.eclipse.dataspaceconnector.core;
```
bscholtes1A, line 1: As discussed, check this one is needed.

core/base/src/main/java/org/eclipse/dataspaceconnector/core/CoreExtension.java
```diff
@@ -0,0 +1,16 @@
+package org.eclipse.dataspaceconnector.core;
```
bscholtes1A, line 1: As discussed, check this one is needed.

extensions/iam/distributed-identity/dummy-credentials-verifier/src/main/java/org/eclipse/dataspaceconnector/iam/verifier/DummyCredentialsVerifierExtension.java
```diff
@@ -14,34 +14,23 @@
...
     }
 
-    @Override
-    public Set<String> provides() {
-        return Set.of(CredentialsVerifier.FEATURE);
-    }
-
-    @Override
-    public Set<String> requires() {
-        return Set.of(IdentityHubClient.FEATURE);
```
bscholtes1A, line 42: No @Inject corresponding to this requires?

extensions/iam/distributed-identity/registration-service/src/main/java/org/eclipse/dataspaceconnector/iam/registrationservice/crawler/CrawlerExtension.java
```diff
@@ -47,17 +47,10 @@
     private static final String ION_CRAWLER_TYPE_SETTING = "edc.ion.crawler.did-type";
     private ServiceExtensionContext context;
     private Scheduler quartzScheduler;
-
-    @Override
-    public Set<String> provides() {
```
bscholtes1A, line 52: No corresponding @Provides ?

extensions/iam/oauth2/oauth2-core/src/main/java/org/eclipse/dataspaceconnector/iam/oauth2/core/impl/Oauth2ServiceImpl.java
```diff
@@ -53,6 +54,7 @@
 /**
  * Implements the OAuth2 client credentials flow and bearer token validation.
  */
+@Feature("oauth2")
```
bscholtes1A, line 57: Shouldnt it be `@Provides(Oauth2Service.class)` ?

spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/store/TransferProcessStore.java
```diff
@@ -24,9 +25,11 @@
 /**
  * Manages persistent storage of {@link TransferProcess} state.
  */
+@Feature(TransferProcessStore.FEATURE)
 public interface TransferProcessStore {
 
-    String FEATURE = "dataspaceconnector:transferprocessstore";
+    @Deprecated
```
bscholtes1A, line 31: `@Deprecated` ??

bscholtes1A (approved):

paullatzelsperger:
- `DummyCredentialsVerifierExtension.java`: 
  > No @Inject corresponding to this requires?

  Acually no, because no IdentityHub instance is used by this extension
- `Oauth2ServiceImpl.java`: 
  > Do we need to declare the Feature within the Implem?

  yes, correct, will delete.
- `TransferProcessStore.java`:
  > @Deprecated ??

  must have been a merge/rebase problem or some other glitch. it's not deprecated of course :)

jimmarino (commented):
Looks good but do InjectionPoint and the related classes need to be in SPI? Can't they be moved to the bootstrap module?

paullatzelsperger:
> Looks good but do InjectionPoint and the related classes need to be in SPI? Can't they be moved to the bootstrap module?

Unfortunately they cannot, because `ServiceExtensionContext.loadServiceExtensions()`, which is in SPI, returns a `List<InjectionContainer>>`

jimmarino:
> > Looks good but do InjectionPoint and the related classes need to be in SPI? Can't they be moved to the bootstrap module?
> 
> Unfortunately they cannot, because `ServiceExtensionContext.loadServiceExtensions()`, which is in SPI, returns a `List<InjectionContainer>>`, so that would introduce a circular dependency between `spi` and `core:base`

OK, not for this PR but why don't we internalize the loadExtensions methods by moving them out of ServiceExtensionContext and into a separate, internal interface used by the bootsrap system (this could extend the former)? The DefaultServiceExtensionContext would implement both interfaces. This would remove those methods and referenced classes from the SPI, which are not intended to be used by extensions?

jimmarino (approved):

PR #478, opened 2022-01-04
data-protocols: add guard to avoid NPE when tokenResult is failed

ndr-brt:


jimmarino:
@ndr-brt is this ready to merge? If you are OK with it, I'll do that.

ndr-brt:
@jimmarino ok

PR #481, opened 2022-01-05
refactor(oauth2-core): only set jwsAlgorithm in constructor

tmberthold:


ndr-brt (approved):

tmberthold:
_Pipeline needs to be restarted due to a timeout connecting to a dependency repo._

Issue #92, opened 2021-09-22
Improve folder structure


MoritzKeppler:
We will have multiple associated services in our repo next to the connector, e.g. registration service for DID functionality.
I suggest to update the folder structure in a way that makes it easier to understand, which code belongs to which service.
Proposal:
```
/docs                       overall docs
/common                     shared libs, like the service loading framework
/services                   all services that form the connector ecosystem
e.g.
/services/connector         the connector itself :-)
/services/did-registration  registration service that crawls ION network
/services/catalog           ...
```
each service can have the following subfolders
```
.../core                core business functionality of the service
.../launcher            one or more bundles / launchers
.../scripts             e.g. helm charts
.../extensions          extensions that can get bundled
.../interfaces (or spi) extension interfaces
```
What do you think?
<sub> Moritz Keppler <moritz.keppler@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

jimmarino:
+1

DominikPinsel:
I think that's great idea. What do you think about restructuring the extension directory, too? 
It would be nice if the directory structure would show which part of the system is extended.

Examples:
- /filesystem/configuration-fs -> /configuration/filesystem
- /filesystem/vault-fs -> /vault/filesystem
- /aws/s3/provision -> /transfer/provision/aws/s3
- /aws/s3/s3-schema -> /schemas/aws/s3
- /azure/transfer-process-store-cosmos -> transfer/store/azure-cosmos
- /azure/vault -> /vault/azure

DominikPinsel:
Would it be fine when I create a draft pull-request with these changes?

PR #482, opened 2022-01-05
Restructure modules 

paullatzelsperger:
This PR has two main areas of change:
- split apart the SPI module into different sub-modules, so that different runtimes (e.g. dataloader, data-plane) only need to have classes that they need on the classpath.
- go through all extensions and make sure they are at the correct place in the folder structure (e.g. a `cosmos` subfolder was created)

I also added BOM files for some features.


spi/catalog-spi/build.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+
+
+dependencies {
+    api(project(":spi:core-spi"))
+    api(project(":spi:contract-spi"))
+}
+
+publishing {
+    publications {
+        create<MavenPublication>("transfer-spi") {
```
ndr-brt, line 27: should be `catalog-api`

spi/contract-spi/build.gradle.kts
```diff
@@ -16,7 +16,16 @@ plugins {
...
+
 dependencies {
-    api(project(":spi"))
-    implementation(project(":common:util"))
+    api(project(":spi:core-spi"))
+}
+
+publishing {
+    publications {
+        create<MavenPublication>("transfer-spi") {
```
ndr-brt, line 26: should be `contract-api`

spi/core-spi/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
...
+    `java-library`
+    `maven-publish`
+}
+
+dependencies {
+    api("com.fasterxml.jackson.core:jackson-core:${jacksonVersion}")
+    api("com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}")
+    api("com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}")
+    api("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jacksonVersion}")
+    api("com.squareup.okhttp3:okhttp:${okHttpVersion}")
```
ndr-brt, line 30: okhttp and failsafe dependencies should be defined in `core-base` and not in "core-spi"

spi/core-spi/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
...
+
+    api("net.jodah:failsafe:${jodahFailsafeVersion}")
+
+    api(project(":core:policy:policy-evaluator"))
```
ndr-brt, line 34: an `spi` module should not depend from a `core` one, in my opinion the `policy-evaluator` module should become a `policy-spi` and `core-spi` should depend on it

ndr-brt (commented):
Looks good, added some comments.
Regarding the `extensions` folder, at the moment we have to way of grouping them: by domain term (e.g. `dataloading`, `policy`, `transfer-function`) and by "technology name" (e.g. `aws`, `azure` etc...).
I think we should choose an unique way to group exception, to avoid confusion, maybe we should open a discussion/issue for this.

bscholtes1A:
> k we should choose an unique way to group exc

I agree. Mixing both will be confusing.

spi/core-spi/build.gradle.kts
paullatzelsperger: A lot of extensions depend on Failsafe and OkHttp. If we move those into `core:base` this would mean, that these extensions require a dependency on `core:base` in addition to `core-spi`.
Just so we're aware of that.

ndr-brt: I think this is correct... without the `core:base` dependency, the `OkHttpClient` and the `RetryPolicy` services cannot be obtained

paullatzelsperger:
> Looks good, added some comments. Regarding the `extensions` folder, at the moment we have to way of grouping them: by domain term (e.g. `dataloading`, `policy`, `transfer-function`) and by "technology name" (e.g. `aws`, `azure` etc...). I think we should choose an unique way to group exception, to avoid confusion, maybe we should open a discussion/issue for this.

Personally I am in favor of splitting this based on technology rather than functionality, because it will make the creation of BOMs rather easy. However, there is already an issue for a broader folder restructuring (#92)

juliapampus:
@paullatzelsperger Does it make sense to create this duplicated naming scheme? E.g. `:extensions:azure:blobstorage:blob-schema` could be change to `:extensions:azure:blobstorage:schema` and `:extensions:azure:cosmos:cosmos-common` to `:extensions:azure:cosmos:common` etc. If this should be addressed in another PR anyway, just ignore my comment ^^

paullatzelsperger:
> @paullatzelsperger Does it make sense to create this duplicated naming scheme? E.g. `:extensions:azure:blobstorage:blob-schema` could be changed to `:extensions:azure:blobstorage:schema` and `:extensions:azure:cosmos:cosmos-common` to `:extensions:azure:cosmos:common` etc. If this should be addressed in another PR anyway, just ignore my comment ^^

In theory that would be nice, but gradle gets confused if it has two modules with the same name, regardless whether they are in different submodules. so e.g. `azure:blobstorage:schema` and `aws:s3:schema` will cause nasty problems and build errors.

ndr-brt (approved):
looks good!

jimmarino (approved):

PR #483, opened 2022-01-06
added initial devcontainer setup files

StoyanStAtanasov:
Add devcontainer setup files for vscode to ease evaluation and experimentation.
Some tweaking may be needed as the files are from the example repo with MIT license https://github.com/microsoft/vscode-remote-try-java/blob/main/LICENSE

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,24 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.195.0/containers/java/.devcontainer/base.Dockerfile
+# [Choice] Java version (use -bullseye variants on local arm64/Apple Silicon): 8, 11, 16, 8-bullseye, 11-bullseye, 16-bullseye, 8-buster, 11-buster, 16-buster
+ARG VARIANT=11-bullseye
+FROM mcr.microsoft.com/vscode/devcontainers/java:0-${VARIANT}
+
+# [Option] Install Maven
+ARG INSTALL_MAVEN="false"
+ARG MAVEN_VERSION=""
+# [Option] Install Gradle
+ARG INSTALL_GRADLE="false"
+ARG GRADLE_VERSION=""
+RUN if [ "${INSTALL_MAVEN}" = "true" ]; then su vscode -c "umask 0002 && . /usr/local/sdkman/bin/sdkman-init.sh && sdk install maven \"${MAVEN_VERSION}\""; fi \
+    && if [ "${INSTALL_GRADLE}" = "true" ]; then su vscode -c "umask 0002 && . /usr/local/sdkman/bin/sdkman-init.sh && sdk install gradle \"${GRADLE_VERSION}\""; fi
+
+# [Choice] Node.js version: none, lts/*, 16, 14, 12, 10
+ARG NODE_VERSION="none"
+RUN if [ "${NODE_VERSION}" != "none" ]; then su vscode -c "umask 0002 && . /usr/local/share/nvm/nvm.sh && nvm install ${NODE_VERSION} 2>&1"; fi
+
+# [Optional] Uncomment this section to install additional OS packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+#     && apt-get -y install --no-install-recommends <your-package-list-here>
+
+# [Optional] Uncomment this line to install global node packages.
+# RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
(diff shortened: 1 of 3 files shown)


StoyanStAtanasov:
I am having problem with eclipse foundation contribution agreement. I have created eclipse account, accepted the new agreement, entered my personal data and installed the validation app on my github account for this fork. Now eclipesfdn/eca is pending. Any ideas?

MoritzKeppler:
thx for your contribution! 
I'm not familiar with devcontainer, but maybe other committers @jimmarino @alexandrudanciu have experience / an opinion?

We need at least two things:
* you need to sign the eclipse contributor's agreement
* files must contain a copyright header (see the other source code files for an example) that reflect the license (in this case it's MIT) and the copyright owner.

alexandrudanciu:
In my opinion, no IDE-specific config files should be managed within a source code repository. We neither want to promote one specific IDE nor do we want to cover all usual suspects.
Another aspect for a rationale is, if the creation of a devcontainer.json represents a high entry barrier for developers. In my opinion, the EDC technology stack is rather lightweight and the auto-generated devcontainer.json specific for the selected OS and programming language should fit.

StoyanStAtanasov:
@MoritzKeppler @alexandrudanciu Thanks for your responses! 
What about non Java developers that want to kick the tires without installing JDKs and specialized IDEs? Maybe just having a Dockerfile that can build and run the connector is a good alternative?
If I uderstand correctly that EDC wants to promote itself to the widest possible audience a good IDE defaluts for VSCode can be a great thing, especially with devcontainer where it will automatically build your dev machine for you and you are presented with quick results.
P.S the defaults did not work for me. Most probably user error :)

paullatzelsperger:
Developing in Java without installing a Java IDE or a JDK seems like a bit of a paradox :)

To my knowledge VSCode really isn't widely used for Java development. While it does work, there are much better, and much better suited, alternatives out there, even free ones. 
I will acknowledge however, that in the end it's a matter of preference, so there really isn't an argument to be made pro one or contra the other.

That also the main reason why - like @alexandrudanciu - I'd argue that as a general rule we **do not** check in IDE-specific files. Else we'd also have to check in `.idea` directories and whatever the Eclipse IDE uses and that will definitely cause problems.

If a low entry barrier is what you're after, please take a look the samples and the onboarding guides. Those should get you up and running quickly. If there's a problem with those, we'll gladly fix/update them.

StoyanStAtanasov:
@paullatzelsperger Thank you, I apreciate your reply. 

Understood: do not check in IDE-specific files
What is the practical alternative place? samples, another repo? What about a Dockerfile?

Rationale:
I believe, there is value to be added for wider range of developers, that are new to the project, which could help adoption. I just don't know how to do it.
"Developing in Java without installing a Java IDE or a JDK seems like a bit of a paradox :)" - Not "develop", but more evaluate, adapt, experiment with.

The point is, of course, to provide a predefined container that will handle everything for any developer. devcontainer is the perfect technology for this, as it will install the JDK and the Java extension and anything else that is needed to start evaluating the project, without messing with your Dev machine. There are also the Github codespaces, that can take advantage of devcontainer.

In the end (I hope) EDC is just another service in the infrastructure (which is probably kubernetes), so it should be usable by non Java centric developer teams.

Onboarding guide: "We'll assume that you've just checked out the EDC code base and have Java 11 installed on your development machine" - I think we can do better.

VSCode may not be the most widely used Java IDE, but it is the most widely used IDE :) After all, it is even in the EDC videos 😉.

paullatzelsperger:
Everyone is free to host a devcontainer configuration in their own repo of course, we're just not going to do so in the EDC repo.

Also, the EDC onboarding is not an introduction into Java or how to set up a JDK. As with any other project we assume a working development environment. That's just the minimum requirement we have. As soon as you want to run Java code, you're going to need a JDK, whether that is inside a dev container or natively installed is your personal business. 

Let me also emphasize that the EDC is *not* a runnable application or service, it rather is a framework upon which applications are built or that can be incorporated into existing applications. 

With that in mind there isn't really _the_ EDC, there's just applications _using_ EDC, so there also is not really a standard API for EDC.

With that being said, there are indeed REST endpoints that we provide out-of-the-box (`extensions/api`), they just have to be included in the build config. Unfortunately we do not yet have an automatically generated API doc for it, but it is on our todo list. For now there are README files. 

I'd argue that VSCode is an editor rather than an IDE, but that's a philosophical argument therefore intrinsically moot.

StoyanStAtanasov:
Thank you guys again for the great feedback!

Should I close the issue?

MoritzKeppler:
thanks for the idea and the discussion!

PR #492, opened 2022-01-11
data-protocols: evaluate ids.webhook.address at startup and not lazily

ndr-brt:
The `ids.webhook.address` should be evaluated and verified on the initialization phase, this will avoid unexpected errors on execution.

### DIFF
--- a/data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/IdsMultipartDispatcherServiceExtension.java
+++ b/data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/IdsMultipartDispatcherServiceExtension.java
@@ -50,6 +50,10 @@ public class IdsMultipartDispatcherServiceExtension implements ServiceExtension 
     public static final String EDC_IDS_ID = "edc.ids.id";
     public static final String DEFAULT_EDC_IDS_ID = "urn:connector:edc";
 
+    @EdcSetting
+    public static final String IDS_WEBOHOOK_ADDRESS = "ids.webhook.address";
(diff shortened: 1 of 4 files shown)


paullatzelsperger:
also, we could check to see if there's a way to supply sane defaults, like, can we find out what the hostname/URL is we're running on?

ndr-brt:
I don't think this is possible... but I will be happy to be proved wrong

ndr-brt:
With this change the builders of the basic launcher fails. Maybe we could set up a default value and print a warning if this has not been configured

paullatzelsperger:
> With this change the builders of the basic launcher fails. Maybe we could set up a default value and print a warning if this has not been configured

Are you sure it's the `launchers/basic` launcher that fails? just built and ran it successfully. 

Generally the `ids.webhook.address` is necessary as soon as contract negotiation is used. The only meaningful default would be `<HOST_WERE_RUNNING_ON>:<PORT>`. If we cannot get that at runtime, we should rather throw an exception if `ids.webhook.address` is not configured.

ndr-brt:
It failed because there wasn't a default value for it, now I added a default value (`http://localhost`) and a warning log.

For what I see, it's a mandatory configuration for the ids data-protocol, so yes, it should give an exception, but I didn't want to complicate the procedure to run the basic connector.
So the choices are:
- give a default value and print a warning
- throw exception if no value is set and modify the readme to make sure that everyone is aware of this before launching the basic connector

paullatzelsperger:
I just saw, that the basic connector has a dependency onto `data-protocols:ids`, which - strictly speaking - it doesn't need IMO. So here's what I would do: 
1. remove that dependency from `launchers/basic`
2. throw an exception if `ids.webhook.address` is not configured
3. add that config value to all other launchers (e.g. `launchers/ids-connector`) and the samples

If this is too much of a hassle I'm also fine with leaving this PR how it is now and solve points 1-3 in another PR

PR #495, opened 2022-01-11
Prevent override of RemoteMessageDispatcher at boot time

bscholtes1A:


paullatzelsperger (approved):

ndr-brt:
Could that be loaded by an upper level extension? After all the implementation class is located in the `core:base` module, I would register it in the `CoreServicesExtension`

jimmarino:
I think this PR should go through independent of the others issues described below.

It seems like core/base is doing a lot of things:

- Registering Jetty and Jersey 
- Registering SchemaRegistryImpl
- Contains NullVaultExtension
- Contains RemoteMessageDispatcherRegistryImpl

Now that we have simplified dependency ordering, can we push the module restructuring further? I have in mind:

- Place Jetty and Jersey in a separate extension that is loaded after the core extensions. Ensuring that downstream extensions require them through @Inject should be a straightforward task.

- Kill SchemaRegistryImpl as it is no longer used (correct, @paullatzelsperger?)

- Introduce a core/boot module. This module would only contain classes related to runtime boostrap (e.g. the extension loading and related infrastructure, monitor) currently in core/base

- The core/base module remains as a service extension that is loaded prior to other core extensions. This would only include RemoteMessageDispatcherRegistryImpl and the ExponentialWaitStrategy services (which it would register). 

- Create a BOM to make referencing these modules easy

If people agree with this assessment, we can open a separate issue for that further modularization.

bscholtes1A:
> I think this PR should go through independent of the others issues described below.
> 
> It seems like core/base is doing a lot of things:
> 
> * Registering Jetty and Jersey
> * Registering SchemaRegistryImpl
> * Contains NullVaultExtension
> * Contains RemoteMessageDispatcherRegistryImpl
> 
> Now that we have simplified dependency ordering, can we push the module restructuring further? I have in mind:
> 
> * Place Jetty and Jersey in a separate extension that is loaded after the core extensions. Ensuring that downstream extensions require them through @Inject should be a straightforward task.
> * Kill SchemaRegistryImpl as it is no longer used (correct, @paullatzelsperger?)
> * Introduce a core/boot module. This module would only contain classes related to runtime boostrap (e.g. the extension loading and related infrastructure, monitor) currently in core/base
> * The core/base module remains as a service extension that is loaded prior to other core extensions. This would only include RemoteMessageDispatcherRegistryImpl and the ExponentialWaitStrategy services (which it would register). RemoteMessageDispatcherRegistryImpl needs to be loaded here since it is required by the TPM and CNM.
> * Create a BOM to make referencing these modules easy
> 
> If people agree with this assessment, we can open a separate issue for that further modularization.

Hi @jimmarino 
do you want these points to be tackled in the present PR? Or can we merge it?

paullatzelsperger:
@bscholtes1A for now lets merge this PR, we can do what @jimmarino said in another issue/PR.

jimmarino:
> @bscholtes1A for now lets merge this PR, we can do what @jimmarino said in another issue/PR.

Yep, it should be done in another PR. I'll open another issue.

jimmarino:
Opened a new issue in #497.

PR #501, opened 2022-01-12
spi: make SELECT_ALL an empty criteria

### LINKED ISSUE #494: AssetSelectorExpression.SELECT_ALL is incompatible with CosmosDB
Running a query e.g. on an asset index implementation on CosmosDB with the `AssetSelectorExpression.SELECT_ALL` an error is returned:
```
Invalid query. Specified parameter name '@*' is invalid. Parameter names should be in the format of symbol '@' followed by a valid identifier. E.g. @param1
```

That's because the `SELECT` all selector defines `*` as left operand, and this is injected in the query as parameter (that in cosmos db is lead by a `@`)

Currently,  `AssetSelectorExpression.SELECT_ALL` is a list of criterion with a single element, but a "select all" in my opinion could be translated in an empty criterion list.
So my proposal is to avoid that `new Criterion("*", "=", "*")`

ndr-brt:
Fix #494


paullatzelsperger (changes requested):
The distinction between an empty query and the `SELECT_ALL` expression was implemented intentionally. The idea was to avoid a situation where someone forgets to specify a query, and this resulting in loading potentially a lot of assets from the index. 
The correct way is therefor to check whether a query is equal to `SELECT_ALL` and handle this case explicitly. 

I have noticed though that the `CosmosAssetIndex` does not in fact check/handle the `SELECT_ALL` case, which it should. That's what this PR should do.

ndr-brt:
What's the difference to do an empty query and a select all one? 
Both can lead to a system overload, if we want to avoid it we should set a validation where the query cannot be empty and cannot be a select all, or we should introduce pagination.
In most of the query languages/dsl (e.g. sql) the "select all" is no criterion at all, I think it's the best way to represent that concept.

paullatzelsperger:
The difference lies in the intent: it should be harder to accidentally fetch all items as this may stress the system. An empty query may produce no results or even raise an exception to indicate user error. 

So if actually all items are to be fetched, users must explicitly specify the `SELECT_ALL` constant. Even in SQL you have to specify the asterisk `SELECT * FROM yourtable;`

Agree on pagination.

ndr-brt:
The `*` in SQL doesn't mean "all the data", it means "all the fields", it's different.

In my opinion we shouldn't make things "harder" but safer and simpler.

Mandatory pagination on every "repository" will do the job.

paullatzelsperger:
I care less about whether the `SELECT_ALL` operation actually translates to `*=*` or anything else, but I do think that it should be an explicit operation and `AssetIndex` implementations should check for it.

ndr-brt:
Sorry but I still can't get this need for a select all criterion, I can't recall any system or tool that have this, why invent something new?
If the resources are all paginated there will be no risks, so it really seems an unnecessary complication to me. 

Will be good to hear other opinions, this PR was opened 15 days ago.

paullatzelsperger:
You're right, pagination will alleviate this. Pagination is required for issue #546 so I'm good with this PR as it is.

paullatzelsperger (approved):

PR #502, opened 2022-01-12
Implementation of the Command Queue

### LINKED ISSUE #386: Possibility of cancelling TransferProcesses
The EDC API should provide a mechanism to cancel running TransferProcesses in a clean way. To achieve this we are currently using the `TransferProcessStore.update` method to update the process state to ERROR. A problem with that, is that `TransferProcessListeners` are not called back, as with any other state change triggered within the `TransferProcessManager`. Additionally having process updates from different threads lead to race conditions.

This has already been provided as feedback for the planned TransferProcessManager redesign #330, but we would appreciate having a temporary solution ASAP. Our suggestion would be to add a `void cancelTransferProcess(String processId)` method to the TransferProcessManager interface which takes care of calling the listeners after moving the state to ERROR (see https://github.com/Agera-CatenaX/EclipseDataSpaceConnector/pull/7 for a solution we are working on as a possible contribution)

### LINKED ISSUE #470: Implement CommandQueue
## Problem statement

We'll run into situations, where persistent objects such as Transfer processes need to be modified from different threads, or even different runtimes, so we need to guarantee strong consistency, otherwise they could be overwritten and we could get an inconsistent state.

This proposal deals with the implementation of a command queue, which realizes the "command"-part of a CQRS-pattern. We'll present this using the example of the `TransferProcessManager` and `TransferProcess` objects, because that's where it'll be used first. We foresee that this construct will be applied to other persistent storages as well, for example the `ContractNegotiationManager`.

The main problem arises from the fact, that persistent entities (ie. the `TransferProcess`) need to be read and updated in a single transaction, thus guaranteeing that `read()` and `update()` operations happen atomically.

Typically in SQL-like environments a common way of doing this is a `SELCT ... FOR UPDATE` statement, which locks the row. Some databases, especially NoSQL-ones like CosmosDB don't have that feature, so we need to make this independent of the underlying database technology.

## Implementation considerations: Commands and Command handlers

The command queue framework consists of these major parts:
- a `CommandQueue`, which holds the command objects and could be based on persistent storage.
- a `Command`, which is a POJO holding all relevant information for the `CommandHandler`, and must be serializable
- a `CommandResult`, which indicates whether a command was successful, is retryable or failed fatally
- a `CommandHandler` which does the actual processing of the command and must be stateless.

## Requirements
- Implementors of a persistence backend must take of proper consistency and locking mechanisms, e.g. throuth a `TransactionContext`, or a persistent lock flag
- The `Command` object is a POJO that must be serializable to JSON.
- Neither the `Command` nor the `CommandHandler` must not store any references to persistent objects (as they could get stale)
- The `CommandHandler` may maintain a reference to collaborating objects (e.g. the `TransferProcessStore`)
- The `CommandHandler` is stateless
- `CommandHandlers` must perform their task on the current thread and cannot delegate to new threads as this might corrupt transaction integrity.
- `CommandHandlers` must validate the command immediately and return ERROR_FATAL if the command has become invalid or unprocessable. The corresponding DB entity could have been deleted in the meantime, etc.

## Limitations 
- No guarantees are made with respect to ordering of the commands, only that every command will get executed eventually.
- Commands depending on other commands are therefore not allowed
- Command handlers should do as little work as possible so as not to lock the entity longer than needed.

This is an amendment to #330, we can use #462 as a basis.

paullatzelsperger:
This PR adds the possibility to add actions (=`Command`s) to the TPM (= TransferProcessManager), which are executed at a certain point in time, namely after the regular state machine has ticked over. 

This ensures that those commands are processed sequentially and without interfering with the state machine. 

Most notable aspects: 
- clear separation between command ("what") and handler ("how")
- reusability of this infrastructure e.g. in the Contract Negotiation framework
- extensible `CommandQueue`
- ships with two commands (and their handlers) out-of-the-box: `DeprovisionTransferCommand` and `CancelTransferCommand`

_Note: In the course of this PR an unrelated utility feature called the `ObjectFactory` emerged, which allows to construct objects using reflection and dependency injection (leveraging EDC's `@Inject` annotation)._

closes #470 
closes #386


paullatzelsperger:
@algattik or @marcgs care to take a look?

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/DeprovisionCommandHandler.java
```diff
@@ -0,0 +1,42 @@
...
+    public DeprovisionCommandHandler() {
+        // used for reflection + dependency injection
+    }
+
+    public DeprovisionCommandHandler(TransferProcessStore store) {
+        this.store = store;
+    }
+
+    @Override
+    public void handle(DeprovisionRequest command) {
```
ndr-brt, line 25: Should this command be launched also by the `deprovision` method on the TPM? Because at the moment there's some duplicated code there

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/BoundedCommandQueue.java
```diff
@@ -0,0 +1,89 @@
...
+ * to add elements to an already full queue will fail and raise an {@link IllegalStateException}.
+ * <p>
+ * This queue is threadsafe, so every operation on it can be assumed atomic.
+ */
+public class BoundedCommandQueue implements CommandQueue {
+    private final ArrayBlockingQueue<Command> queue;
+    private final ReentrantLock lock;
+
+    public BoundedCommandQueue(int bound) {
+        lock = new ReentrantLock();
```
ndr-brt, line 26: the lock is not needed, as the `ArrayBlockingQueue` already implements a lock mechanism

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/BoundedCommandQueue.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+        return element.getFuture();
+    }
+
+    @Nullable
+    @Override
+    public Command dequeue() {
+        try {
+            lock.lock();
+            if (queue.isEmpty()) { //avoid blocking
```
ndr-brt, line 49: The `poll` method already does this "if empty then null otherwise dequeue"

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/BoundedCommandQueue.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.stream.IntStream;
+
+/**
+ * In-memory implementation of the {@link CommandQueue} that is backed by a bounded queue. This means that attempting
+ * to add elements to an already full queue will fail and raise an {@link IllegalStateException}.
+ * <p>
+ * This queue is threadsafe, so every operation on it can be assumed atomic.
+ */
+public class BoundedCommandQueue implements CommandQueue {
+    private final ArrayBlockingQueue<Command> queue;
```
ndr-brt, line 22: I would define it as a `Queue<Command>`, since it's the interface that we need

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/AsyncTransferProcessManager.java
```diff
@@ -278,6 +291,26 @@ private void run() {
...
+    private int processCommandQueue() {
+        var batchSize = 5;
+        var commands = commandQueue.dequeue(batchSize);
+        commands.forEach(command -> {
+            var commandResult = commandRunner.runCommand(command);
+            if (commandResult.failed()) {
+                //re-queue if possible
+                if (command.canRetry()) {
+                    monitor.warning(format("Could not process command [%s], will retry. error: %s", command.getClass(), commandResult.getFailureMessages()));
+                    commandQueue.enqueue(command);
```
ndr-brt, line 303: I fear that the command cannot be re-enqueued as its future would be already `exceptionally complete` so, every other completion on it would not work. 
Personally I would completely remove the `CompletableFuture` from the command/commandrunner, also because it's not serializable so completely useless in a persisted queue scenario.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/CommandHandlerRegistryImpl.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Map;
+
+public class CommandHandlerRegistryImpl implements CommandHandlerRegistry {
+    private final Map<Class<? extends Command>, CommandHandler<?>> registrations;
+
+    public CommandHandlerRegistryImpl() {
+        registrations = new HashMap<>();
+    }
+
+    public CommandHandlerRegistryImpl(Map<Class<? extends Command>, CommandHandler<?>> registrations) {
```
ndr-brt, line 17: not used constructor

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/AsyncTransferProcessManager.java
```diff
@@ -278,6 +291,26 @@ private void run() {
...
+                } else {
+                    monitor.severe(format("Command [%s] has exceeded its retry limit, will discard now", command.getClass()));
+                }
+            } else {
+                monitor.debug(format("Successfully processed command [%s]", command.getClass()));
+            }
+        });
+        return 0;
```
ndr-brt, line 311: typo?

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/DelegatingTransferProcessManager.java
```diff
@@ -30,6 +33,12 @@ public TransferInitiateResult initiateProviderRequest(DataRequest dataRequest) {
         return dataRequest.isSync() ? syncManager.initiateProviderRequest(dataRequest) : asyncManager.initiateProviderRequest(dataRequest);
     }
 
+    @Override
+    public CompletableFuture<Void> enqueueCommand(Command command) {
+        return asyncManager.enqueueCommand(command);
+        //todo: should the sync manager also have a command queue?
```
ndr-brt, line 39: in my opinion the `Sync` process manager should be merged with the async, adding an if condition on the `initiate` methods and setting different behaviours for different `DataRequest` type, because the empty-implementation methods on the sync one are telling us that maybe that was not the best design. Should be addressed on another issue, but I think that's why some questions like this comes to the surface

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/CommandExtension.java
```diff
@@ -0,0 +1,46 @@
...
+        factory = new ReflectiveObjectFactory(new InjectorImpl(), new InjectionPointScanner(), context);
+
+        CommandHandlerRegistryImpl registry = new CommandHandlerRegistryImpl();
+        context.registerService(CommandHandlerRegistry.class, registry);
+
+        registerDefaultCommands(registry);
+    }
+
+    private void registerDefaultCommands(CommandHandlerRegistryImpl registry) {
+        registry.register(factory.constructInstance(cast(CancelTransferCommandHandler.class)));
```
ndr-brt, line 42: do we need all the `ObjectFactory` effort since we can simply register our command handlers implementation like the one at row 43? Less annotations and no unused constructors needed.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/commands/CancelTransferCommand.java
```diff
@@ -0,0 +1,15 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.commands;
+
+import org.eclipse.dataspaceconnector.spi.transfer.command.Command;
+
+public class CancelTransferCommand extends Command {
```
ndr-brt, line 5: We could define a `TransferProcessCommand` abstraction, that already contains the `transferProcessId`

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/DeprovisionCommandHandler.java
paullatzelsperger: Those methods are now marked to be removed and so they should be. I'll replace them with adding commands.

marcgs:
IMO this change deserves a sample or integration test showing the command functionality working E2E.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/BoundedCommandQueue.java
paullatzelsperger: It's fine to declare the field as `Queue`, however supplying anything other than an `ArrayBlockingQueue` might change the behaviour of the `BoundedCommandQueue`, so we need to keep that in mind.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/AsyncTransferProcessManager.java
paullatzelsperger: not a typo, actually forgot to return the correct value :/

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/CommandExtension.java
paullatzelsperger: we do not need it, it kind of "fell out" of the development as a by-product, so I wanted to show how it's used. Maybe a sample would be a better place for it though.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/commands/CancelTransferCommand.java
paullatzelsperger: thought of that too. Might be worth the effort. That could also handle the "TP-not-found-case"

spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/command/CommandHandler.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * CommandHandlers are ephemeral objects that receive a {@link Command} object and act on it.
+ * They must have an empty default constructor because they will get instantiated through reflection by the {@link ObjectFactory}. Also, they MUST
+ * declare all their fields as {@link org.eclipse.dataspaceconnector.spi.system.Inject}.
+ *
+ * @param <T> The concrete type of {@link Command}
+ * @see ObjectFactory
+ */
+public interface CommandHandler<T extends Command> {
```
marcgs, line 13: What is the motivation for keeping command data and logic separated (or as mentioned in the PR description "clear separation between command ("what") and handler ("how"))? Keeping both data and behaviour in the same command class seems like a good encapsulation practice to me. The current implementation reminds me of an [anemic domain model](https://martinfowler.com/bliki/AnemicDomainModel.html).

ndr-brt: The `command` is a "declaration of intent", it is a request of an action, it can be represented by a simple data object. 
Keeping the commands separated from the handlers in our case is fundamental if we want to make the command serializable, to permit the implementation of a persistent queue for example.

An anemic domain model is a model where the domain model objects (like `TransferProcess`) are treated like POJO and all the logic (like, changing state etc...) is delegated to other services, definitely not our case.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/AsyncTransferProcessManager.java
paullatzelsperger: Yup, that's why it has the `transient` modifier. We might still need a way of notifying the original creator of the command (i.e. the "client") about its completion. Classic callbacks have the same problem of non-serializability.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/DelegatingTransferProcessManager.java
ndr-brt: Opened an issue about: #509

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/CancelTransferCommandHandler.java
```diff
@@ -0,0 +1,40 @@
...
+import org.eclipse.dataspaceconnector.spi.system.Inject;
+import org.eclipse.dataspaceconnector.spi.transfer.command.CommandHandler;
+import org.eclipse.dataspaceconnector.spi.transfer.store.TransferProcessStore;
+import org.eclipse.dataspaceconnector.transfer.core.commandqueue.commands.CancelTransferCommand;
+
+import static java.lang.String.format;
+
+public class CancelTransferCommandHandler implements CommandHandler<CancelTransferCommand> {
+    @Inject
+    private TransferProcessStore store;
```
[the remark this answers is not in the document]
paullatzelsperger: The handlers will have to use the store, because that's where the modification will take place albeit in a synchronized and deterministic way. 
However, your comment touches on another important question, which is "how do commands get into the system?". There are basically two options: 
1. Use `CommandQueue.enqueue()`
2. Use `TransferProcessManager.enqueueCommand()`

Both of those options suffer the same drawback, i.e. exposing more API surface than necessary. I chose to go with option 2 because then we could make the queue invisible (= not registered in the context) altogether.

spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/command/CommandHandler.java
paullatzelsperger: > IMO this change deserves a sample or integration test showing the command functionality working E2E. Besides some additional documentation would be required for developers to understand that commands should be used instead of directly using `TransferProcessStore` for modifying TransferProcesses.

Yup, a sample demonstrating the command queue is definitely a great idea. Will create one. Also, documentation about this new feature is non-existent at this time, so that'll go on my list too :)

marcgs: `The command is a "declaration of intent", it is a request of an action, it can be represented by a simple data object.
Keeping the commands separated from the handlers in our case is fundamental if we want to make the command serializable, to permit the implementation of a persistent queue for example.`

Naming seems a bit off in this case... `Command` implies that there is behaviour in it (it is a behavioural design pattern in the end). Consider renaming to match the real purpose of each class.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/CancelTransferCommandHandler.java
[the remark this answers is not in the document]
marcgs: A clear and concise public API for developers using EDC will definitely help preventing errors, so I would go with the second option too. 

In this PR we are introducing a mechanism to perform operations on TransferProcesses through commands, but it is still possible for an EDC developer to modify TPs using the `TransferProcessStore`. My point was that we should try to reduce the risk of this erroneous usage, if not possible by reducing visibility then by clearly documenting that this is a class internal to the framework.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/BoundedCommandQueue.java
ndr-brt: the `drainTo(Collection<? super E> c, int maxElements)` method could be used instead, it's part of the `BlockingQueue` interface

spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/command/CommandHandler.java
ndr-brt: I think you are talking about the classic GoF command pattern, but this is not an application of that pattern, it's an excerpt from the CQRS pattern, which provides the existence of `commands` as data objects and `command handlers` as behaviour classes. I find it very clever and neat, I used it a lot in the past.
What names would you suggest for these two?

marcgs: Ok I see. `Command` and `CommandHandler` are just very broad, as a developer I would not know what exactly I can do with it just by looking at the code + javadoc. As mentioned above a sample and documentation on how this feature should be used will definitely help me understanding (as well as other EDC developers). I think it is key to make it clear that commands are required to perform operations on persistent objects (like TransferProcesses) to prevent concurrency issues. This might be enough and no class renaming is necessary then.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/AsyncTransferProcessManager.java
```diff
@@ -123,43 +125,8 @@ public TransferInitiateResult initiateProviderRequest(DataRequest dataRequest) {
...
-            monitor.info("Request already deprovisioning or deprovisioned.");
-        } else {
-            monitor.info("starting to deprovision data request " + processId);
-            process.transitionCompleted();
-            process.transitionDeprovisionRequested();
-            transferProcessStore.update(process);
-        }
-
-        return Result.success(TransferProcessStates.from(process.getState()));
+    public void enqueueCommand(Command command) {
```
marcgs, line 128: Have you considered implementing lifecycle operations in the `TransferProcessManager` (provisionInitialProcesses, sendOrProcessProvisionedRequests, ...) or at least parts of it as commands too?

paullatzelsperger: yes we have, and we decided against it for various reasons. chief amongst those are that the "normal" state machine transitions should not be influenced by the command queue. We also want to guarantee fairness and that the SM advances, even it there are loads of commands in the queue.

spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/command/CommandHandler.java
paullatzelsperger: I added the sample `samples/04.2-modify-transferprocess`. Does that help in understanding? Also, are you ok with the naming?

samples/04.2-modify-transferprocess/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+Also, most clients will already have a reference to the `TransferProcessManager`, so little change needs to be done. Instead simply
+do:
+```java
+tpm.enqueueCommand(new CheckTimeoutCommand(3, TransferProcessStates.IN_PROGRESS, Duration.ofSeconds(10)));
+```
+
+that will eventually get processed by the `TransferProcessManager`, resulting in log output similar to this: 
+
+```bash
```
marcgs, line 102: I would add the commands required to run the sample:
```
 ./gradlew samples:04.2-modify-transferprocess:provider:build
java -Dedc.fs.config=samples/04.2-modify-transferprocess/consumer/config.properties -jar samples/04.2-modify-transferprocess/consumer/build/libs/consumer.jar
```

marcgs (approved):

samples/04.2-modify-transferprocess/watchdog/src/main/java/org/eclipse/dataspaceconnector/samples/sample042/CheckTimeoutCommand.java
```diff
@@ -0,0 +1,32 @@
+package org.eclipse.dataspaceconnector.samples.sample042;
+
+import org.eclipse.dataspaceconnector.spi.transfer.command.Command;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcessStates;
+
+import java.time.Duration;
+
+public class CheckTimeoutCommand extends Command {
```
marcgs, line 8: `CheckTransferProcessTimeoutCommand` is a more informative name what is really being handled by this command

spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/command/CommandHandler.java
marcgs: It certainly does, thanks for the efforts @paullatzelsperger . Naming is fine but as `Command` and `CommandHandler` are such generic concepts I would move these classes out of the `transfer-spi` module and into the `core-spi`. WDYT?

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/commands/CancelTransferCommand.java
```diff
@@ -0,0 +1,9 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.commands;
+
+public class CancelTransferCommand extends TransferProcessCommand {
```
jimmarino, line 3: Can you pls add documentation, e.g. the semantics of this command?

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/commands/DeprovisionRequest.java
```diff
@@ -0,0 +1,8 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.commands;
+
+public class DeprovisionRequest extends TransferProcessCommand {
```
jimmarino, line 3: Can you pls add documentation, e.g. the semantics of this command?

spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/command/CommandHandler.java
```diff
@@ -0,0 +1,18 @@
+package org.eclipse.dataspaceconnector.spi.transfer.command;
+
+import org.eclipse.dataspaceconnector.spi.ObjectFactory;
+
+/**
+ * CommandHandlers are ephemeral objects that receive a {@link Command} object and act on it.
+ * They must have an empty default constructor because they will get instantiated through reflection by the {@link ObjectFactory}. Also, they MUST
```
jimmarino, line 7: I don't think it should be a requirement to have a default ctor or require that handler instances be instantiated reflectively via the Injection infrastructure. It should be possible to register handler instances in the runtime.

jimmarino:
@paullatzelsperger can we separate out the Injector-related commits into another PR?

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/ObjectFactory.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Factory object that is used to generate instances of objects based on their class.
+ * One way to implement this is to generate a new instance on demand using reflection and dependency injection.
+ */
+@FunctionalInterface
+@Feature("edc:core:object-factory")
+public interface ObjectFactory {
+    /**
+     * Gets (and potentially creates) a new instance of a commandHandler
```
jimmarino, line 13: Outdated doc?

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/ObjectFactory.java
```diff
@@ -0,0 +1,20 @@
...
+     *
+     * @param clazz The object's class
+     * @return An instance of the object, potentially newly created, or null if none could be found.
```
jimmarino, line 16: Would there ever be a case where null is returned? Shouldn't a runtime exception be thrown instead?

spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/command/CommandRunner.java
```diff
@@ -0,0 +1,38 @@
...
+        if (handler == null) {
+            command.increaseErrorCount();
+            return Result.failure("No command handler found for command type " + commandClass);
+        }
+        try {
+            handler.handle(command);
+            return Result.success();
+        } catch (RuntimeException ex) {
+            command.increaseErrorCount();
+            return Result.failure(ex.getMessage());
```
jimmarino, line 34: The stacktrace should be sent to the monitor

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/commands/TransferProcessCommand.java
```diff
@@ -0,0 +1,15 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.commands;
+
+import org.eclipse.dataspaceconnector.spi.transfer.command.Command;
+
+public class TransferProcessCommand extends Command {
```
jimmarino, line 5: Please add doc

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/CancelTransferCommandHandler.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    @Override
+    public Class<CancelTransferCommand> getType() {
+        return CancelTransferCommand.class;
+    }
+
+    @Override
+    protected void modify(TransferProcess process) {
+        process.transitionError("Cancelled");
```
jimmarino, line 20: Should the TP be transitioned to an error state or should we have a canceled state?

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/TransferProcessCommandHandler.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public void handle(TransferProcessCommand command) {
+        var transferProcessId = command.getTransferProcessId();
+        var transferProcess = store.find(transferProcessId);
+        if (transferProcess == null) {
+            throw new EdcException(format("Could not find TransferProcess with ID [%s]", transferProcessId));
+        } else {
+            modify(transferProcess);
+            store.update(transferProcess);
```
jimmarino, line 26: What happens in the case that the implementation of modify(..) does not change the TP? Should it still be saved regardless or can modify return a boolean indicating that the TP was updated?

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/commands/DeprovisionRequest.java
```diff
@@ -0,0 +1,8 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.commands;
```
jimmarino, line 1: license needs to be added

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/commands/CancelTransferCommand.java
```diff
@@ -0,0 +1,9 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.commands;
```
jimmarino, line 1: license needs to be added

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/commands/TransferProcessCommand.java
```diff
@@ -0,0 +1,15 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.commands;
```
jimmarino, line 1: license needs to be added

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/CancelTransferCommandHandler.java
```diff
@@ -0,0 +1,22 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.handlers;
```
jimmarino, line 1: license needs to be added

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/DeprovisionCommandHandler.java
```diff
@@ -0,0 +1,22 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.handlers;
```
jimmarino, line 1: license needs to be added

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/TransferProcessCommandHandler.java
```diff
@@ -0,0 +1,31 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.handlers;
```
jimmarino, line 1: license needs to be added

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/BoundedCommandQueue.java
```diff
@@ -0,0 +1,57 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue;
```
jimmarino, line 1: license needs to be added

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/CommandHandlerRegistryImpl.java
```diff
@@ -0,0 +1,27 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue;
```
jimmarino, line 1: license needs to be added

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/CommandExtension.java
```diff
@@ -0,0 +1,38 @@
+package org.eclipse.dataspaceconnector.transfer.core;
```
jimmarino, line 1: license needs to be added

core/transfer/src/test/java/org/eclipse/dataspaceconnector/transfer/core/command/BoundedCommandQueueTest.java
```diff
@@ -0,0 +1,103 @@
+package org.eclipse.dataspaceconnector.transfer.core.command;
```
jimmarino, line 1: license needs to be added

core/transfer/src/test/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/CancelTransferCommandHandlerTest.java
```diff
@@ -0,0 +1,53 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.handlers;
```
jimmarino, line 1: license needs to be added

core/transfer/src/test/java/org/eclipse/dataspaceconnector/transfer/core/command/TestCommand.java
```diff
@@ -0,0 +1,6 @@
+package org.eclipse.dataspaceconnector.transfer.core.command;
```
jimmarino, line 1: license needs to be added

core/transfer/src/test/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/DeprovisionCommandHandlerTest.java
```diff
@@ -0,0 +1,63 @@
+package org.eclipse.dataspaceconnector.transfer.core.commandqueue.handlers;
```
jimmarino, line 1: license needs to be added

spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/command/CommandHandler.java
paullatzelsperger: not a bad idea, as we might re-use them later in the contract negotiation manager.

spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/command/CommandHandler.java
paullatzelsperger: it certainly is :)

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/TransferProcessCommandHandler.java
paullatzelsperger: Sounds a bit strange to have a `TransferProcessCommand` that does not actually modify anything, but I added a boolean return value to indicate the need for an update.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/commandqueue/handlers/CancelTransferCommandHandler.java
paullatzelsperger: that would certainly be more appropriate. Will do in #515

paullatzelsperger:
> @paullatzelsperger can we separate out the Injector-related commits into another PR?

I assume you mean the `ObjectFactory` etc.? I can remove it from this PR, and re-add it in another one.

Issue #497, opened 2022-01-11
Further Modularize core/base


jimmarino:
It seems like core/base is doing a lot of things:

- Registering Jetty and Jersey
- Registering SchemaRegistryImpl
- Contains NullVaultExtension
- Contains RemoteMessageDispatcherRegistryImpl

Now that we have simplified dependency ordering, can we push the module restructuring further? I have in mind:

- Place Jetty and Jersey in a separate extension that is loaded after the core extensions. Ensuring that downstream extensions require them through @Inject should be a straightforward task.

- Kill SchemaRegistryImpl as it is no longer used (correct, @paullatzelsperger?)

- Introduce a core/boot module. This module would only contain classes related to runtime bootstrap (e.g. the extension loading and related infrastructure, monitor) currently in core/base

- The core/base module remains as a service extension that is loaded prior to other core extensions. This would only include RemoteMessageDispatcherRegistryImpl and the ExponentialWaitStrategy services (which it would register). RemoteMessageDispatcherRegistryImpl needs to be loaded here since it is required by the TPM and CNM.

- Create a BOM to make referencing these modules easy

jimmarino:
@ndr-brt is this something you have cycles to help with? If so, ping me and we can discuss.

bscholtes1A:
Hi @jimmarino 
concerning this point:
> The core/base module remains as a service extension that is loaded prior to other core extensions. This would only include RemoteMessageDispatcherRegistryImpl and the ExponentialWaitStrategy services (which it would register). RemoteMessageDispatcherRegistryImpl needs to be loaded here since it is required by the TPM and CNM.

There is today a `CoreServicesExtension` which :
- registers the `HealthCheckService`
- register the parser in the `PrivateKeyResolver`
- register the `RetryPolicy`
- register the `OkHttpClient`

If I am properly following, you suggest that `CoreServicesExtension` registers only the `RemoteMessageDispatcherRegistryImpl` and the `ExponentialWaitStrategy`. If yes, where should we put all the `CoreServicesExtension` logic described just above?

Benjamin.

jimmarino:
I think those are OK to keep in core

jimmarino:
Here are a couple of additional items we need to do:

- Move WebService our of core-spi and into its own spi
- More ResponseStatus and ResponseFailure into their own package in core-spi. These classes are useful outside of the transfer implementation, e.g. from the data plane, which should not have visibility into the transfer services
[...]

PR #508, opened 2022-01-13
extract jersey and jetty extensions

ndr-brt:
Ref #497 
I ended up with this, I separated jersey from jetty just because they were two different services (`WebService` and `JettyService`).
Maybe they will stay better together and named as `Jetty`, because at the end the server is lunched by the JettyService, so maybe it deserves to extend the `WebService` interface.
Open to a discussion


jimmarino (commented):
LGTM, but can we move the Jetty and Jersey modules to extension/http? That will be consistent with how the other extensions are structured and allow us to easily create a BOM.

ndr-brt:
@jimmarino you're right, created http subfolder and added a BOM

PR #511, opened 2022-01-14
build system: avoid deprecated gradle.kts features to specify main class

arosetti:
This is my first and not very significant contribution :)

Here is a warning-free way of specifying the main class: [JavaApplication].(https://docs.gradle.org/current/javadoc/org/gradle/api/plugins/JavaApplication.html)


ndr-brt (approved):
looks good :+1:

PR #493, opened 2022-01-11
Removes all IDS-REST modules and related code

paullatzelsperger:
This PR removes all java modules related to IDS REST and related code, most notably:
- ` ids-api-transfer`: previously used for initiating data transfers. Superseded by the Contract Negotiation + Transfer via IDS Multipart.
- `ids-api-catalog`: previously used for querying a connector's catalog. Superseded by the Federated Catalog.
- `DataRequestMessageSender` and `QueryMessageSender`: related message senders for the aforementioned endpoints.
- `PolicyRegistry`: removed as policies are now attached to `ContractOffer` objects.

_Note: Our "numbered" Samples have been checked and updated, however all samples in the `other` folder have not and might not work anymore. Respective developers have been notified and absent an action on their part those samples will get removed subsequently!_

Closes #359

PR #512, opened 2022-01-14
data-protocols: removed deprecated IdsPolicyService and related extension

ndr-brt:
I think we missed this in #493


bscholtes1A (approved):

paullatzelsperger (commented):
It might be used in downstream projects, but we'll just have to remove it there too then.

paullatzelsperger:
@jimmarino i think we should merge this once all conflicts are resolved.

ndr-brt:
conflict resolved

ndr-brt:
There's a test that sometimes fails randomly for a timeout exception in `MultipartDispatcherIntegrationTest`, however I forced push to re-execute the CI, let's see

PR #521, opened 2022-01-17
extensions: make audience an intrinsic attribute of the identity service

### LINKED ISSUE #520: Error: "Token audience did not match required audience: null"
I'm trying to make two connectors communicate using DAPS, but, on the validation of the token on the provider side I get this error:

`Token audience did not match required audience: null`

Looking at the code I think that's because a `null` is passed as audience value to the `verifyJwtToken` of the IdentityService:
Should that value read from the configuration instead?

https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/ce440245189ca48f2a10a63c51c7f8b223660756/data-protocols/ids/ids-api-multipart-endpoint-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/controller/MultipartController.java#L98

EDIT: not strictly releated, but I noticed this `DapsService` interface and implementation that are not used by anyone, can they just go away?
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/6193ffa2fcc1e9b081851ac26570e5da456b572b/data-protocols/ids/ids-spi/src/main/java/org/eclipse/dataspaceconnector/ids/spi/daps/DapsService.java

ndr-brt:
Fix #520 . This pr brings:
* audience is not something you need to provide to the identity service to validate a token but it gets loaded from the configuration of the identity service
* Removed `DapsService` as it is unused
* Bump omejdn version for integration tests


paullatzelsperger (approved):
Seems fine to me.

PR #524, opened 2022-01-18
contract negotiation: add contract negotiation listener

### LINKED ISSUE #352: Listener for contract negotiation
Like the TransferProcessManagerImpl, the Provider- and ConsumerContractNegotiationManagerImpl should be observable and provide the possibility to notify listeners about state changes of ContractNegotiation instances. Therefore, a ContractNegotiationListener interface should be added to the spi module. The ContractNegotiationManagerImpls should extend Observable and notify all listeners when a state is changed.

Then, the ContractNegotiationIntegrationTest can be refactored to use listeners instead of the currently used SignalingInMemoryContractNegotiationStore.

ronjaquensel:
This PR makes the `Provider- and ConsumerContractNegotiationManagers` observable by providing the possibility to register listeners. The listeners will be notified about every state change on a `ContractNegotiation` instance. The PR also refactors the `ContractNegotiationIntegrationTest` to use listeners instead of signaling stores to determine when the desired end state has been reached.

Closes #352

### DIFF
--- a/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
+++ b/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
@@ -16,7 +16,6 @@ package org.eclipse.dataspaceconnector.contract.negotiation;
 
 import org.eclipse.dataspaceconnector.contract.common.ContractId;
 import org.eclipse.dataspaceconnector.spi.contract.negotiation.ConsumerContractNegotiationManager;
-import org.eclipse.dataspaceconnector.spi.contract.negotiation.ContractNegotiationListener;
 import org.eclipse.dataspaceconnector.spi.contract.negotiation.ContractNegotiationObservable;
 import org.eclipse.dataspaceconnector.spi.contract.negotiation.NegotiationWaitStrategy;
 import org.eclipse.dataspaceconnector.spi.contract.negotiation.response.NegotiationResult;
@@ -42,7 +41,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.function.Consumer;
 
 import static java.lang.String.format;
 import static org.eclipse.dataspaceconnector.contract.common.ContractId.DEFINITION_PART;
@@ -486,15 +484,6 @@ public class ConsumerContractNegotiationManagerImpl extends ContractNegotiationO
                 }
             }
         }
-    }
-    
-    /**
-     * Invokes a given action on all registered {@link ContractNegotiationListener}s.
-     *
-     * @param action the action to invoke.
-     */
-    private void invokeForEach(Consumer<ContractNegotiationListener> action) {
-        getListeners().forEach(action);
     }
 
     /**
--- a/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
+++ b/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
@@ -148,20 +148,11 @@ public class ConsumerContractNegotiationManagerImpl extends ContractNegotiationO
         Result<ContractOffer> result = validationService.validate(token, contractOffer, latestOffer);
         negotiation.addContractOffer(contractOffer); // TODO persist unchecked offer of provider?
         if (result.failed()) {
-            //if (result.isCounterOfferAvailable()) {
-            //    negotiation.addContractOffer(result.getCounterOffer());
-            //    monitor.debug("[Consumer] Contract offer received. A counter offer is available.");
-            //    negotiation.transitionOffering();
-            //    negotiationStore.save(negotiation);
-            //    invokeForEach(l -> l.consumerOffering(negotiation));
-            //} else {
-            // If no counter offer available + validation result invalid, decline negotiation.
             monitor.debug("[Consumer] Contract offer received. Will be rejected.");
             negotiation.setErrorDetail("Contract rejected."); //TODO set error detail
             negotiation.transitionDeclining();
             negotiationStore.save(negotiation);
             invokeForEach(l -> l.declining(negotiation));
-            //}
         } else {
             // Offer has been approved.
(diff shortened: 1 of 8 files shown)


core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -146,19 +150,24 @@ public NegotiationResult offerReceived(ClaimToken token, String negotiationId, C
             //    negotiation.addContractOffer(result.getCounterOffer());
             //    monitor.debug("[Consumer] Contract offer received. A counter offer is available.");
             //    negotiation.transitionOffering();
+            //    negotiationStore.save(negotiation);
```
ndr-brt, line 153: If this commented out code is there because we miss something I would open an issue and delete that block.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -404,15 +425,17 @@ private int declineContractOffers() {
             if (response.isCompletedExceptionally()) {
                 process.transitionDeclining();
                 negotiationStore.save(process);
+                invokeForEach(l -> l.declining(process));
                 monitor.debug(format("[Consumer] Failed to send contract rejection. ContractNegotiation %s stays in state %s.",
                         process.getId(), ContractNegotiationStates.from(process.getState())));
                 continue;
```
ndr-brt, line 431: Not introduced by this PR but could this code be refactored to avoid the `continue` statement? 
Pretty old-fashioned and dangerous.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -452,6 +475,15 @@ private void run() {
             }
         }
     }
+    
+    /**
+     * Invokes a given action on all registered {@link ContractNegotiationListener}s.
+     *
+     * @param action the action to invoke.
+     */
+    private void invokeForEach(Consumer<ContractNegotiationListener> action) {
```
ndr-brt, line 484: this could stay in the `ContractNegotiationObservable` abstract class, but also in `Observable` directly

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java
```diff
@@ -297,13 +308,15 @@ private int checkProviderOffering() {
             if (response.isCompletedExceptionally()) {
                 negotiation.transitionOffering();
                 negotiationStore.save(negotiation);
+                invokeForEach(l -> l.providerOffering(negotiation));
                 monitor.debug(format("[Provider] Failed to send contract offer with id %s. ContractNegotiation %s stays in state %s.",
                         currentOffer.getId(), negotiation.getId(), ContractNegotiationStates.from(negotiation.getState())));
                 continue;
```
ndr-brt, line 314: same as below about the `continue` statement

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java
```diff
@@ -399,6 +414,7 @@ private int checkConfirming() {
             if (response.isCompletedExceptionally()) {
                 negotiation.transitionConfirming();
                 negotiationStore.save(negotiation);
+                invokeForEach(l -> l.confirming(negotiation));
                 monitor.debug(format("[Provider] Failed to send contract agreement with id %s. ContractNegotiation %s stays in state %s.",
                         agreement.getId(), negotiation.getId(), ContractNegotiationStates.from(negotiation.getState())));
                 continue;
```
ndr-brt, line 420: uhm... there are a lot of `continue`, maybe this refactor deserves another issue to avoid overwhelming this?

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
ronjaquensel: Good call, I'll move the method to `Observable`, then the negotiation managers and the async transfer process manager can share it.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
ronjaquensel: This code was a first attempt at introducing counter offers to the negotiation, but since we currently don't have a service or hook for creating counter offers, it was commented out. I'll remove the commented out code and check if there is already an issue for integrating counter offers. If not, I'll create one.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java
ronjaquensel: What's so bad/dangerous about `continue` statements? But sure, could be refactored in another PR.

ndr-brt: they lead to a not-so-readable code flow, at least for an hi-level programming language. In this case, for example, an `else` condition would give better meaning on what the code it's doing.

Looking better at this code, the real error is using the `isCompletedExceptionally` on a future that could be still not completed, better use the `whenComplete`, otherwise that condition could never be meet. Will open an issue for this, sorry for the flooding

ronjaquensel: Makes sense, thanks for opening the issue.

ndr-brt (approved):

paullatzelsperger:
Looks good to me. However, we will eventually run into the same consistency problems as we have with the `TransferProcessManager`, so the addition of the `CommandQueue` will be necessary.

paullatzelsperger (approved):

PR #529, opened 2022-01-19
data address: add null checks and adapt unit tests

### LINKED ISSUE #452: Add Unit Tests to `DataAddress` (and fix possible issues)
The class `DataAddress` is currently missing some Unit Tests. At the same time I assume there might be some issues within the implementation.

Here a list of things to address within this issue. But maybe there are more:

### 1. `KeyName` and `Type` Setter

I would assume the setter are only working, when the corresponding property was set before in the builder.

```java
public void setType(String type) {
        Objects.requireNonNull(type);
        properties.replace(TYPE, type);
}
```

```java
public void setKeyName(String keyName) {
        Objects.requireNonNull(keyName);
        properties.replace(KEYNAME, keyName);
}
```

### 2. Required Property of the Builder

The builder does not always check for required properties. It only validates that if there is at least one property. Looking at the null-check I assume this is an implementation error.

```java
public DataAddress build() {
            if (!address.getProperties().isEmpty()) {
                Objects.requireNonNull(address.getType(), "type");
            }
            return address;
}
```

### 3. Builder method `properties(Map<String, ?> properties)` converts object to String using `toString()`

The `DataAddress` has a `Map<String, String` which is filled by the builder method `properties(Map<String, ?> properties)`.
To convert the objects into a String the `toString()` method is called. This looks not very safe and looking at the comment I assume there is already a situation, where this doesn't work properly. Maybe change it to `properties(Map<String, String> properties)`?

```java
public Builder properties(Map<String, ?> properties) {
            // ArtifactRequestMessageImpl#urifyObjects (line 176): this "feature" converts every string starting with
            // 'http' to become a URI.
            // Thus, sometimes the <String, String> map can contain ... you guessed it ... URIs :(
            properties.forEach((k, v) -> address.properties.put(k, v.toString()));
            return this;
}
```

bcronin90:
Closes #452 

Check for unacceptable null keys/values in data addresses. Update unit tests to account for this.

Brendan Cronin brendan.cronin@daimler.com, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)


extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetLoaderLoaderTest.java
```diff
@@ -90,14 +90,13 @@ void setup() {
...
-        return Asset.Builder.newInstance().id(id).name(name).version("1").build();
+        return Asset.Builder.newInstance().id(id).name(name).version("1").contentType("type").build();
     }
 
     private DataAddress createDataAddress(Asset asset) {
         return DataAddress.Builder.newInstance()
-                .type("test-asset")
+                .type(asset.getContentType())
                 .keyName("test-keyname")
-                .properties(flatten(asset))
```
ndr-brt, line 100: now the flatten method is unused and can be deleted

extensions/in-memory/assetindex-memory/src/test/java/org/eclipse/dataspaceconnector/metadata/memory/InMemoryAssetIndexTest.java
```diff
@@ -177,14 +177,13 @@ private Asset createAsset(String name) {
...
     private Asset createAsset(String name, String id) {
-        return Asset.Builder.newInstance().id(id).name(name).version("1").build();
+        return Asset.Builder.newInstance().id(id).name(name).version("1").contentType("type").build();
     }
 
     private DataAddress createDataAddress(Asset asset) {
         return DataAddress.Builder.newInstance()
-                .type("test-asset")
                 .keyName("test-keyname")
-                .properties(flatten(asset))
```
ndr-brt, line 187: now the flatten method is unused and can be deleted

ndr-brt (approved):

jimmarino:
@paullatzelsperger can you take a look at this?

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/DataAddress.java
```diff
@@ -81,28 +84,23 @@ public Builder type(String type) {
...
             address.properties.put(key, value);
             return this;
         }
 
-        public Builder properties(Map<String, ?> properties) {
-            // ArtifactRequestMessageImpl#urifyObjects (line 176): this "feature" converts every string starting with
-            // 'http' to become a URI.
-            // Thus, sometimes the <String, String> map can contain ... you guessed it ... URIs :(
-            properties.forEach((k, v) -> address.properties.put(k, v.toString()));
+        public Builder properties(Map<String, String> properties) {
```
paullatzelsperger, line 92: has the problem, that the ArtifactRequestMessageImpl converts strings to URIs, been taken care of elsewhere?

bcronin90: This was really only a problem in some tests, where assets were flattened and passed directly to the properties() method. I don't think that currently serves any purpose. In any other case using Map<String, String> has not been an issue.

The idea with this change is that we shouldn't blindly cast anything to String. It'd be cleaner to just let anything that would generate non-String properties just typecast them itself. Which currently no component does.

So my current solution would be to not solve it at all. Does that make sense?

launchers/data-loader-cli/src/test/resources/assets.json
```diff
@@ -1 +1 @@
...
...:"someval","type":"test-dataaddress-4"},"type":"test-dataaddress-4"}},{"asset":{"properties":{"asset:prop:name":"test-asset-5","asset:prop:version":"1.0","asset:prop:id":"test-asset-5"}},"dataAddress":{"properties":{"someprop":"someval","type":"test-dataaddress-5"},"type":"test-dataaddress-5"}},{"asset":{"properties":{"asset:prop:name":"test-asset-6","asset:prop:version":"1.0","asset:prop:id":"test-asset-6"}},"dataAddress":{"properties":{"someprop":"someval","type":"test-dataaddress-6"},"type":"test-dataaddress-6"}},{"asset":{"properties":{"asset:prop:name":"test-asset-7","asset:prop:version":"1.0","asset:prop:id":"test-asset-7"}},"dataAddress":{"properties":{"someprop":"someval","type":"test-dataaddress-7"},"type":"test-dataaddress-7"}},{"asset":{"properties":{"asset:prop:name":"test-asset-8","asset:prop:version":"1.0","asset:prop:id":"test-asset-8"}},"dataAddress":{"properties":{"someprop":"someval","type":"test-dataaddress-8"},"type":"test-dataaddress-8"}},{"asset":{"properties":{"asset:prop:name":"test-asset-9","asset:prop:version":"1.0","asset:prop:id":"test-asset-9"}},"dataAddress":{"properties":{"someprop":"someval","type":"test-dataaddress-9"},"type":"test-dataaddress-9"}}]
```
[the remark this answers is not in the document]
bcronin90: I just realize I never commented on these. Thanks for the thorough review. Since the points are all fixed, I'll set them to resolved. Please notify me if there's anything I missed.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/DataAddress.java
paullatzelsperger: This didn't just occur in tests, also in actual running code. I'm fine with leaving it as it is, as long as the end-to-end tests work, which unfortunately has to be done by hand. 

We cannot not solve it if the problem persists.

bcronin90: That's a valid concern. Is there a readme for those tests? I'm not sure where to look.

MoritzKeppler: would be great to document which tests need to be executed in CONTRIBUTION.md so that we don't have to remind anybody and can just refer to the docs, @paullatzelsperger 

One observation: the properties method does not show a consistent behaviour over all builders we have (sometimes it adds values, sometime it overrides the whole map, sometime it string-object but stored as string-string, sometimes string-object and stored as string-object) - so this change feels like a step to a more consistent behaviour.

paullatzelsperger: It would be even better to have those e2e tests run automatically on every gradle build, which is currently being worked on. It will be able to run e.g. a file transfer from withing a JUnit test. If we document this now (which is fine with me), we'll have to remove it again soon.

On the other topic: I agree, Builders are somewhat inconsis tent, but that will require a broader refactoring. I propose:
- have a single final field (as opposed to invoking the object's CTor in the `build()` method
- if there are extensible properties, we have an `addProperty()` which adds a prop, and a `properties()` which replaces the entire map
WDYT?

paullatzelsperger: OK, just ran several tests on my local installation, all went fine. So I guess the original problem doesn't exist anymore. Will approve the PR.

paullatzelsperger (approved):

PR #530, opened 2022-01-19
core: evaluate futures correctly

### LINKED ISSUE #526: Bad use of isCompletedExceptionally method
I noticed, on `ConsumerContractNegotiationManagerImpl` and `ProviderContractNegotiationManagerImpl`

that there are some usages of the method `isCompletedExceptionally` on a future, e.g.:
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/6193ffa2fcc1e9b081851ac26570e5da456b572b/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java#L277

Looking at the `isCompletedExceptionally` documentation:
> Returns true if this CompletableFuture completed exceptionally, in any way. Possible causes include cancellation, explicit invocation of completeExceptionally, and abrupt termination of a CompletionStage action.
> Returns:
> true if this CompletableFuture completed exceptionally

But at this point we cannot guarantee that the future is already completed (in any manner) without calling `get` or `join` (and we don't want to). 
We should use the async method `whenComplete` to set the behavior to execute on completion.

ndr-brt:
Fix #526

### DIFF
--- a/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
+++ b/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
@@ -281,10 +281,12 @@ public class ConsumerContractNegotiationManagerImpl implements ConsumerContractN
                     .whenComplete((response, throwable) -> {
                         if (throwable == null) {
                             process.transitionRequested();
+                            negotiationStore.save(process);
                             monitor.debug(String.format("[Consumer] ContractNegotiation %s is now in state %s.",
                                     process.getId(), ContractNegotiationStates.from(process.getState())));
-                            negotiationStore.save(process);
                         } else {
+                            process.transitionRequesting();
+                            negotiationStore.save(process);
                             String message = format("[Consumer] Failed to send contract offer with id %s. ContractNegotiation %s stays in state %s.",
                                     offer.getId(), process.getId(), ContractNegotiationStates.from(process.getState()));
                             monitor.debug(message, throwable);
@@ -311,10 +313,12 @@ public class ConsumerContractNegotiationManagerImpl implements ConsumerContractN
                     .whenComplete((response, throwable) -> {
                         if (throwable == null) {
                             process.transitionOffered();
+                            negotiationStore.save(process);
                             monitor.debug(String.format("[Consumer] ContractNegotiation %s is now in state %s.",
                                     process.getId(), ContractNegotiationStates.from(process.getState())));
-                            negotiationStore.save(process);
                         } else {
+                            process.transitionOffering();
+                            negotiationStore.save(process);
                             String message = format("[Consumer] Failed to send contract offer with id %s. ContractNegotiation %s stays in state %s.",
                                     offer.getId(), process.getId(), ContractNegotiationStates.from(process.getState()));
                             monitor.debug(message, throwable);
@@ -371,10 +375,12 @@ public class ConsumerContractNegotiationManagerImpl implements ConsumerContractN
                     .whenComplete((response, throwable) -> {
                         if (throwable == null) {
                             process.transitionApproved();
+                            negotiationStore.save(process);
                             monitor.debug(String.format("[Consumer] ContractNegotiation %s is now in state %s.",
                                     process.getId(), ContractNegotiationStates.from(process.getState())));
-                            negotiationStore.save(process);
                         } else {
+                            process.transitionApproving();
+                            negotiationStore.save(process);
                             String message = format("[Consumer] Failed to send contract agreement with id %s. ContractNegotiation %s stays in state %s.",
                                     agreement.getId(), process.getId(), ContractNegotiationStates.from(process.getState()));
                             monitor.debug(message, throwable);
@@ -411,16 +417,17 @@ public class ConsumerContractNegotiationManagerImpl implements ConsumerContractN
                     .whenComplete((response, throwable) -> {
                         if (throwable == null) {
                             process.transitionDeclined();
+                            negotiationStore.save(process);
                             monitor.debug(String.format("[Consumer] ContractNegotiation %s is now in state %s.",
                                     process.getId(), ContractNegotiationStates.from(process.getState())));
-                            negotiationStore.save(process);
                         } else {
+                            process.transitionDeclining();
+                            negotiationStore.save(process);
                             String message = format("[Consumer] Failed to send contract rejection. ContractNegotiation %s stays in state %s.",
                                     process.getId(), ContractNegotiationStates.from(process.getState()));
                             monitor.debug(message, throwable);
                         }
                     });
-
         }
 
         return processes.size();
--- a/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
+++ b/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
@@ -45,6 +45,10 @@ import static java.lang.String.format;
 import static org.eclipse.dataspaceconnector.contract.common.ContractId.DEFINITION_PART;
 import static org.eclipse.dataspaceconnector.contract.common.ContractId.parseContractId;
 import static org.eclipse.dataspaceconnector.spi.contract.negotiation.response.NegotiationResult.Status.FATAL_ERROR;
+import static org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiationStates.CONSUMER_APPROVING;
+import static org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiationStates.CONSUMER_OFFERING;
+import static org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiationStates.DECLINING;
+import static org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiationStates.REQUESTING;
 
 /**
  * Implementation of the {@link ConsumerContractNegotiationManager}.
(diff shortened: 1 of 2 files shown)


core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -269,22 +273,23 @@ public NegotiationResult declined(ClaimToken token, String negotiationId) {
...
-            negotiationStore.save(process);
+            sendOffer(offer, process, ContractOfferRequest.Type.INITIAL)
+                    .whenComplete((response, throwable) -> {
+                        if (throwable == null) {
+                            process.transitionRequested();
+                            monitor.debug(String.format("[Consumer] ContractNegotiation %s is now in state %s.",
+                                    process.getId(), ContractNegotiationStates.from(process.getState())));
+                            negotiationStore.save(process);
+                        } else {
+                            String message = format("[Consumer] Failed to send contract offer with id %s. ContractNegotiation %s stays in state %s.",
```
juliapampus, line 288: The line `process.transitionRequesting();` is missing in the else block. It there a reason why?

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -298,22 +303,23 @@ private int sendContractOffers() {
...
-            negotiationStore.save(process);
+            sendOffer(offer, process, ContractOfferRequest.Type.COUNTER_OFFER)
+                    .whenComplete((response, throwable) -> {
+                        if (throwable == null) {
+                            process.transitionOffered();
+                            monitor.debug(String.format("[Consumer] ContractNegotiation %s is now in state %s.",
+                                    process.getId(), ContractNegotiationStates.from(process.getState())));
+                            negotiationStore.save(process);
+                        } else {
+                            String message = format("[Consumer] Failed to send contract offer with id %s. ContractNegotiation %s stays in state %s.",
```
juliapampus, line 318: See above. `process.transitionOffering();` missing

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -361,18 +367,19 @@ private int approveContractOffers() {
...
-            negotiationStore.save(process);
+            dispatcherRegistry.send(Object.class, request, process::getId)
+                    .whenComplete((response, throwable) -> {
+                        if (throwable == null) {
+                            process.transitionApproved();
+                            monitor.debug(String.format("[Consumer] ContractNegotiation %s is now in state %s.",
+                                    process.getId(), ContractNegotiationStates.from(process.getState())));
+                            negotiationStore.save(process);
+                        } else {
+                            String message = format("[Consumer] Failed to send contract agreement with id %s. ContractNegotiation %s stays in state %s.",
```
juliapampus, line 378: `process.transitionApproving();`

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -298,22 +303,23 @@ private int sendContractOffers() {
...
-                monitor.debug(format("[Consumer] Failed to send contract offer with id %s. ContractNegotiation %s stays in state %s.",
-                        offer.getId(), process.getId(), ContractNegotiationStates.from(process.getState())));
-                continue;
-            }
-
-            process.transitionOffered();
-            monitor.debug(String.format("[Consumer] ContractNegotiation %s is now in state %s.",
-                    process.getId(), ContractNegotiationStates.from(process.getState())));
-            negotiationStore.save(process);
+            sendOffer(offer, process, ContractOfferRequest.Type.COUNTER_OFFER)
```
bscholtes1A, line 310: Thanks a lot for this PR! Could we also take opportunity of this change to display the throwable message within the logs? Otherwise when sendOffer fails then you do not get any information why.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -400,19 +407,20 @@ private int declineContractOffers() {
...
-            }
+            dispatcherRegistry.send(Object.class, rejection, process::getId)
+                    .whenComplete((response, throwable) -> {
+                        if (throwable == null) {
+                            process.transitionDeclined();
+                            monitor.debug(String.format("[Consumer] ContractNegotiation %s is now in state %s.",
+                                    process.getId(), ContractNegotiationStates.from(process.getState())));
+                            negotiationStore.save(process);
+                        } else {
+                            String message = format("[Consumer] Failed to send contract rejection. ContractNegotiation %s stays in state %s.",
```
juliapampus, line 418: `process.transitionDeclining();`

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java
```diff
@@ -292,20 +295,19 @@ private int checkProviderOffering() {
...
-                    negotiation.getId(), ContractNegotiationStates.from(negotiation.getState())));
+            dispatcherRegistry.send(Object.class, contractOfferRequest, () -> null)
+                    .whenComplete((response, throwable) -> {
+                        if (throwable == null) {
+                            negotiation.transitionOffered();
+                            negotiationStore.save(negotiation);
+                            monitor.debug(String.format("[Provider] ContractNegotiation %s is now in state %s.",
+                                    negotiation.getId(), ContractNegotiationStates.from(negotiation.getState())));
+                        } else {
+                            String message = format("[Provider] Failed to send contract offer with id %s. ContractNegotiation %s stays in state %s.",
```
juliapampus, line 306: See comments on `ConsumerContractnegotiationManagerImpl`. For all changes in this class.

juliapampus (changes requested):

bscholtes1A (approved):

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
ndr-brt: I removed it here as in the other cases because I don't see the point to set the state as it is already (in this case the `save` method was already missing, so setting it to `transitionRequesting` is definitely not useful)

ndr-brt:
@juliapampus I removed all the state set after an exception because in all the cases it was set at the same state as it were.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
juliapampus: We added this for retry cases - that are not yet implemented. The transition method increases the state count when a process enteres the same state repeatedly. This way, we can abort the process as soon as a certain number of retries is reached.

juliapampus:
Missed your comment here, see [above](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/530#discussion_r787750477). @ndr-brt Not sure whether to then keep it or just add it again as soon as we implement the described logic.

juliapampus:
@jimmarino @bscholtes1A What do you think?

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
ndr-brt: got it, I was asking myself what's the scope of stateCount, since it's not used by anyone. should that be addressed in a new issue? In the meantime I will reintroduce "state tick"

ndr-brt:
@juliapampus makes sense, I reintroduced them, but we should address an issue for that, because is not clear what those `stateCount` are for, since those fields are not used by anyone.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -269,22 +273,25 @@ public NegotiationResult declined(ClaimToken token, String negotiationId) {
...
+            sendOffer(offer, process, ContractOfferRequest.Type.INITIAL)
+                    .whenComplete((response, throwable) -> {
+                        if (throwable == null) {
+                            process.transitionRequested();
+                            negotiationStore.save(process);
+                            monitor.debug(String.format("[Consumer] ContractNegotiation %s is now in state %s.",
+                                    process.getId(), ContractNegotiationStates.from(process.getState())));
+                        } else {
+                            process.transitionRequesting();
+                            negotiationStore.save(process);
```
bscholtes1A, line 289: @ndr-brt Beware that you are deviating from the logic which is currently limplemented here.

In the current implementation, the `negotiationStore.save(process);` is not called in case of an exception, due to the `continue` statement. 

@juliapampus was it a miss in the current implem?

ndr-brt: yes, I added it by purpose. Without saving the process, no `stateCount` gets incremented. 
I think this was an oversight.

bscholtes1A: > yes, I added it by purpose. Without saving the process, no `stateCount` gets incremented. I think this was an oversight.

Yeap seems correct. As soon as @juliapampus confirms, it should ready for merge.

juliapampus (approved):

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
jimmarino: The state count should increase since it is transitioning back into the same state. As @juliapampus mentioned, we just haven't yet implemented logic for a transition threshold before moving the process into an error state. That should be a configurable threshold. For the TPM, we should discuss if the threshold is the same for all states or not. We also will need to handle this for contract negotiation as well.

jimmarino:
> @jimmarino @bscholtes1A What do you think?

I was going to defer to you since you are more familiar with this code than I am. I agree with your comments on state transition thresholds.

ndr-brt:
Seems we are ready for merge

PR #536, opened 2022-01-20
Feature/416 tp error after failing transitions

### LINKED ISSUE #416: Reliability - Critical - Configurable timeout for TransferProcess
TransferProcesses might get stuck in a state for a long time, or even indefinitely when errors occur. The `TransferProcesManager` should monitor this situation and react to it after a configurable threshold by moving the process to an ERROR state, effectively taking the TransferProcess out of the state machine processing loop.

The threshold should be configurable on a TransferProcess basis, as appropriate thresholds may diverge vastly depending on the nature of the transfer itself (moving few KB VS several GB). A sensible default value should be used in case no configuration is available for a given TransferProcess.

ndr-brt:
Fix #416.

* Added `stateTimeoutThreshold` that will used to set state to ERROR after that threshold is crossed, on state transition
* Add `stateCountThreshold` that will used to set state to ERROR after that threshold is crossed, on state transition

### DIFF
--- a/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/CoreTransferExtension.java
+++ b/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/CoreTransferExtension.java
@@ -16,6 +16,8 @@ package org.eclipse.dataspaceconnector.transfer.core;
 
 import org.eclipse.dataspaceconnector.core.CoreExtension;
 import org.eclipse.dataspaceconnector.core.base.ExponentialWaitStrategy;
+import org.eclipse.dataspaceconnector.spi.EdcException;
+import org.eclipse.dataspaceconnector.spi.EdcSetting;
 import org.eclipse.dataspaceconnector.spi.command.CommandHandlerRegistry;
 import org.eclipse.dataspaceconnector.spi.command.CommandQueue;
 import org.eclipse.dataspaceconnector.spi.command.CommandRunner;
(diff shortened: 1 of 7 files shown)


paullatzelsperger:
How would we ever set the state threshold? Users don't create TransferProcesses. Would that be a global config? If so, would there be _one threshold per state_, or the same for all states?

I'm ok with this being a fallback cleanup, so as to not zombie `TransferProcesses`. 

A fancier solution could be to implement a dedicated watchdog extension, which contains a `Watchdog` class, that implements the `TransferProcessListener` and uses the `CommandQueue`. That `Watchdog` could then surveil all TransferProcesses that violate whatever watchdog policy is configured and kick them using the `CommandQueue`. 

@ndr-brt @margs if you agree we could do this as an improvement issue.

ndr-brt:
you're right, the issue confused me about this, because we talk about configuring threshold on TransferProcess basis, but, as you said, TPs are not created by users, because they are an internal concept.
Personally I would rely on `stateCount` rather on the timeout, trying to avoid stale status in the state machine itself, e.g., every `check*` method should always transate the TP state in some way (currently it's not like that)

The Watchdog process would be the last resort.

jimmarino:
> you're right, the issue confused me about this, because we talked about configuring threshold on TransferProcess basis, but, as you said, TPs are not created by users, because they are an internal concept. Personally I would rely on `stateCount` rather on the timeout, trying to avoid stale status in the state machine itself, e.g., every `check*` method should always transate the TP state in some way (currently it's not like that)
> 
> The Watchdog process would be the last resort.

I was thinking this is a TPM, not TP configuration. For me, this is a characteristic of the state machine and allowable transitions, and should not be handled by a watchdog.

paullatzelsperger:
I'm fine with having a general maximum time or amount a TP may remain in a state. 
A watchdog should be optional, but seems like a good concept if a more elaborate or more specific handling of stale processes is required.

ndr-brt:
I make the TPM thresholds configurable. Does this PR covers the requirements of the issue #416 or there's something missing?

ndr-brt:
Closed as we need to complete the state machine review discussed in #549 first

PR #540, opened 2022-01-21
Feature: Support for Portable Transactions

jimmarino:
Implements support for portable transactions. For details, see #541.


paullatzelsperger:
hey @jimmarino, there are a few remarks/questions I had: 

* would it be possible to merge `transaction-local-spi` and `transaction-spi` (and later: `transaction-distributed-spi`)? seems more compact that way
* `transaction-local-core` should then just be named `transaction-local`
* personally, I'd prefer adding a stub for distributed transactions, even if it initially contains no code and every method throws an `UnsupportedOperationException`, as this indicates that we didn't just leave it out. Eventually, they will get filled, preferably by an Atomikos based impl. 
* Can you add an example how a complete vertical would look like? I'm thinking of a SQL-based `AssetIndex` that uses a Atomikos-backed data source. This would show which extensions are involved, and what an implementor of any particular storage system would have to do.
* What do you think about an improvement to the configuration mechanism, where the configuration for each datasource could be namespaced and implicitly named, e.g. like this: 
   ```
   edc.datasource.firstdatasource.user=...
   edc.datasource.firstdatasource.password=...
   //...
   edc.datasource.seconddatasource.user=...
   edc.datasource.seconddatasource.password=...
   ```
   which would then automatically recognize that there are two datasources and configure them accordingly. Much like Spring does it.

jimmarino:
> hey @jimmarino, there are a few remarks/questions I had:
> 
> * would it be possible to merge `transaction-local-spi` and `transaction-spi` (and later: `transaction-distributed-spi`)? seems more compact that way
> * `transaction-local-core` should then just be named `transaction-local`

+1 on both

> * personally, I'd prefer adding a stub for distributed transactions, even if it initially contains no code and every method throws an `UnsupportedOperationException`, as this indicates that we didn't just leave it out. Eventually, they will get filled, preferably by an Atomikos based impl.

Yes, I'm planning on doing that. I'll create a separate PR and do it after this one since this one is starting to cover a lot of ground.

> * Can you add an example how a complete vertical would look like? I'm thinking of a SQL-based `AssetIndex` that uses a Atomikos-backed data source. This would show which extensions are involved, and what an implementor of any particular storage system would have to do.

Good Idea.

> * What do you think about an improvement to the configuration mechanism, where the configuration for each datasource could be namespaced and implicitly named, e.g. like this:
>   ```
>   edc.datasource.firstdatasource.user=...
>   edc.datasource.firstdatasource.password=...
>   //...
>   edc.datasource.seconddatasource.user=...
>   edc.datasource.seconddatasource.password=...
>   ```

I was thinking of the same exact scheme! I was planning on adding that with Atomikos.

ndr-brt (commented):
Fine for me, an e2e example as Paul suggested would be great.

One little thing, it's not necessary to define test dependencies on the modules since they are already inherited from the root build file.

jimmarino:
> Fine for me, an e2e example as Paul suggested would be great.
>
Let's put that e2e sample in a separate PR to keep the size of this one down. I'm also going to do the same for the global trx impl.

> One little thing, it's not necessary to define test dependencies on the modules since they are already inherited from the root build file.

Yeah, thanks, copy-and-paste, I'll update now.

paullatzelsperger (approved):

ndr-brt (approved):

jimmarino:
The failing tests are caused by another PR and are unrelated to this PR. Given that, I'm going to merge this as there are additional PRs waiting for this one.

Issue #538, opened 2022-01-21
Contract negotiation loop


hlange:
Hi,
I have a problem with the sample 4.0. I fixed the issue with keyName like mentioned in the discussions. But this generated new problems because after everything gone through and the file was moved from one directory to a different, the CPU load of one core gone up to 100%.

Because I thought maybe this could be because it was not the newest version I clone the current version. This generated a complete new issue because the negotiation was running in a loop.

Consumer:
![grafik](https://user-images.githubusercontent.com/12150189/150515320-793bd0f6-466a-4e2d-a458-bc5e27c73ddc.png)

Provider:
![grafik](https://user-images.githubusercontent.com/12150189/150515377-96bf420d-b206-4139-b55c-4e0e1acafa5b.png)

By running the following command:
`curl -X POST -H "Content-Type: application/json" -d @samples/04.0-file-transfer/contractoffer.json "http://localhost:9191/api/negotiation?connectorAddress=http://localhost:8181/api/ids/multipart"`

And I added keyName to ConsumerApiController.java

Best regards,
Henrik

ndr-brt:
Thanks for finding out, here I see two problems:
* the sample 04 was renamed to 04.0 but its README was not updated accordingly
* on the consumer side ContractNegotiation `REQUESTING` state has two meanings:
  * ContractNegotiation created and waiting to be sent to the provider
  * ContractNegotiation sent to the provider and waiting for response

This causes the infinite loop, because every `REQUESTING` negotiation is sent to the provider, but it should not

ndr-brt:
I noticed strange behavior into the contract negotiation repositories (stores).
In fact, both of the "production" (I mean "production code", not "production ready") implementations contain this logic:
```
if (ContractNegotiationStates.UNSAVED.code() == negotiation.getState()) {
            if (ContractNegotiation.Type.CONSUMER.equals(negotiation.getType())) {
                negotiation.transitionRequesting();
            } else {
                negotiation.transitionRequested();
            }
        }
```

this is dangerous because this kind of domain logic should stay in the domain classes, in this case in `ContractNegotiation` probably. Otherwise every time someone implement a `ContractNegotiationStore` should report that logic in their implementation, and that would bring to duplication.

However my proposal is to define an additional state between UNSAVED and REQUESTING called INITIALIZED. The negotiations on consumer state in that state will be sent to provider and moved to PROCESSING. That would resolve this bug

[...]

PR #545, opened 2022-01-24
Feature/538 avoid negotiation loop

ndr-brt:
Ref #538 

This fix the referred issue introducing an `INITIAL` state that goes between `UNSAVED` and `REQUESTING`.
However, there are still some problems, probably introduced (or, better, unearthed) by the correct evaluation of the futures.

Before, for every state there was a single intermediate state: 
 * `REQUESTED` -> `REQUESTING`
 * `PROVIDER_OFFERED` -> `PROVIDER_OFFERING`
 * ...

But now we would need another intermediate states, like:
`INITIAL`: the negotiation is initialized, and it's retrievable by the state machine loop for being processed
`REQUESTING`: the negotiation has been retrieved from the store and sent to the provider, it should be not retrieved anymore by the state machine loop, otherwise it would be sent multiple times
`REQUESTED`: the request has been successfully sent to the provider.

To better describe these async transitions there could be 3 states for every "real state":
* `TO_BE_xxxED`
* `xxxING`
* `xxxED`
(replace xxx with, e.g. "REQUEST", "PROVIDER_OFFER", etc...)

This 3-state pattern need to be implemented for every state transition, otherwise there will be multiple and uncontrollable state changes.

I ask thoughts to those who worked on the state machines @jimmarino @paullatzelsperger @juliapampus @ronjaquensel , somebody else?

### DIFF
--- a/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
+++ b/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
@@ -32,8 +32,8 @@ import org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.Cont
 import org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractOfferRequest;
 import org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractRejection;
 import org.eclipse.dataspaceconnector.spi.types.domain.contract.offer.ContractOffer;
+import org.jetbrains.annotations.NotNull;
 
-import java.time.Instant;
 import java.time.LocalDate;
 import java.util.Objects;
 import java.util.UUID;
@@ -41,6 +41,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.BiConsumer;
 
 import static java.lang.String.format;
 import static org.eclipse.dataspaceconnector.contract.common.ContractId.DEFINITION_PART;
@@ -49,7 +50,7 @@ import static org.eclipse.dataspaceconnector.spi.contract.negotiation.response.N
 import static org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiationStates.CONSUMER_APPROVING;
 import static org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiationStates.CONSUMER_OFFERING;
 import static org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiationStates.DECLINING;
-import static org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiationStates.REQUESTING;
+import static org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiationStates.INITIAL;
 
 /**
  * Implementation of the {@link ConsumerContractNegotiationManager}.
@@ -103,12 +104,10 @@ public class ConsumerContractNegotiationManagerImpl extends ContractNegotiationO
                 .protocol(contractOffer.getProtocol())
                 .counterPartyId(contractOffer.getConnectorId())
                 .counterPartyAddress(contractOffer.getConnectorAddress())
-                .state(0)
-                .stateCount(0)
-                .stateTimestamp(Instant.now().toEpochMilli())
                 .build();
 
         negotiation.addContractOffer(contractOffer.getContractOffer());
+        negotiation.transitionInitial();
         negotiationStore.save(negotiation);
         invokeForEach(l -> l.requesting(negotiation));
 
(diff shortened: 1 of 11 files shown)


ronjaquensel:
Hi @ndr-brt,

I think it makes sense to introduce this 3-state-pattern, to account for the time where the other party processes the request. I just ran sample 4 on your branch and introducing the INITIAL state already seems to break the infinite loop (now only the agreement is confirmed multiple times, but at least not infinitely), so introducing the intermediary state for all states should solve the issue.

One thing we have to think about though is the state count. Currently the state count is increased when a negotiation is transitioned to the same state as before. Since we will now always have an intermediary state, the transition will e.g. be `requesting -> initial -> requesting` instead of `requesting -> requesting`, therefore the state count will not be increased as is. I don't think this is crucial right now, as we are not using a retry strategy in the negotiation managers at the moment, but this should definitely be addressed later on.

ndr-brt:
@ronjaquensel thank you for the feedback. 
I think that, when we do an async computation on the entitity (like in the "requesting" case), we should not check that entity anymore, it goes to a state that's "processing, waiting for the async computation to be completed", it will change to another state on the future completion, the state count should not increase, instead there should be a process that checks if there are entities in a "processing" state for some (configurable) time and that it applies some logic to avoid that stale state in ways that could be, e.g. retry it for some times and kill it after N retry.
I think this pattern would handle failing situations nicely.

anyway, I shared some thorough thoughts in this discussion #549

ndr-brt:
Also if this doesn't resolve all the problems, at least it will make the sample 04 work, I think it's important to merge it prior to monday since it will be shown at the workshop session in the afternoon.

florianrusch-cluetec:
I will test it regarding the sample 04 today

ndr-brt:
@ronjaquensel @florianrusch-cluetec 
I managed to fix the "multiple confirmations" by adding an intermediate state between `CONFIRMING` and `CONFIRMED`, it's what usually I call "an ugly patch" but does the job, in the wait of the state machine review, this will permit a flawless demo on monday, I tried the sample 04.0 and it worked, other feedbacks are welcome

ronjaquensel:
@ndr-brt I just tested sample 4 on your branch again and looks good. On the provider side I only get
```
DEBUG 2022-01-27T13:06:18.5138138 [Provider] ContractNegotiation initiated. db6b4616-ac5d-40c7-bc2c-2d20eef716e5 is now in state REQUESTED.
DEBUG 2022-01-27T13:06:18.5208097 [Provider] Contract offer received. Will be approved.
DEBUG 2022-01-27T13:06:18.5218082 [Provider] ContractNegotiation db6b4616-ac5d-40c7-bc2c-2d20eef716e5 is now in state CONFIRMING.
DEBUG 2022-01-27T13:06:19.127198 Response received from connector. Status 200
DEBUG 2022-01-27T13:06:19.1462043 [Provider] ContractNegotiation db6b4616-ac5d-40c7-bc2c-2d20eef716e5 is now in state CONFIRMED.

```
and on the consumer side I only get:
```
DEBUG 2022-01-27T13:06:13.7506063 [Consumer] ContractNegotiation initiated. a203bac2-ba50-42b7-8d36-88d3e4f98038 is now in state INITIAL.
DEBUG 2022-01-27T13:06:18.5628357 Response received from connector. Status 200
DEBUG 2022-01-27T13:06:18.6398487 [Consumer] ContractNegotiation a203bac2-ba50-42b7-8d36-88d3e4f98038 is now in state REQUESTED.
DEBUG 2022-01-27T13:06:19.1111958 [Consumer] Contract agreement received. Validation successful.
DEBUG 2022-01-27T13:06:19.1121933 [Consumer] ContractNegotiation a203bac2-ba50-42b7-8d36-88d3e4f98038 is now in state CONFIRMED.
```

florianrusch-cluetec:
Same for me ;-)

DominikPinsel:
I just tried using the fix from this branch to run sample 5 and I get some exceptions.

This is from the provider (spammed in a loop)
```java
DEBUG 2022-01-27T14:48:23.191292 [Provider] Failed to send contract agreement with id 1:1dff842f-85b9-4fff-8a19-d6f3da12d1c6. ContractNegotiation 80bb5f2f-884a-4648-bf6c-577594168def stays in state CONFIRMING.
org.eclipse.dataspaceconnector.spi.EdcException: Received an error from connector:500
	at org.eclipse.dataspaceconnector.ids.api.multipart.dispatcher.sender.IdsMultipartSender.lambda$send$0(IdsMultipartSender.java:209)
	at org.eclipse.dataspaceconnector.ids.core.message.FutureCallback.onResponse(FutureCallback.java:45)
```

And this is from the consumer (spammed in a loop)
```java
Caused by: java.lang.IllegalStateException: Cannot transition from state CONFIRMED to CONFIRMED
	at org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiation.transition(ContractNegotiation.java:373)
	at org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiation.transitionConfirmed(ContractNegotiation.java:287)
	at org.eclipse.dataspaceconnector.contract.negotiation.ConsumerContractNegotiationManagerImpl.confirmed(ConsumerContractNegotiationManagerImpl.java:213)
	at org.eclipse.dataspaceconnector.ids.api.multipart.handler.ContractAgreementHandler.handleRequest(ContractAgreementHandler.java:123)
	at org.eclipse.dataspaceconnector.ids.api.multipart.controller.MultipartController.request(MultipartController.java:120)
	at jdk.internal.reflect.GeneratedMethodAccessor83.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory.lambda$static$0(ResourceMethodInvocationHandlerFactory.java:52)
```

ndr-brt:
@DominikPinsel this is the typical hen-or-the-egg problem, we have to understand who of the two connector started flooding requests. 
From what I see, the provider find the process in state CONFIRMING, try to send it to the consumer but it receive 500 because it's already CONFIRMED. I don't know if in the IDS there's a way to tell to the peer "already done, never mind" so that the process could be considered completed. 
Another way to resolve it is to permit the transitions from CONFIRMED to CONFIRMED without throwing error, but that will clash with the stateCount pattern to detect failing states (not yet merged however).

florianrusch-cluetec:
I've tested it again and found maybe another problem were I'm not sure if it depends on the issues or fixes we already discussed here.

If the provider can't find the file he logs this in a loop:
```
INFO 2022-01-28T09:48:16.250353 Copying data from File to File
SEVERE 2022-01-28T09:48:16.250997 Error processing transfer request. Setting to retry: d861aab5-ee59-409d-8c97-e86394c24f45
```


But if I read the logs correctly, the consumer logs that the transfer process was completed:
```
DEBUG 2022-01-28T09:42:29.373446 [Consumer] ContractNegotiation initiated. 7979516d-df08-4da6-b015-3ea6a2a2bcdc is now in state INITIAL.
DEBUG 2022-01-28T09:42:31.244869 Response received from connector. Status 200
DEBUG 2022-01-28T09:42:31.353274 [Consumer] ContractNegotiation 7979516d-df08-4da6-b015-3ea6a2a2bcdc is now in state REQUESTED.
DEBUG 2022-01-28T09:42:35.266974 [Consumer] Contract agreement received. Validation successful.
DEBUG 2022-01-28T09:42:35.267726 [Consumer] ContractNegotiation 7979516d-df08-4da6-b015-3ea6a2a2bcdc is now in state CONFIRMED.

INFO 2022-01-28T09:48:07.643573 Received request for file test-document against provider http://localhost:8181/api/ids/multipart
DEBUG 2022-01-28T09:48:10.567869 Response received from connector. Status 200
INFO 2022-01-28T09:48:10.574536 Object received: org.eclipse.dataspaceconnector.ids.api.multipart.dispatcher.message.MultipartRequestInProcessResponse@36b1394a
DEBUG 2022-01-28T09:48:15.547119 Process 2ce3e43a-4180-4d9d-9c17-dd8e4949ff75 is now COMPLETED
```

Does this relates to the issues we had before?

ndr-brt:
@florianrusch-cluetec this problem was already in place, it's beacause the data request is not managed (so no provisioning needed), and there's no status checker implemented:
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/69285ee2baac58e5f87dd6f1769f0172c500d59d/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/AsyncTransferProcessManager.java#L369

so the transfer process set it to completed.

I guess, to improve this, we could set the data request as managed, add a "FileResourceDefinition" that will create an empty file on the destination and a FileStatusChecker that will check if the file is not empty. Should work, let me try it out

florianrusch-cluetec:
> I guess, to improve this, we could set the data request as managed, add a "FileResourceDefinition" that will create an empty file on the destination and a FileStatusChecker that will check if the file is not empty. Should work, let me try it out

That sounds like a workaround we'd better avoid or not? 🤣

ndr-brt:
> > I guess, to improve this, we could set the data request as managed, add a "FileResourceDefinition" that will create an empty file on the destination and a FileStatusChecker that will check if the file is not empty. Should work, let me try it out
> 
> That sounds like a workaround we'd better avoid or not? rofl

totally :D I think we are ok with this fix, setting an existing file on the provider side is a mandatory pre-requisite for the sample

ronjaquensel (approved):

Issue #541, opened 2022-01-21
Support for Portable Transactions


jimmarino:
# Overview

This issue outlines the requirements for, and the implementation of, transaction support in the EDC.

The main requirements for transactions are:

1. It must be possible to support transactional operations (create, update, delete, enqueue, dequeue) on data in the runtime. 

2. Transactional operations may span multiple services contributed to the runtime by different extensions. 

3. Transactional operations may be composed differently based on use cases and code execution paths. For example, consider the following situations:

> Service A ---> Service 1 ([work])

In this case, Service A invokes Service 1 in insolation. Service 1 demarcates the work it performs within a transaction, denoted with [..].

> Service A [----> Service 1 (work), Service 2 (work)]

In this case, all services demarcate their work within transactions. Since Service A defines a transactional boundary that spans invocations to Service 1 and Service 2, their transactions join the parent Service A transaction 
  
4. Transaction support most offer a portable programming model, whether the underlying implementation is based on a JTA transaction manager, a local resource such as a JDBC connection pool, or a host environment such as Spring.

# Core Services

## TransactionContext

The transaction implementation will be centered on a `TransactionContext` registered with the runtime service context:

```
public interface TransactionContext {

    /**
     * Executes the code within a transaction.
     */
    void execute(TransactionBlock block);

    /**
     * Defines a block of transactional code.
     */
    @FunctionalInterface
    interface TransactionBlock {

        /**
         * Executes the transaction block.
         */
        void execute();

    }
}
```

The `TransactionContext` provides a portable programming model by abstracting the specifics of transaction management. It also gives the added benefit of handling joining, completion, and rollback operations, which are repetitive and error-prone tasks. The following demonstrates how code interacts with the `TransactionContext`, including the semantics of nested transaction demarcation:

```
        transactionContext.execute(() -> {
            // top level work...
            transactionContext.execute(() -> {
                // nested work joins the parent transaction...
            });
            // transaction work completes
        });
    }
```
Underlying transaction subsystems provide an implementation of `TransactionContext`. There will initially be an implementation that supports local transactions. A JTA-based version can also be implemented in a subsequent PR.

Note that begin(), commit(), rollback() and setRollbackOnly() methods could be added to `TransactionContext` in a future PR. This may be useful for bulk loading scenarios where batches must be transactionally committed. In general, however, transaction blocks should be preferred as it alleviates user-code from needing to manage transaction operations.

## DataSourceRegistry

[...]

PR #555, opened 2022-01-26
Add filtering for properties and making them hierarchical

### LINKED ISSUE #546: Improve configuration system: add namespaced and dynamic properties
The configuration subsystem should be able to handle config entries, that are namespaced and dynamic. For example, when configuring data sources we'll need configuration values like 
```
edc.datasource.default.user=...
edc.datasource.default.password=...
edc.datasource.default.driverClassName=...
//...
edc.datasource.anotherDs.user=...
edc.datasource.anotherDs.password=...
edc.datasource.anotherDs.someConfigProp=...
```
In the example, the extension that instantiates datasources would read all entries starting with `edc.datasource`, group them by their "name" (here: `default` and `anotherDs`) and configures them. 
The configuration properties could be different, e.g. datasources - different drivers might need different properties, so that needs to be handled generically.

This issue is an amendment to #541, where it will likely be used first.

paullatzelsperger:
This PR extends the configuration feature and adds an option to filter by prefix. There also is a new helper method that will then interpret the resulting properties hierarchically.
Having this feature makes configuring multiple instances of a single thing (e.g. datasources) much easier.

Closes #546

### DIFF
--- a/common/util/src/main/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctions.java
+++ b/common/util/src/main/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctions.java
@@ -56,9 +56,6 @@ public class ConfigurationFunctions {
         return hierarchical(properties, autoRoot(properties.keySet()));
     }
 
-    private static String autoRoot(Set<String> keySet) {
-        return StringUtils.getCommonPrefix(new ArrayList<>(keySet));
-    }
 
     /**
      * Takes a list of flat properties and converts them into a hierarchical map of entries starting from a given root.
@@ -97,13 +94,11 @@ public class ConfigurationFunctions {
             if (k.contains(".")) {
                 var group = k.split("\\.")[0];
                 var key = k.replace(group + ".", "");
-                if (groupedList.containsKey(group)) {
-                    groupedList.get(group).put(key, v);
-                } else {
-                    var m = new HashMap<String, Object>();
-                    m.put(key, v);
-                    groupedList.put(group, m);
-                }
+
+                var m = groupedList.computeIfAbsent(group, s -> new HashMap<>());
+
+                groupedList.get(group).put(key, v);
+                m.put(key, v);
             } else {
                 var m = groupedList.computeIfAbsent(root, s -> new HashMap<>());
                 m.put(k, v);
@@ -111,5 +106,9 @@ public class ConfigurationFunctions {
         });
 
         return groupedList;
+    }
+
+    private static String autoRoot(Set<String> keySet) {
+        return StringUtils.getCommonPrefix(new ArrayList<>(keySet));
     }
 }
--- a/common/util/src/main/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctions.java
+++ b/common/util/src/main/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctions.java
@@ -14,6 +14,16 @@
 
 package org.eclipse.dataspaceconnector.common.configuration;
 
+import org.eclipse.dataspaceconnector.common.string.StringUtils;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
 /**
  * Common configuration functions used by extensions.
  */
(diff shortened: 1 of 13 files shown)


common/util/src/main/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctions.java
```diff
@@ -32,4 +42,74 @@ public static String propOrEnv(String key, String defaultValue) {
...
+     * @param properties A flat map of properties
+     * @return A "map of maps", each of which containing the sub-properties of one node.
+     * @see ConfigurationFunctions#hierarchical(Map, String)
+     * @see StringUtils#getCommonPrefix(List)
+     */
+    public static Map<String, Map<String, Object>> hierarchical(Map<String, Object> properties) {
+        return hierarchical(properties, autoRoot(properties.keySet()));
+    }
+
+    private static String autoRoot(Set<String> keySet) {
```
jimmarino, line 59: Can you slide this below the public methods?

jimmarino (approved):
LGTM

common/util/src/test/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctionsTest.java
```diff
@@ -0,0 +1,61 @@
...
+                "user", "test-user2",
+                "password", "test-pwd2",
+                "sub.property2", "bar",
+                "driverClassName", "org.company.another.Driver"));
+    }
+
+    @Test
+    void hierarchical_withRoot_deeperLevel() {
+
+        var hierarchy = ConfigurationFunctions.hierarchical(allProperties, "edc.datasource.default");
```
ndr-brt, line 53: This is not clear to me, why in the test above the key for getting the property is only the part next to the root, instead here there's the need to report all the root key?
This hierarchy should countain only the elements in `edc.datasource.default` (user, password, driverClassName)

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/ConfigurationExtension.java
```diff
@@ -27,4 +29,11 @@
     @Nullable
     String getSetting(String key);
 
+    /**
+     * Gets all configuration settings that start with a certain prefix (= filtering).
+     *
+     * @param prefix the prefix
+     * @return A map that contains all config entries that start with the prefix.
+     */
+    Map<String, String> getSettingsWithPrefix(String prefix);
```
ndr-brt, line 38: Why this `settingsWithPrefix` is needed?, should be enough to get all the settings on which to apply the `hierarchical` function. 
Filtering can be implemented by the client

ndr-brt (commented):
Nice start, this could lead to introduce some kind of `Config` object that would offer things like `getString`, `getInt` but also `getPath` that would return another `Config` related to the path specified. No need to introduce these concepts right now, just some idea here and there.

I added a pair of comments

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/ConfigurationExtension.java
paullatzelsperger: the reason for this is that we might not know the exact set of properties. This was developed with configuring JDBC drivers in mind, which can have different properties. 
With this, it is sufficient to state the prefix, which would be `edc.datasource.<DS_NAME>` and that'll give you all properties that are configured.

ndr-brt: what about an interface like:
```
Map<String, Object> getConfig(String path)
```
that will directly give an hierarcal structure of the configuration at that path so to permit the client to avoid creating hierarcy by itself? (as said in the main comment, the return type could be a `Config` class that wraps the contents and offer a better interface than `Map`)

common/util/src/test/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctionsTest.java
paullatzelsperger: We have two situations: 
1) there is an item between `root` and the elements. Here, that item would be `default`
2) there is no item between `root` and the elements

Now since both situations must return a map-of-maps, in the second case I need to introduce some sort of "artificial" root key, which is the entire prefix. 

Solving this without the "artificial root key" would require returning a tree structure instead of a map-of-maps which has 2 or 3 levels, depending on which situation we're in. Potentially even more.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/ConfigurationExtension.java
paullatzelsperger: I assume you meant to type `Config getConfig(String path)`?

common/util/src/test/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctionsTest.java
ndr-brt: This behavior is a little tricky to understand at first sight, or better, it's not what I could expect from that method.
Would be better to return `Map<String, Object>` and leave to the client to check the type of the value, otherwise, the `Config` class would resolve all those issue nicely

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/ConfigurationExtension.java
ndr-brt: Yes

paullatzelsperger: As I said, the client might not know about the structure, as this may be handled in a generic way. Then again, I might not have completely understood your idea :)

But since @jimmarino is waiting for this feature to get merged for his PR #541, which in turn blocks a few other features: would you be ok to merge this as-is and let's come up with a proposal that uses your `Config` variant?

ndr-brt: sure

paullatzelsperger: opened #557

ndr-brt (approved):

PR #558, opened 2022-01-26
rename single implementers to "*Impl" convention

### LINKED ISSUE #556: Define a naming standard for single implementers
As discussed in #553 , at the moment there are two ways to name single implementations of an interface:
- `<interface name>Impl`
- `Default<interface name>`

This is confusing, so, as discussed we should choose a way.
@paullatzelsperger proposed:
> to me there is a slight difference between a Default and an Impl. The way I see it, the Default* convention should eventually supersede the InMemory* stuff. So for example by default, for every registry there is an im-mem implementation (which it essentially a map) which is called Default...Registry. It is conceivable, that someone wants to implement a registry based on some storage other than memory.
The Impl should be used for something that really only ever has one implementation, such as the BlobStoreApiImpl or the DataFlowManagerImpl.

I agree with this, and I think we need also to describe this in the `architecture-principles` markdown file

ndr-brt:
Fix #556


paullatzelsperger (approved):

jimmarino (approved):
LGTM

PR #561, opened 2022-01-27
Feature/557 enhance hierarcal config parsing

### LINKED ISSUE #557: Introduce Config object: enhance hierarchical config parsing
As the hierarchical parsing of configs introduced in #546 may not be 100% intuitive for everyone, lets see if we can come up with a `Config` object, that encapsulates a configuration and would also wrap the tree-like traversing of configuration properties.

ndr-brt:
Fix #557

The main concepts in this PR are:
#### Config
an interface that describes a configuration object that can manipulate settings, return them in different types (at the moment `String`, `Integer`, `Long` are implemented).
Provides a `getConfig` method that permits to navigate the tree, and a `plus` method that permit to sum it with another `Config` object.
#### ConfigImpl
The implementation of `Config`, all the logic and functionality is described by the tests on this class.
#### ConfigFactory
A factory that permits to instantiate a `Config` giving a `Map<String, String>` or a `Properties` instance.


spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/Config.java
```diff
@@ -0,0 +1,119 @@
...
+    Config getConfig(String path);
+
+    /**
+     * Sums two Config objects returning a config that will own the union of the entries (in case of duplicates, other's one will subdue).
+     * The root path will be reset.
+     *
+     * @param other another Config object
+     * @return a Config that's the sum of the current and the other
+     */
+    Config plus(Config other);
```
paullatzelsperger, line 96: To me, `plus` implies some sort of mathematical summation operation - maybe `merge` or `union` would be a more appropriate name?

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/Config.java
```diff
@@ -0,0 +1,119 @@
...
+     *     group.key = value
+     * </pre>
+     * and the rootPath is "group", this method will return a map that contains this entry:
+     * <pre>
+     *     key = value
+     * </pre>
+     *
+     * @return a map containing the config entries relative to the current root path
+     */
+    Map<String, String> getRelativeEntries();
```
paullatzelsperger, line 118: Here, an overload accepting the `rootPath` as parameter might be worth considering, just as a means of convenience.

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java
```diff
@@ -273,6 +233,19 @@ private void addDefaultExtensions(List<ServiceExtension> loadedExtensions) {
         baseDependencies.forEach(se -> loadedExtensions.add(0, se));
     }
 
+    private Config loadConfig() {
```
paullatzelsperger, line 236: I think the order of precedence should be config props < env props < system props, meaning system props overwrite env, and so on. I know that this probably wasn't the case before, but since you're touching it now... :)

paullatzelsperger (changes requested):
Overall a very clean and elegant solution, love it. 
A few minor things commented directly in the code.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/Config.java
ndr-brt: done

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/Config.java
ndr-brt: I avoided it because it can be obtained using `getConfig(path).getRelativeEntries()` and it seemed more straightforward.
To be honest, I don't think those `getEntries`/`getRelativeEntries` will be used a lot, because it will be enough to pass the `Config` object that will give all the beef.

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java
ndr-brt: makes sense. done

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/Config.java
paullatzelsperger: fine with me.

paullatzelsperger (approved):

jimmarino:
@ndr-brt can you fix the conflicts and I'll merge?

ndr-brt:
@jimmarino done

jimmarino:
@ndr-brt looks like something is broken. Can you check again?

ndr-brt:
Got it, that was because Atomikos extension wasn't ready for the new config handling, introduced the missing features, should be ok now.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/Config.java
ndr-brt: At the end you were right, I had to overload the `getRelativeEntries` method :)

PR #567, opened 2022-01-27
docs: sequence diagrams http data transfer

MoritzKeppler:
document state of discussion
diagrams are subject to change

Signed-off-by: Moritz Keppler <moritz.keppler@daimler.com>


docs/diagrams/data-flow/data-flow-api-endpoint.puml
```diff
@@ -0,0 +1,153 @@
...
+    participant CBE as "Backend Service"
+    participant CCP as "Control Plane" #Business
+    participant CDP as "Data Plane" #Business
+end box
+
+box Provider #ProviderColor
+    participant PDP as "Data Plane" #Business
+    participant PCP as "Control Plane" #Business
+    participant PBE as "Backend Service"
+end box
```
florianrusch-cluetec, line 41: If you want to make it clearer that the EDC and BDS are two different services, as we properly would suggest, you could split the Consumer and Provider boxes into two each.

```suggestion
box "Consumer\n controlled zone" #ConsumerColor
    participant CBE as "Backend Service"
end box

box Consumer EDC #ConsumerColor
    participant CCP as "Control Plane" #Business
    participant CDP as "Data Plane" #Business
end box

box Provider EDC #ProviderColor
    participant PDP as "Data Plane" #Business
    participant PCP as "Control Plane" #Business
end box

box "Provider\n controlled zone" #ProviderColor
    participant PBE as "Backend Service"
end box
```

This would look like this:
image

docs/diagrams/data-flow/data-flow-api-endpoint.puml
```diff
@@ -0,0 +1,153 @@
...
+
+title
+ **Access to REST API**
+ <sub>Control Plane and Date Plane are part of the EDC.</sub>
+ <sub>Precondition: contract 'access to api' iuncluding constraints (e.g. how often, for which purpose etc.) is signed</sub>
```
florianrusch-cluetec, line 46:
```suggestion
 <sub>Precondition: contract 'access to api' including constraints (e.g. how often, for which purpose etc.) is already signed</sub>
```

docs/diagrams/data-flow/data-flow-api-endpoint.puml
```diff
@@ -0,0 +1,153 @@
...
+
+ Case 1a: Each API call matches one TransferProcess (http push)
+ <sub>Provider is pushing data to the Consumer</sub>
+end title
+
+CBE ->> CCP : getAsset(query params, int destination)
```
florianrusch-cluetec, line 52: Do you think of something when you defined the destination to be an int?

docs/diagrams/data-flow/data-flow-api-endpoint.puml
```diff
@@ -0,0 +1,153 @@
...
+activate CCP
+CCP ->> CDP : provision ext endpoint
+CCP ->> PCP : start data transfer(asset, contract, query params, ext endpoint)
```
florianrusch-cluetec, line 55: Should we be more specific at this point? Or is it widely known that it is more or less the same?
```suggestion
CCP ->> PCP : start data transfer(asset, contractAggreement, query params, ext endpoint)
```

docs/diagrams/data-flow/data-flow-api-endpoint.puml
florianrusch-cluetec: Ahhh this means internal 🙈
Maybe adding a dot would make it clearer, thats these are abbreviations: `int.` and `ext.`

docs/diagrams/data-flow/data-flow-api-endpoint.puml
```diff
@@ -0,0 +1,153 @@
...
+== Working ==
+note right of CCP : stays in state\nSTREAMING
+note left of PCP : stays in state\nSTREAMING
+CBE -> CDP : API Call (GET/PUT/POST/DELETE/...)
+note left of CDP : Dataplane\nacts as proxy.\nCould adjust calls
+activate CDP
+CDP -> PDP : API Call
+note right of PDP : acts as proxy\nsee Consumer
+activate PDP
+PDP -> PBE : API Call
```
florianrusch-cluetec, line 136: I would keep it like in the other pictures:
```suggestion
PDP -> PBE : pull data
```

florianrusch-cluetec (changes requested):
Looks very good! I just have some small comments.

florianrusch-cluetec:
The storage of data in the backend data services was previously designed in such a way that these could also be data caches inside or outside the BDS. This is still missing in your graphics, whereby this would make the whole thing even more complicated. So I would say, let's leave them out of these images for now or?

docs/diagrams/data-flow/data-flow-api-endpoint.puml
```diff
@@ -0,0 +1,153 @@
...
+CDP ->> CCP : notify: data received
+deactivate CDP
+CCP ->> CDP : deprovision
+PCP ->> PBE : deprovision\n(if needed)
+deactivate CCP
+deactivate PCP
+
+
+
+newpage Case 1b: Each API call matches one TransferProcess (http pull)\n<sub>Consumer is pulling data from Provider</sub>
```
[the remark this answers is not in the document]
dr-markus: In step 6 the Backend Data Service would actually call the end of the data plane to proxy the data through to the receiving data plane. Same in case of streaming - the data plane of the consumer has to wait as the DP will no know when date is finally provided.

docs/diagrams/data-flow/data-flow-api-endpoint.puml
MoritzKeppler: but it's not a "pull" data. Can also be a POST call, a DELETE call etc.  Think of any REST interface.

MoritzKeppler:
> The storage of data in the backend data services was previously designed in such a way that these could also be data caches inside or outside the BDS. This is still missing in your graphics, whereby this would make the whole thing even more complicated. So I would say, let's leave them out of these images for now or?

yes, wanted to keep those details out of the diagram. Backend could be a complete, complex structure of data caches, data lakes, business applications....

docs/diagrams/data-flow/data-flow-api-endpoint.puml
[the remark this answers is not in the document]
MoritzKeppler: hm, interesting question.  Would make sense if more than one call will be made to the provider, similar to diagram 2.
Wonder if it is sufficient to have it like shown in diagram 2 or if we a mix, i.e. a data plane that is actively copying data instead of just being a proxy.

PR #577, opened 2022-01-28
Fix Sample 5: IdsId parsing for URNs, closes #562

### LINKED ISSUE #562: Fix Sample 5: ContractAgreement not found
When the Consumer sends an `ArtifactRequestMessage` to the Provider, that contains a contract agreement. The same agreement exists on provider and consumer side. But the Provider is not able to identify the corresponding ContractAgreement and rejects the data transfer and the `ArtifactRequestHandler`.

The underlying problem is, that the id is incorrectly converted from the IDS-id.
```
IDS-ID: urn:contract:1:e0002542-6a46-4896-804a-112e237df9ac
```

is converted to
```
EDC-ID: 1
```
but should be 
```
EDC-ID: 1:e0002542-6a46-4896-804a-112e237df9ac
```

The issue resides within the the `IdsIdParser`, that does not expect an URN within an URN:
```
public class IdsIdParser {

    public static final String DELIMITER = ":";

    public static IdsId parse(String urn) {

       /* code */

        String[] parts = urn.split(DELIMITER, 3);

       /* more code */

       String idValue = parts[2];

       return IdsId.Builder.newInstance().type(type).value(idValue).build();
    }
}
``` 

DominikPinsel:

### DIFF
--- a/data-protocols/ids/ids-spi/src/main/java/org/eclipse/dataspaceconnector/ids/spi/IdsIdParser.java
+++ b/data-protocols/ids/ids-spi/src/main/java/org/eclipse/dataspaceconnector/ids/spi/IdsIdParser.java
@@ -1,27 +1,33 @@
 package org.eclipse.dataspaceconnector.ids.spi;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 /**
  * ID / URI parser for IDS resources.
  */
 public class IdsIdParser {
+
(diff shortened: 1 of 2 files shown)


DominikPinsel:
Just saw that `String[] parts = urn.split(DELIMITER, 3);` works as intended and I have probably mixed up this with another issue yesterday, where the id was just "1".
Will close this PR and the issue.

PR #582, opened 2022-01-31
extensions: fix typo

ndr-brt:
This will fix bug reported here: #581


ndr-brt:
I also removed a duplicated service registration of `JettyService`, that caused a warning printed on the console

jimmarino (approved):

PR #595, opened 2022-02-01
extensions: introduce nbf jwt claim leeway configuration

### LINKED ISSUE #594: OAuth2 notBefore (nbf) validation
I'm having issues with the `notBefore` validation, because in an environment that I deployed seems like the DAPS server timestamp is some seconds ahead the connector's one so I'm getting the **Token is not valid yet** error, that is caused by the `checkRule` in `Oauth2ValidationRule` 

https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/6cd65111e81d7cfd189e935bd9f6beb28259c345/extensions/iam/oauth2/oauth2-core/src/main/java/org/eclipse/dataspaceconnector/iam/oauth2/core/impl/Oauth2ValidationRule.java#L37

The [RFC 7591](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.5) reports that
> Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew.

and also

> Use of this claim is OPTIONAL

So there could be two proposals:
* introducing a configurable leeway for the `nbf` claim validation.
* introducing a configurable switch to disable `nbf` claim validation.

I'm more inclined to the first but any opinion is appreciated.

ndr-brt:
Fix #594


extensions/iam/oauth2/oauth2-core/README.md
```diff
@@ -4,12 +4,13 @@ This extension provides an `IdentityService` implementation based on the OAuth2
...
+| `edc.oauth.provider.audience`     | Provider audience                                                                  | false     | id of the connector             |
+| `edc.oauth.provider.jwks.url`     | URL from which well-known public keys of Authorization server can be fetched       | false     | http://localhost/empty_jwks_url | 
+| `edc.oauth.public.key.alias`      | Alias of public associated with client certificate                                 | true      | null                            |
+| `edc.oauth.private.key.alias`     | Alias of private key (used to sign the token)                                      | true      | null                            |
+| `edc.oauth.provider.jwks.refresh` | Interval at which public keys are refreshed from Authorization server (in minutes) | false     | 5                               |
+| `edc.oauth.client.id`             | Public identifier of the client                                                    | true      | null                            |
+| `edc.oauth.validation.nbf.leeway` | Leeway in seconds to remedy clock skew on notBefore claim validation               | false     | 10                              |
```
tmberthold, line 16: Obvious to us since we know the problem, but perhaps clearly note that the seconds are added to the Connector's time and not subtracted in standard case (unless someone enters a negative int).

something like: `Leeway in seconds added to current time to remedy clock skew on notBefore claim validation`

tmberthold (changes requested):

bscholtes1A (approved):

tmberthold (approved):

ndr-brt:
Thanks for reviewing, I take the opportunity to divide classes in two packages: `jwt` and `identity`, that are more meaningful than the former `impl` one.

bscholtes1A (approved):

ndr-brt:
@jimmarino I think this can be merged

tmberthold (approved):

PR #597, opened 2022-02-01
Optional injection points are not reported as errors anymore

paullatzelsperger:
`@Inject(required=false)` injection points were reported as errors, which is wrong. This PR fixes that.

### DIFF
--- a/core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java
+++ b/core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java
@@ -22,10 +22,9 @@ import org.eclipse.dataspaceconnector.spi.monitor.Monitor;
 import org.eclipse.dataspaceconnector.spi.system.ConfigurationExtension;
 import org.eclipse.dataspaceconnector.spi.system.EdcInjectionException;
 import org.eclipse.dataspaceconnector.spi.system.Feature;
-import org.eclipse.dataspaceconnector.spi.system.FieldInjectionPoint;
-import org.eclipse.dataspaceconnector.spi.system.Inject;
 import org.eclipse.dataspaceconnector.spi.system.InjectionContainer;
 import org.eclipse.dataspaceconnector.spi.system.InjectionPoint;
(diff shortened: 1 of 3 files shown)


bscholtes1A (approved):

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java
```diff
@@ -281,7 +283,12 @@ private void addDefaultExtensions(List<ServiceExtension> loadedExtensions) {
 
         var injectFields = Arrays.stream(ext.getClass().getDeclaredFields())
                 .filter(f -> f.getAnnotation(Inject.class) != null)
-                .map(f -> new FieldInjectionPoint<>(ext, f, getFeatureValue(f.getType())));
+                .map(f -> {
+                    var isRequired = f.getAnnotation(Inject.class).required();
+                    var fip = new FieldInjectionPoint<>(ext, f, getFeatureValue(f.getType()));
```
ndr-brt, line 288: This method is equal to `InjectPointScanner.getInjectionPoints`, so the duplication could be removed making this method calling the `InjectPointScanner`.

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java
```diff
@@ -281,7 +283,12 @@ private void addDefaultExtensions(List<ServiceExtension> loadedExtensions) {
...
+                    fip.setRequired(isRequired);
```
ndr-brt, line 289: I would add the `required` parameter to the constructor to avoid the set methods and leaving the `FieldInjectionPoint` object immutable

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java
paullatzelsperger: good catch!

ndr-brt (approved):

PR #598, opened 2022-02-01
DataManagementApi: Creates Auth SPI

### LINKED ISSUE #583: DataManagementApi: Create re-usable authentication/authorization SPI
_This issue only deals with the SPI, i.e. only interfaces. An implementation is explicitly NOT required._

We want to handle authentication/authorization in a centralized way, therefore we need two extensions:
  - `extensions:api:auth-spi`: contains interfaces (`AuthorizationService`) that the API controllers can then use. This module must also contain an implementation of the `ContainerRequestFilter` that performs the auth using the `AuthorizationService`
  - `api:auth-tokenbased`: implements the auth service interface, uses token-based authentication. This should also replace the `HttpApiKeyAuthContainerRequestFilter`.

paullatzelsperger:
This PR lays the groundwork for a central authentication/authorization framework for the DataManagementApi. It adds a request filter that delegates out to an `AuthorizationService`

Closes #583


denisneuling (changes requested):
Hey Paul, 

It looks like you're introducing authentication functionality (not to confuse with authorization), so I'd propose renaming

* AuthorizationFailedException to **Authentication**FailedException
* AuthorizationExtension to **Authentication**Extension
* AuthorizationRequestFilter to **Authentication**RequestFilter
* AuthorizationService to **Authentication**Service
* TokenBasedAuthorizationExtension to TokenBased**Authentication**Extension
* TokenBasedAuthorizationService to TokenBased**Authentication**Service 

paullatzelsperger:
@denisneuling Sounds reasonable.

extensions/api/auth-spi/src/main/java/org/eclipse/dataspaceconnector/api/auth/AuthenticationExtension.java
```diff
@@ -0,0 +1,37 @@
...
+    private AuthenticationService authenticationService;
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        //        if (authenticationService == null) {
+        //            context.getMonitor().warning("No AuthenticationService was registered. Will deny all requests!");
+        //            authenticationService = (headers) -> false;
+        //        }
+
+        // todo: this extension will eventually go away - the request filter should be registered by every API separately
```
ndr-brt, line 35: Accoring to this todo the extension will go away, but since it does nothing I think it can go away right now

extensions/api/auth-spi/src/main/java/org/eclipse/dataspaceconnector/api/auth/AuthenticationRequestFilter.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) {
+        var headers = requestContext.getHeaders();
+
+        var isAuthenticated = authenticationService.isAuthenticated(headers.entrySet().stream().collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
+
+        if (!isAuthenticated) {
+            throw new NotAuthorizedException();
```
ndr-brt, line 38: shouldn't be `AutenticationFailedException`?

extensions/api/auth-tokenbased/src/main/java/org/eclipse/dataspaceconnector/api/auth/TokenBasedAuthenticationService.java
```diff
@@ -0,0 +1,54 @@
...
+    public boolean isAuthenticated(Map<String, List<String>> headers) {
+
+        Objects.requireNonNull(headers, "headers");
+
+        var apiKey = headers.keySet().stream()
+                .filter(k -> k.equalsIgnoreCase(API_KEY_HEADER_NAME))
+                .map(headers::get)
+                .findFirst();
+
+        return apiKey.map(this::checkApiKeyValid).orElseThrow(() -> new AuthenticationFailedException(API_KEY_HEADER_NAME + " not found"));
```
ndr-brt, line 48: Won't be better to avoid throwing the exception here and leave the filter to throw it in case the return of this is false?

paullatzelsperger: The idea was to leave the decision, when an Authentication has failed (as opposed to: not authorized) up to the impl. of the `AuthenticationService`. E.g. a token-based Auth Service might require an `X-Api-Key` header, whereas an OAuth2-based impl. may required different headers.

extensions/api/auth-spi/src/main/java/org/eclipse/dataspaceconnector/api/auth/AuthenticationExtension.java
paullatzelsperger: that's correct. That will be done during issue #645

extensions/api/auth-spi/src/main/java/org/eclipse/dataspaceconnector/api/auth/AuthenticationRequestFilter.java
paullatzelsperger: I think there is a difference between not being able to perform an authentication (`AuthenticationFailedException`) and the corresponding entity not being _authorized_ (`NotAuthorizedException`).

paullatzelsperger: to quote [Wikipedia](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors):
> Http 401: Similar to 403 Forbidden, but specifically for use when authentication is required and has failed or has not yet been provided

So the 401 is returned when the auth service throws an `AuthenticationFailedException`, but the 403 is returned, when the authentication header was understood but the "server refuses action".

jimmarino (approved):

ndr-brt (approved):

denisneuling (approved):

Issue #584, opened 2022-01-31
DataManagementApi: expose separate Jetty context


paullatzelsperger:
The Data Management API should be exposable on a versioned root path `/api/v1/` and even on a different port. That should be optional though. We therefore need another Jetty context handler.
The version can be hard-coded for now.

PR #600, opened 2022-02-01
http: add support for multiple ports

bcronin90:
DRAFT

This PR will include changes to jetty extension to allow for different ports for different servlets.

Brendan Cronin brendan.cronin@daimler.com, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)


paullatzelsperger:
There is people working on a similar feature (#584) that encompasses what you did here. So what you're effectively doing is doubling the work.
Please coordinate with the other people in this project and stick to the milestone plan.

PR #601, opened 2022-02-01
Configurable Jetty Contexts and ports

### LINKED ISSUE #584: DataManagementApi: expose separate Jetty context

paullatzelsperger:
This PR adds features to expose certain controllers/endpoints on different Jetty contexts. This is helpful to scope e.g. request filters, so that they only fire on those controllers.
Also, this PR adds a feature to specify a port per context, e.g. exposing a management API.

Depends on #561 for config parsing.

Enables #583

Closes #584

### DIFF
--- a/extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/JerseyRestService.java
+++ b/extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/JerseyRestService.java
@@ -24,20 +24,22 @@ import org.glassfish.jersey.media.multipart.MultiPartFeature;
 import org.glassfish.jersey.server.ResourceConfig;
 import org.glassfish.jersey.servlet.ServletContainer;
 
-import java.util.HashSet;
-import java.util.Set;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 import static java.util.stream.Collectors.toSet;
 import static org.glassfish.jersey.server.ServerProperties.WADL_FEATURE_DISABLE;
 
 public class JerseyRestService implements WebService {
-    private static final String API_PATH = "/api/*";
+    private static final String DEFAULT_API_PATH = "/*";
 
     private final JettyService jettyService;
     private final TypeManager typeManager;
     private final Monitor monitor;
 
-    private final Set<Object> controllers = new HashSet<>();
+    private final Map<String, List<Object>> controllers = new HashMap<>();
     private final CorsFilterConfiguration corsConfiguration;
 
     public JerseyRestService(JettyService jettyService, TypeManager typeManager, CorsFilterConfiguration corsConfiguration, Monitor monitor) {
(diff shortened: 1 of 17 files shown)


extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/JerseyRestService.java
```diff
@@ -49,34 +51,53 @@ public JerseyRestService(JettyService jettyService, TypeManager typeManager, Cor
 
     @Override
     public void registerController(Object controller) {
```
bcronin90, line 53: This is probably just here to be backwards-compatible, right? In that case it could be @Deprecated

bcronin90: Never mind, saw it below.

jimmarino (approved):

extensions/http/jetty/src/test/java/org/eclipse/dataspaceconnector/extension/jetty/JettyConfigurationTest.java
```diff
@@ -0,0 +1,119 @@
...
+        ThrowableAssert.ThrowingCallable doubleDefaultConfig = () -> JettyConfiguration.createFromConfig(null, null, ConfigFactory.fromMap(Map.of(
+                "web.http.port", "1234",
+                "web.http.default.port", "8888"
+        )));
+
+        assertThatThrownBy(doubleDefaultConfig).isInstanceOf(IllegalArgumentException.class).hasMessageStartingWith("A port mapping for port '8888' already exists");
+    }
+
+    @Test
+    void createFromConfig_multiPathConfig() {
```
ndr-brt, line 80: Is this case really needed?

extensions/http/jetty/src/test/java/org/eclipse/dataspaceconnector/extension/jetty/JettyConfigurationTest.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Test
+    void createFromConfig_implicitAndExplicitDefault_shouldThrowException() {
+        ThrowableAssert.ThrowingCallable doubleDefaultConfig = () -> JettyConfiguration.createFromConfig(null, null, ConfigFactory.fromMap(Map.of(
+                "web.http.port", "1234",
+                "web.http.default.port", "8888"
+        )));
+
+        assertThatThrownBy(doubleDefaultConfig).isInstanceOf(IllegalArgumentException.class).hasMessageStartingWith("A port mapping for port '8888' already exists");
```
ndr-brt, line 76: Is the message correct? the problem is that the "default" mapping could be defined, not that the port 8888 is already defined (because it isn't)

extensions/http/jetty/src/main/java/org/eclipse/dataspaceconnector/extension/jetty/JettyConfiguration.java
```diff
@@ -14,20 +14,91 @@
...
     }
 
-    public int getHttpPort() {
-        return httpPort;
+    public static JettyConfiguration createFromConfig(String keystorePassword, String keymanagerPassword, Config config) {
+        var jettyConfig = new JettyConfiguration(keystorePassword, keymanagerPassword);
+
+        var subConfig = config.getConfig(WEB_HTTP_PREFIX);
+
+        Map<String, Map<String, String>> tempMappings = new HashMap<>();
```
ndr-brt, line 50: This logic (without supporting multipathconfig, that in my opinion are not really useful as I commented in the test class) could be handled by the `Config` object adding a simple `isLeaf` method on it, that tells you if there's only one entry and its path equals root path and using this code snipped instead:

```
Stream.concat(Stream.of(subConfig), subConfig.partition())
                .filter(it -> !it.isLeaf())
                .map(c -> {
                    String currentNode = c.currentNode();
                    if (currentNode.equals("default")) throw new IllegalArgumentException("A 'default' port mapping cannot be defined");

                    var name = currentNode.equals("http") ? "default" : currentNode;
                    return new PortMapping(name, c.getInteger("port", DEFAULT_PORT), c.getString("path", DEFAULT_PATH));
                })
                .forEach(jettyConfig::portMapping);
```

extensions/http/jetty/src/test/java/org/eclipse/dataspaceconnector/extension/jetty/JettyConfigurationTest.java
paullatzelsperger: it should probably be renamed to `contextAliasWithDotsInIt`, as the intention was to test that dots in the middle don't change the result.

extensions/http/jetty/src/main/java/org/eclipse/dataspaceconnector/extension/jetty/JettyConfiguration.java
paullatzelsperger: I quickly threw this into `ConfigImpl` assuming that's what you meant:
```java
    @Override
    public boolean isLeaf() {
        return entries.size() == 1 && entries.keySet().stream().allMatch(rootPath::equals);
    }
```
The only thing which then won't work is something like `web.http.this.is.a.long.port`, where `this.is.a.long` would previously have been regarded as the context alias.
That's a minor drawback though, so I'm good with adopting your solution.

ndr-brt: yes, it does, is the 
```var name = currentNode.equals("http") ? "default" : currentNode;```
line that does the job

File: extensions/http/jetty/src/test/java/org/eclipse/dataspaceconnector/extension/jetty/JettyConfigurationTest.java (hunk 2 above)
Comment (ndr-brt) [reply extensions/http/jetty/src/test/java/org/eclipse/dataspaceconnector/extension/jetty/JettyConfigurationTest.java]: @paullatzelsperger yes, but why we should need that dots in between? the dots represent a configuration "level" so if there we want to list different port mappings, they should stay at the same level. In this case the property should be configured as:
```web.http.this_is_longer.port```

File: extensions/http/jetty/src/main/java/org/eclipse/dataspaceconnector/extension/jetty/JettyConfiguration.java (hunk 4 above)
Comment (paullatzelsperger) [reply extensions/http/jetty/src/main/java/org/eclipse/dataspaceconnector/extension/jetty/JettyConfiguration.java:50]: @ndr-brt are you OK with merging this as-is and investigate an improvement down the road? This PR is required for a bunch of others, and MS2 is fast approaching :)

Comment (ndr-brt) [reply extensions/http/jetty/src/main/java/org/eclipse/dataspaceconnector/extension/jetty/JettyConfiguration.java:50]: ok

**Review by ndr-brt — APPROVED** [review]

File: extensions/http/jetty/src/main/java/org/eclipse/dataspaceconnector/extension/jetty/JettyService.java [hunk 5]
```diff
@@ -56,26 +60,37 @@ public JettyService(JettyConfiguration configuration, KeyStore keyStore, Monitor
...
+                } else {
+                    connector = httpServerConnector(mapping.getPort());
+                    monitor.info("HTTP context '" + mapping.getName() + "' listening on port " + mapping.getPort());
+                }
+                connector.setName(mapping.getName());
+                server.addConnector(connector);
+
+                ServletContextHandler handler = createHandler(mapping);
+                handlers.put(mapping.getPath(), handler);
+                server.setErrorHandler(new JettyErrorHandler());
```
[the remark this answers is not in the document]
bcronin90: Oh, merged already. Never mind.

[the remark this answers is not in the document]
paullatzelsperger: I will address it in a small bug fix PR. We were a bit pressed for time :)

PR #612, opened 2022-02-03
Introduce a contract agreement end to end test

ndr-brt:
I think I understood that someone's already working on this, but I wanted to give a shot on this.

#### What this PR brings
A simple real end to end test (`IdsEndToEndTest`) that relies on two connectors loaded through `docker-compose` and starts a contract negotiation between them. When the consumer approves the contract agreement, the test is completed.
The test is run with a new workflow (`e2e-tests.yaml`) that can be launched without external dependencies, since all the stores are in memory.
The workflow creates a specific docker image for the test, then starts a docker-compose environment, at the end, launches the tests inside the e2e-test launcher.

#### Where it can be improved
At the moment no file transfer are executed, I guess that to have a real transfer we should wait the introduction of the data plane, otherwise we could introduce something specific for the test that will transfer something through http, but I don't know how that would be valuable.
Asset and ContractDefinition are created into the extension itself, a real E2E test should take care of create the resources on the test execution, but at the moment there aren't api to do that.

Feedback are welcome


jimmarino:
I'm going to loop in @algattik here because he is working on enhancing the JUnit runner to support launching multiple instances in the same VM. The idea is by having such a runner we would be able to simulate very fine-grained conditions and scenarios.

algattik:
Right, we have a proto working and will share ~ tomorrow. @cpeeyush

ndr-brt:
Looking forward to it, eventually I will close this

jimmarino:
> Looking forward to it, eventually I will close this

@ndr-brt I merged #665 so we should be good-to-go with writing end-to-end tests.

ndr-brt:
Yes, maybe in another PR

PR #621, opened 2022-02-03
Minor SPI package cleanup

jimmarino:
This PR makes the following package adjustments in core-spi:

1. WaitStrategy moves from the top-level SPI package to the retry subpackage.
2. ObjectFactory and ReflectiveObjectFactory move from the top-level SPI package to the system subpackage. This removes package cycles.
3. Observable moves from the top-level SPI package to the observe subpackage.

### DIFF
--- a/spi/core-spi/src/test/java/org/eclipse/dataspaceconnector/spi/ObservableTest.java
+++ b/spi/core-spi/src/test/java/org/eclipse/dataspaceconnector/spi/ObservableTest.java
@@ -1,5 +1,6 @@
 package org.eclipse.dataspaceconnector.spi;
 
+import org.eclipse.dataspaceconnector.spi.observe.Observable;
 import org.junit.jupiter.api.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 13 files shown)


spi/core-spi/src/test/java/org/eclipse/dataspaceconnector/spi/ObservableTest.java
```diff
@@ -1,5 +1,6 @@
 package org.eclipse.dataspaceconnector.spi;
```
ndr-brt, line 1: this should go into the `observe` package

ndr-brt (approved):

spi/core-spi/src/test/java/org/eclipse/dataspaceconnector/spi/ObservableTest.java
jimmarino: ah missed that one, will do

bscholtes1A (approved):

paullatzelsperger (approved):

PR #624, opened 2022-02-04
spi: apply composition on the Observable component over the existing abstraction

ndr-brt:
Currently, `Observable` is an abstract class that can be extended, same for `TransferProcessObservable` and `ContractNegotiationObservable`, and it's extended by `TransferProcessManagerImpl` and the two `ContractNegotiationManagerImpl`.
This is not a good practice, becase it gives to the `*Manager`s an additional behavior when it's better to have a collaborator that could take care of such behavior and, the reason that made me approach to this refactor: it makes impossible to extract common logic between the two `ContractNegotiationManager` because, as you would know, it's possible to extend only one superclass in java.

This PR transform `Observable` to an interface, same for `TransferProcessObservable` and `ContractNegotiationObservable` and provides a single implementation for it. On the managers then the observable is a collaborator that it's created and injected at the object instantiation phase.


core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractServiceExtension.java
```diff
@@ -118,18 +119,21 @@ private void registerServices(ServiceExtensionContext context) {
 
         var waitStrategy = context.hasService(NegotiationWaitStrategy.class) ? context.getService(NegotiationWaitStrategy.class) : new ExponentialWaitStrategy(DEFAULT_ITERATION_WAIT);
 
+        ContractNegotiationObservableImpl observable = new ContractNegotiationObservableImpl();
```
ronjaquensel, line 122: The observable should also be registered in the context.

ndr-brt: agree. done

paullatzelsperger (approved):

Issue #542, opened 2022-01-23
Specification of the Data Management API 


paullatzelsperger:
This Issue deals with creating a specification of the Data Management API based on the design considerations and recommendations stated in the [attached document](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/files/7920322/data_management_api.md). I didn't want to paste directly here, to avoid walls of text. 

The TL;DR:
### 1. Endpoints vs Use cases
Following the REST principles, every endpoint should modify the _state_ of an object rather than trigger _behaviour_. For example, to initiate a data transfer the endpoints should be `POST http://your-connector.com/api/v1/transfer-request` rather than `POST http://your-connector.com/api/v1/transfer/start-transfer`
Consequently, uses cases (as outlined in the [attached document](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/files/7920322/data_management_api.md)) might not have a 1-to-1 translation into API endpoints, they might even involve several consecutive API calls.

### 2. `copy/clone` instead of `update`
every modification triggers the creation of the domain object. 
### 3. support pagination
every `GET` endpoint should support pagination, sorting and filtering

### 4. Desired outcome
We should have a (or several) YAML files containing the specification of each of the endpoints. In addition, we should have `Controller` classes that implement that spec, even if most of them are just stubs returning dummy data for the immediate future. So, by referencing the `data-management`-BOM, a controller implicitly exposes the entire API. 
_Note: we will **NOT** consider the standalone deployment scenario, especially the communication channel, in this issue_

### 5. Module structure
Every domain object should be reflected in the module
```bash
extensions/api/data-management/asset                      //includes DataAddress
extensions/api/data-management/contractdefinition
extensions/api/data-management/contractnegotiation
extensions/api/data-management/transferprocess
extensions/api/data-management/policy                      //includes PolicyTemplate
```
a BOM should exist for the entire API.

jimmarino:
Good start. One quick comment: 

extensions/api/data-management/asset 

In my mind doesn't model an asset but an "asset/data address pair". We should come up with a different name for it. You could probably make an argument that "Asset" without the data address is one representation of this "pair" but I think it is a separate thing. For example, the data address of an asset can change without that semantically resulting in a new asset.

paullatzelsperger:
We've called that an `AssetEntry` in previous implementations. That's also what the `DataLoader` utilizes at the moment. WDYT?

[...]

PR #630, opened 2022-02-07
Added generic exception mapper

### LINKED ISSUE #586: DataManagementApi: create centralized exception handler
Controller classes should not return individual HTTP Status Codes themselves, thus they must not return `Response` objects. Instead they should return the DTOs directly or throw exceptions.

All exceptions are caught in one place and mapped to HTTP status codes. This can be achieved by implementing the `ExceptionMapper` interface (check [this example](https://mincong.io/2018/12/03/exception-handling-in-jax-rs/)). 

The following mapping should be implemented: 
- 400: `IllegalArgumentException`, should be thrown whenever query params etc. are missing or invalid. May also include `NullpointerException`.
- 401: `AuthorizationService` (Issue #583) throws `AuthorizationFailedException`, no authentication header present
- 403: `AuthorizationService` throws `NotAuthorizedException`, supplied auth header does not have the proper permissions
- 404: `ObjectNotFoundException`
- 409: `ObjectExistsException`, usually when creating an object whos ID already exists
- 423: `ObjectNotModifiableException`, `IllegalStateException`, thrown when e.g. an Asset cannot be deleted anymore, because it's already used in a contract.
- 501: `UnsupportedOperationException`
- 503: any other unhandled exception

paullatzelsperger:
closes #586

### DIFF
--- a/extensions/api/api-core/build.gradle.kts
+++ b/extensions/api/api-core/build.gradle.kts
@@ -1,0 +1,43 @@
+/*
+ *  Copyright (c) 2021 Daimler TSS GmbH
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Daimler TSS GmbH - Initial API and Implementation
+ *
+ */
+
+
+val infoModelVersion: String by project
+val rsApi: String by project
+val jerseyVersion: String by project
+
+plugins {
+    `java-library`
+}
+
+dependencies {
+    api(project(":spi"))
+    implementation(project(":common:util"))
+    testImplementation(project(":core:base"))
+
+    implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
+    testImplementation("org.glassfish.jersey.core:jersey-common:${jerseyVersion}")
+    testImplementation("org.glassfish.jersey.core:jersey-server:${jerseyVersion}")
+
+    testImplementation(testFixtures(project(":launchers:junit")))
+}
+
+publishing {
+    publications {
+        create<MavenPublication>("api-core") {
+            artifactId = "api-core"
+            from(components["java"])
+        }
+    }
+}
--- a/extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/model/Transformer.java
+++ b/extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/model/Transformer.java
@@ -1,0 +1,8 @@
+package org.eclipse.dataspaceconnector.api.datamanagement.contractdefinition.model;
+
+public interface Transformer<DTO, DOMAINOBJECT> {
+
+    DTO convertToDto(DOMAINOBJECT domainobject);
+
+    DOMAINOBJECT convertToObject(DTO dto);
+}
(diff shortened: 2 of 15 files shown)


extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/model/Transformer.java
```diff
@@ -0,0 +1,8 @@
+package org.eclipse.dataspaceconnector.api.datamanagement.contractdefinition.model;
```
ndr-brt: Probably not part of this PR

extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/exception/EdcApiExtension.java
```diff
@@ -0,0 +1,24 @@
...
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.api.exception;
+
+import org.eclipse.dataspaceconnector.spi.EdcException;
+
+//marker class
+public abstract class EdcApiExtension extends EdcException {
```
ndr-brt, line 20: Probably a typo, should be called `EdcApiException`

extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/exception/mappers/EdcApiExceptionMapper.java
```diff
@@ -0,0 +1,55 @@
...
+import org.eclipse.dataspaceconnector.api.exception.AuthorizationFailedException;
+import org.eclipse.dataspaceconnector.api.exception.NotAuthorizedException;
+import org.eclipse.dataspaceconnector.api.exception.ObjectExistsException;
+import org.eclipse.dataspaceconnector.api.exception.ObjectNotFoundException;
+import org.eclipse.dataspaceconnector.api.exception.ObjectNotModifiableException;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Provider
```
ndr-brt, line 29: Would be better to register it to the `resourceConfig`explicitly?
It will make this process more readable (those kind of annotation hide details and that's not a great thing)

extensions/api/api-core/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
+/*
```
ndr-brt, line 1: This module should be declared as dependency of `control-api` and `observability-api` modules

extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/exception/mappers/EdcApiExceptionMapper.java
```diff
@@ -0,0 +1,55 @@
...
+public class EdcApiExceptionMapper implements ExceptionMapper<Throwable> {
+    private final Map<Class<? extends Throwable>, Integer> exceptionMap;
+
+    public EdcApiExceptionMapper() {
+        exceptionMap = new HashMap<>(Map.of(
```
ndr-brt, line 34: the `HashMap` wrapping is not needed, `Map.of` already did the job.

extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/model/Transformer.java
paullatzelsperger: already deleted. was a copy-paste error.

extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/exception/EdcApiExtension.java
paullatzelsperger: of course :D

extensions/api/api-core/build.gradle.kts
paullatzelsperger: The upcoming feature of configurable jetty contexts (#601) will allow that different APIs are exposed under different ports/context paths, so the `ExceptionMapper` will have to be registered explicitly in every one of these. Declaring the dependency onto `api-core` will only register it for the default context.

Slightly unrelated: `control-api` and `datamanagement-api` (#542) have a considerable overlap, so I think it makes sense to collate them into one API and expose them on the same context.


Also, for the `observability-api` I think this mapper really doesn't provide much value.

ndr-brt (approved):

Issue #626, opened 2022-02-04
Integration test harness supporting multiple runtimes


algattik:
An integration test harness that supports multiple runtimes is driven from multiple needs:

- Stabilizing samples that run multiple connectors, which have been breaking frequently.
- Testing system behavior when multiple connectors interact, e.g. the contract negotiation process.
- Testing system behavior upon component failure.
- Providing a test facility for factoring out application components to separate runtimes (e.g. https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/463).

Key drivers for the choice are:

- Fast and efficient run in CI.
- Fast "inner loop" (i.e. ability to quickly rerun tests after changing code) and debuggability for developers.
- Use of existing frameworks, stability and portability.

Tasks:
- [ ] Prototype and propose an approach for building such tests.
- [ ] Implement test for sample 04.0.

PR #636, opened 2022-02-07
Decision record on Integration testing (#626)

algattik:
Relates to #626.

This is a proposal for a decision on how to go forward with the Integration test framework.

After we agree on this decision we can contribute a PR based on the spike code, with proper structure (e.g. integrate with EdcExtension) and JavaDoc.

### DIFF
--- a/docs/developer/decision-records/2022-02-03-integration-testing/README.md
+++ b/docs/developer/decision-records/2022-02-03-integration-testing/README.md
@@ -1,0 +1,302 @@
+# Integration testing
+
+## Decision
+
+Extend the existing `EdcExtension` JUnit facility (that can currently run a single EDC runtime and supports stubbing and extension of runtime services) as follows:
+
+- Load more than one EDC runtime.
+- Load EDC runtimes in separate Class Loaders.
+- Run each EDC runtime Class Loader with its effective runtime class path, based on its module's Gradle configuration.
+
+## Rationale
+
+The need to provide an integration test harness that supports multiple runtimes emerges from multiple needs:
+
+- Stabilizing [samples](../../samples) that run multiple connectors, which have been breaking frequently.
+- Testing system behavior when multiple connectors interact, e.g. the contract negotiation process.
+- Testing system behavior upon component failure.
+- Providing a test facility for factoring out application components to separate runtimes (e.g. [DPF](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/463)).
+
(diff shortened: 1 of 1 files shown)


jimmarino:
Hi @algattik nice write-up (at first, I didn't realize it was in the README). 

I like the idea of using a particular project classpath to load runtime instances in isolated classloaders. However, there seems to be a lot of configuration required in the build script to enable that. Could we wrap that in a Gradle plugin? I've done this on previous projects and the only tricky part of this is making the plugin available within the same repository. That would give us the advantage of accessing the Gradle API to obtain the classpath and also make setup of the different runtimes user friendly.

In the previous projects, we created a separate plugin repo but that was tedious. Could we instead look into a way to reference the plugin as part of the plugin classpath via a local URL and have modules that rely on the plugin declare it as a task dependency?

algattik:
Actually all that is needed is to add once to the **root** build.gradle.kts these 5 lines.

```kotlin
// build.gradle.kts (under allprojects)
     tasks.register("printClasspath") {​​​​​​​
     doLast {​​​​​​​
        println("${​​​​​​​sourceSets["main"].runtimeClasspath.asPath}​​​​​​​");
        }​​​​​​​
    }​​​​​​​
```

Then we just build once a quite simple class (that could extend from EdcExtension to inherit all the mocking behavior) and calls `Runtime.getRuntime().exec(root +"/gradlew -q "+ moduleName +":printClasspath")​​​​​​​`
and runs the thread and classloader. After that we can use that class instead of the current EdcExtension to get very much the behavior as of EdcExtension, but for >= 1 instance.

jimmarino:
> Actually all that is needed is to add once to the **root** build.gradle.kts these 5 lines.
> 
> ```kotlin
> // build.gradle.kts (under allprojects)
>      tasks.register("printClasspath") {​​​​​​​
>      doLast {​​​​​​​
>         println("${​​​​​​​sourceSets["main"].runtimeClasspath.asPath}​​​​​​​");
>         }​​​​​​​
>     }​​​​​​​
> ```
> 
> Then we just build once a quite simple class (that could extend from EdcExtension to inherit all the mocking behavior) and calls `Runtime.getRuntime().exec(root +"/gradlew -q "+ moduleName +":printClasspath")​​​​​​​` and runs the thread and classloader. After that we can use that class instead of the current EdcExtension to get very much the behavior as of EdcExtension, but for >= 1 instance.

Ah, OK it looked like it required more. This is great, looking forward to the harness!

Issue #634, opened 2022-02-07
ContractDefinitionStore: support paging


paullatzelsperger:
#593 requires paging support, so the `ContractDefinitionStore` should add a method:
```java
 Stream<ContractDefinition> findAll(PagingSpec spec);
```

(the existing parameterless `findAll` can remain)

[edit] changed return type to `Stream`

Issue #633, opened 2022-02-07
ContractNegotiationStore: support paging 


paullatzelsperger:
#593 requires paging, so the ContractNegotiationStore should add a method 
```java
Collection<ContractNegotiation> findAll(PagingSpec pagingSpec);
```

Issue #632, opened 2022-02-07
AssetIndex: add paging support


paullatzelsperger:
#593 requires paging, so the following method needs to be added to the `AssetIndex`: 
```java
Collection<Asset> findAll(PagingSpec spec);
```

PR #638, opened 2022-02-07
Adds `PagingSpec` class

### LINKED ISSUE #635: Create PagingSpec object
For many api-related "get-all" calls we need pagination. Therefore a `PagingSpec` object should be created with the following properties:
- `offset` defaults to 0. cannot be negative.
- `limit` defaults to 50. cannot be negative. 
- `filter` contains a filter expression, defaults to null
- `sort` sort order, either `ASC` or `DESC` defaults to `DESC`
- `sortBy` field to sort by. name of the field. overrides natural sorting (e.g. by ID)

paullatzelsperger:
Simply adds the `PagingSpec` object, which will be used in #632, #634, #633

Closes #635


spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/PagingSpec.java
```diff
@@ -0,0 +1,111 @@
+package org.eclipse.dataspaceconnector.spi.query;
+
+import java.util.List;
+import java.util.regex.Pattern;
+
+public class PagingSpec {
```
jimmarino, line 6: Could you change the name to QuerySpec since it seems to be about querying?

jimmarino (approved):
If you make the naming change, LGTM

PR #641, opened 2022-02-08
Adds API for `ContractDefinition` objects

### LINKED ISSUE #588: DataManagementApi: Create API for ContractDefinition
_This issue only deals with the API, i.e. a stub for the `*Controller` class. An implementation is explicitly NOT to be done._

## API Endpoints
  - `POST /api/v1/contractdefinitions`: creates a new ContractDefinition, accepts `ContractDefinitionDto` in body, returns `void`
  - `GET /api/v1/contractdefinitions`: gets all `ContractDefinition`, accepts paging spec, returns `List<ContractDefinitionDto>`
  - `GET /api/v1/contractdefinitions/{id}`: gets one particular contract-def as `ContractDefinitionDto`
  - `DELETE /api/v1/contractdefinitions/{id}`: delete a ContractDefinition. Throws `IllegalStateException` if not possible.

## DTOs
```java
public class ContractDefinitionDto {
  private String accessPolicyId; 
  private String contractPolicyId;
  private List<Criterion> criteria;
  private String id;
  //setters, getters, etc.
}
```

paullatzelsperger:
Closes #588


extensions/api/data-management/contractdefinition/build.gradle.kts
```diff
@@ -0,0 +1,45 @@
...
+dependencies {
+    api(project(":spi"))
+    implementation(project(":common:util"))
+    implementation(project(":extensions:api:api-core"))
+    implementation(project(":core"))
+
+    implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
+    testImplementation(testFixtures(project(":launchers:junit")))
+    testImplementation(project(":extensions:in-memory:contractdefinition-store-memory"))
+    testImplementation(project(":extensions:http:jetty"))
```
ndr-brt, line 34: The simpler `:extensions:http` could be used instead

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/ContractDefinitionApiExtension.java
```diff
@@ -0,0 +1,37 @@
...
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtensionContext;
+
+public class ContractDefinitionApiExtension implements ServiceExtension {
+    @Inject
+    private WebService webService;
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        //avoid jetty throwing exceptions down the road
+        if (!context.getConfig().getEntries().containsKey("web.http.data.port")) {
```
ndr-brt, line 29: an `hasKey` method could be added on the `Config` object

extensions/api/data-management/contractdefinition/src/test/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/ContractDefinitionsApiControllerIntegrationTest.java
```diff
@@ -0,0 +1,102 @@
...
+import static org.mockito.Mockito.mock;
+
+public class ContractDefinitionsApiControllerIntegrationTest {
+
+    private static int port;
+    private OkHttpClient client;
+
+    @BeforeAll
+    static void prepareWebserver() {
+        port = 1024 + new Random().nextInt(10000);
```
ndr-brt, line 45: I've seen different methods to create port number, a test util method that does that shared by all the modules could be introduced.

ndr-brt (approved):
Ok for me, just some minor comments

ronjaquensel (approved):

extensions/api/data-management/contractdefinition/src/test/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/ContractDefinitionsApiControllerIntegrationTest.java
paullatzelsperger: good idea. Ideally, that method would also make sure that only unoccupied ports are returned.

paullatzelsperger: I added a method `getFreePort(int)` to the `TestUtils` class in `:common:util`

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/model/ContractDefinitionDto.java
```diff
@@ -0,0 +1,104 @@
...
+
+        public Builder criteria(List<Criterion> criteria) {
+            dto.criteria = criteria;
+            return this;
+        }
+
+        public Builder id(String id) {
+            dto.id = id;
+            return this;
+        }
```
florianrusch-cluetec, line 98: Is there a reason why you set the attributes directly and not use the setter methods?

paullatzelsperger: we would have to have (private) setter methods only for that reason, which makes little sense.
There shouldn't be public setters (unless there are good and specific reasons), since we perform object creation with Builders.

paullatzelsperger: ... and I just saw that there are in fact setter :)
IntelliJ must have shown preemptive obedience. fixed.

florianrusch-cluetec: Okaay. But you have public setters (e.g., lines 38-40). Or are you and I talking about different things?
What I meant was this:

```java
        public Builder accessPolicyId(String accessPolicyId) {
            dto.setAccessPolicyId (accessPolicyId);
            return this;
        }
```

paullatzelsperger: seems like a comment race condition. check my comment above

paullatzelsperger: anyway, thanks for noticing!

florianrusch-cluetec: All right, then I'm happy and quiet ;-) Thanks for the explanation!

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/model/ContractDefinitionDto.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.List;
+
+@JsonDeserialize(builder = ContractDefinitionDto.Builder.class)
+public class ContractDefinitionDto {
+    private String accessPolicyId;
+    private String contractPolicyId;
+    private List<Criterion> criteria;
+    private String id;
+
+    private ContractDefinitionDto() {
```
florianrusch-cluetec, line 31: Another question: wouldn't it make sense to have a constructer which takes the "normale" object as a parameter and then creates a new DTO object out of it? It's then lot easier to transform a normal object into an DTO.

I your case:
```java
public ContractDefinitionDto(ContractDefinition contractDefinition) {
    accessPolicyId = contractDefinition.getAccessPolicyId();
    ...
} 
```

paullatzelsperger: No, DTOs should not accept the domain object in constructors, because then the DTO has a runtime dependency onto `:spi`. The way we will tackle this is by using transformers, that convert DTOs into domain objects and vice versa.

florianrusch-cluetec: Ahh okay and we would like to avoid this in reason of? (Sorry for the noob question)

paullatzelsperger: Couple of reasons: the conversion might need additional validation (especially from DTO->BO), there might be additional dependencies that the transformer requires, e.g. when the DTO is actually a composition of several BOs, transformers may be injectable, etc. 
pick your favourite :)

florianrusch-cluetec: The question was misleading - sorry! Transformers makes absolute sense for me. The question was regarding the dependency onto `:spi`

paullatzelsperger: it increases decoupling, also DTOs don't need to know about "their" BO.

florianrusch-cluetec: Ahhh okay, thought there would be a bigger reason for that which I don't know. Thanks for the explanation! :)

paullatzelsperger: I'd say clean architecture is big enough of a reason :)

extensions/api/data-management/contractdefinition/src/test/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/ContractDefinitionApiControllerTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+
+    @ParameterizedTest
+    @ValueSource(strings = { "id > id1", "id < id1", "id like id1", "id inside id1" })
+    void getAll_paging_filterWithInvalidOperator(String filter) {
+        //todo: implement
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = { "id>id1", "id<id1" })
```
florianrusch-cluetec, line 50: Is there a particular reason why you moved this out to a separate test? They are the same operators as in the above test or not?

extensions/api/data-management/build.gradle.kts
```diff
@@ -0,0 +1,37 @@
...
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Daimler TSS GmbH - Initial API and Implementation
+ *
+ */
```
florianrusch-cluetec, line 13: Wrong copyright?

extensions/api/data-management/contractdefinition/src/test/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/ContractDefinitionApiControllerTest.java
paullatzelsperger: it's a subtle difference, but in above variant there are spaces, making it valid, whereas this is invalid. If you look at the `QuerySpec`, it'll make sense.

florianrusch-cluetec: Understood. So for example instead of `"id>id1"` there could also be something like `idlikeid1` 😄  You just choose `<`, `>` because those are "more" real operators. It's more an example/test that there MUST be spaces inside the filter. But then why are there some with spaces and some without in line 37 - shouldn't it be consistent across "all" operators (I know there is currently only the `=` as an operator)?

Other than that, it's fine for me, and you can leave it as is.

paullatzelsperger: The equality operator is a bit special, because it is parsed explicitly and there is an option to translate it to a "contains" operator automatically, assuming that that's probably far more common than an accurate `equals`. 

However, the more I think about this, the less merit I see in this. I may change this in another PR.

PR #642, opened 2022-02-08
Feature/616 fetch entities async

### LINKED ISSUE #616: [state machines review] every state change after an async operation should fetch the entity again

### The state is updated without fetching the entity again after an asynchronous operation
**Problem**
Sometimes, after an async call result evaluation (inside the `whenComplete` block), the state is updated and the entity saved to the store, but using the entity that was fetched from `nextForState`, this could lead to problems if in the meantime another async operation happens (e.g. a message is received from a peer)
**Solution**
After an async operation the entity should be fetched again before being updated and stored.
</div>

ndr-brt:
Fix #616 

### Context
At the moment, there's a not safe behaviour in `Transfer` and `ContractNegotiation` managers, to update and save, after the completion of an async operation, the entity that was fetched before the async operation. This could lead to states overrides and unexpected errors.

### What brings
The entity will be fetched again on the completion of every async operation.

### DIFF
--- a/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
+++ b/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
@@ -114,8 +114,7 @@ public class TransferProcessManagerImpl implements TransferProcessManager {
     private Monitor monitor;
     private Telemetry telemetry;
 
-    private TransferProcessManagerImpl() {
-    }
+    private TransferProcessManagerImpl() { }
 
     public void start(TransferProcessStore processStore) {
         transferProcessStore = processStore;
@@ -237,7 +236,7 @@ public class TransferProcessManagerImpl implements TransferProcessManager {
         observable.invokeForEach(l -> l.deprovisioned(transferProcess));
     }
 
-    private void transitionRequestAck(String processId) {
+    private void transitionToRequestAck(String processId) {
         TransferProcess transferProcess = transferProcessStore.find(processId);
         if (transferProcess == null) {
             monitor.severe("TransferProcessManager: no TransferProcess found for acked request");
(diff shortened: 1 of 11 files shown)


ronjaquensel (approved):

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
```diff
@@ -477,19 +468,26 @@ private boolean processInitial(TransferProcess process) {
...
+                        onProvisionComplete(process.getId(), responses);
+                    } else {
+                        transitionToError(process.getId(), throwable.getCause(), "Error during provisioning");
+                    }
+                });
 
         return true;
     }
 
+    private void transitionToError(String id, Throwable throwable, String message) {
```
paullatzelsperger, line 483: Is reloading really absolutely necessary for `transitionError`? A transition to `ERROR` should be possible from every other state, so it shouldn't matter. I'd be ok with leaving it in for having a consistent programming model though.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
```diff
@@ -523,29 +515,38 @@ private void processProviderRequest(TransferProcess process, DataRequest dataReq
...
                 .whenComplete((o, throwable) -> {
                     if (o != null) {
                         monitor.info("Object received: " + o);
-                        if (dataRequest.getTransferType().isFinite()) {
-                            process.transitionInProgress();
-                        } else {
-                            process.transitionStreaming();
-                        }
-                        transferProcessStore.update(process);
+                        transitionToInProgress(transferProcessStore.find(process.getId()));
```
paullatzelsperger, line 535: Any particular reason why you're performing the `transferProcessStore.find(...)` directly here as opposed to: inside the `transitionToInProgress` method?

paullatzelsperger (commented):
looks good

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
ndr-brt: Yeah, I'm not that happy with this solution, I did like this to have the opportunity to use the `transitionToInProgress` also from `processProviderRequest` and `processAckRequested`. Added a null check though.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
ndr-brt: Probably not necessary, just for consistency. Also added a null check, probably not necessary but just for consistency too.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
ndr-brt: I think the if condition could be moved in a method inside `TransferProcess`. Doing this the method `transitionToInProgress` would be superfluous.

ndr-brt: @paullatzelsperger I added a method on `TransferProcess` that handles the choice between "in progress" and "streaming". What do you think?

ndr-brt:
@paullatzelsperger @jimmarino can this be merged?

paullatzelsperger (approved):

PR #619, opened 2022-02-03
Feature/614 dont consider unprocessed entities

ndr-brt:
Fix #614

PR #652, opened 2022-02-09
Add command queue for negotiation managers

### LINKED ISSUE #617: [state machines review] introduce command queue on contract negotiation manager
Like it was done on Transfer Process Manager, to avoid "out of the loop entity updates"

### There are actions that don't follow the loop order
**Problem**
There are interface methods that permit the modification of the entities outside the state machine/command queue loop (e.g. `confirmed` in `ConsumerContractNegotiationManager`). This could lead to errors in the case where that entity is already leased by another thread.
**Solution**
Every operation on the entities should be executed by the loop. Every of those method should create and enqueue a command that will be executed in the loop.

I'd like to gather some impressions by everyone to eventually open some issues.</div>

ronjaquensel:
Closes #617 

Waiting for #619 to be merged first.


core/base/src/main/java/org/eclipse/dataspaceconnector/core/base/BoundedCommandQueue.java
```diff
@@ -22,49 +23,45 @@
...
- * In-memory implementation of the {@link CommandQueue} that is backed by a bounded queue. This means that attempting
- * to add elements to an already full queue will fail and raise an {@link IllegalStateException}.
- * <p>
- * This queue is threadsafe, so every operation on it can be assumed atomic.
- */
-public class BoundedCommandQueue implements CommandQueue {
-    private final BlockingQueue<Command> queue;
-
-    public BoundedCommandQueue(int bound) {
+public abstract class BoundedCommandQueue<C extends Command> implements CommandQueue<C> {
```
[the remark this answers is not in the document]
ronjaquensel: Good point. I'll refactor this for the BoundedCommandQueue as well as the CommandRunner.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java
```diff
@@ -484,6 +520,16 @@ public Builder dispatcherRegistry(RemoteMessageDispatcherRegistry dispatcherRegi
...
             return this;
         }
+        
+        public Builder commandQueue(ContractNegotiationCommandQueue commandQueue) {
+            manager.commandQueue = commandQueue;
+            return this;
+        }
+        
+        public Builder commandRunner(ContractNegotiationCommandRunner commandRunner) {
+            manager.commandRunner = commandRunner;
```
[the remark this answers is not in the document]
ronjaquensel: I don't think it can. The CommandRunner requires a CommandHandlerRegistry in its constructor. Since the managers do not have a reference to a CommandHandlerRegistry, we don't have one available in the builder. And even if we did, we could not be sure that the registry would have been set before the CommandQueue.

core/base/src/main/java/org/eclipse/dataspaceconnector/core/base/BoundedCommandQueue.java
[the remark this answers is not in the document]
ronjaquensel: Done :)

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java
```diff
@@ -275,6 +288,29 @@ private void run() {
             }
         }
     }
+    
+    private int processCommandQueue() {
```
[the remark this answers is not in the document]
ronjaquensel: Done :)

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java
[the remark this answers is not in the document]
ndr-brt: oh, you're right, my oversight

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
```diff
@@ -85,7 +85,7 @@
  * If no processes need to be transitioned, the transfer manager will wait according to the the defined {@link TransferWaitStrategy} before conducting the next iteration.
  * A wait strategy may implement a backoff scheme.
  */
-public class TransferProcessManagerImpl implements TransferProcessManager {
+public class TransferProcessManagerImpl extends CommandQueueProcessor<TransferProcessCommand> implements TransferProcessManager {
```
ndr-brt, line 88: Personally I would prefer composition over abstraction especially on specific features like these. 
An internal field instantiated on declaration would do the job nicely, and `CommandQueueProcessor` could be not abstract.

paullatzelsperger:
Hey @ronjaquensel admitting that I haven't looked at the code yet, so maybe you already did, but still:
there is an excellent [PR](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/619) from @ndr-brt about how processed/unprocessed entities are counted. In my mind it makes absolute sense to adopt this into the `*NegotiationManagers`? Maybe another PR?

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
ronjaquensel: Hi, I refactored it. Could you check and see if this is that what you had in mind?

ronjaquensel:
Hi @paullatzelsperger, thanks for the hint. I took a look at the PR and it looks like it is already adopted for the existing methods in the *NegotiationManagers, so only the processing of the CommandQueue would have to be refactored. Probably would make sense to do that in one of the two PRs. We could leave this one open until the other one is merged, then I can include the changes here.

core/base/src/main/java/org/eclipse/dataspaceconnector/core/base/BoundedCommandQueue.java
```diff
@@ -22,49 +23,45 @@
...
+    public void enqueue(C element) {
         //add will throw an IllegalStateException if the queue exceeds its capacity
         queue.add(element);
     }
-
+    
     @Nullable
     @Override
     public Command dequeue() {
         return queue.poll();
```
denisneuling, line 43: I'd recommend using [BlockingQueue#poll(time, unit)](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html#poll(long,%20java.util.concurrent.TimeUnit)) which gives us blocking functionality out of the box. Duration and TimeUnit could be set via the BoundedCommandQueue constructor

ndr-brt: Actually I think this method can be removed since it's not used by anyone, corrently the `dequeue(amount)` is used

ndr-brt (approved):

paullatzelsperger:
I'll take a quick look

ndr-brt (approved):
looks good

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/command/CommandHandler.java
```diff
@@ -7,7 +7,6 @@
  * @param <T> The concrete type of {@link Command}
  */
 public interface CommandHandler<T extends Command> {
```
paullatzelsperger, line 9: I know it was me who forgot in the first place, but could you copy the copyright from `CommandHandlerRegistry` over :)?

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/command/CommandRunner.java
```diff
@@ -11,18 +25,18 @@
  *     <li>convert exception into a {@link Result}, increase error count</li>
  * </ol>
  */
-public class CommandRunner {
-    private final CommandHandlerRegistry commandHandlerRegistry;
-    private final Monitor monitor;
+public class CommandRunner<C extends Command> {
+    protected final CommandHandlerRegistry<C> commandHandlerRegistry;
```
paullatzelsperger, line 29: Why does this need to be `protected`?

spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/TransferProcessManager.java
```diff
@@ -39,6 +39,6 @@
     /**
      * Adds a single Command to the queue which is processed sequentially.
      */
-    void enqueueCommand(Command command);
+    void enqueueCommand(TransferProcessCommand command);
```
paullatzelsperger, line 42: Not sure this restriction is a good idea- what happens if someone just wants to add a command that operates on several TransferProcesses, like "count all in State XYZ"?

spi/contract-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/contract/negotiation/command/ContractNegotiationCommand.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ */
+package org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.command;
+
+import org.eclipse.dataspaceconnector.spi.command.Command;
+
+/**
+ * Sub-type of {@link Command} for modifying ContractNegotiations.
+ */
+public abstract class ContractNegotiationCommand extends Command {
```
paullatzelsperger, line 21: Maybe I'm missing something, but why do we need the `ContractNegotiationCommand` and the `SingleNegotiationCommand`? Analogously, the `TransferProcessCommand`...

paullatzelsperger (changes requested):
comments in code

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/command/CommandHandler.java
ronjaquensel: Done

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/command/CommandRunner.java
ronjaquensel: It doesn't. CommandRunner was an abstract class in between, missed it when I refactored it back. Changed it now.

spi/contract-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/contract/negotiation/command/ContractNegotiationCommand.java
ronjaquensel: I created the `ContractNegotiationCommand` as a super-class for all commands targeting ContractNegotiations. Same for the `TransferProcessCommand` and TransferProcesses. The idea was that these two classes can be used for typing, so e.g. the `TransferProcessManager` has a `CommandQueue<TransferProcessCommand>`, to avoid that a command modifying a ContractNegotiation gets executed within the transfer state machine. The `Single*Commands` always target a specific TransferProcess/ContractNegotiation, whereas another class extending the TransferProcessCommand could also modify multiple TransferProcesses. Does that make sense?

spi/contract-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/contract/negotiation/command/ContractNegotiationCommandHandlerRegistry.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+package org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.command;
+
+import org.eclipse.dataspaceconnector.spi.command.CommandHandlerRegistry;
+
+/**
+ * Sub-type of the {@link CommandHandlerRegistry} for registering {@link ContractNegotiationCommand}
+ * handlers.
+ */
+public interface ContractNegotiationCommandHandlerRegistry extends CommandHandlerRegistry<ContractNegotiationCommand> {
```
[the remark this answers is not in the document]
ronjaquensel: In general, I think the `CommandHandlerRegistry<>` would suffice. Only problem I see is that the `CommandHandlerRegistry<TransferProcessCommand>` and `CommandHandlerRegistry<ContractNegotiationCommand>` are provided by different extensions. Is there a way to use an interface with generics in the `@Provides` annotiation?

PR #654, opened 2022-02-09
Avoid duplication between runtimes

### LINKED ISSUE #374: ServiceExtension shutdown duplication
I noticed that the process of shutting down ServiceExtensions is duplicated in 4 different classes:
`BaseRuntime` (the right place)
`RegistrationServiceRuntime` (that does the same things at the BaseRuntime, so it could be substituted)
`EdcConnectorConsumerRuntime` (pretty similar to BaseRuntime too, probably it can be substituted without regressions)
`EdcExtension` (I think this could be a BaseRuntime extension)

ndr-brt:
Fix #374


jimmarino (approved):
LGTM

paullatzelsperger (approved):

PR #666, opened 2022-02-10
fix: resolve minor issues in contract negotiation process

juliapampus:


core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/common/ContractValidationException.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ *  Contributors:
+ *       Fraunhofer Institute for Software and Systems Engineering - initial API and implementation
+ *
+ */
+package org.eclipse.dataspaceconnector.contract.common;
+
+import org.eclipse.dataspaceconnector.spi.EdcException;
+
+public class ContractValidationException extends EdcException {
```
jimmarino, line 18: This class seems very specific to contract validation as opposed to being a common exception type. Instead of moving it to that package, I think we may be able to get rid of it altogether (additional comments to follow).

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/validation/ContractValidationServiceImpl.java
```diff
@@ -52,29 +54,22 @@ public ContractValidationServiceImpl(ParticipantAgentService agentService, Suppl
     @Override
     @NotNull
     public Result<ContractOffer> validate(ClaimToken token, ContractOffer offer) {
```
jimmarino, line 56: I wonder if the use of ContractValidationException can be simplified? Most of its uses immediately catch and unwrap the message to pass back as a result:

1. IMO the original lines 55-64 read easier and eliminate the need to wrap the error message, raise it as an exception, and immediately catch it. Related to this, getContractDefinitionByTokens and validateContractId are confusing to me because they actually create the agent reference from a parsed id and parse (not validate) the id.  If the original code is restored, there is no need to throw the exception since a result can be returned directly.

2. The introduction of the validateMandatoryAttributes method seems to be the main change. Shouldn't that be one of the first things done when the method is entered? If there are only four items to validate, then either it could be inlined with a failure result returned directly or extracted to a method with a validation return type similar to other areas of the codebase.

Other than those two specific points, looks good to me.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -248,6 +242,15 @@ public NegotiationResult declined(ClaimToken token, String negotiationId) {
         return NegotiationResult.success(negotiation);
     }
 
+    private ContractNegotiation findContractNegotiationById(String negotiationId) {
```
ndr-brt, line 245: Should this method used also on `declined` method on provider side?

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/common/ContractValidationException.java
juliapampus: I will remove it.

ndr-brt (approved):
Looks good, added a question and I agree with jim on the use of validation exceptions

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/validation/ContractValidationServiceImpl.java
juliapampus: > If the original code is restored, there is no need to throw the exception since a result can be returned directly.

Maybe to explain the motivation behind my changes: I wanted to encapsulate these calls as they later need to be used in every method and this would cause a lot of code duplication.

> Shouldn't that be one of the first things done when the method is entered?

It is?

> If there are only four items to validate, then either it could be inlined with a failure result returned directly or extracted to a method with a validation return type similar to other areas of the codebase.

Again, this may be extended by some values and needs to be called in every method. Therefore, I put it in a method.

I see your points, I will make some changes and re-request your review. Thanks!

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
juliapampus: Yes, you are right, I will add it. Thanks!

ndr-brt (approved):

jimmarino (approved):

ndr-brt (approved):

juliapampus:
I just rebased the branch locally, will merge it afterwards.

PR #671, opened 2022-02-11
Adds a CANCELLED state to the TransferProcess

### LINKED ISSUE #515: TransferProcessManager: introduce new state CANCELLED
On order to accomodate cancelled TPs there needs to be another state CANCELLED. 
All states **except these** are legal transition start states:
- COMPLETED
- ERROR
- ENDED

No transitions can have CANCELLED as start state, i.e. resuming is not supported.

paullatzelsperger:
Adds a new state CANCELLED to the transferprocess.

Closes #515

### DIFF
--- a/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/command/handlers/CancelTransferCommandHandler.java
+++ b/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/command/handlers/CancelTransferCommandHandler.java
@@ -15,6 +15,7 @@ package org.eclipse.dataspaceconnector.transfer.core.command.handlers;
 
 import org.eclipse.dataspaceconnector.spi.transfer.store.TransferProcessStore;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcess;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcessStates;
 import org.eclipse.dataspaceconnector.transfer.core.command.commands.CancelTransferCommand;
 
 public class CancelTransferCommandHandler extends TransferProcessCommandHandler<CancelTransferCommand> {
@@ -30,7 +31,14 @@ public class CancelTransferCommandHandler extends TransferProcessCommandHandler<
 
     @Override
     protected boolean modify(TransferProcess process) {
-        process.transitionError("Cancelled");
+        var state = process.getState();
+        if (state == TransferProcessStates.COMPLETED.code() ||
+                state == TransferProcessStates.ERROR.code() ||
+                state == TransferProcessStates.CANCELLED.code() ||
+                state == TransferProcessStates.ENDED.code()) {
+            return false;
+        }
+        process.transitionCancelled();
         return true;
     }
 }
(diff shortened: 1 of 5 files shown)


core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/command/handlers/CancelTransferCommandHandler.java
```diff
@@ -30,7 +31,14 @@ public CancelTransferCommandHandler(TransferProcessStore store) {
 
     @Override
     protected boolean modify(TransferProcess process) {
-        process.transitionError("Cancelled");
+        var state = process.getState();
+        if (state == TransferProcessStates.COMPLETED.code() ||
```
ndr-brt, line 35: a `canBeCancelled` or `isInFinalState` method on `TransferProcess` could make this more readable and better encapsulate the knowledge.

core/transfer/src/test/java/org/eclipse/dataspaceconnector/transfer/core/command/handlers/CancelTransferCommandHandlerTest.java
```diff
@@ -47,10 +51,31 @@ void handle() {
...
+
+        verify(storeMock).find(anyString());
+        verify(storeMock).update(tp);
+        verifyNoMoreInteractions(storeMock);
+
     }
 
+    @ParameterizedTest
+    //COMPLETED, ENDED, ERROR, CANCELLED -> values must be compile-time constant
+    @ValueSource(ints = { 800, 1100, -1, 1200 })
```
ndr-brt, line 65: this would be more readable:
```
@EnumSource(value = TransferProcessStates.class, names = {"COMPLETED", "ENDED", "ERROR", "CANCELLED"})
```

spi/transfer-spi/src/test/java/org/eclipse/dataspaceconnector/spi/types/domain/transfer/TransferProcessTest.java
```diff
@@ -142,6 +146,26 @@ void verifyProvisioningComplete() {
         assertTrue(process.provisioningComplete());
     }
 
+    @ParameterizedTest
+    @ValueSource(ints = { 0, 100, 200, 300, 400, 500, 600, 700, 850, 900, 1000, 1200 })
```
ndr-brt, line 150: Same as before, the `@EnumSource` could do a good job here, in this case the `mode = EXPLICIT` could be used to make it more succint

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/command/handlers/CancelTransferCommandHandler.java
paullatzelsperger: I'm hesitant to add that method for just that one usage. If we ever need to check the same condition again, I'm all for encapsulating it though.

ndr-brt: As a rule of thumb, the less the states are valued outside the `TransferProcess`, the better. This is a way to stay away from an anemic model.
But ok, I agree on waiting for duplication.

ndr-brt (approved):

jimmarino (commented):
Can you update the doc on `TransferProcess` to add the CANCELLED state, including valid transitions?

PR #675, opened 2022-02-11
extensions: implement S3BucketReader

ndr-brt:
## Context
In the process of deploying an EDC to transfer data from an S3 bucket to another we noticed that the data cannot be read because the `S3BucketReader` is a dummy implementation.

## What brings
An implementation of `S3BucketReader`. a `S3ClientProvider` was introduced upfront to permit testability.

### DIFF
--- a/.github/workflows/integrationtests.yaml
+++ b/.github/workflows/integrationtests.yaml
@@ -96,6 +96,13 @@ jobs:
           S3_SECRET_ACCESS_KEY: password
         run: ./gradlew extensions:aws:s3:s3-provision:check
 
+      - name: AWS S3 StatusChecker Test
+        env:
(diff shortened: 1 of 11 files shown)


bscholtes1A:
Hi @ndr-brt 
sorry for the late review. 
This `DataOperator` will soon go to trash and be replaced by the DPF. 
Thus instead of what you ve done I would suggest to go directly for an implementation of the `DataSource` interface ;)

ndr-brt:
@bscholtes1A It seems a bit too much for this PR, I suggest to merge this if it's ok and then open another issue to provide an s3 implementation for the `DataSource`/`DataSink`. Will open it right now

bscholtes1A (approved):

PR #688, opened 2022-02-14
Distributed tracing with OpenTelemetry

### LINKED ISSUE #355: Support for correlating trace headers between incoming and outgoing requests
To monitor the E2E transaction flow in our scenario we’d like to build the distributed traces across Client, Consumer, Provider and API backend.

Currently the EDC framework doesn’t support forwarding the [_traceparent_ header](https://www.w3.org/TR/trace-context/#traceparent-header):

- between http request incoming to Consumer API endpoint and the IDS message sent to the provider.
- between http request sent to the provider from the consumer and the final data transfer operation or backend API request.

Examples of an expected behavior:

- When I send a http request to consumer API with _traceparent_ http header I expect the outgoing request to the provider to also contain _traceparent_ header with the same _trace id_ attribute.
- If the http request sent from the consumer to the provider contains _traceparent_ http header I expect the final data transfer request to also contain _traceparent_ header with the same _trace id_ attribute.

marcgs:
closes #355 
authors: @marcgs @ouphi

Introduced [OpenTelemetry](https://opentelemetry.io/) for tracing. 
The main focus of this PR is to make sure that the contract negotiation and the data transfer flows can be traced end to end.

OpenTelemetry stores the traceContext in the current thread. Whenever new threads are started, this context information is lost and a separate trace context is used by the new thread. A [mechanism](https://opentelemetry.io/docs/instrumentation/java/manual/#context-propagation) exists to propagate the context to overcome this limitation.

- Added a Telemetry class to easily get the current tracing context and propagate it to the current thread.
- The TransferProcess and ContractNegotiation entities are used as trace context carriers. Asynchronous workers propagate the trace context from the carriers to the current thread.
- Provided a sample explaining how to use the open telemetry java agent to enable tracing and visualising the traces with Jaeger.

See below screenshots of the traces that we can observe in Jaeger. These traces will be extended with further [custom spans](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/manual-instrumentation.md#creating-spans-around-methods-with-withspan) in follow-up PRs.

## Contract Negotiation Trace
contract_negotiation

## Transfer Process Trace
transfer_process


core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
```diff
@@ -640,9 +653,9 @@ public TransferProcessManagerImpl build() {
             Objects.requireNonNull(manager.dataProxyManager, "DataProxyManager cannot be null!");
             Objects.requireNonNull(manager.proxyEntryHandlers, "ProxyEntryHandlerRegistry cannot be null!");
             Objects.requireNonNull(manager.observable, "Observable cannot be null");
-            
+
```
paullatzelsperger, line 656: Since the `telemetry` field is not nullable, there should be a check here

extensions/in-memory/negotiation-store-memory/build.gradle.kts
```diff
@@ -12,15 +12,15 @@
...
 plugins {
     `java-library`
 }
 
-
 dependencies {
     api(project(":spi"))
-
-
+    api("io.opentelemetry:opentelemetry-extension-annotations:${openTelemetryVersion}")
```
paullatzelsperger, line 23: this is not needed.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/telemetry/Telemetry.java
```diff
@@ -0,0 +1,84 @@
...
+     * var prevTraceContext = telemetry.getCurrentTraceContext();
+     * try (Scope scope = setCurrentTraceContext(traceCarrier)) {
+     *   assert telemetry.getCurrentTraceContext() == traceCarrier.getTraceContext()
+     * }
+     * assert telemetry.getCurrentTraceContext() == prevTraceContext;
+     * }</pre>
+     *
+     * @param carrier The trace context carrier
+     */
+    public Scope setCurrentTraceContext(TraceCarrier carrier) {
```
paullatzelsperger, line 52: A setter (or setter-like) method with a return value seems a bit odd. In addition, the return value isn't used anyway, so we could just make this a normal setter.

Alternatively, seeing that the only usage of this method makes use of the Scope being an `AutoCloseable`, I'd suggest renaming this method and returning an `AutoCloseable`

spi/core-spi/build.gradle.kts
```diff
@@ -25,6 +25,8 @@ dependencies {
     api("com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}")
     api("com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}")
     api("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jacksonVersion}")
+    api("io.opentelemetry:opentelemetry-api:${openTelemetryVersion}")
```
paullatzelsperger, line 28: this can be converted to an `implementation` dependency, if the `Telemetry` class provides a parameterless constructor, which could internally initialize with `noop()`:
```java
public Telemetry() {
    this.openTelemetry = OpenTelemetry.noop();
}
```

ndr-brt:
I have one question before diving into a deep review:
shall the telemetry become a mandatory core feature?
Why it has not been implemented as a `ServiceExtension`?

marcgs:
> I have one question before diving into a deep review: shall the telemetry become a mandatory core feature? Why it has not been implemented as a `ServiceExtension`?

@ndr-brt Observability (and in this case tracing) is a cross-cutting concern that needs to be supported in the EDC core.  Most of the work in this PR is actually adapting EDC core to propagate the standard [trace headers](https://www.w3.org/TR/trace-context/) so that requests can be traced properly across consumer-provider interactions like the ones displayed in the sample screenshots provided in this PR. Hope this answers your question.

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/ExtensionLoader.java
```diff
@@ -89,6 +92,19 @@ public static void loadVault(ServiceExtensionContext context) {
         return loadMonitor(loader.stream().map(ServiceLoader.Provider::get).collect(Collectors.toList()));
     }
 
+    public static @NotNull Telemetry loadTelemetry() {
+        var loader = ServiceLoader.load(OpenTelemetry.class);
+        var openTelemetries = loader.stream().map(ServiceLoader.Provider::get).collect(Collectors.toList());
+        return new Telemetry(loadOpenTelemetry(openTelemetries));
+    }
+
+    protected static @NotNull OpenTelemetry loadOpenTelemetry(List<OpenTelemetry> openTelemetries) {
```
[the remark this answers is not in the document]
marcgs: This would required mocking the static method, correct? I see this is only done in one other place in the codebase with `mockito-inline`, correct?

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -571,9 +580,9 @@ public ConsumerContractNegotiationManagerImpl build() {
             Objects.requireNonNull(manager.commandQueue, "commandQueue");
             Objects.requireNonNull(manager.commandRunner, "commandRunner");
             Objects.requireNonNull(manager.observable, "observable");
-    
+
```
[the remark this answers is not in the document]
marcgs: There is actually a default implementation set in the Builder:
```
private Builder() {
            manager = new ConsumerContractNegotiationManagerImpl();
            manager.telemetry = new Telemetry(OpenTelemetry.noop()); // default noop implementation
        }
```

But since users can still override it with null, I will add the check too. Thanks!

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
marcgs: Done, thanks!

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java
```diff
@@ -532,9 +541,9 @@ public ProviderContractNegotiationManagerImpl build() {
             Objects.requireNonNull(manager.commandQueue, "commandQueue");
             Objects.requireNonNull(manager.commandRunner, "commandRunner");
             Objects.requireNonNull(manager.observable, "observable");
```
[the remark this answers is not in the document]
marcgs: Done, thanks!

extensions/in-memory/negotiation-store-memory/build.gradle.kts
marcgs: You are right, thanks! This dependency will be needed if we introduce custom spans later on.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/telemetry/Telemetry.java
marcgs: Actually this function is does not need to be exposed anymore, as we use the `contextPropagationMiddleware` method instead. I have made it private and renamed it to `propagateTraceContext`.

spi/core-spi/build.gradle.kts
marcgs: Thanks for the hint. I needed to add a second "implementation" dependency in the core/boot module for it to work as the `ExtensionLoader` still needs access to the open telemetry API.

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/ExtensionLoader.java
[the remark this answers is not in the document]
paullatzelsperger: Yeah I guess it's fine. Can be resolved.

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/ExtensionLoader.java
```diff
@@ -89,6 +92,19 @@ public static void loadVault(ServiceExtensionContext context) {
...
+        if (openTelemetries.size() > 1) {
+            throw new IllegalStateException(String.format("Found %s OpenTelemetry implementations. Please provide only one OpenTelemetry service provider.", openTelemetries.size()));
```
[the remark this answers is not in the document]
paullatzelsperger: The difference is though, that there can be multiple Monitors, but there can only be one `Telemetry`. 
Essentially we'll end up with two options:
- mock the static `load` method of the ServiceLoader
- perform the check in the `loadTelemetry` method, and pass a single instance to `loadOpenTelemetry`

Both of which are not ideal. At the very least I would put a comment onto the `loadOpenTelemetry` method explaining that the List can only ever contain at most a single item.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/telemetry/Telemetry.java
paullatzelsperger: Ah I just saw - `Scope#close()` does not throw an exception, while `AutoCloseable#close()` throws an `Exception`. So it's better to return `Scope` instead of `AutoCloseable`.

paullatzelsperger (approved):

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/ExtensionLoader.java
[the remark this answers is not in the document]
marcgs: I tried static mocking but things get nasty quickly, with the need of mocking Streams of `ServiceLoader.Provider`. IMO the sole purpose of having the `loadOpenTelemetry` method is to be able to test the logic verifying that at most a custom implementation is provided, and if none the default `GlobalOpenTelemetry` is used. I think renaming the method to `selectOpenTelemetryImpl` might help understanding this purpose.

jimmarino:
@ndr-brt and @bscholtes1A can you guys take a look?

ndr-brt (approved):

PR #690, opened 2022-02-14
Feature/613 lock entities

### LINKED ISSUE #613: [state machine review] Introduce locks on entities
We need to introduce the possibility to lock an entity when it's waiting for a response from an async operation (e.g. provision for TransferProcess) so that it won't be fetched by the `nextForState` method.

This will fix this discussion point:

### The state is not moved to an "not to be processed" state before executing an async operation
**Problem**
This is the problem described here: #538 . When an entity is fetched from the `nextForState` and an async operation is called, it should be set in a state that's not fetched by the state machine loop, otherwise that can cause useless loops (as described in the issue)
**Solution**
Define 3 type of states:
* To be processed state: an entity with this state is fetched by the `nextForState`, processed and moved to a processing state (async operation), to another to be processed state (sync operation) or to a final state (nothing else should be done)
* Processing state: an entity with this state is waiting for an asynchronous response, it's not fetched by the `nextForState`.
* Final state: an entity with this state has reached the terminus and there should stay, e.g. states like `COMPLETED`, `ENDED`, `ERROR`, ...
</div>

ndr-brt:
Fix #613


core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java
```diff
@@ -333,21 +331,25 @@ private boolean processProviderOffering(ContractNegotiation negotiation) {
...
+
         //TODO protocol-independent response type?
         dispatcherRegistry.send(Object.class, contractOfferRequest, () -> null)
                 .whenComplete((response, throwable) -> {
+                    var contractNegotiation = negotiationStore.find(negotiation.getId());
                     if (throwable == null) {
-                        negotiation.transitionOffered();
-                        negotiationStore.save(negotiation);
-                        observable.invokeForEach(l -> l.providerOffered(negotiation));
+                        contractNegotiation.transitionOffered();
```
ronjaquensel, line 342: I think `contractNegotiation` should be checked for null here (like in the `TransferProcessManagerImpl`), otherwise might lead to a NullPointerException. Same for other methods in both *NegotiationManagers.

ndr-brt: Yes, you're right, I didn't do that because it was already done in #642, I didn't want to replicate too much logic here since I'll have enough conflicts to fix when it will be the time for rebase

ronjaquensel: Makes sense, I missed that it was done on the other branch.

ronjaquensel (approved):

ndr-brt:
With @paullatzelsperger agreed that the lock mechanism should be handled with the lease on cosmos db, so that an entity is leased when it's fetched with `nextForState` and the lease will be released by the `update` operation.
In this way, it will simplier and the behavior will be like a SQL "select for update".
I will remove the `locked` flag logic from here and the lease thing will be implemented by paul in another PR

PR #693, opened 2022-02-14
Added support for QuerySpec to ContractNegotiationStore

### LINKED ISSUE #633: ContractNegotiationStore: support paging

paullatzelsperger:
Adds a `queryNegotiations(QuerySpec querySpec)` method to the `ContractNegotiationStore, effectively adding support for paging, filtering and sorting.

Closes #633


extensions/in-memory/negotiation-store-memory/src/main/java/org/eclipse/dataspaceconnector/negotiation/store/memory/CriterionToPredicateConverter.java
```diff
@@ -0,0 +1,52 @@
...
+import org.eclipse.dataspaceconnector.spi.query.CriterionConverter;
+import org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiation;
+
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.function.Predicate;
+
+import static org.eclipse.dataspaceconnector.negotiation.store.memory.ContractNegotiationFunctions.property;
+
+class CriterionToPredicateConverter implements CriterionConverter<Predicate<ContractNegotiation>> {
```
jimmarino, line 27: Documentation, please

extensions/in-memory/negotiation-store-memory/src/main/java/org/eclipse/dataspaceconnector/negotiation/store/memory/ContractNegotiationFunctions.java
```diff
@@ -0,0 +1,42 @@
...
+class ContractNegotiationFunctions {
+
+    /**
+     * Utility function to get value of a field from a {@link ContractNegotiation}.
+     *
+     * @param contractNegotiation The {@link ContractNegotiation} object
+     * @param key                 The name of the field
+     * @return The field's value. Returns null if the field does not exist.
+     */
+    static <T> T property(ContractNegotiation contractNegotiation, String key) {
```
ndr-brt, line 31: this method could stay in `ContractNegotiation` class

extensions/in-memory/negotiation-store-memory/src/main/java/org/eclipse/dataspaceconnector/negotiation/store/memory/CriterionToPredicateConverter.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Objects;
+import java.util.function.Predicate;
+
+import static org.eclipse.dataspaceconnector.negotiation.store.memory.ContractNegotiationFunctions.property;
+
+/**
+ * Converts a {@link Criterion} into a {@link Predicate} for use with an {@link InMemoryContractNegotiationStore}.
+ * Currently, only the "=" and "in" operators are supported (cf. {@link Criterion#getOperator()}).
+ */
+class CriterionToPredicateConverter implements CriterionConverter<Predicate<ContractNegotiation>> {
```
ndr-brt, line 31: There's another `CriterionToPredicateConverter`, and these classes are pretty much the same. Eventually the `property` method that I commented above could be part of a `ReflectionUtils` or something like that and give the property for any kind of object it receives.

extensions/in-memory/negotiation-store-memory/src/main/java/org/eclipse/dataspaceconnector/negotiation/store/memory/InMemoryContractNegotiationStore.java
```diff
@@ -116,6 +123,54 @@ public void delete(String processId) {
         });
     }
 
+    @Override
+    public Stream<ContractNegotiation> queryNegotiations(QuerySpec querySpec) {
+        return readLock(() -> {
+            Stream<ContractNegotiation> negotiationStream = processesById.values().stream();
+            // filter
+            if (CollectionUtil.isNotEmpty(querySpec.getFilterExpression())) {
```
ndr-brt, line 131: This if condition is not needed if the `filterExpression` field is initiated by default with an empty list (that's a good practice overall) because the "empty list" case is already handled by the reduce's identity

paullatzelsperger: The `filterExpression` could still be `null`, specifically when `null` is passed in the Builder's `filter()` method. 
`CollectionUtil.isNotEmpty()` checks for null _and_ emptiness.

extensions/in-memory/negotiation-store-memory/src/main/java/org/eclipse/dataspaceconnector/negotiation/store/memory/CriterionToPredicateConverter.java
paullatzelsperger: I'll create a `ReflectionUtils` class, we'll probably need this in several other places as well.
Then, a base `CriterionToPredicateConverter` will become easy.

extensions/in-memory/negotiation-store-memory/src/main/java/org/eclipse/dataspaceconnector/negotiation/store/memory/ContractNegotiationFunctions.java
paullatzelsperger: moved to `ReflectionUtil` and I also made it more generic.

extensions/in-memory/negotiation-store-memory/src/main/java/org/eclipse/dataspaceconnector/negotiation/store/memory/InMemoryContractNegotiationStore.java
ndr-brt: Why someone would set the `filter` as null? A null check could be added on the builder however, like it was done for offset and limit.

paullatzelsperger: I'd imagine the `QuerySpec` to be created based on eg. query params of a REST controller, so it seems convenient to just create the `QuerySpec` object directly from the params. 

The `offset` and `limit` are a bit different, because not checking for `null` may cause NPEs during auto-boxing/-unboxing directly in the builder methods.

That said, I don't mind having an `if` statement in the builder method :)

PR #694, opened 2022-02-14
Open Telemetry custom spans

### LINKED ISSUE #643: Tracing - OpenTelemetry custom spans and tags
Issue #355 introduces support for distributed tracing in EDC. This allows EDC developers to correlate requests between EDC components like the consumer and provider connector. Although this is already a first good step, additional tracing information on internal EDC operations would be valuable. An example of this could be the TransferProcess state changes so that the whole lifecycle of the TransferProcess can be observed in traces.
 
Enhance existing tracing introduced in #355 with custom spans and tags for relevant EDC internal operations, so that they can be visualized in monitoring tools.
 
Tasks:
- [ ]  Decide which additional internal EDC operations are relevant for tracing
- [ ]  Implement additional spans and tags with open telemetry

marcgs:
closes #643 
authors: @ouphi @marcgs 

This PR enriches tracing as introduced in #688 with custom spans providing information that will help identify EDC issues.

Custom spans can be added using Open Telemetry's [@WithSpan](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/manual-instrumentation.md#creating-spans-around-methods-with-withspan) annotation. In this PR we suggest to introduce such custom spans for the following operations:

- Consumer-provider connector communication (methods called by REST handlers)
- State handling methods (methods called by the ContractNegotiation and TransferProcess state machine)
- Resource provisioning and deprovisioning
- TransferProcess data flows

## Sample ContractNegotiation Trace:
contract_negotiation

## Sample Transfer Process Trace
transfer_process

⚠️ The transfer process trace reveals that TransferProcesses are never moved to the "complete" state on the provider side. This seems to be a bug in EDC.


jimmarino (approved):

marcgs:
@paullatzelsperger What is your take on the issue visible on the transfer process trace? It seems Transfer Processes are never moved to completed on the provider side when running sample 04.

paullatzelsperger:
> @paullatzelsperger What is your take on the issue visible on the transfer process trace? It seems Transfer Processes are never moved to completed on the provider side when running sample 04.

This happens, because at this time only the `StatusChecker` can move a transferprocess to `COMPLETE`. Status checkers are currently only used on the Consumer side. The working assumption until now was that there will be some sort of external signal that "finishes" the transfer on the provider side. 

I am aware of this, but so far didn't act on it, since the `InlineDataFlowController` will likely get substituted altogether by the DPF. Should we decide to keep it, e.g. for demos, samples, quick-starting new devs, we can utilize the `CommandQueue` from inside the `InlineDataFlowController` to `COMPLETE` the command after the transfer finishes.

paullatzelsperger (approved):
Looks good to me. Just out of curiosity: how does correlation work? Meaning: how is it determined which method invocations "belong" together?

marcgs:
> Looks good to me. Just out of curiosity: how does correlation work? Meaning: how is it determined which method invocations "belong" together?

Open telemetry keeps the tracing context information in the current thread. Thanks to the previous work done in #688 this context is propagated and persisted in the corresponding business entity carrier (TransferProcess or ContractNegotiation) on creation, and then back into the current thread when processing for that entity resumes.

Issue #683, opened 2022-02-12
Fix modularity drift


jimmarino:
## Overview
During the M2 development cycle, we have introduced substantial drift in the modular runtime architecture the EDC is built on. I'm planning on fixing this ASAP, before the M2 release.

This issue will describe the areas of drift that have been uncovered and give an overview of how they will be fixed. The issue will also start to codify EDC modularity rules and suggest how they may be enforced as part of the build system.

## Modularity Drift

M2 has introduced a number of dependency tangles and improper coupling between modules. 

1. There are many dependencies on core modules throughout the codebase, in particular core:base. No other module should depend on core:base, core:transfer, core:contract, or core:boot other than launcher or BOM modules

2. The core:base module is used as an extension and a library module. It should only be an extension. Implementation classes that are used by other modules (e.g. by services they provide or for testing) need to be moved out to an appropriate module. 

3. The core modules themselves have improper dependencies that couple various runtimes. For example, core:base relies on the spi BOM module, which pulls in spi:contract and spi:transfer. The core:base module should only include essentials common to all runtimes (e.g. Connector, Data Plane Server, Catalog, etc.). These couplings create a tangled dependency graph. 

## Remedy

These issues will be fixed by doing the following:

### I. SPI module cleanup

**core:policy:policy-evaluator** 

- [ ] Module core:policy:policy-evaluator should be moved out of core because spi modules (e.g. contract-spi, contract-transfer) relies on it. This module should be moved to spi and renamed policy-spi-model

**spi:web-spi** 

- [ ] Module spi:web-spi move out of spi and up to a higher level

**core:spi**

- [ ] Clean up core-spi system package by creating sub-packages for injection, configuration, extensions

### II. Core module cleanup

**core:base**

The core:base module is used as a library and an extension. This needs to be fixed by removing its use as a library.

- [x] core:base relies on the spi BOM, which drags in spi:transfer and spi:contract. Remove this.

- [x] `ExponentialWaitStrategy` implements `TransferWaitStrategy`, `NegotiationWaitStrategy` should not do that because it couples base to spi:transfer and spi:contract. `TransferWaitStrategy` and `ContractWaitStrategy` should only be used as extension point interfaces and should not be referenced by clients (they should reference `WaitStrategy`). 

- [x] Move `ExponentialWaitStrategy` to core-spi since it is accessed by other core modules. The core:contract and core:transfer modules will create wait strategies for their managers directly.

[...]

PR #695, opened 2022-02-14
Fix modularity drift

jimmarino:
Implements the Section 2 of #683


core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractNegotiationCommandExtension.java
```diff
@@ -30,23 +29,18 @@
...
+
     @Override
     public void initialize(ServiceExtensionContext context) {
-        CommandHandlerRegistry registry = context.getService(CommandHandlerRegistry.class, true);
-        if (registry == null) {
-            registry = new CommandHandlerRegistryImpl();
-            context.registerService(CommandHandlerRegistry.class, registry);
-        }
-        
+        CommandHandlerRegistry registry = context.getService(CommandHandlerRegistry.class);
```
ndr-brt, line 38: This could injected with `@Inject` annotation

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/TransferProcessCommandExtension.java
```diff
@@ -36,12 +35,7 @@
 
     @Override
     public void initialize(ServiceExtensionContext context) {
-        CommandHandlerRegistry registry = context.getService(CommandHandlerRegistry.class, true);
-        if (registry == null) {
-            registry = new CommandHandlerRegistryImpl();
-            context.registerService(CommandHandlerRegistry.class, registry);
-        }
-
+        CommandHandlerRegistry registry = context.getService(CommandHandlerRegistry.class);
```
ndr-brt, line 38: Should be injected via `@Inject`

extensions/policy/ids-policy/src/test/java/org/eclipse/dataspaceconnector/ids/policy/MockPolicyContext.java
```diff
@@ -0,0 +1,50 @@
...
+package org.eclipse.dataspaceconnector.ids.policy;
+
+import org.eclipse.dataspaceconnector.spi.contract.agent.ParticipantAgent;
+import org.eclipse.dataspaceconnector.spi.contract.policy.PolicyContext;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class MockPolicyContext implements PolicyContext {
```
ndr-brt, line 23: It's equal to `PolicyContextImpl`

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/TransferProcessCommandExtension.java
jimmarino: OK, let's do that in a subsequent PR because there is some other cleanup I want to do here that is related.

extensions/policy/ids-policy/src/test/java/org/eclipse/dataspaceconnector/ids/policy/MockPolicyContext.java
jimmarino: OK, let's remove that in a subsequent PR

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractNegotiationCommandExtension.java
jimmarino: OK, let's do that in a subsequent PR because there is some other cleanup I want to do here that is related.

ndr-brt (approved):

extensions/azure/cosmos/assetindex-cosmos/build.gradle.kts
```diff
@@ -17,13 +17,18 @@ plugins {
...
+val jodahFailsafeVersion: String by project
 
 dependencies {
     api(project(":spi"))
     api(project(":common:util"))
     api(project(":extensions:azure:cosmos:cosmos-common"))
     api(project(":extensions:dataloading"))
+
     implementation("com.azure:azure-cosmos:${cosmosSdkVersion}")
+    implementation("com.squareup.okhttp3:okhttp:${okHttpVersion}")
```
paullatzelsperger, line 30: didn't find a usage for this, so maybe it can go away?

extensions/azure/cosmos/contract-negotiation-store-cosmos/build.gradle.kts
```diff
@@ -17,14 +17,15 @@ plugins {
 }
 
 val cosmosSdkVersion: String by project
+val jodahFailsafeVersion: String by project
 
 dependencies {
     api(project(":spi"))
```
paullatzelsperger, line 23: does this really need the entire SPI or could we get away with `:spi:contract-spi`?

extensions/policy/ids-policy/build.gradle.kts
```diff
@@ -21,7 +21,6 @@ plugins {
 
 
 dependencies {
-    api(project(":core:contract"))
     api(project(":spi"))
```
paullatzelsperger, line 24: does this really need the entire SPI?

paullatzelsperger (commented):
Couple of comments/questions in the code. 

It feels like moving code from `core:base` into `:spi:*` rather than simply depending onto `core:base` isn't improving much, as it effectively turns the SPI into a library module. For instance the `ConsoleMonitor`, `BoundedCommandQueue` (which are implementation classes!), the `CommandProcessor`... aren't extensible interfaces.
I think its okay-ish for now, but down the road we may need a `commons` module that acts as something of a "pool for commonly used classes" (for lack of a better word). And since we already have `commons`...

jimmarino:
> Couple of comments/questions in the code.
> 
> It feels like moving code from `core:base` into `:spi:*` rather than simply depending onto `core:base` isn't improving much, as it effectively turns the SPI into a library module. For instance the `ConsoleMonitor`, `BoundedCommandQueue` (which are implementation classes!), the `CommandProcessor`... aren't extensible interfaces. I think its okay-ish for now, but down the road we may need a `commons` module that acts as something of a "pool for commonly used classes" (for lack of a better word). And since we already have `commons`...

We can't put the impl classes in commons util because they depend on the SPI. The problem with relying on core:base is that the former is a service extension, not a library, which is broken. We can look later to pull out individual libraries but deal with specific areas such as commands. I don't think we should have a general library, however, as that will turn into the equivalent of a dumping ground.

I don't have a problem with small-ish implementation classes in SPI (e.g. ConsoleMonitor) but we can revisit this later once the depdency tangles are cleaned up.

extensions/azure/cosmos/assetindex-cosmos/build.gradle.kts
jimmarino: I didn't look too much at third-party libraries in this PR. I'll remove this one though.

extensions/azure/cosmos/fcc-node-directory-cosmos/build.gradle.kts
```diff
@@ -17,13 +17,16 @@ plugins {
 }
 
 val cosmosSdkVersion: String by project
+val jodahFailsafeVersion: String by project
 
 dependencies {
     api(project(":spi"))
```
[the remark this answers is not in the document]
jimmarino: I missed that, will fix

extensions/azure/cosmos/transfer-process-store-cosmos/build.gradle.kts
```diff
@@ -17,13 +17,16 @@ plugins {
 }
 
 val cosmosSdkVersion: String by project
+val jodahFailsafeVersion: String by project
 
 dependencies {
     api(project(":spi"))
```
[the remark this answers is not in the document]
jimmarino: Ditto, fixed

extensions/azure/cosmos/contract-negotiation-store-cosmos/build.gradle.kts
jimmarino: Ditto, fixed

extensions/azure/cosmos/contract-definition-store-cosmos/build.gradle.kts
```diff
@@ -17,13 +17,16 @@ plugins {
 }
 
 val cosmosSdkVersion: String by project
+val jodahFailsafeVersion: String by project
 
 dependencies {
     api(project(":spi"))
```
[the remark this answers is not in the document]
jimmarino: ditto, fixed

extensions/api/observability/build.gradle.kts
```diff
@@ -25,7 +25,6 @@ plugins {
 dependencies {
     api(project(":spi"))
```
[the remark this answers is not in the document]
jimmarino: Turns out it needs web-spi as well, fixed.

ndr-brt:
@jimmarino I have a question. If a common-logic class emerged (e.g. between `core-transfer` and `core-contract` modules) where it should be put? in the spi? you talked about individual libraries, but where they could stay? in the `commons` path?
Because I'm looking to extracting a generic implementation ok the state machine loop but I cannot find a good place to put it in, `spi` seems not a great choice.

PR #697, opened 2022-02-15
Propagate exception with Atomikos (#696)

### LINKED ISSUE #696: AtomikosTransactionContext fails to propagate exception
Upon exception, the transaction is rolled back, but the exception is not propagated up, hence the caller flow is not interrupted.

algattik:
Closes #696


jimmarino (approved):
Thanks!

PR #699, opened 2022-02-15
dpf: Add public Api to DPF and make HttpDataSource more generic

bscholtes1A:


extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/transfer/DataPlanePublicApiController.java
```diff
@@ -0,0 +1,113 @@
...
+import static org.eclipse.dataspaceconnector.dataplane.api.common.ResponseFunctions.notAuthorizedErrors;
+import static org.eclipse.dataspaceconnector.dataplane.api.common.ResponseFunctions.validationError;
+import static org.eclipse.dataspaceconnector.dataplane.api.common.ResponseFunctions.validationErrors;
+import static org.eclipse.dataspaceconnector.dataplane.core.schema.DataRequestSchema.METHOD;
+import static org.eclipse.dataspaceconnector.dataplane.core.schema.DataRequestSchema.QUERY_PARAMS;
+
+/**
+ * Controller for synchronous data pull APIs of the data plane.
+ */
+@Consumes({MediaType.APPLICATION_JSON})
```
MoritzKeppler, line 55: Popose to document more API details using the swagger annotations https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Annotations
So far we missed to do so but I think it would help a lot, especially a short sentence about who is thought to call an endpoint in which situation.
thx a lot in advance!

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/transfer/DataPlanePublicApiController.java
```diff
@@ -0,0 +1,113 @@
...
+
+    public DataPlanePublicApiController(TokenValidationClient tokenValidationClient, DataPlaneManager dataPlaneManager, Monitor monitor, TypeManager typeManager) {
+        this.tokenValidationClient = tokenValidationClient;
+        this.dataPlaneManager = dataPlaneManager;
+        this.monitor = monitor;
+        this.typeManager = typeManager;
+    }
+
+    @GET
+    @Path("validate/{token}")
```
MoritzKeppler, line 74: wouldn't it be better to pass the token as part of the http headers?
* pathes often get logged and we should avoid to have security information in logfiles.
* also the token could get quite long
* and we could handle various pathes belonging to the same contract in future (query params are not sufficient in every case)
What do you think @bscholtes1A ?

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/transfer/DataPlanePublicApiController.java
```diff
@@ -0,0 +1,113 @@
...
+            return internalErrors(transferResult.getFailureMessages());
+        }
+        return Response.ok().entity(Map.of("data", stream.toString())).build();
+    }
+
+    private DataFlowRequest createDataFlowRequest(ClaimToken claims, MultivaluedMap<String, String> queryParams, String method) {
+        var dataAddress = typeManager.readValue(claims.getClaims().get("dad"), DataAddress.class); // TODO: use constant here
+        return DataFlowRequest.Builder.newInstance()
+                .processId(UUID.randomUUID().toString()) // TODO: map the transfer process id into the token?
+                .destinationDataAddress(dataAddress)
```
MoritzKeppler, line 107: isn't it the sourceAddress instead of the destinationAddress that we have to provide?

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/transfer/DataPlanePublicApiController.java
```diff
@@ -0,0 +1,113 @@
...
+    public Response getData(@PathParam("token") String token, @Context UriInfo uriInfo, @Context Request request) {
+        // validate and decode input token
+        var tokenValidationResult = tokenValidationClient.validate(token);
+        if (tokenValidationResult.failed()) {
+            return notAuthorizedErrors(tokenValidationResult.getFailureMessages());
+        }
+
+        // create request and perform validation
+        var dataFlowRequest = createDataFlowRequest(tokenValidationResult.getContent(), uriInfo.getQueryParameters(), request.getMethod());
+        var validationResult = dataPlaneManager.validate(dataFlowRequest);
```
MoritzKeppler, line 84: as we already call the control plane: would it also be an option to retrieve more details like policies we have to fulfill? e.g. "log every request"
And if so: could we remove those details from the token?

jimmarino: @MoritzKeppler Yes, we need to do this, probably M3 b/c it needs to be thought through a bit more.

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/transfer/DataPlanePublicApiController.java
jimmarino: @paullatzelsperger mentioned he will do this with the API work he is engaged in.

Issue #432, opened 2021-12-15
DevOps - Useful - provide Helm charts


algattik:
Providing sample charts would accelerate implementation and favour use of best practices.

ndr-brt:
In Catena-x we have a specific repository to do that: https://github.com/catenax-ng/product-edc
Probably they won't fit better here as every EDC deployments depends from the extensions needed.

@algattik if you agree I would close this

algattik:
Closing as implemented in #701

Issue #608, opened 2022-02-02
Settings as environment variables not working


DominikPinsel:
Currently the  environment variables settings are searched using the default setting keys (e.g. edc.web.port).
I assume the setting key should be made uppercase and the dots should be replaced by underscore.

From `DefaultServiceExtensionContext` 

```java
    @Override
    public String getSetting(String key, String defaultValue) {
        String value;
        for (ConfigurationExtension extension : configurationExtensions) {
            value = extension.getSetting(key);
            if (value != null) {
                return value;
            }
        }
        value = System.getProperty(key);
        if (value != null) {
            return value;
        }
        value = System.getenv(key); /* e.g. key="edc.web.port" is not a environment variable */
        return value != null ? value : defaultValue;
    }
```

<sub>Dominik Pinsel <dominik.pinsel@daimler.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md) </sub>

ndr-brt:
Good catch. Actually you posted an old piece of code, now the settings are obtained from a `Config` instance. Fixing this is a matter to map the environment variable keys to the "java properties" format before creating the config object here:
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/32936a4eb6ecba5eb33617c75ea2cd119897757a/core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java#L246

DErCouch:
Suggest to mark system properties and system environment variables like the following, cause they are different in their meaning

1. System Property -> lowercase separated by dot (e.g., os.name)
2. System Environment -> uppercase separated by underscore (e.g., JAVA_HOME) 

Therefore in my opinion the better place to address this is directly in the ConfigFactory:

https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/32936a4eb6ecba5eb33617c75ea2cd119897757a/core/base/src/main/java/org/eclipse/dataspaceconnector/core/config/ConfigFactory.java#L31-L41

paullatzelsperger:
> Good catch. Actually you posted an old piece of code, now the settings are obtained from a `Config` instance. Fixing this is a matter to map the environment variable keys to the "java properties" format before creating the config object here:
> 
[...]

PR #701, opened 2022-02-16
Helm chart with minikube tests (#432)

algattik:
A [Helm Chart](https://helm.sh) is provided to be used to deploy the EDC runtime in Kubernetes environments.

### Chart design

The chart is highly configurable:

- Configurable runtime based on Docker image of base EDC, Data Plane Framework Server, or custom runtime.
- Configurable environment variables (as [ConfigMap](https://kubernetes.io/docs/concepts/configuration/configmap/)) for runtime settings, following [12-factor principles](https://12factor.net/config).
- Configurable secrets (injected as environment variables into pods).
- Configurable labels and annotations.
- Configurable ingress.

The chart was generated using `helm create` in order to follow best practice for components, configurability and ingress definitions, and adapted to fit the EDC runtime.

A generic Dockerfile was created that builds a Docker image using any runtime JAR. It is designed to be used as-is and take configuration from environment variables. Alternatively, implementors can extend the Dockerfile to package a configuration file.

### Testing

The chart must be tested in CI, like any other code, to ensure it remains valid as either the chart or EDC code evolves.

A GitHub worklow installs minikube (a local lightweight kubernetes installation) and deploys two releases of the chart, for the Consumer and Provider EDC connector of sample 04.0. The workflow then triggers file copy and verifies that the destination file was created on the Provider pod.

### DIFF
--- a/launchers/generic/Dockerfile
+++ b/launchers/generic/Dockerfile
@@ -1,0 +1,22 @@
+#
+#  Copyright (c) 2022 Microsoft Corporation
+#
+#  This program and the accompanying materials are made available under the
+#  terms of the Apache License, Version 2.0 which is available at
+#  https://www.apache.org/licenses/LICENSE-2.0
+#
+#  SPDX-License-Identifier: Apache-2.0
+#
+#  Contributors:
+#       Microsoft Corporation - initial API and implementation
+#
+#
+
+FROM openjdk:11-jre-slim
+ARG JAR
+
+COPY $JAR app.jar
+
+# Use "exec" for Kubernetes graceful termination (SIGINT) to reach JVM.
+ENTRYPOINT [ "sh", "-c", \
+    "exec java $JVM_ARGS -jar app.jar"]
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -118,3 +118,38 @@ jobs:
           killall java      
 
       # we cannot check sample 5 currently because we'd need repo secrets for that (client id,...)
+
+
+  Helm-Chart:
+    runs-on: ubuntu-latest
+    name: Test Helm chart in minikube
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-java@v2
+        with:
+          distribution: 'temurin'
+          java-version: '11'
+          # Do not use cache, as it would cause the test not to run on cache hit
+
+      - name: Install and start minikube
+        run: |
+          curl -fLo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
+          chmod +x minikube
+          ./minikube start --wait all
+
+      - name: Build JARs
+        run: ./gradlew samples:04.0-file-transfer:consumer:shadowJar samples:04.0-file-transfer:provider:shadowJar --no-daemon
+
+      - name: Test
+        run: |
+          eval $(minikube -p minikube docker-env)
+          samples/04.0-file-transfer/integration-tests/minikube/minikube-test.sh
+
+      - name: Kubernetes logs
+        if: always()
+        run: |
+          set -x
+          kubectl get pods
+          kubectl logs deployment/provider-dataspace-connector --tail=200
+          kubectl logs deployment/consumer-dataspace-connector --tail=200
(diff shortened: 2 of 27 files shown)


jimmarino (approved):

launchers/generic/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN apt-get update && apt-get install -y \
+  bash \
+  && rm -rf /var/lib/apt/lists/*
+
+COPY $JAR app.jar
+
+# Use "bash" (instead of default "sh") to support environment variables containing a dot. 
+# Use "exec" for Kubernetes graceful termination (SIGINT) to reach JVM.
+ENTRYPOINT [ "bash", "-c", \
+    "exec java $JVM_ARGS -jar app.jar"]
```
florianrusch-zf, line 27: Do we really have environment variables with a dot? Thought we have some kind mapping mechanism which transforms between screaming snake case and the dot case. E.g `STUFF_TEST` to `stuff.test`

algattik: This mapping exists only for settings fetched with `propOrEnv` e.g. the location of the config file `edc.fs.config`, not for other settings:

```
$ env web_http_port=9192 WEB_HTTP_PORT=9192 java -jar samples/04.0-file-transfer/consumer/build/libs/consumer.jar
INFO 2022-02-17T08:53:52.549321 HTTP context 'default' listening on port 8181


$ env web.http.port=9192 java -jar samples/04.0-file-transfer/consumer/build/libs/consumer.jar
INFO 2022-02-17T08:54:00.39245 HTTP context 'default' listening on port 9192
```

florianrusch-zf: Interesting. Do you or anyone else know if there is already a plan to align this to have some consistent ways to set the variables?

DominikPinsel: In issue #608 there is a discussion to make environment variables always snake case.

paullatzelsperger: I think the following should happen:
- internally in the runtime we only use the dot notation
- the `Config` mechanism checks for properties in the dot notation and env-vars in the uppercase-underscore notation
- `propOrEnv` should not be used anywhere but tests

ndr-brt: to be fair, that conversion was not in place before, it has only been implemented in the `propOrEnv`. However an issue about that was already opened: #608

florianrusch-zf: Thank you guys! Didn't want to start a discussion here, so lets move over with this topic to the other issue 😉

paullatzelsperger: doesn't matter now, i didn't want to assign blame :) Your comment on #608 is correct and should be done so.

jimmarino: How about fixing the setting properties by normalizing them as suggested by @ndr-brt and @paullatzelsperger and then realigning this PR with that?

jimmarino: I'm OK to merge this as-is and fix the other issue ASAP. We can then convert shells in a subsequent PR if wanted.

ndr-brt (commented):
I would avoid the `charts` folder in an already crowded root folder, I would put it into `resources`

paullatzelsperger:
Moving the `charts` folder to `resources` as @ndr-brt suggested seems reasonable and will probably have little impact on the rest of the code. Other than that I'm OK with this.

algattik:
> I would avoid the `charts` folder in an already crowded root folder, I would put it into `resources`

Done. I would still recommend to put it in the root though, so that “casual” users find it

ndr-brt:
> > I would avoid the `charts` folder in an already crowded root folder, I would put it into `resources`
> 
> Done. I would still recommend to put it in the root though, so that “casual” users find it

We can document that in the README, probably is the first place where a "casual" user take a look at.

algattik:
> > > I would avoid the `charts` folder in an already crowded root folder, I would put it into `resources`
> > 
> > 
> > Done. I would still recommend to put it in the root though, so that “casual” users find it
> 
> We can document that in the README, probably is the first place where a "casual" user take a look at.

I have already :) thanks

paullatzelsperger (approved):

ndr-brt (approved):

PR #713, opened 2022-02-20
Code clean up related to DataPlane implementation

### LINKED ISSUE #653: Centralize token creation/validation services
There are several places in the codebase where we have to deal with tokens, especially for `Oauth2` extension and in the `DataPlane`. 

In PR #639 we initiated the `token-services` extension, which provides generic services to generate and validate tokens. These services are the ones used by the `DataPlane`

These services are largely inspired of what is already done in the `Oauth2` extension, especially for the token validation (see `ValidationRule` interface). 

The objective would be here to centralize the token management stuff so that both `Oauth2` and the `DataPlane` could rely on a common code base and set of interfaces when dealing with tokens.

bscholtes1A:
Closes #653


jimmarino (approved):
LGTM. @bscholtes1A  shouldn't core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/SyncTransferProcessManager.java 

be removed? It looks like the empty file was left.

bscholtes1A:
> LGTM. @bscholtes1A shouldn't core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/SyncTransferProcessManager.java
> 
> be removed? It looks like the empty file was left.

No idea why it shows as an empty file. On my side this file does not appear anymore.

Issue #618, opened 2022-02-03
DPF - Azure Object Storage Extension


zeier:
# Overview

This issue adds the Azure Object Storage extension for stream copy to the Data Plane Framework (DPF) #463 to read data from and write data to Azure Blob Storage.

### Data Movement

- Extend the existing DPF Transfer function controller to:
  - Read data from Azure Blob Storage
  - Write data into Azure Blob Storage
- Azure SDK streams operations are used to perform data operations. This will support large datasets as well.
- End-to-end testing on Azure
- Test execution to validate builds

### In-Place Copy and Parallelization (Out of Scope)

In-place copy and parallelization are out of scope for this issue.

#773 will provide another extension that uses Azure Data Factory for optimized transfers.

Additional optimizations (parallelization and server-side copy) without usage of Azure Data Factory could be implemented in a subsequent step if needed.

zeier:
Updated to reflect usage of Azure Data Factory instead of implementation of parallelization and server-side copy to optimize throughput and latency.

algattik:
Item left is to share a thread pool across all DataSink implementations. We will raise a PR for this.

zeier:
Made delineation between this issue and #773 clearer by mentioning that Azure Data Factory is out of scope for this issue.

Issue #463, opened 2021-12-21
Data Plane Framework


jimmarino:
# Overview

This issue will detail the architecture and design of the Data Plane Framework (DPF), which will provide a mechanism for implementing an extensible `DataFlowController` that supports a myriad of deployment topologies.

## Scope

The Data Plane Framework is designed to only work with finite data transfers and small payload, latent non-finite transfers such as events. High-volume or latency-sensitive streaming (non-finite) data transfers should be handled by other `DataFlowController` implementations that delegate to specialized third-party infrastructure such as Kafka.

## Principles

1. **Minimal state.** All state pertaining to a transfer process will be maintained by the Connector Control Plane as part of the `TransferProcess`. The only state that will be maintained by the DPF is if a transfer process has been completed. This requires the Control Plan to issue retries in the event of failure.

2. **No transformations.** The DPF is not an ETL tool. There will be no facilities for data transformations or processing. This is expected to handled by the Control Plane as part of the provisioning state.

3. **Do not reinvent the wheel.** The DPF will rely on existing data transfer technology such as S3, Azure Object Storage, FTP, etc. As a rule of thumb: the DPF should not contain any wire protocol implementations.

4. **Flexible Deployment.** It must be possible to:

    - Deploy the DPF to a K8S cluster
    - Deploy the DPF remotely from the Control Plane
    - Deploy the DPF in the same process as the Control Plane for demo and testing purposes

5. **Extensible.** The DPF will be built using the EDC modularity and extensibility system.     

### Data Plane Manager

The `DataPlaneManager` (DPM) will enqueue `DataFlowRequests` requests, which may originate from an ingress such as an HTTP endpoint. The queue implementation will be a bounded, in-memory data structure that provides support for backpressure when the system is overloaded. However, since the queue is in-memory, the client control plane submitting the request will be responsible for re-submitting a request in the event of failure:

The `DataPlaneManager` will look similar to the following:

```
/**
 * Manages the execution of data plane requests.
 */
public interface DataPlaneManager {

    /**
     * Initiates a transfer for the data flow request. This method is non-blocking with respect to processing the request.
     */
    void initiateTransfer(DataFlowRequest dataRequest);

    /**
     * Performs a data transfer using the supplied data source.
     */
    CompletableFuture<TransferResult> transfer(DataSource source, DataFlowRequest request);

    /**
     * Performs a data transfer using the supplied data sink.
     */
[...]

PR #719, opened 2022-02-21
DPF - Azure Object Storage Extension (#618)

algattik:
Add the Azure Object Storage extension to the Data Plane Framework (DPF) #463 to read data from and write data to Azure Blob Storage.

In scope:
- source to read a single blob from Azure storage.
- sink to write one or multiple entries into Azure storage.
- integration test verifying copy between 2 storage accounts in storage emulator.

Out of scope for this PR:
- reading multiple blobs from Azure storage, e.g. enumerating a container or using wildcards.
- optimisations such as parallel or server-side copy.

Introduced an abstract class `ParallelSink` with the logic shared across implementations supporting parallel writes (HTTP and Blob, presumably in the future most if not all sinks).


extensions/azure/data-plane-azure-storage/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azurestorage/adapter/BlobAdapterFactory.java
```diff
@@ -0,0 +1,51 @@
...
+        this.blobstoreEndpoint = blobstoreEndpoint;
+    }
+
+    public BlobAdapter getBlobAdapter(String accountName, String containerName, String blobName, String sharedKey) {
+        var blobServiceClient = new BlobServiceClientBuilder()
+                .credential(new StorageSharedKeyCredential(accountName, sharedKey))
+                .endpoint(createEndpoint(accountName))
+                .buildClient();
+
+        BlockBlobClient blockBlobClient = blobServiceClient
```
jimmarino, line 38: Minor nit: in the same class it would be good to be consistent with the use of var or not

extensions/azure/data-plane-azure-storage/src/test/java/org/eclipse/dataspaceconnector/azure/dataplane/azurestorage/pipeline/AzureStorageDataSinkTest.java
```diff
@@ -0,0 +1,124 @@
...
+import static org.eclipse.dataspaceconnector.azure.dataplane.azurestorage.pipeline.AzureStorageTestFixtures.createContainerName;
+import static org.eclipse.dataspaceconnector.azure.dataplane.azurestorage.pipeline.AzureStorageTestFixtures.createRequest;
+import static org.eclipse.dataspaceconnector.azure.dataplane.azurestorage.pipeline.AzureStorageTestFixtures.createSharedKey;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+class AzureStorageDataSinkTest {
+
+    static Faker faker = new Faker();
```
jimmarino, line 36: Do we need this dependency?

extensions/data-plane/data-plane-spi/src/main/java/org/eclipse/dataspaceconnector/dataplane/spi/pipeline/ParallelSink.java
```diff
@@ -0,0 +1,102 @@
...
+ * Writes data in parallel.
+ */
+public abstract class ParallelSink implements DataSink {
+    protected String requestId;
+    protected int partitionSize = 5;
+    protected ExecutorService executorService;
+    protected Monitor monitor;
+
+    @Override
+    public CompletableFuture<TransferResult> transfer(DataSource source) {
```
jimmarino, line 41: Shouldn't we have test cases for this class even though subclasses may have them?

jimmarino (commented):
Great stuff! I'm good with this if my comments are addressed.

extensions/data-plane/data-plane-spi/src/main/java/org/eclipse/dataspaceconnector/dataplane/spi/pipeline/ParallelSink.java
algattik: Yes, I created an item in our backlog already. However I believe this is a separate item from this PR (here I only factored out existing code from HttpDataSink)

extensions/azure/data-plane-azure-storage/src/test/java/org/eclipse/dataspaceconnector/azure/dataplane/azurestorage/pipeline/AzureStorageDataSinkTest.java
algattik: It's quite a good practice to make test data variable so that we don't program against a test case, and also makes it clearer what test data is arbitrary vs special values. Faker was already used in EDC tests.

extensions/azure/data-plane-azure-storage/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azurestorage/adapter/BlobAdapterFactory.java
algattik: Fixed

extensions/data-plane/data-plane-spi/src/main/java/org/eclipse/dataspaceconnector/dataplane/spi/pipeline/ParallelSink.java
jimmarino: Prior, HttpDataSink tests covered that. ParalellSink is now refactored out in this PR so it needs to have test coverage before this is merged.

extensions/azure/data-plane-azure-storage/src/test/java/org/eclipse/dataspaceconnector/azure/dataplane/azurestorage/pipeline/AzureStorageDataSinkTest.java
jimmarino: Hmm, I didn't see it was included in the logger extension. For these tests, I don't see the value in pulling in another dependency for this but since it is already being used, let's just keep it.

extensions/data-plane/data-plane-spi/src/main/java/org/eclipse/dataspaceconnector/dataplane/spi/pipeline/ParallelSink.java
algattik: HttpDataSink extends this new class, that just factors out a method from it, so coverage is not changed. ParallelSink has 100% coverage. Is your concern specifically that it should have coverage from tests *within its own module?* I can look into that.

PR #721, opened 2022-02-21
Run locally CosmosDB tests

### LINKED ISSUE #360: Integration-Tests: replace external systems with self-contained systems
Integration tests should use self-contained systems (e.g. in a docker container) rather than actual cloud infrastructure.
At the moment we have tests running against Azure CosmosDB and AWS S3.
The following solutions should be looked into: 
- CosmosDB Emulator: https://docs.microsoft.com/en-us/azure/cosmos-db/linux-emulator?tabs=ssl-netstd21
- AWS S3: https://hub.docker.com/r/minio/minio/

Lets make sure we're not violating any licenses etc. in the process :)

Linked to #246 

#377 contributes

_Note: we want to create an E2E-Enviroment, where we actually deploy connector instances in the cloud, which could then indeed use the "real" cloud backends again. That is more of a long-term goal though._

ndr-brt:
Fix #360

### Context
Currently, it's not possible to run the CosmosDB integration tests locally, making working to that code hard and error prone.

### What brings
I explored the [CosmosDB Emulator](https://docs.microsoft.com/en-us/azure/cosmos-db/linux-emulator) tool, with this PR will be possible to run the CosmosDB tests by running a local instance of CosmosDB emulator as explained in the [azure-test readme](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/compare/main...ndr-brt:feature/360-cosmosdb-it-local?expand=1#diff-42c99770082a466df8a3e99d227b3129ddeff4e89e89bc221dbbb53c3e1e3845).

The scope was to make those running also on Actions, but seems like there are some performance problems, as the tests start to fail after some time (e.g. https://github.com/ndr-brt/DataSpaceConnector/runs/5277051610), so for the moment I think the improvement is enough.


paullatzelsperger:
@algattik maybe you and @ndr-brt could sync on this topic, lest we duplicate workstreams.

algattik:
Beautiful work @ndr-brt !! Regarding the GH action, the issue you experienced in not stability but emulator capacity, when running with `AZURE_COSMOS_EMULATOR_PARTITION_COUNT=1` See https://docs.microsoft.com/azure/cosmos-db/emulator-command-line-parameters#set-partitioncount

extensions/azure/azure-test/src/testFixtures/java/org/eclipse/dataspaceconnector/azure/testfixtures/CosmosTestClient.java
```diff
@@ -0,0 +1,107 @@
...
+            var keystorePath = System.getProperty("java.home") + separator + "lib" + separator + "security" + separator + "cacerts";
+            var keystore = KeyStore.getInstance(KeyStore.getDefaultType());
+
+            var keystorePassword = "changeit".toCharArray();
+            keystore.load(new FileInputStream(keystorePath), keystorePassword);
+
+            var certificate = getCertificateFrom(endpoint);
+            keystore.setCertificateEntry("cosmosdb", certificate);
+
+            try (var output = new FileOutputStream(keystorePath)) {
```
algattik, line 58: I get this. Would need to run as root.

```
Caused by:
        java.io.FileNotFoundException: /usr/lib/jvm/java-11-openjdk-amd64/lib/security/cacerts (Permission denied)
            at java.base/java.io.FileOutputStream.open0(Native Method)
            at java.base/java.io.FileOutputStream.open(FileOutputStream.java:298)
            at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:237)
            at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:126)
            at org.eclipse.dataspaceconnector.azure.testfixtures.CosmosTestClient.localClient(CosmosTestClient.java:58)
            ... 2 more
```

algattik: This seems to work though

```

            var trustStorePath = someTempFile;
            try (var output = new FileOutputStream(trustStorePath)) {
                keystore.store(output, keystorePassword);
            }
            System.setProperty("javax.net.ssl.trustStore", trustStorePath);
```

extensions/azure/azure-test/src/testFixtures/java/org/eclipse/dataspaceconnector/azure/testfixtures/CosmosTestClient.java
```diff
@@ -0,0 +1,107 @@
...
+                keystore.store(output, keystorePassword);
+            }
+
+            return new CosmosClientBuilder()
+                    .key("C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==")
```
algattik, line 63: document where this is coming from?

ndr-brt: I wasn't able to find something on the official documentation, but it can be found on the emulator github repository, though there's no mentions to the fact that's a fixed value. Added that link as referral

extensions/azure/azure-test/src/testFixtures/java/org/eclipse/dataspaceconnector/azure/testfixtures/CosmosTestClient.java
ndr-brt: Good catch, I use sdkman so my jdk is in the user folder and I have no problems with permissions.

ndr-brt:
> the issue you experienced is emulator capacity, when running with `AZURE_COSMOS_EMULATOR_PARTITION_COUNT=1` See https://docs.microsoft.com/azure/cosmos-db/emulator-command-line-parameters#set-partitioncount

I tried with other values but I noticed that the bigger the partition count is, the more the errors I get (or at least I think, yesterday was a tweak-and-fail day so maybe I did the wrong assumptions)

algattik:
> I tried with other values but I noticed that the bigger the partition count is, the more the errors I get (or at least I think, yesterday was a tweak-and-fail day so maybe I did the wrong assumptions)

Could you point to a branch? We could give it a try as well

ndr-brt:
> > I tried with other values but I noticed that the bigger the partition count is, the more the errors I get (or at least I think, yesterday was a tweak-and-fail day so maybe I did the wrong assumptions)
> 
> Could you point to a branch? We could give it a try as well

The branch is the one used for this PR, but at the end I erased the `integrationtest.yml` file, need to try it again

ndr-brt:
Do we need this check on `testFixtures` code? This is needed to get the certificate from cosmosdb emulator.
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/721/files#annotation_2814408626

ndr-brt:
out of 10 pipelines, 3 succedeed, 6 failed and one staled: 
https://github.com/ndr-brt/DataSpaceConnector/actions/runs/1880912127

I propose to go on with this PR to give the possibility to run tests locally and then fix this github action thing later

algattik:
@ndr-brt this should get you going with a Windows agent. https://github.com/agera-edc/DataSpaceConnector/commit/dcae76b4ac88c65794f79067ea14a88a9603dac2
Note that Windows emulator doesn't seem to offer the emulator.pem file for download, so I tweaked the trust store part.

Windows is not as fast, but I've enabled Gradle cache so at least subsequent builds on the same config will be a bit faster.

I've reached out to the Linux emulator developers so that hopefully we can go back to a Linux agent later on.

ndr-brt:
> @ndr-brt this should get you going with a Windows agent. [agera-edc@dcae76b](https://github.com/agera-edc/DataSpaceConnector/commit/dcae76b4ac88c65794f79067ea14a88a9603dac2) Note that Windows emulator doesn't seem to offer the emulator.pem file for download, so I tweaked the trust store part.
> 
> Windows is not as fast, but I've enabled Gradle cache so at least subsequent builds on the same config will be a bit faster.
> 
> I've reached out to the Linux emulator developers so that hopefully we can go back to a Linux agent later on.

Nice, will integrate this right now

ndr-brt:
Since the integration tests with the cosmos db emulator were shaky also running them on a windows machine, I bring them back to use the `COSMOS_KEY` variable and run against an azure environment.

The PR is ready to be merged.

.github/workflows/integrationtests.yaml
```diff
@@ -36,6 +36,7 @@ jobs:
         run: ./gradlew extensions:azure:blobstorage:blob-provision:check
 
   Azure-CosmosDB-Integration-Test:
+
     # run only on upstream repo (requires secret access)
     if: github.repository_owner == 'eclipse-dataspaceconnector'
```
algattik, line 41:
```suggestion
    if: ${{ secrets.COSMOS_KEY }}
```

This way forks can also run this

algattik:
nit: should we run `container.deleteAllItemsByPartitionKey` also Before tests? In case a test has crashed?

algattik (approved):

algattik:
Thanks a lot for the extra effort you put in. Do you have more feedback on the issue with Windows emulator so we can relay this internally?

algattik:
> An alternative is to make tests able to run considering that other tests could have created entries in the database, but that need some other work, that could be addressed in another PR

OK. My question was just, if a test has crashed without cleanup, can we make sure the next run still passes simply by deleting all records also *before* each test. If it turns out this is extra effort, it's surely also good enough as-is.

ndr-brt:
> Thanks a lot for the extra effort you put in. Do you have more feedback/logs on the issue with Windows emulator so we can relay this internally?

Seems like the instance sometimes doesn't start, unfortunately I don't have a windows machine to try it out.



> > An alternative is to make tests able to run considering that other tests could have created entries in the database, but that need some other work, that could be addressed in another PR
> 
> OK. My question was just, if a test has crashed without cleanup, can we make sure the next run still passes simply by deleting all records also _before_ each test. If it turns out this is extra effort, it's surely also good enough as-is, since such crashes are probably rare.

Yes, probably it won't happen often. My opinion is that the tests should take for granted that the database could be dirty.

PR #724, opened 2022-02-22
CosmosDB: `nextForState` disallows re-leasing

### LINKED ISSUE #615: [state machines review] cosmosdb nextForState should not retrieve leased or locked entities
At the moment the `nextForState` cosmosdb implementation for the stores retrieves all the entities that are leased by the connector that own the lease, it should not do that to avoid unneeded processing on entities that are already processed (probably waiting for an async operation to be completed.

paullatzelsperger:
## What this PR changes/adds:

Some entities stored in a CosmosDB, specifically `TransferProcess` and `ContractNegotiation` objects, have an exclusive lock on them, a so-called `Lease`. 
This `Lease` is acquired during the `nextForState` call (which actually is a stored procedure) and is only broken during a `save`/`update` or a `delete`.
Also, some cleanup was done regarding duplicate classes.

## Why it does that

Due to recent changes in the `TransferProcessManager` and `ContractNegotiationManager`s, which now use async calls, it's necessary to "hide" an entity until the async thread processing it has finished and explicitly breaks the `Lease`. Otherwise two subsequent tick-overs of the state machine could attempt to process an entity that's already being processed, thereby blocking other entities from being processed.

## Further notes

While changing the `nextforState` a few related cleanup tasks were also undertaken:
- the `Lease` class was collated and moved to `cosmos-common` 
- the `lease` property of both the `TransferProcessDocument` and the `ContractNegotiationDocument` was extracted into a base class (`LeasableCosmosDocument`)
- the default `Lease` duration time base was changed from seconds to milliseconds
- some raw read methods in i-tests were extracted into helper methods.
- packages have been corrected in `cosmos-common`
- stored procedures (`nextforState.js` and `lease.js`) have been unified and moved to `cosmos-common` as resources
- invoking the `lease` SPROC is now handled centrally by the `LeaseContext` class
- much of the rather large changeset is only package declarations and imports

## Linked Issue(s)

Closes #615 

## Checklist


paullatzelsperger:
@bscholtes1A I removed you as a reviewer since @ndr-brt and @algattik are currently working in the same area anyway. 
If you are interested, any feedback if ofc welcome :)

ndr-brt (commented):
I was wondering, at this point some state changes in `TransferProcessManager` will be no more necessary, like `transitionProvisioning` before calling the provision manager.
I noticed that in `ContractNegotiationManagers` these "state changes before an async operation" aren't put in place.
Probably in another PR we should align them.

extensions/azure/cosmos/contract-definition-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/definition/store/CosmosContractDefinitionStore.java
```diff
@@ -40,7 +40,8 @@ public CosmosContractDefinitionStore(CosmosDbApi cosmosDbApi, TypeManager typeMa
     }
 
     @Override
-    public @NotNull Collection<ContractDefinition> findAll() {
+    public @NotNull
```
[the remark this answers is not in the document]
paullatzelsperger: Looks like this was caused by the auto formatter. Will take a look

extensions/azure/cosmos/cosmos-common/src/main/java/org/eclipse/dataspaceconnector/azure/cosmos/util/LeaseContext.java
```diff
@@ -0,0 +1,91 @@
...
+    public LeaseContext usingRetry(List<RetryPolicy<Object>> retryPolicies) {
+        retryPolicy = Failsafe.with(retryPolicies);
+        return this;
+    }
+
+    /**
+     * Breaks the exclusive Lock on a document
+     *
+     * @param documentId  The database ID of the document
+     * @param leaseHolder A string identifying the holder of the lease, e.g. a connector runtime name
```
algattik, line 66: how about making `leaseHolder` a field?

extensions/azure/cosmos/cosmos-common/src/test/java/org/eclipse/dataspaceconnector/cosmos/azure/CosmosDbApiImplIntegrationTest.java
```diff
@@ -8,6 +8,8 @@
 import com.azure.cosmos.models.CosmosDatabaseResponse;
 import com.azure.cosmos.models.CosmosItemRequestOptions;
 import com.azure.cosmos.models.PartitionKey;
+import org.eclipse.dataspaceconnector.azure.cosmos.CosmosDbApi;
+import org.eclipse.dataspaceconnector.azure.cosmos.CosmosDbApiImpl;
```
algattik, line 12: spurious (also in other classes)

extensions/azure/cosmos/contract-definition-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/definition/store/CosmosContractDefinitionStore.java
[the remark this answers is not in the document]
paullatzelsperger: There are quite many of the `public @NotNull ...` occurrences throughout the code base. I don't want to change them all in this PR, but it should be done in another one together with revising the auto-formatting guidelines for IntelliJ

extensions/azure/cosmos/cosmos-common/src/main/java/org/eclipse/dataspaceconnector/azure/cosmos/util/LeaseContext.java
paullatzelsperger: The original idea was to have this as unopinionated and agnostic as possible, but since the lock exclusivity targets entire runtimes rather than threads within a runtime, the upside of having sleeker method signatures probably outweighs the downside of having a fatter factory method.

paullatzelsperger:
> I was wondering, at this point some state changes in TransferProcessManager will be no more necessary, like transitionProvisioning before calling the provision manager.

I think it still is necessary - the lease could expire before the provisioning is done, and then we need to guarantee idempotency (which - I know - is another area of concern).

> I noticed that in ContractNegotiationManagers these "state changes before an async operation" aren't put in place. Probably in another PR we should align them.
Yep, I'll create the issue.

extensions/azure/cosmos/contract-definition-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/definition/store/CosmosContractDefinitionStore.java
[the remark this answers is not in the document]
ndr-brt: Actually, there's no functional differences between
```
public @NotNull Collection<ContractDefinition> findAll()
```
and
```
@NotNull
public Collection<ContractDefinition> findAll()
```

but I find the latter more readable

ndr-brt:
> > I noticed that in ContractNegotiationManagers these "state changes before an async operation" aren't put in place. Probably in another PR we should align them.
> 
> Which ones exactly do you mean?

I think those are pretty much all of them. In these cases after a nextForState the state is not changed before calling the async operation.

https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/39ee6ab6a96483dbc6e2979e47dbda0f6211d97c/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java#L329

https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/39ee6ab6a96483dbc6e2979e47dbda0f6211d97c/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java#L477

https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java#L344

https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java#L393

extensions/azure/cosmos/contract-negotiation-store-cosmos/src/test/java/org/eclipse/dataspaceconnector/contract/negotiation/store/CosmosContractNegotiationStoreIntegrationTest.java
```diff
@@ -316,35 +374,25 @@ void findAll_noQuerySpec() {
...
     }
 
     @Test
     void findAll_verifyPaging_pageSizeLargerThanCollection() {
 
-        var all = IntStream.range(0, 10).mapToObj(i -> generateDocument())
-                .peek(d -> container.createItem(d))
-                .map(ContractNegotiationDocument::getId)
-                .collect(Collectors.toList());
+        var all = IntStream.range(0, 10).mapToObj(i -> generateDocument()).peek(d -> container.createItem(d)).map(ContractNegotiationDocument::getId).collect(Collectors.toList());
```
algattik, line 387: nit: in all these tests, it seemed more legible to split chain on multiple lines

algattik (approved):

paullatzelsperger:
> > > I noticed that in ContractNegotiationManagers these "state changes before an async operation" aren't put in place. Probably in another PR we should align them.
> > 
> > 
> > Which ones exactly do you mean?
> 
> I think those are pretty much all of them. In these cases after a nextForState the state is not changed before calling the async operation.
> 
> https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/39ee6ab6a96483dbc6e2979e47dbda0f6211d97c/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java#L329
> 
> https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/39ee6ab6a96483dbc6e2979e47dbda0f6211d97c/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java#L477
> 
> https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java#L344
> 
> https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java#L393

The states will be revised in an upcoming Issue/Discussion. The current working assumption is that pessimistic locking (as implemented here) together with idempotency (e.g. in the `Provisioner`) should be sufficient.

extensions/azure/cosmos/contract-negotiation-store-cosmos/src/test/java/org/eclipse/dataspaceconnector/contract/negotiation/store/CosmosContractNegotiationStoreIntegrationTest.java
paullatzelsperger: I tend to agree for >3 chained statements.

paullatzelsperger:
@ndr-brt anything to add?

paullatzelsperger:
@jimmarino or @juliapampus could we merge this?

PR #730, opened 2022-02-23
Integrate IDS serializer, closes #236

### LINKED ISSUE #236: Use Serializer instead of ObjectMapper for Infomodel objects
In the IDS modules the class `de.fraunhofer.iais.eis.ids.jsonld.Serializer` should be used to map Infomodel objects to and from JSON. Currently, an ObjectMapper is used to perform these operations. Thus, the required context information is missing in the JSON and it cannot be deserialized by an Infomodel serializer.

Also, the MultipartController takes a RequestMessage as parameter, so that the header part is automatically mapped to a RequestMessage object upon request. This mapping should also be performed by the Infomodel serializer.

DominikPinsel:

## What this PR changes/adds:

This pull request integrates the IDS Serializer from the Fraunhofer repository. As the serializer adds message validation it was necessary to update/fix the creation of most IDS message.

## Why it does that

The IDS serializer is necessary to send and receive IDS compliant messages.

## Further notes

The biggest change happened for the negotiation, where the whole negotiation process was not based on IDS message ids.  As correlation IDs for most negotiation messages became mandatory in the IDS serializer the negotation managers where updated, to use them as well. To make this possible the `ContractOffer` got a properties map, to track corresponding message IDs.

## Linked Issue(s)

Closes #236 

## Checklist


juliapampus:
Hi @DominikPinsel, after talking to @ronjaquensel and @tmberthold, I want to highlight the following issue we always had with the IDS Infomodel and Serializer libraries: Setting the serializer version as fixed, forces the component to only ensure the compatibility to one Infomodel version, as we do not have any compatibility table. This means an incoming message with 4.2.7 could be deserialized with serializer 4.2.8, but maybe not with 4.2.6 or 5.0.0, although the handling of the IDS objects stays the same. This implies that we need to reject messages of another version in order to prevent exceptions. This raises some issues regarding the compatibility to other IDS components both central services and connectors. Maybe we could find a way to support multiple serializer versions? In this context, the EDC could recognize what version the incoming message has and take the appropriate serializer version to deserialize objects. This would solve syntax problems. Nevertheless, we also have to keep in mind and implement ways of handling missing properties. So if our message processing depends on an attribute that a newer version of the Infomodel may not cover, we need to check and implement workarounds. Thus, every attribute from an IDS input has to be checked for existence before processing it.

jimmarino:
Before code is written I would highly recommend we discuss the requirements and technical approach that should be taken. @juliapampus brings up a number of issues we should discuss upfront.

DominikPinsel:
> Before code is written I would highly recommend we discuss the requirements and technical approach that should be taken. @juliapampus brings up a number of issues we should discuss upfront.

Yea sure. Maybe we can do this in this PR, so that it's transparent for everyone.

For the discussion I'd assume these two points are given
- to be IDS compliant our IDS messages must be JSON-LD
- for the serialization and deserialization we will use the IDS-Serializer, which is forked in our EclipseDataspaceConnector organization, too.

Do you agree?

jimmarino:
> > Before code is written I would highly recommend we discuss the requirements and technical approach that should be taken. @juliapampus brings up a number of issues we should discuss upfront.
> 
> @jimmarino Yea sure. Maybe we can do this in this PR, so that it's transparent for everyone.
> 
> For the discussion I'd assume these two points are given
> 
> * to be IDS compliant our IDS messages must be JSON-LD
> * for the serialization and deserialization we will use the IDS-Serializer, which is forked in our EclipseDataspaceConnector organization
> 
> Do you agree?

That's not my point and this type of work should not begin with a PR.

juliapampus:
@DominikPinsel @jimmarino I apologize for starting a discussion here, that was not my intention. We can turn my comment into a discussion if you like. I think adding one serializer at all (without major structural changes) also fits directly into a PR like this or not? This is actually just a change of the object mapper and adding mandatory attributes.

jimmarino:
> @DominikPinsel @jimmarino I apologize for starting a discussion here, that was not my intention. We can turn my comment into a discussion if you like. I think adding one serializer at all (without major structural changes) also fits directly into a PR like this or not? This is actually just a change of the object mapper and adding mandatory attributes.

@juliapampus that's exactly aligned with my thoughts. Adding a serializer should be a simple change that fits into a defined issue and focused PR. This PR currently includes gratuitous changes to 75 files all over the codebase and its intention is unclear. I think it may be more efficient to open an Issue (isn't there already one for the serializer work?) and then create a focused PR to address it.

DominikPinsel:
> > @DominikPinsel @jimmarino I apologize for starting a discussion here, that was not my intention. We can turn my comment into a discussion if you like. I think adding one serializer at all (without major structural changes) also fits directly into a PR like this or not? This is actually just a change of the object mapper and adding mandatory attributes.
> 
> @juliapampus that's exactly aligned with my thoughts. Adding a serializer should be a simple change that fits into a defined issue and focused PR. This PR currently includes gratuitous changes to 75 files all over the codebase and its intention is unclear. I think it may be more efficient to open an Issue (isn't there already one for the serializer work?) and then create a focused PR to address it.

I will try to come up with a list of smaller issues, that all need to be resolved before it adding a serializer becomes a "simple change". Like already said, most message that were created by the code are currently not  IDS compliant (which is validated by the serializer). As these information is most of the time not next to the message creating code, many files needed to be updated, with the most complex changed in the contract negotiation.

juliapampus:
I opened a discussion for the multi-serializer topic: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/discussions/767

PR #731, opened 2022-02-23
Use correct scope for obtaining token in IdsMultipartSender

ronjaquensel:
## What this PR changes/adds:

It changes the scope used to request a token in the `IdsMultipartSender`. Previously, the ID of the recipient connector was used here instead of `idsc:IDS_CONNECTOR_ATTRIBUTES_ALL`.

## Why it does that

As defined [here](https://github.com/International-Data-Spaces-Association/IDS-G/tree/main/Components/IdentityProvider/DAPS/#request-call-to-get-a-token), the DAPS expects the scope to be `idsc:IDS_CONNECTOR_ATTRIBUTES_ALL` for a token request. 

In order to limit the token to certain connectors, the `aud` field should be used as stated [here](https://github.com/International-Data-Spaces-Association/IDS-G/tree/main/Components/IdentityProvider/DAPS/#request-token-that-is-handed-in-at-daps-side). Though at the moment, this feature is not supported by the DAPS.

## Further notes

Changing the scope for the token request eliminates the need for the abstract method `retrieveRemoteConnectorId(M request)`, as the result was only used for the token request. Therefore, it has been removed form the `IdsMultipartSender` and all its implementations. Tests have been adjusted accordingly.


denisneuling (approved):
@ronjaquensel, thanks for clarifying usage of the `idsc:IDS_CONNECTOR_ATTRIBUTES_ALL` scope - for me looks plausible. Approving.

PR #735, opened 2022-02-23
Fix unstable tests in CI

### LINKED ISSUE #728: Unstable tests in CI
Running 100 CI builds shows 17 failures mostly in unit tests.

It seems the introduction of Jacoco (#702) has increased test time by ~50%, which might cause some timeouts or race conditions to be hit. *(NB: in another PR we will disable Jacoco outside of CI, to improve the developer experience)*

Problems seem to be in:
- Multiple tests using OkHttp client. Default timeouts are quite short.
- LoaderManagerImplTest#batchSizeNotReachedWithinTimeframe, which relies on a 100 ms timeout.

marcgs:
## What this PR changes/adds:

This PR introduces diverse fixes for unstable tests.

## Why it does that

It seems the introduction of Jacoco (https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/702) has increased test time by ~50%, which might cause some timeouts or race conditions to be hit.

## Further notes

- Multiple tests using OkHttp client have been fixed. Default timeouts are quite short and they were increased.
- `LoaderManagerImplTest#batchSizeNotReachedWithinTimeframe` was using a queue with 3 elements, which was not testing what was intended ("Verify that the loader manager waits one pass when the queue does not yet contain sufficient elements"), as 3 is exactly the limit for the LoaderManager to start processing. This processing took sometimes longer than the configured 100ms timeout, making the test fail.

This should solve 16 out of 17 failures observed in 100 CI runs.

## Linked Issue(s)

Closes #728

### DIFF
--- a/common/util/src/testFixtures/java/org/eclipse/dataspaceconnector/common/testfixtures/TestUtils.java
+++ b/common/util/src/testFixtures/java/org/eclipse/dataspaceconnector/common/testfixtures/TestUtils.java
@@ -14,6 +14,8 @@
 
 package org.eclipse.dataspaceconnector.common.testfixtures;
 
+import okhttp3.OkHttpClient;
+
 import java.io.File;
 import java.io.IOException;
 import java.net.ServerSocket;
@@ -25,6 +27,7 @@ import java.nio.file.Paths;
 import java.util.Objects;
 import java.util.Random;
 import java.util.Scanner;
+import java.util.concurrent.TimeUnit;
 
 import static java.lang.String.format;
 import static org.junit.jupiter.api.Assertions.fail;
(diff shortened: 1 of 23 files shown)


common/util/src/testFixtures/java/org/eclipse/dataspaceconnector/common/testfixtures/TestUtils.java
```diff
@@ -133,4 +136,17 @@ public static String tempDirectory() {
...
         }
     }
+
+    /**
+     * Create an {@link OkHttpClient.Builder} suitable for using in unit tests. The client configured with long timeouts
+     * suitable for high-contention scenarios in CI.
+     *
+     * @return an {@link OkHttpClient.Builder}.
+     */
+    public static OkHttpClient.Builder testOkHttpClientBuilder() {
```
paullatzelsperger, line 146: Can we make this return a `OkHttpClient` instead of the `Builder`? In those cases where the builder is actually needed, which are few, we can always call `.newBuilder()`.

paullatzelsperger (commented):
LGTM other than one minor thing

common/util/src/testFixtures/java/org/eclipse/dataspaceconnector/common/testfixtures/TestUtils.java
marcgs: Thanks for the hint. Will be cleaner indeed.

paullatzelsperger (approved):

PR #758, opened 2022-02-24
DPF - HTTP extension integration tests (#618) 

cpeeyush:
## What this PR changes/adds:

- Integration tests for DPF Http extension.
- Transfer result endpoint in DataPlaneTransferController.
- Transfer results are stored in-memory DataPlaneStore.
- Usage mock-server to create in-memory http source and sink for DPF HTTP extension and assertions.
- Updated DPF HttpTestFixtures to reuse for UTs and Systems tests.

## Why it does that

- To improve test coverage of DPF HTTP extension.
- Multi layer testing using integration/system tests.

## Further notes
Existing HttpTestFixtures class modified to reuse test fixtures in UTs and Integration Tests.

## Linked Issue(s)

Relates to #618 

## Checklist


extensions/data-plane/integration-tests/src/test/java/org/eclipse/dataspaceconnector/dataplane/http/DataPlaneHttpIntegrationTests.java
```diff
@@ -0,0 +1,582 @@
...
+        // Verify HTTP Source server called exactly once.
+        httpSourceClientAndServer.verify(
+                getRequest(),
+                VerificationTimes.once()
+        );
+
+        // Verify HTTP Sink server called exactly once.
+        // TODO: reading the code, it is expected that the sink is only
+        //  called once, but it's actually called twice, we need to figure out why that is.
+        //  When the sink returns an error, it is only called once.
```
zeier, line 388: @jimmarino, @paullatzelsperger is the current behavior that the sink is called twice in this scenario as expected? If so, could you please quickly explain why?

jimmarino (approved):
LGTM let's wait for @zeier and if he OKs it, we can merge

extensions/data-plane/integration-tests/src/test/java/org/eclipse/dataspaceconnector/dataplane/http/DataPlaneHttpIntegrationTests.java
jimmarino: In general, it depends based on the value of ResponseStatus. Is the dropped connection a fatal error or retryable. 

@cpeeyush Can you describe the scenario being tested?

zeier: I'm OK with removing the comment and merging when the behavior is as expected.

cpeeyush: @jimmarino Here in a dropped connection scenario situations is that mock sink server will drop the connection without returning any response. 

And as we see in the code HttpSink client does not have a retry policy configured while posting data that's why we were looking for the reason that why sink mock server is called twice here after server drops connection.

Doing more research I found that it is happening because okhttp client has retryOnConnectionFailure=true by default and that is causing this retry on this connection dropping from server. So as per the current [okhttp client config](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/65d204f865937686b5c3149b37ab2c36a3368cc0/core/base/src/main/java/org/eclipse/dataspaceconnector/core/CoreServicesExtension.java#L138-L142) it is an expected behavior. I have added a comment in test explaining this.

PR #760, opened 2022-02-24
Update ids-connector launcher

### LINKED ISSUE #757: Update ids-connector launcher
The current `ids-connector` launcher contains all `:data-protocols:ids` modules, enabling it to communicate with other connectors via IDS multipart messages. But at the moment, it contains the `:iam:iam-mock` extension for an Identity Provider. As in the IDS context, we currently always find a DAPS as the Identity Provider, the `:iam:daps` extension should be used here instead. Then, the documentation should also be updated. It should describe which configuration parameters to provide and maybe also how to set up a local DAPS instance for testing.

ronjaquensel:
## What this PR changes/adds:

It updates the `ids-connector` launcher to use a DAPS as the Identity Provider. In addition, it updates the documentation of the launcher to explain which configurations have to be made to run an IDS connector.

## Why it does that

So far, the Identity Provider has been mocked in the `ids-connector` launcher. As currently, only the DAPS is used as an Identity Provider in the IDS context, it makes sense to integrate this here.

## Further notes

As not everybody might have access to/a certificate for a publicly hosted DAPS, the documentation will also show how to set up a local DAPS instance for testing the communication.

## Linked Issue(s)

Closes #757


DominikPinsel (approved):
Wow! Great work :) Sorry I didn't see this earlier. I must really look out better for these GitHub notifications!

launchers/ids-connector/gradle.properties
```diff
@@ -0,0 +1,30 @@
+javaVersion=11
```
ndr-brt, line 1: Is this file needed?

launchers/ids-connector/settings.gradle.kts
```diff
@@ -0,0 +1,182 @@
+/*
```
ndr-brt, line 1: is this file needed?

ndr-brt (commented):
looks great, I only don't understand why the `gradle.properties` and `settings.gradle.kts` are copied into this directory, since the gradle build command is launched from the root of the project

launchers/ids-connector/gradle.properties
ronjaquensel: Hi, it was the only way I got the Dockerfile working. Even if I called the Docker build command from the project root, the build still failed until the files were moved to the launcher directory. Maybe I also missed something there since I'm not familiar with Gradle and Docker. If we can get it working another way, I would remove the files again.

ronjaquensel: That makes sense. I'll try it out and then update the README and remove the files.

ronjaquensel: though this might be different when calling Docker run form the project root. I'll try it out and adjust it if necessary.

paullatzelsperger:
@juliapampus good to merge? enough people approved it

PR #763, opened 2022-02-24
Refactor `data-plane-transfer` module and add documentation

### LINKED ISSUE #463: Data Plane Framework

### LINKED ISSUE #585: Data Plane Framework (DPF) API
## Overview

- As mentioned in #463, DPF is (or can be) a separated runtime that _lives_ independently of the Control Plane, e.g. in a dedicated POD, as a sidecar... Thus it should expose API to support the following types of data transfer:

1. client pull: consumer actively pulls data from the consumer,
2. provider push: data is pushed asynchronously from the provider to the consumer,
3. event streaming: same as (2) except that data source is now a stream: events are pushed continuously to the consumer 

Supporting these three configuration requires two kinds of API in the data plane. 

- **Control Plane API** used to trigger the data transfer from the control plane (push)
- **Data Transfer API** used to request data (pull) 

### Sequence

- Consumer control plane sends artifact request to the provider control plane which returns back a signed token containing the data address (obtained from the `Asset`) and the `contractId`,
- It is now up to the consumer to hit the public API exposed by the Provider DPF with the previous token in order to collect the data,
- Token is forwarded to the validation façade exposed by the provider control plane, in order to decode the signed token, 
- If token is valid then data address is returned back to the DPF which then hits the provider data source to retrieve the data.
- Data is then passed returned in the response.

client_pull drawio

There will be a `DataPlaneResolver` that will resolve the incoming request into a public URL and the data plane to use for the transfer.

### Additional requirements

- at least `GET`, `POST` and `PUT` verbs should be supported
- it should be possible to convey query parameters to the data source.

Controller for public API exposed by the DPF would then look like this:

```
public interface DataPlanePublicApiController {

    @GET
    @Path("/{token}")
    Object get(@PathParam("token") String token, @Context UriInfo uriInfo) {
       ...
    }

    @POST
    @Path("/{token}")
    Object get(@PathParam("token") String token, @Context UriInfo uriInfo) {
       ...
    }

    @PUT
    @Path("/{token}")
    Object get(@PathParam("token") String token, @Context UriInfo uriInfo) {
       ...
    }
}
```
 
### Sequence

- Consumer control plane sends artifact request to the provider control plane which acknowledge the consumer that its request is being processed, i.e. policy and other checks have been successfully performed...
- Provider control plane then hits its DPF API to initiate the flow (it is recommended to leverage on http function extension here) => request contains both the source and destination data addresses,
- The provider DPF then hits the data source and push back the data provided destination data address, which can be an endpoint exposed by the consumer DPF or any other endpoint (S3 bucket, Azure blobstorage...).

provider_push drawio

Controller for the _internal_ DPF API should be as follows

```
public interface ControlPlaneApiController {

    @POST
    Response initiateFlow(DataRequest request) {
       ...
    }

   @POST
   void cancelTransfer(String processId) {
      ...
   }
}
```

All push data transfers are by default trackable and persisted in a dedicated store maintained by the data plane (in-memory, cosmos...)

### Sequence

- Consumer control plane sends artifact request to the provider control plane which acknowledge the consumer that its request is being processed, i.e. policy and other checks have been successfully performed...
- Provider control plane then hits its DPF API to initiate the flow, which triggers the creation of a `Subscriber` (if none was already existing for this data source),
- `Subscriber` persists the provision message in a store for recovery purpose,
- `Subscriber` then consumes from the event source (e.g. Kafka topic...) and convey the event as a `DataSource` to the pipeline along with the destination data address which is used to create the `Sink`. **Note that there can be more than one destination data address!**,
- Event is then propagated to the `DataSink`, which can for example be an event system (e.g. Azure EventGrid) to which the consumer applications can subscribe, a multicast sink that pushes event to consumer through http, or even a combination of both (`CompositeSink`)
- In case of a crash of the Data Plane, the persistent store is used to restart the `Subscriber` instances.

streaming drawio

bscholtes1A:
## What this PR changes/adds:

This PR refactors the `data-plane-transfer` module introduced in #711 and add some fixes related to end-to-end testing.
It also adds documentation for the `data-plane-transfer` module.

## Why it does that

No documentation was added before.

## Linked Issue(s)

Fixes #585, fixes #463


ndr-brt (approved):
Great! There's a typo in the diagram, the arrow that connects the data plane to the validation api on the provider has the number 3 but I think it should be 7 or so

bscholtes1A:
> Great! There's a typo in the diagram, the arrow that connects the data plane to the validation api on the provider has the number 3 but I think it should be 7 or so

Good catch, fixed

extensions/data-plane/README.md
```diff
@@ -0,0 +1,4 @@
+# Data Plane module
+
+A comprehensive description of the Data Plane can be
+found [here](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/463).
```
MoritzKeppler, line 4: would prefer to have the documentation as part of the repo (under version control; issues are not part of code checked out etc).

MoritzKeppler:
thx for the contribution, really like to get more docs!
two observations: 
The PR contains more than just docs, would be great if you could adjust the description accordingly.
Are the linked issues solved completely with this PR?
<sub> Moritz Keppler <moritz.keppler@daimler.com>, Daimler TSS GmbH, [legal info/cImpressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

extensions/data-plane/README.md
bscholtes1A: Documentation imported into the repo.

bscholtes1A:
> thx for the contribution, really like to get more docs! two observations: The PR contains more than just docs, would be great if you could adjust the description accordingly. Are the linked issues solved completely with this PR? Moritz Keppler [moritz.keppler@daimler.com](mailto:moritz.keppler@daimler.com), Daimler TSS GmbH, [legal info/cImpressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)

 - Code modifications are fix following the end-to-end testing done with Andrea. I added a comment in the body of the PR to indicate that. 
 - This PR contributes to both of these issues but does not close them.

bscholtes1A:
> The docs provide valuable insides into the 'hows'! Would be great to also learn more about the 'whys' and potential changes in the future (maybe as comments in the code or as part of the Readme). Especially the reasoning behind the architectural decisions (e.g. why is the contractId included in the token).

Hi @MoritzKeppler
I tried to improve a bit. Could you please check again?

extensions/data-plane-transfer/data-plane-transfer-core/src/main/java/org/eclipse/dataspaceconnector/transfer/dataplane/core/security/DataEncrypterImpl.java
```diff
@@ -0,0 +1,15 @@
+package org.eclipse.dataspaceconnector.transfer.dataplane.core.security;
+
+import org.eclipse.dataspaceconnector.transfer.dataplane.spi.security.DataEncrypter;
+
+public class DataEncrypterImpl implements DataEncrypter {
```
ndr-brt, line 5: Maybe this could be called `NoopDataEncrypter` and used as default implementation for `DataEncrypter`

extensions/data-plane-transfer/data-plane-transfer-sync/src/main/java/org/eclipse/dataspaceconnector/transfer/dataplane/sync/DataPlaneTransferSyncExtension.java
```diff
@@ -0,0 +1,153 @@
...
+     * {@link org.eclipse.dataspaceconnector.spi.types.domain.edr.EndpointDataReference} which describes the endpoint serving the data.
+     * Note that default protocol for exchange is `ids-multipart`.
+     */
+    private void registerDataFlowController(ServiceExtensionContext context) {
+        var dataFlowManager = context.getService(DataFlowManager.class);
+        dataFlowManager.register(new HttpProviderProxyDataFlowController(context.getConnectorId(), dispatcherRegistry));
+    }
+
+    private static @NotNull String getMandatorySetting(ServiceExtensionContext context, String name) {
+        var setting = context.getSetting(name, null);
```
ndr-brt, line 147: `context.getConfig().getString(name)` already throws an exception if the config is null

extensions/data-plane-transfer/data-plane-transfer-sync/src/main/java/org/eclipse/dataspaceconnector/transfer/dataplane/sync/DataPlaneTransferSyncConfiguration.java
```diff
@@ -0,0 +1,21 @@
+package org.eclipse.dataspaceconnector.transfer.dataplane.sync;
+
+import org.eclipse.dataspaceconnector.spi.EdcSetting;
+
+import java.util.concurrent.TimeUnit;
+
+public final class DataPlaneTransferSyncConfiguration {
```
ndr-brt, line 7: if it is only a constants container, could be defined as interface

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/edr/EndpointDataReferenceClaimsSchema.java
```diff
@@ -22,5 +22,6 @@
 public interface EndpointDataReferenceClaimsSchema {
     String DATA_ADDRESS_CLAIM = "dad";
     String CONTRACT_ID_CLAM = "cid";
```
ndr-brt, line 24: typo

extensions/data-plane/data-plane-spi/src/main/java/org/eclipse/dataspaceconnector/dataplane/spi/pipeline/OutputStreamDataSinkFactory.java
```diff
@@ -0,0 +1,31 @@
...
+    @Override
+    public @NotNull Result<Boolean> validate(DataFlowRequest request) {
+        return canHandle(request) ? Result.success(true) :
+                Result.failure(String.format("%s: Cannot handle destination data address with type: %s",
+                        getClass().getSimpleName(), request.getDestinationDataAddress().getType()));
+    }
+
+    @Override
+    public DataSink createSink(DataFlowRequest request) {
+        return null;
```
ndr-brt, line 29: A null-object implementation of `DataSink` could avoid potential NPE

ndr-brt (approved):
Looks good

PR #776, opened 2022-02-28
CI: Use Cosmos DB Emulator

### LINKED ISSUE #360: Integration-Tests: replace external systems with self-contained systems

algattik:
## What this PR changes/adds:

Use Cosmos DB Emulator in CI. The integration test pipeline currently runs only in upstream main, as it requires access to an Azure Cosmos DB instance. This now runs the pipeline in every PR.

## Why it does that

This allows fully validating changes that affect the Cosmos DB related extensions *before* merging to upstream main, rather than *after*.

## Further notes

One test had a logical issue due to incomplete data cleanup, and was fixed.

Pipeline was run 100 times without failures.

Use gradle-build-action as it does a better job of caching than setup-java, to accelerate the pipeline in most runs (from 12 min without cache to 6 min)

## Linked Issue(s)

Closes #360

## Checklist

### DIFF
--- a/.github/workflows/integrationtests.yaml
+++ b/.github/workflows/integrationtests.yaml
@@ -1,9 +1,6 @@
 name: Integration Tests
 
-on:
-  push:
-    branches: [ main ]
-  pull_request_target:
+on: [ push, pull_request ]
 
 jobs:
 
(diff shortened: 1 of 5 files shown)


ndr-brt (approved):
That's an huge improvement! Thanks

paullatzelsperger (approved):
This should work as intended, but on a semi-related topic I think we should substitute using the `accessPolicyId` as partitionKey in the `ContractDefinitionDocument` with an explicit one. Will create an issue for it.

PR #784, opened 2022-03-01
CosmosDB: Use static partition key for all `CosmosDocuments`

### LINKED ISSUE #780: CosmodDB: refactor partition keys
Currently we have several different approaches for partition keys throughout the code base, e.g. the `ContractDefinitionDocument` currently (ab-)uses the `accessPolicyId` as partition key, others use a configured value, etc.

Although the [official documentation](https://docs.microsoft.com/en-us/azure/cosmos-db/partitioning-overview#choose-partitionkey) (thanks @algattik for the link) states it is good practice to use an item's ID as partition key, we **cannot** do that, because currently there is no way to execute stored procedures across logical partitions.

Partition keys need to be harmonized so that all `CosmosDocument`s use a static partition key.

_Should be implemented after #776 is merged._

**Downstream projects may want to recreate their cosmos-db containers**

paullatzelsperger:
## What this PR changes/adds:

Harmonizes the usage of partition keys across all subclasses of `CosmosDocument`: partition keys must be static and should be passed in from "the outside", e.g. the connector ID or a configured value. 
This effectively causes all documents to be put into a single logical partition. 

## Why it does that

There was a variety of approaches to partition keys, e.g. the `ContractDefinitionDocument` used an `accessPolicyId`. 

This could produce unexpected results, even bugs, e.g. updating the access policy would move the document into another logical partition, even causing it to get ignored by the SPROC.

## Further notes

- Our initial plan to use the document ID turned out to be impossible, because there is no way to execute stored procedures across logical partitions, specifically the `nextForState` SPROC would then become useless.
- I-tests have been reverted to creating and deleting containers in the `@BeforeEach` and `@AfterEach` respectively, because deleting by partition key only seems to be available in the emulator, not in the actual instance.
- In practice, only subclasses of `LeaseableCosmosDocument` neet to have a shared partition key, because only those are currently queried with `nextForState`. However from a modelling perspective the fact that a document is "leasable" and it being queryable with an SPROC seem to be orthogonal concerns. This might be something we could improve on in the future should the need arise.

## Linked Issue(s)

Closes #780 

## Checklist

### DIFF
--- a/extensions/azure/cosmos/contract-definition-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/definition/store/model/ContractDefinitionDocument.java
+++ b/extensions/azure/cosmos/contract-definition-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/definition/store/model/ContractDefinitionDocument.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2020, 2021 Microsoft Corporation
+ *  Copyright (c) 2020 - 2022 Microsoft Corporation
  *
  *  This program and the accompanying materials are made available under the
  *  terms of the Apache License, Version 2.0 which is available at
@@ -24,9 +24,10 @@ import org.eclipse.dataspaceconnector.spi.types.domain.contract.offer.ContractDe
 public class ContractDefinitionDocument extends CosmosDocument<ContractDefinition> {
 
     @JsonCreator
-    public ContractDefinitionDocument(@JsonProperty("wrappedInstance") ContractDefinition contractDefinition) {
+    public ContractDefinitionDocument(@JsonProperty("wrappedInstance") ContractDefinition contractDefinition,
+                                      @JsonProperty("partitionKey") String partitionKey) {
         //todo: lets think about whether this a good partition key
-        super(contractDefinition, contractDefinition.getAccessPolicy().getUid());
+        super(contractDefinition, partitionKey);
     }
 
 
--- a/extensions/azure/cosmos/contract-definition-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/definition/store/model/ContractDefinitionDocument.java
+++ b/extensions/azure/cosmos/contract-definition-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/definition/store/model/ContractDefinitionDocument.java
@@ -26,7 +26,6 @@ public class ContractDefinitionDocument extends CosmosDocument<ContractDefinitio
     @JsonCreator
     public ContractDefinitionDocument(@JsonProperty("wrappedInstance") ContractDefinition contractDefinition,
                                       @JsonProperty("partitionKey") String partitionKey) {
-        //todo: lets think about whether this a good partition key
         super(contractDefinition, partitionKey);
     }
 
--- a/extensions/azure/cosmos/contract-definition-store-cosmos/src/test/java/org/eclipse/dataspaceconnector/contract/definition/store/TestFunctions.java
+++ b/extensions/azure/cosmos/contract-definition-store-cosmos/src/test/java/org/eclipse/dataspaceconnector/contract/definition/store/TestFunctions.java
@@ -23,13 +23,14 @@ import java.util.UUID;
 
 public class TestFunctions {
 
-    public static final String PARTITION_KEY = "test-ap-id1";
+    public static final String ACCESS_POLICY_ID = "test-ap-id1";
+    public static final String CONTRACT_POLICY_ID = "test-cp-id1";
 
     public static ContractDefinition generateDefinition() {
         return ContractDefinition.Builder.newInstance()
                 .id(UUID.randomUUID().toString())
-                .contractPolicy(Policy.Builder.newInstance().id("test-cp-id1").build())
-                .accessPolicy(Policy.Builder.newInstance().id(PARTITION_KEY).build())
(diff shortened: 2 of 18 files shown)


paullatzelsperger:
> I noticed that `CosmosTransferProcessStore` gets the partitionKey from the configuration, I was wondering if it could be the case to uniform this. Personally I would remove it from the configuration and adhere to the convention to set it as the `connectorId` value, since having it configurable does not bring any useful feature, since it's static for all the documents.

How about this: let's use the connector id by default, and add the possibility to override with a config value? That would give us the option to have a couple of connector instances work "together", even if they don't share a common ID as they would in a cluster.

paullatzelsperger:
> What's the reason to use the connectorId here? Sharing a collection across applications seems like an antipattern. It seems to also waste storage space, and potentially make migrations more complex.

How else would we have multiple connector instances work on the same collection of e.g. transfer processes? 
I'm beginning to think stored procedures may not be the ideal tool for this.

algattik:
> How else would we have multiple connector instances work on the same collection of e.g. transfer processes?

Yes that's what I think would be an antipattern. Applications should not share a (logical) database. If you want two instances to share one Cosmos DB instance for admin/cost reasons, use different *collections*.

paullatzelsperger:
> Yes that's what I think would be an antipattern. Applications should not share a (logical) database. If you want two instances to share one Cosmos DB instance for admin/cost reasons, use different _collections_.

Logically, several replicas of the EDC runtime would still be "one application", no? Specifically, two replicas of the EDC should be able to work on the same collection of `TransferProcesses`. Of course, there would be a separate collection for `ContractNegotiations`, etc.

paullatzelsperger:
@ndr-brt @algattik apart from the question, whether there is a more suitable alternative to SPROCs, which should probably be another discussion, do you see anything else that needs to be addressed?

ndr-brt (approved):

extensions/azure/cosmos/contract-definition-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/definition/store/model/ContractDefinitionDocument.java
```diff
@@ -24,9 +24,10 @@
 public class ContractDefinitionDocument extends CosmosDocument<ContractDefinition> {
 
     @JsonCreator
-    public ContractDefinitionDocument(@JsonProperty("wrappedInstance") ContractDefinition contractDefinition) {
+    public ContractDefinitionDocument(@JsonProperty("wrappedInstance") ContractDefinition contractDefinition,
+                                      @JsonProperty("partitionKey") String partitionKey) {
         //todo: lets think about whether this a good partition key
```
algattik, line 29: update this comment?

extensions/azure/cosmos/contract-negotiation-store-cosmos/src/test/java/org/eclipse/dataspaceconnector/contract/negotiation/store/CosmosContractNegotiationStoreIntegrationTest.java
```diff
@@ -77,8 +86,30 @@ static void deleteDatabase() {
...
     }
 
+    private static void uploadStoredProcedure(CosmosContainer container, String name) {
+        // upload stored procedure
+        var is = Thread.currentThread().getContextClassLoader().getResourceAsStream(name + ".js");
+        if (is == null) {
+            throw new AssertionError("The input stream referring to the " + name + " file cannot be null!");
+        }
+
+        Scanner s = new Scanner(is).useDelimiter("\\A");
```
algattik, line 96: use var throughout? (or don't use var at L91)

extensions/azure/cosmos/contract-negotiation-store-cosmos/src/test/java/org/eclipse/dataspaceconnector/contract/negotiation/store/CosmosContractNegotiationStoreIntegrationTest.java
```diff
@@ -77,8 +86,30 @@ static void deleteDatabase() {
...
+        String body = s.hasNext() ? s.next() : "";
```
algattik, line 97: do we want to upload an empty string? If yes needs comment

extensions/azure/cosmos/contract-definition-store-cosmos/src/test/java/org/eclipse/dataspaceconnector/contract/definition/store/TestFunctions.java
```diff
@@ -20,7 +34,7 @@ public static ContractDefinition generateDefinition() {
                 .build();
     }
 
-    public static ContractDefinitionDocument generateDocument() {
-        return new ContractDefinitionDocument(generateDefinition());
+    public static ContractDefinitionDocument generateDocument(String partitionKey) {
```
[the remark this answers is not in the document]
paullatzelsperger: we do, but test classes need control over the key, since they also instantiate the `CosmosContractDefinitionStore` and pulling out the `TEST_PARTITION_KEY` from `TestFunctions` feels kind of awkward.

algattik (approved):

paullatzelsperger:
the AWS Integration Test fails because of issue #795. That appears to be unrelated to this changeset, and there is already a PR for it (#842) so I think we can still merge this.

jimmarino:
Merged despite failing test as that is unrelated and will be fixed in a subsequent PR.

PR #787, opened 2022-03-02
Extract StateMachine component

ndr-brt:
## What this PR changes/adds:

Extracts a component called `StateMachineLoop` that handles the execution of the loop that takes care of the state transitions in the `TransferProcess`/`ConsumerContractNegotiation`/`ProviderContractNegotiation`

## Why it does that

Principally to avoid duplication since this pattern was used 3 times in different places, but with this we'll gain some other advantages:
* Fully tested component that will permit to handling eventually new features introduction more easily
* Graceful shutdown, that would wait for some time for letting processe before shutting down everything.
* Better readability for the `*Manager` classes.

## Further notes

I put this component in the `common` folder inside a `-lib` module because seemed like the right position. 
I remember a discussion about this in #695 but there wasn't an official statement.

## Linked Issue(s)

Relates mainly to #549 , does not resolve any issue but it will help with further developments on this topic.

## Checklist


jimmarino:
Looping in @algattik here as well because we had a discussion about leveraging state machines for certain data sink implementations in the DPF. This is really good stuff and I think putting it in a library is the right way to do it since it will allow sharing regardless of the runtime (e.g. Connector, DPF, etc.)

common/state-machine-lib/src/main/java/org/eclipse/dataspaceconnector/common/statemachine/EntitiesProcessor.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package org.eclipse.dataspaceconnector.common.statemachine;
+
+/**
+ * Interface that declares an abstraction for a component that process some entities and return the number of the processed ones.
+ * Used by {@link StateMachineLoop} to decide whether to apply wait strategy in loop iteration
+ *
+ */
+public interface EntitiesProcessor {
```
jimmarino, line 22: Should we label this @Functional ?

jimmarino, line 22: I also wonder if we should rename this now to something more generic to reflect that it is a processing stage or step?

jimmarino (approved):
LGTM, let's talk about names but this looks great.

common/state-machine-lib/src/main/java/org/eclipse/dataspaceconnector/common/statemachine/StateMachineLoop.java
```diff
@@ -0,0 +1,149 @@
...
+
+import static java.lang.String.format;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Handles a loop that processes entities continuously.
+ * On every iteration it runs all the set processors sequentially,
+ * applying a wait strategy in the case no entities are processed on the iteration.
+ */
+public class StateMachineLoop {
```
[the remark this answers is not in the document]
ndr-brt: Done

common/state-machine-lib/src/main/java/org/eclipse/dataspaceconnector/common/statemachine/EntitiesProcessor.java
ndr-brt: At first I called it `Processor` but it seemed too generic, I agree on look for a better name for it but at the moment I don't have better ideas.

common/state-machine-lib/src/main/java/org/eclipse/dataspaceconnector/common/statemachine/StateMachine.java
```diff
@@ -0,0 +1,149 @@
...
+     *
+     * @return true if it's active, false otherwise
+     */
+    public boolean isActive() {
+        return active.get();
+    }
+
+    private Runnable loop() {
+        return () -> {
+            while (active.get()) {
```
algattik, line 97: nit: Just FYI, for #790, we will have to change this to use scheduled executor rather than endless loop.

common/state-machine-lib/src/test/java/org/eclipse/dataspaceconnector/common/statemachine/StateMachineTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Test
+    void shouldExitWithAnExceptionIfProcessorExitsWithAnUnrecoverableError() {
+        var processor = mock(EntitiesProcessor.class);
+        when(processor.run()).thenThrow(new Error("unrecoverable"));
+        var loop = StateMachine.Builder.newInstance("test", monitor, waitStrategy)
+                .processor(processor)
+                .build();
+
+        assertThat(loop.start()).succeedsWithin(1, SECONDS);
```
algattik, line 102: With parallel builds and Jacoco we might have sporadic slow tests. Is there a strong reason for such short timeouts?

algattik (approved):
> Looping in @algattik here as well because we had a discussion about leveraging state machines for certain data sink implementations in the DPF. This is really good stuff and I think putting it in a library is the right way to do it since it will allow sharing regardless of the runtime (e.g. Connector, DPF, etc.)

Indeed yesterday I was thinking "wouldn't it be nice if..." and today @ndr-brt has done it. Strong impact on technical debt. Hats off.

common/state-machine-lib/src/main/java/org/eclipse/dataspaceconnector/common/statemachine/StateMachine.java
ndr-brt: Yes, I saw it, so it will be easier to that since all this logic will be in a single class. :+1:

common/state-machine-lib/src/test/java/org/eclipse/dataspaceconnector/common/statemachine/StateMachineTest.java
ndr-brt: I keep them shorts because I felt like it's a fast operation and it shouldn't take more than a second to complete also in a slow test environment. Let's see how it will behave.

ndr-brt: and, another thing that could be improved, I had to modify the monitor implementation to have the output, I think we should have the logs enabled by default and maybe disable them on the CI

common/state-machine-lib/src/main/java/org/eclipse/dataspaceconnector/common/statemachine/EntitiesProcessor.java
jimmarino: @paullatzelsperger do you have an idea for a name? `StateMachineProcessor`?

paullatzelsperger: those are champagne problems to have :D

jimmarino: @ndr-brt looks good, can you just ping me when you are ready with this?

ndr-brt:
@jimmarino ready to go

PR #793, opened 2022-03-02
fix: replace api/ids/multipart with api/ids/data

### LINKED ISSUE #791: Fix wrong naming of ids multipart endpoint
The IDS spec defines a REST endpoint named `/api/ids/data` for processing ids multipart messages. Nevertheless, the edc currently works with `/api/ids/multipart`.

juliapampus:
## What this PR changes/adds:

The REST endpoint that processes IDS multipart messages is changed from `api/ids/multipart` to `api/ids/data`.

## Why it does that

Due to the IDS specification, other IDS components expect to send IDS multipart messages to either `ids/data` or `ids/infrastructure`. Changed the current endpoint accordingly.

## Linked Issue(s)

Closes #791 

## Checklist


paullatzelsperger (commented):
Seems you caught every occurrence. A couple of questions though:
- Could we extract the `/api/ids/data` path into a constant or a utility method?
- Does the spec allow for versioning, like `/api/v1/ids/data`? 
- Could you link the relevant part of the spec either here or in the Issue?
- Maybe for another PR: do you think we should create an explicit Jetty context alias for this?

juliapampus:
> Could we extract the `/api/ids/data` path into a constant or a utility method?

I guess this does not belong to the fix, so why not open another PR for this?

Does the spec allow for versioning, like `/api/v1/ids/data`?

I think, sure.

> Could you link the relevant part of the spec either here or in the Issue?

I wish this would be that easy, because I guess you cannot access certain links. I changed the issue. However, seems like it must not have the `ids` prefix. But I think this is not bad, is it?

> Maybe for another PR: do you think we should create an explicit Jetty context alias for this?

Sry for the stupid questions: For this endpoint or all IDS endpoints? Are we doing this with other APIs as well? What would be the benefit?

juliapampus:
> What prefix do you mean?

I mentioned that the specified path is `/ids/data` but after checking again I saw that it is only `/data`. But I think `/ids` + `/data` is fine, especially for distinguishing between ids and non-ids endpoints.

> The benefit is that we can expose those paths on separate jetty ports, thus allowing us to introduce separate security zones etc. It also allows us to register RequestFilters and ExceptionMappers etc. targeted at just some APIs. We do this already for the Data Management API, because we cannot have the AuthenticationRequestFilter to interfere with IDS communication, for example. We can either leave all IDS APIs on the "default" mapping, or we can create a separate one, which could have a root path of /api/ids/data

Okay, thanks for explaining. Well, maybe it makes sense to introduce an extra mapping then. However, I would not set `/api/ids/data` as root but `/api/ids/**`. This way, we automatically include a possible IDS REST API (I think this discussion is not finished yet).

paullatzelsperger:
> However, I would not set /api/ids/data as root but /api/ids/**

I'd say that depends whether IDS Multipart and a potential IDS REST implementation will use the same set of (Jetty) resources. 

Also, I'd put a version string in, thus `/api/v1/ids/**` or `/api/ids/v1/**`, depending on whether we want to increase version number for IDS APIs separately or together with all other APIs (I am in favor of "together" btw.).

juliapampus:
@paullatzelsperger Added versioning to the current endpoints. 

While doing that, I have discovered the classes `ArtifactRequestExecutor` and `DescriptionRequestExecutor` in the samples > other > commandline. Are they used anywhere? They execute POST requests on endpoints (`/api/ids/description`, `/api/ids/request`) that do not and should not exist.

paullatzelsperger:
> @paullatzelsperger Added versioning to the current endpoints.
> 
> While doing that, I have discovered the classes `ArtifactRequestExecutor` and `DescriptionRequestExecutor` in the samples > other > commandline. Are they used anywhere (not in code, checked this, but any demos or similar)? They execute POST requests on endpoints (`/api/ids/description`, `/api/ids/request`) that do not and should not exist.

Not that I know of. Seems like the slipped through the cracks when we switched to IDS Multipart. To be honest, I'd even delete the entire `:samples:other:commandline` module, it's probably out of date anyways.

paullatzelsperger (approved):

PR #794, opened 2022-03-03
System tests: Refactored from RestAssured to Gatling

### LINKED ISSUE #783: System tests for EDC core: folder structure and performance test setup
In #665 and #701 we built a system test around sample 4 to test end-to-end flow across application instances. However, system tests should have a different lifecycle than samples. They should be able to be scaled up for performance testing. Samples should remain as simple as possible for new users to understand the behaviour of EDC.

- Move system test out of the samples directory. Create new projects under root `system-tests` directory. System tests should use their own set of `launchers` to configure EDC applications to run (e.g. sample 4 consumer EDC instance and provider EDC instance). Do not "pollute" the root  `launchers` directory with system-test specific stuff.
- Wrap client call with Gatling to build an illustrative performance test set up. The load test should run as fast as possible (i.e. for a single query), but be configurable to run higher load tests manually.

Note: DPF is not yet integrated in the system test.

Note: there is no new test code or logic to write (except to bind into Gatling).

algattik:
## What this PR changes/adds:

- Moved system tests out of the samples directory, and made them independent of sample code
- Decision record on Gatling for performance and functional tests
- Refactored RestAssured tests into Gatling
 
## Why it does that

As explained in the decision record, Gatling allows scaling up performance tests later on.

## Further notes

Gatling uses SLF4J. I had to do to a fix in the EDC SLF4J implementation so that it works when not initialised through the EDC runtime.

Test runtime controllers were copied from sample 4.0.

## Linked Issue(s)

Closes #783 

## Checklist


paullatzelsperger (approved):
Very cool stuff! Once we're actually using load/stress tests, it might make sense to use a gradle task for that, so that we can run it only at specific times like a nightly build etc.

PR #797, opened 2022-03-03
DataManagementApi: merges remaining ControlApi endpoints into DMgmt

### LINKED ISSUE #628: ControlAPI: Refactoring
# Refactoring Control API

The following changes are proposed to incorporate the current control API into the data management API:

* -[x] Create a new `ContractController` in `:extensions:api:data-management:contract-agreement`
   - `GET /contracts/{id}`: returns a `ContractAgreementDto` for the given ID
   - `GET /contracts`: returns all contract agreements, includes paging

* -[x] add the following methods to the `ContractNegotiationController`:
   - `GET /contractnegotiations/{id}/agreement`: returns `ContractAgreementDto` if one exists for the given negotiation
   - `POST /contractnegotiations`: initiates a new contract negotiation accepting a `NegotiationInitiateDto` in the body
   
* -[x] add the following methods to the `AssetController`:
   - `POST /asset/{id}/transfer`: creates a transfer request for an asset, i.e. initiates a data transfer, accepts a `TransferRequestDto`

## DTOs
- `ContractAgreementDto`: same as `ContractAgreement` BO, but carries an `assetId` and a `policyId` instead of the actual objects
- `NegotiationInitiateDto`: contains only:
  - `connectorAddress`: e.g. `<URL>/api/v1/ids/multipart`
  - `protocol`: nullable, must be `"ids-multipart"`. we can also leave it out.
  - `connectorId`: ids-ID of the initiating connector
  - `offerId`: ID of the contract offer
- `TransferRequestDto`: is essentially the same as a `DataRequest` minus the Asset ID, since that's already in the path.

## Further notes
for compatibility with downstream projects, playgrounds, demos etc. the current Control API should remain in place for a time, but the code will be marked deprecated

paullatzelsperger:
## What this PR changes/adds:

Moves the remaining endpoints, that were previously located in the `:extensions:api:control` module, into the DataManagement API. Essentially this adds 4 capabilities:
- initiate contract negotiation
- initiate data transfer
- get contract for negotiation (if exists)
- get contract

## Why it does that

We want to retire the Control API and move all its functionality into the Data Management API to avoid scattering the APIs.
Also this will then provide a common approach to authentication, error handling and deployment.

## Further notes

- the `ContractAgreementDto` exists twice, once in `extensions:api:data-management:contractnegotiation` and once in `extensions:api:data-management:contractagreement`. This is _not_ an oversight, I did this on purpose, because a) I simply wanted to avoid creating a separate model lib for just this class and b) different use cases warrant separate DTOs. Also, the earlier one has the `negotiationId` property to indicate its relation to a negotiation.
- Downstream projects that still use the "old" Control API should make the switch ASAP!

## Linked Issue(s)

Closes #628 

## Checklist


extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/AssetApiExtension.java
```diff
@@ -1,14 +1,15 @@
 /*
- * Copyright (c) 2022 ZF Friedrichshafen AG
+ *  Copyright (c) 2020 - 2022 Microsoft Corporation
```
DominikPinsel, line 2: Maybe just list all the copyright holders in a comma separated list

paullatzelsperger: whups, thanks for pointing out. auto save actions...

ndr-brt (approved):

juliapampus (approved):

Issue #814, opened 2022-03-04
DataManagementApi: implement soft-delete


paullatzelsperger:
All relevant entities mentioned in #811 should have a `deleted` field marking them such. 
Deleting an entity will then set the flag to `true` instead of purging it from the store, causing all queries to not include that entity anymore.

This involves a significant amount of work:
- introduce a `Deletable` interface providing the `isDeleted()` flag, which all aforementioned entities/business-objects must implement
- adapt all data stores and the queries inside them to not include entities marked as deleted.
- adapt and extend all tests
- possibly add queries to all stores allowing to specifically query deleted items

_This is not yet ready for development - further architecting is required!_

ndr-brt:
Question. It's clear that the `query(QuerySpec)` method should not return the deleted object (unless explicitly specified e.g. through a query param), but what about the `findById(String)`?
Probably it will return the entity that will be marked as deleted. This question is to understand how to handle deletion. E.g. an asset that is referenced in a contract agreement could be soft deleted? I think so, because from the deletion onwards it won't be accessible anymore from any query, but the agreement will keep to be valid.
Otherwise the Asset could not be deleted until the agreement is still valid (or not deleted).

PR #816, opened 2022-03-04
Feature: SQL contract definition

### LINKED ISSUE #460: Extension - ContractDefinitionStore - Integrate the PostgreSQL library
Issue #438 provided a library that supports PostgreSQL persistence. Integrate this library in a new `ContractDefinitionStore` extension.

bcronin90:
## What this PR changes/adds

Adds an implementation for ContractDefinitionStore, ContractDefinitionLoader via SQL based on a ContractDefinitionSchema.

## Why it does that

To allow for SQL persistence of Contract Definitions.

## Further notes

none

## Linked Issue(s)

Closes #460 #461

Brendan Cronin brendan.cronin@mercedes-benz.com, Daimler TSS GmbH, [legal info/Impressum](https://github.com/mercedes-benz/daimler-foss/blob/master/LEGAL_IMPRINT.md)


bcronin90:
eclipsefdn/eca failed due to our corporate email change. Should be solved now.

bcronin90:
@jimmarino @paullatzelsperger Does #542 impact the requirements of this PR in any way?

paullatzelsperger:
@bcronin90 eventually the functionality will change, but don't worry about that now. #814 isn't yet ready for development, and it might not make it into Milestone 3.

extensions/sql/contract-definition/loader/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoader.java
```diff
@@ -0,0 +1,118 @@
...
+
+import java.sql.Connection;
+import java.util.Objects;
+import javax.sql.DataSource;
+
+import static org.eclipse.dataspaceconnector.sql.SqlQueryExecutor.executeQuery;
+
+public class SqlContractDefinitionLoader implements ContractDefinitionLoader {
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
paullatzelsperger, line 33: should be passed in through the constructor since we're using a centrally configured OM

extensions/sql/contract-definition/schema/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/schema/SqlContractDefinitionTables.java
```diff
@@ -0,0 +1,23 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Daimler TSS GmbH - Initial API and Implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.sql.contractdefinition.schema;
+
+public final class SqlContractDefinitionTables {
```
paullatzelsperger, line 17: Minor nit: typically we're using interfaces to `store` constants, you're already doing it for config keys :)

extensions/sql/contract-definition/loader/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoader.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    private String writeObject(Object object) {
+        try {
+            String className = object.getClass().getName();
+            String content = OBJECT_MAPPER.writeValueAsString(object);
+
+            Envelope envelope = new Envelope(className, content);
+
+            return OBJECT_MAPPER.writeValueAsString(envelope);
```
paullatzelsperger, line 74: why are you wrapping JSON inside JSON? First, I don't understand the need for having the classnames in the database, and second, having JSON in a STRING field would probably complicate things like indexing and (full-text) searching. 
And even if you were hell-bent on having JSON in the database, why not just use a `Map<String, Object` with one entry, and serialize that?

extensions/sql/contract-definition/loader/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/LazyDataSource.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.PrintWriter;
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.SQLFeatureNotSupportedException;
+import java.util.Objects;
+import java.util.function.Supplier;
+import java.util.logging.Logger;
+import javax.sql.DataSource;
+
+class LazyDataSource implements DataSource {
```
paullatzelsperger, line 26: Is there any particular reason why you would want the DataSource lazily initialized? Also, since it actually _wraps_ a datasource, it should be named `LazyDataSourceWrapper` to make this visible.

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/Limit.java
```diff
@@ -0,0 +1,72 @@
...
+ *  Contributors:
+ *       Daimler TSS GmbH - Initial API and Implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.sql.contractdefinition.store;
+
+import org.jetbrains.annotations.Nullable;
+
+class Limit {
```
paullatzelsperger, line 19: That looks like a class we could move next to and make use of in the `QuerySpec`, WDYT?

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
```diff
@@ -0,0 +1,244 @@
...
+        } catch (Exception exception) {
+            throw new EdcPersistenceException(exception);
+        }
+    }
+
+    @Override
+    public void save(Collection<ContractDefinition> definitions) {
+        Objects.requireNonNull(definitions);
+
+        String query = String.format("INSERT INTO %s (%s, %s, %s, %s) VALUES (?, ?, ?, ?)",
```
paullatzelsperger, line 100: Nit: Seems like refactoring the query into a constant somewhere will make it more impervious to refactoring, since you're using it here and in the `*loader`.

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
```diff
@@ -0,0 +1,244 @@
...
+                SqlContractDefinitionTables.CONTRACT_DEFINITION_COLUMN_SELECTOR);
+
+        transactionContext.execute(() -> {
+            try (Connection connection = dataSource.getConnection()) {
+                definitions.forEach((definition) -> executeQuery(connection, query,
+                        definition.getId(),
+                        writeObject(definition.getAccessPolicy()),
+                        writeObject(definition.getContractPolicy()),
+                        writeObject(definition.getSelectorExpression())));
+            } catch (Exception exception) {
```
paullatzelsperger, line 114: Unneccessary, because the `SqlQueryExecutor` already wraps every `Exception` in an `EdcPersistenceException`

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
```diff
@@ -0,0 +1,244 @@
...
+                throw new EdcPersistenceException(exception);
+            }
+        });
+    }
+
+    @Override
+    public void save(ContractDefinition definition) {
```
paullatzelsperger, line 121: Why not just call `save(Collections.singletonList(definition))`?

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
```diff
@@ -0,0 +1,244 @@
...
+            try (Connection connection = dataSource.getConnection()) {
+                executeQuery(connection, query,
+                        id);
+            } catch (Exception exception) {
+                throw new EdcPersistenceException(exception);
+            }
+        });
+    }
+
+    private String writeObject(Object object) {
```
paullatzelsperger, line 188: see `SqlContractDefinitionLoader`

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
```diff
@@ -0,0 +1,244 @@
...
+                    .selectorExpression(readObject(resultSet.getString(SqlContractDefinitionTables.CONTRACT_DEFINITION_COLUMN_SELECTOR)))
+                    .build();
+        }
+    }
+
+    private static final class Envelope {
+        private String className;
+        private String content;
+
+        public Envelope() {
```
paullatzelsperger, line 219: not needed. Jackson can use any constructor with `@JsonCreator`. But as I mentioned, I think the `Envelope` class in itself is superfluous.

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
```diff
@@ -0,0 +1,244 @@
...
+
+            Envelope envelope = new Envelope(className, content);
+
+            return OBJECT_MAPPER.writeValueAsString(envelope);
+        } catch (Exception e) {
+            throw new EdcException(e.getMessage(), e);
+        }
+    }
+
+    enum ContractDefinitionMapper implements ResultSetMapper<ContractDefinition> {
```
paullatzelsperger, line 201: This can be converted into a simple method and a method reference in line 67 and 88 which makes the code much cleaner.

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStoreServiceExtension.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Inject
+    private DataSourceRegistry dataSourceRegistry;
+
+    @Inject
+    private TransactionContext transactionContext;
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        String dataSourceName = context.getConfig().getString(ConfigurationKeys.DATASOURCE_NAME);
```
paullatzelsperger, line 38: micro-nit: use var

paullatzelsperger (changes requested):
There are a few things that can be improved (see code), other than that it looks promising. Also, please use `var` wherever possible. The only real concerns I have is regarding flyway, and they are two-fold:
1. The flyway-extension shouldn't be inside the `contract-definition` directory. Seems like that feature is independent of the `ContractDefinition`, it rather cuts across all SQL data stores. 
2. Using Flyway **must** be optional. Some people may want to use alternatives like Liquibase. Renaming the `schema` module to `extensions/sql/migration/flyway` would make this possible and more visible at the same time.

extensions/sql/contract-definition/loader/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoaderTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+        String query = String.format("SELECT count(*) FROM %s WHERE %s=?",
+                SqlContractDefinitionTables.CONTRACT_DEFINITION_TABLE,
+                SqlContractDefinitionTables.CONTRACT_DEFINITION_COLUMN_ID);
+
+        Long definitionCount = executeQuery(getDataSourceRegistry().resolve(DATASOURCE_NAME).getConnection(),
+                (rs) -> rs.getLong(1),
+                query, "id").iterator().next();
+
+        Assertions.assertEquals(1, definitionCount);
```
paullatzelsperger, line 98: Just as a side note, we're using AssertJ, that would make this `assertThat(definitionCount).isEqualTo(1)`

extensions/sql/contract-definition/store/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/AbstractSqlContractDefinitionStoreTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.eclipse.dataspaceconnector.spi.transaction.TransactionContext;
+import org.eclipse.dataspaceconnector.spi.transaction.datasource.DataSourceRegistry;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+@ExtendWith(EdcExtension.class)
+public class AbstractSqlContractDefinitionStoreTest {
```
paullatzelsperger, line 30: Is there much use of having an abstract base class if there's only one inheritor?

bcronin90:
> There are a few things that can be improved (see code), other than that it looks promising. Also, please use `var` wherever possible. The only real concerns I have is regarding flyway, and they are two-fold:
> 
>     1. The flyway-extension shouldn't be inside the `contract-definition` directory. Seems like that feature is independent of the `ContractDefinition`, it rather cuts across all SQL data stores.
> 
>     2. Using Flyway **must** be optional. Some people may want to use alternatives like Liquibase. Renaming the `schema` module to `extensions/sql/migration/flyway` would make this possible and more visible at the same time.

I'm adapting all the smaller changes right now. On flyway though:
1) We've kept the flyway migrations in each separate sql extension to allow for separate migrations for each schema.
2) The (extension currently called) schema is a dependency for the other sql extensions (in this case loader and store). So if we want to generalize this we probably have to define an SPI for it, that is then implemented via flyway, right?

jimmarino:
> eclipsefdn/eca failed due to our corporate email change. Should be solved now.

It still looks like it is failing

jimmarino (changes requested):
When PRs introduce significant new libraries and tools such as Flyway, let's make sure they detail why they do so and the design decisions behind them. That makes it easier for reviewers when they look at a PR and also serves as a technical record that can be later added to the documentation. 

I'm all for database migration tools but I think we need to tackle this holistically and include them as distinct tooling for operators (i.e. not at startup in the same extension). Can you please remove all of the Flyway-related code and we can tackle database migration for a later milestone?

extensions/sql/contract-definition/schema/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+    api(project(":extensions:transaction:transaction-datasource-spi"))
+    api(project(":extensions:transaction:transaction-spi"))
+
+    implementation("org.flywaydb:flyway-core:${flywayVersion}")
+
+    testImplementation(testFixtures(project(":launchers:junit")))
+    testImplementation(project(":core:base"))
+    testImplementation(project(":extensions:sql:pool:apache-commons-pool"))
+    testImplementation(project(":extensions:transaction:transaction-local"))
+    testImplementation("com.h2database:h2:2.1.210")
```
jimmarino, line 34: Please put versions in the Gradle properties file

extensions/sql/contract-definition/loader/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoaderServiceExtension.java
```diff
@@ -0,0 +1,47 @@
...
+    private DataSourceRegistry dataSourceRegistry;
+
+    @Inject
+    private TransactionContext transactionContext;
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        String dataSourceName = context.getConfig().getString(ConfigurationKeys.DATASOURCE_NAME);
+
+        DataSource dataSource = new LazyDataSource(() -> Objects.requireNonNull(
```
jimmarino, line 39: Please remove the use of LazyDataSource and just pass the registry. Adding wrappers makes the code more difficult to follow and the LazyDataSource is just a wrapper.

extensions/sql/contract-definition/schema/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/schema/ConfigurationKeys.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Defines configuration keys used by the SqlContractDefinitionSchemaServiceExtension.
+ */
+public interface ConfigurationKeys {
+
+    /**
+     * Name of the datasource to use for accessing contract definitions.
+     */
+    @EdcSetting(required = true)
```
jimmarino, line 27: This is duplicated in the other module

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/LazyDataSource.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.PrintWriter;
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.SQLFeatureNotSupportedException;
+import java.util.Objects;
+import java.util.function.Supplier;
+import java.util.logging.Logger;
+import javax.sql.DataSource;
+
+class LazyDataSource implements DataSource {
```
jimmarino, line 26: Please remove the use of these wrappers.

extensions/sql/contract-definition/store/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/AbstractSqlContractDefinitionStoreTest.java
```diff
@@ -0,0 +1,58 @@
...
+    private final AtomicReference<ServiceExtensionContext> contextRef = new AtomicReference<>();
```
jimmarino, line 31: Why is this needed, particularly the atomic reference and methods below? If it is needed, volatile will avoid the extra code.

extensions/sql/contract-definition/store/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+plugins {
+    `java-library`
+    `java-test-fixtures`
+    `maven-publish`
+}
+
+dependencies {
+    api(project(":spi"))
```
jimmarino, line 22: I don't think this module needs to rely on the entire core SPI and expose it to downstream modules

extensions/sql/contract-definition/store/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+    api(project(":extensions:dataloading"))
+    api(project(":extensions:transaction:transaction-spi"))
+    api(project(":extensions:transaction:transaction-datasource-spi"))
+
```
jimmarino, line 26: Do all the above need to be API dependencies? Why not impl?

extensions/sql/contract-definition/store/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+    implementation(project(":extensions:sql:contract-definition:schema"))
+    implementation(project(":extensions:sql:common"))
+
+    testImplementation(testFixtures(project(":launchers:junit")))
+    testImplementation(project(":core:base"))
+    testImplementation(project(":extensions:sql:pool:apache-commons-pool"))
+    testImplementation(project(":extensions:transaction:transaction-local"))
+    testImplementation("com.h2database:h2:2.1.210")
```
jimmarino, line 34: Remove version and use properties file

extensions/sql/contract-definition/loader/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/ConfigurationKeys.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Defines configuration keys used by the SqlContractDefinitionLoaderServiceExtension.
+ */
+public interface ConfigurationKeys {
+
+    /**
+     * Name of the datasource to use for accessing contract definitions.
+     */
+    @EdcSetting(required = true)
+    String DATASOURCE_NAME = "edc.contractdefinition.datasource.name";
```
jimmarino, line 28: This is repeated

extensions/sql/contract-definition/loader/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/AbstractSqlContractDefinitionLoaderTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+@ExtendWith(EdcExtension.class)
+class AbstractSqlContractDefinitionLoaderTest {
+
+    private final AtomicReference<ServiceExtensionContext> contextRef = new AtomicReference<>();
```
jimmarino, line 32: See my comments on the similart class in store.

extensions/sql/contract-definition/loader/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+plugins {
+    `java-library`
+    `java-test-fixtures`
+    `maven-publish`
+}
+
+dependencies {
+    api(project(":spi"))
```
jimmarino, line 22: See my comments in the build file in the store module

extensions/sql/contract-definition/loader/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoader.java
```diff
@@ -0,0 +1,118 @@
...
+                SqlContractDefinitionTables.CONTRACT_DEFINITION_COLUMN_SELECTOR);
+
+        transactionContext.execute(() -> {
+            try (Connection connection = dataSource.getConnection()) {
+                executeQuery(connection, query,
+                        definition.getId(),
+                        writeObject(definition.getAccessPolicy()),
+                        writeObject(definition.getContractPolicy()),
+                        writeObject(definition.getSelectorExpression()));
+            } catch (Exception exception) {
```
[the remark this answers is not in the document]
denisneuling: The `java.sql.DataSource` throws a `javaxsql.SQLException` which needs to be catched here.

extensions/sql/contract-definition/loader/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoader.java
denisneuling: Since the central ObjectMapper might be configured to serialize data in a way to be incompatible to what's expected to be written/read into/from database, I'd argue for have a dedicated **private** ObjectMapper instance here. Also the ContractDefinitionLoader/Store abstracts away the burden to take care of how data is serialized.

extensions/sql/contract-definition/schema/src/main/resources/org/eclipse/dataspaceconnector/sql/contractdefinition/schema/V0_0_0__Init_Database_Schema.sql
```diff
@@ -0,0 +1,37 @@
...
+-- table: edc_contract_definitions
+CREATE TABLE IF NOT EXISTS edc_contract_definitions
+(
+    contract_definition_id VARCHAR
+(
+    255
+) NOT NULL,
+    access_policy VARCHAR
+(
+    65535
```
denisneuling, line 23: Instead of `VARCHAR(65535)` why not simply using `VARCHAR(MAX)`?

extensions/sql/contract-definition/loader/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoader.java
paullatzelsperger: can you give an example of such an incompatibility? the central OR mapper uses `@JsonType`s for example. 

Apart from that i don't see any sense in serializing objects such as contract definitions to JSON, only to store them in a relational database. The discussion is therefore moot and the ObjectMapper should go away altogether.

extensions/sql/contract-definition/loader/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoader.java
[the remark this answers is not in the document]
paullatzelsperger: > The (extension currently called) schema is a dependency for the other sql extensions (in this case loader and store). So if we want to generalize this we probably have to define an SPI for it, that is then implemented via flyway, right?

If that's the case you have a design problem: the other sql extensions should never depend on a specifc tool like flyway. Furthermore, a schema and a migration tool are orthogonal concerns and should not be in the same extension. All code pertaining to a schema (table name constants, etc.) could be put into an `-spi`.

paullatzelsperger:
@bcronin90 
> The (extension currently called) schema is a dependency for the other sql extensions (in this case loader and store). So if we want to generalize this we probably have to define an SPI for it, that is then implemented via flyway, right?

If that's the case you have a design problem: the other sql extensions should never depend on a specifc tool like flyway particularly because a schema and a migration tool are orthogonal concerns and should not be in the same extension. All code pertaining to a schema (table name constants, etc.) could be put into an `-spi`.

extensions/sql/contract-definition/loader/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoader.java
jimmarino: The TypeManager is designed specifically for the case of polymorphic deserialization in the runtime and *must* be used by extensions for this mechanism to work. Please do not use a private ObjectMapper.

extensions/sql/contract-definition/schema/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/schema/SqlContractDefinitionSchemaServiceExtension.java
```diff
@@ -0,0 +1,70 @@
...
+    public void start() {
+        var dataSource = Objects.requireNonNull(
+                dataSourceRegistry.resolve(dataSourceName),
+                String.format("DataSource %s could not be resolved", dataSourceName));
+
+        var flywayDataSource = new FlywayDataSource(dataSource);
+        var schemaHistoryTableName = getSchemaHistoryTableName(dataSourceName);
+
+        var flyway = Flyway.configure()
+                .dataSource(flywayDataSource)
```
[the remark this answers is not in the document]
denisneuling: Nope, Flyway tries to set autocommit=true which in our case with the `org.eclipse.dataspaceconnector.transaction.local.ConnectionWrapper` would simply throw an `UnsupportedOperationException`. Workaround here would be to instantiate a new datasource just for the migration task. Anyway to my knowledge flyway isn't wanted to be introduced within that PR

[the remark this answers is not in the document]
ndr-brt: yeah, right, I added the comment then I saw this connection wrapper thing and I forgot to remove it.

extensions/sql/contract-definition/store/build.gradle.kts
jimmarino: I think API scope is appropriate when the current module exposes contained SPI classes with its public signature, otherwise it is an implementation detail.

paullatzelsperger:
side note: we now have checks in place that require a certain PR title format. in this case renaming to e.g. `Feature: SQL contract definition` should suffice.

paullatzelsperger:
In preparation for getting this mergeable: could you please mark every conversation as "resolved" if the respective issue has been addressed?
Once that's done, please rebase on main, resolve any conflicts and let reviewers know that they can re-review.

extensions/sql/contract-definition/loader/build.gradle.kts
```diff
@@ -0,0 +1,48 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Daimler TSS GmbH - initial API and implementation
+ *
+ */
+
+plugins {
+    `java-library`
+    `java-test-fixtures`
```
paullatzelsperger, line 17: this is not needed.

extensions/sql/contract-definition/store/build.gradle.kts
```diff
@@ -0,0 +1,48 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Daimler TSS GmbH - initial API and implementation
+ *
+ */
+
+plugins {
+    `java-library`
+    `java-test-fixtures`
```
paullatzelsperger, line 17: this is not needed.

paullatzelsperger (commented):
I observed a weird behavior: both the `loader` and the `store` module reported compile errors, they couldn't find classes that were in `sql:contract-definition:spi`. After some digging i found out that since both those module depend on the `sql:contract-definition:spi` AND they also depend on the global `:spi`, there is a name conflict and gradle seems to get confused about classpaths.

Therefore the quick solution is to not depend on the whole `:spi`, but rather `:spi:core-spi` and `spi:contract-spi` as per @jimmarino's comment, check this [gist](https://gist.github.com/a75d8b2b7b6ba6ec329e75a30eb0ef09)

extensions/sql/contract-definition/loader/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoaderTest.java
```diff
@@ -0,0 +1,138 @@
...
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.eclipse.dataspaceconnector.sql.SqlQueryExecutor.executeQuery;
+
+@ExtendWith(EdcExtension.class)
+class SqlContractDefinitionLoaderTest {
```
paullatzelsperger, line 49: Why wouldn't you just instantiate the `SqlContractDefinitionLoader` with explicitly instantiated parameters for `DataSource`, `TransactionContext` etc? Testing the loader against H2 should not require instantiating the `ConfigurationExtension` and spinning up the entire DI container plus registering a local extension to initialize the database.
This test has to be rewritten IMO.

extensions/sql/contract-definition/loader/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoaderTest.java
```diff
@@ -0,0 +1,138 @@
...
+                throw new RuntimeException(e);
+            }
+        });
+    }
+
+    @Test
+    @DisplayName("Context Loads, tables exist")
+    void contextLoads() throws SQLException {
+        var query = String.format("SELECT 1 FROM %s", SqlContractDefinitionTables.CONTRACT_DEFINITION_TABLE);
+        executeQuery(dataSourceRegistry.resolve(ConfigurationKeys.DATASOURCE_NAME).getConnection(), query);
```
paullatzelsperger, line 104: this test is missing an assertion. Tests without assertions are almost always pointless. You should at least assert the result of the query (empty, non-empty,...)

extensions/sql/contract-definition/loader/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/loader/SqlContractDefinitionLoaderTest.java
```diff
@@ -0,0 +1,138 @@
...
+                SqlContractDefinitionTables.CONTRACT_DEFINITION_TABLE,
+                SqlContractDefinitionTables.CONTRACT_DEFINITION_COLUMN_ID);
+
+        var definitionCount = executeQuery(
+                dataSourceRegistry.resolve(ConfigurationKeys.DATASOURCE_NAME).getConnection(),
+                (rs) -> rs.getLong(1),
+                query, "id").iterator().next();
+
+        assertThat(definitionCount).isEqualTo(1);
+    }
```
paullatzelsperger, line 137: a few test cases are not covered: `accept_alreadyExists`, `accept_differentId_sameParameters`, `accept_exceptionDuringSql", etc. Please add.

extensions/sql/contract-definition/store/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStoreTest.java
```diff
@@ -0,0 +1,214 @@
...
+            put(ConfigurationKeys.DATASOURCE_SETTING_DRIVER_CLASS, org.h2.Driver.class.getName());
+            put(ConfigurationKeys.DATASOURCE_SETTING_NAME, ConfigurationKeys.DATASOURCE_NAME);
+        }
+    };
+    private volatile DataSourceRegistry dataSourceRegistry;
+    private volatile TransactionContext transactionContext;
+    private volatile ContractDefinitionStore contractDefinitionStore;
+
+    @BeforeEach
+    void setUp(EdcExtension extension) {
```
paullatzelsperger, line 64: refer to my comment in `SqlContractDefinitionLoaderTest`.

extensions/sql/contract-definition/store/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStoreTest.java
```diff
@@ -0,0 +1,214 @@
...
+        var spec = QuerySpec.Builder.newInstance()
+                .limit(limit)
+                .offset(20)
+                .build();
+
+        var definitionsRetrieved = contractDefinitionStore.findAll(spec).collect(Collectors.toList());
+
+        assertThat(definitionsRetrieved).isNotNull();
+        assertThat(definitionsRetrieved.size()).isEqualTo(limit);
+    }
```
paullatzelsperger, line 213: again, this test misses quite a few test cases. `update_whenNotExists`, `saveAll_whenSomeExist`, `saveAll_whenAllExist`, etc.

paullatzelsperger (changes requested):
Test must be reworked, as there are fundamental issues with them.

paullatzelsperger:
Also, can you please provide the `schema.sql` file in such away that people can easily consume and use it when they set up their database?

bcronin90:
> Also, can you please provide the `schema.sql` file in such away that people can easily consume and use it when they set up their database?

Certainly! Currently the idea is to provide a readme for database setup. Alternatively we could use a migration tool to provide the initial setup, as was the case earlier. I'm open to both approaches.

paullatzelsperger:
> > Also, can you please provide the `schema.sql` file in such away that people can easily consume and use it when they set up their database?
> 
> Certainly! Currently the idea is to provide a readme for database setup. Alternatively we could use a migration tool to provide the initial setup, as was the case earlier. I'm open to both approaches.

just put the `schema.sql` file somewhere, maybe in `extensions/sql/contract-definitions/resources`, which people can use easily to configure their database. Migration tools are great but some discussion needs to happen before we decide on adding any particular one to the code base.

bcronin90:
@paullatzelsperger Ready for re-review.

Please ignore the lack of readme and schema.sql for now. Will add tomorrow.

extensions/sql/contract-definition/store/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStoreTest.java
```diff
@@ -0,0 +1,268 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    @AfterEach
+    void tearDown() throws Exception {
+        connectionPool.close();
+    }
+
+    private ContractDefinition getContractDefinition(String id, String contractId, String policyId) {
```
paullatzelsperger, line 92: nit: private methods should be at the end of the class.

extensions/sql/contract-definition/store/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStoreTest.java
```diff
@@ -0,0 +1,268 @@
...
+    void saveOne_alreadyExist() {
+        var definition = getContractDefinition("id", "contract", "policy");
+        contractDefinitionStore.save(definition);
+
+        assertThatThrownBy(() -> contractDefinitionStore.save(definition)).isInstanceOf(EdcPersistenceException.class);
+    }
+
+    @Test
+    @DisplayName("Save a single Contract Definition that is identical to an existing contract definition except for the id")
+    void saveOne_sameIdDifferentParameters() {
```
paullatzelsperger, line 152: probably `differentId...`

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStoreServiceExtension.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.eclipse.dataspaceconnector.dataloading.ContractDefinitionLoader;
+import org.eclipse.dataspaceconnector.spi.contract.offer.store.ContractDefinitionStore;
+import org.eclipse.dataspaceconnector.spi.system.Inject;
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtension;
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtensionContext;
+import org.eclipse.dataspaceconnector.spi.transaction.TransactionContext;
+import org.eclipse.dataspaceconnector.spi.transaction.datasource.DataSourceRegistry;
+
+public class SqlContractDefinitionStoreServiceExtension implements ServiceExtension {
```
paullatzelsperger, line 26: both the `ContractDefintionLoader` and the `ContractDefinitionStore` must be declared in the `@Provides` annotation. Refer to the [docs](docs/developer/dependency_resolution.md)

bcronin90:
> minor things, but once they are in, I think we're OK to merge

Should be fixed now

ndr-brt (approved):

Issue #825, opened 2022-03-07
State machine two-state transitions.


ndr-brt:

paullatzelsperger:
As a pattern there must be an `-ING` state and an `-ED` state, e.g. `PROCESSING` and `PROCESSED`. Issue #818 relates here too.

algattik:
See also https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/336#issuecomment-1060026575 which can build on this.

PR #831, opened 2022-03-07
core: apply 2-state transition pattern to TransferProcessManager

ndr-brt:
## What this PR changes/adds

Migrate all the state transition to the 2-state pattern as described in #825 for `TransferProcessManager`
(PRs for `ContractNegotiationManager`s are about to come)

## Why it does that

Use the 3-state pattern could lead to problems, described in #825

## Further notes

now all the states (except `ERROR`, `CANCELLED`, `ENDED` and `STREAMING`) have their `process*` method that permit them to be transitioned to another state.

the state `REQUESTED_ACK` has been removed, a new `REQUESTING` state has been introducted.
the state `DEPROVISIONING_REQ` has been removed.

currently, the `deprovisioning` method on the listeners is called by the `processDeprovisioning`, but it feels like it would be better to have it called by the `DeprovisionRequestHandler`

I had to implement a lease mechanism into `InMemoryTransferProcessStore`

I had to raise the `provisionWait` on `AbstractQueueProvisioner` to make `DemoPushStreamTransferTest` run on the CI, it was working locally but not on Actions. It was pretty flaky also before this PR. (e.g: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/runs/5478021065 on main branch)

## Linked Issue(s)

#825 (not ready to be closed yet)

## Checklist


core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
```diff
@@ -145,6 +151,183 @@ public void enqueueCommand(TransferProcessCommand command) {
...
+     * Process PROVISIONED transfer<br/>
+     * If CONSUMER, set it to REQUESTING, if PROVIDER initiate data transfer
+     *
+     * @param process the PROVISIONED transfer fetched
+     * @return if the transfer has been processed or not
+     */
+    @WithSpan
+    private boolean processProvisioned(TransferProcess process) {
+        if (CONSUMER == process.getType()) {
+            process.transitionRequesting();
```
algattik, line 204: nit: How about calling sendConsumerRequest here? This state seems to have only the effect of "wasting" a cycle. See https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/336#issuecomment-1060026575. This can also be done in a subsequent PR for that issue, so we keep the scope of this PR.

ndr-brt: This was done to keep the pattern of having a `*-ING` and `*-ED` state, and this helps also setting up a clearer state separation.
e.g. at the moment there's no retry pattern applied on this "request", but, when we'll introduce it, on a failure the state of the transfer process will be set to `REQUESTING` to be retried again. Without this state, we would have to set it to `PROVISIONED`, but it will become unclear and could lead to errors in the case that new states pop up between `PROVISIONED` and `REQUESTED`.
It's like a step, step before we were talking about provision, in this step we're talking about request to the provider.

ndr-brt:
@algattik the minikube test is failing, but it works locally, do you know what could be the cause?

ndr-brt:
I think understood the problem, the `FileTransferLocalSimulation` does not wait for the file to be transferred, just for the transfer process to be in completed state (in the sample 04 there's no status checker so the manager puts the process in `COMPLETED` directly without checking if the file is transferred).
I added a `StatusChecker` to verify if the transfer is completed for real

spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/transfer/TransferProcess.java
```diff
@@ -197,60 +209,52 @@ public void transitionInProgressOrStreaming() {
...
-            transition(TransferProcessStates.IN_PROGRESS, TransferProcessStates.REQUESTED, TransferProcessStates.PROVISIONED);
+            transition(IN_PROGRESS, REQUESTED, PROVISIONED);
         }
     }
 
     public void transitionStreaming() {
         if (type == Type.CONSUMER) {
             // the consumer must first transition to the request/ack states before in progress
-            transition(TransferProcessStates.STREAMING, TransferProcessStates.REQUESTED_ACK);
+            transition(STREAMING, REQUESTED);
```
paullatzelsperger, line 222: Since we're using pessimistic locking with expiring leases and assuming idempotency, I think a transition into the same state should always be allowed, except maybe for terminal states like CANCELLED, ERROR,  COMPLETED. This is probably the same for all `transition`-methods

samples/other/streaming/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/protocols/common/AbstractQueuedProvisioner.java
```diff
@@ -32,7 +32,7 @@
 
     private final AtomicBoolean active = new AtomicBoolean();
 
-    private long provisionWait = 100;
+    private long provisionWait = 5000;
```
paullatzelsperger, line 35: this would be a candidate for the "Further Notes" section in the PR description explaining why this timeout was increased.

paullatzelsperger (commented):
From your PR description:
> currently, the provisioning method on the listeners is called by the processProvisioning, but it feels like it would be better to have it called by the DeprovisionRequestHandler

Is that a typo: `provisioning` should probably be `deprovisioning`, right? Assuming that's the case that would mean that the `TransferProcessObservable` "leaks" all the way through the `ProvisionManager` into the `DeprovisionRequestHandler`, which sounds awkward

ndr-brt:
@paullatzelsperger yes, it was a typo, I fixed it and also the other remarks, thanks

ndr-brt:
ready to review/merge

samples/other/streaming/src/main/java/org/eclipse/dataspaceconnector/transfer/demo/protocols/stream/DemoTopicManager.java
```diff
@@ -49,6 +51,7 @@ public DemoTopicManager(Monitor monitor) {
         if (container == null) {
             return Result.failure("Topic not found: " + topicName);
         } else if (!container.destination.getAccessToken().equals(accessToken)) {
+            monitor.severe(format("Cannot connect to %s because accessToken is not valid: expected: %s - actual: %s", topicName, container.destination.getAccessToken(), accessToken));
```
algattik, line 54: should not log secrets?

ndr-brt: Right, removed.

ndr-brt:
@paullatzelsperger @algattik is ok for you?

extensions/in-memory/transfer-store-memory/src/main/java/org/eclipse/dataspaceconnector/transfer/store/memory/InMemoryTransferProcessStore.java
```diff
@@ -63,14 +63,44 @@ public String processIdForTransferId(String id) {
...
+                    value.removeAll(toBeLeased);
+                }
+                return value;
+            });
+
+            return toBeLeased.stream().map(TransferProcess::copy).collect(toList());
         });
     }
 
+    /*
```
paullatzelsperger, line 82: nit: remove commented code

paullatzelsperger (approved):

PR #852, opened 2022-03-08
ci: check pull request title formatting

juliapampus:
## What this PR changes/adds

Adds an action to check that the pull request title matches the conventional commit format we agreed on to use in the `CONTRIBUTING.md`.

## Why it does that

I know that the pull request title is not the final commit to the main branch, as this could be changed on squash merge. Nevertheless, tbh, no one ever changes it. And having the title correctly formatted would make it unnecessary as well. 

This check enables and simplifies the consistent naming of the commit that subsequently ends up in the main branch - just one small step towards the right direction.

## Checklist


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -17,10 +17,12 @@ Closes # <-- _insert Issue number if one exists_
 
 ## Checklist
 
+- [ ] formatted title correctly? (_see [here](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/CONTRIBUTING.md#submit-a-pull-request)
```
paullatzelsperger, line 20: we could also reference the styleguide directly, i.e. https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/styleguide.md

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -17,10 +17,12 @@ Closes # <-- _insert Issue number if one exists_
...
+      for formatting details_)
 - [ ] added appropriate tests?
 - [ ] performed checkstyle check locally?
 - [ ] added/updated copyright headers?
 - [ ] documented public classes/methods?
 - [ ] added/updated relevant documentation?
-- [ ] added relevant details to the changelog? (skip with label `no-changelog`)
-- [ ] linked GitHub project? (see the section on the right-hand side -->)
+- [ ] added relevant details to the changelog? (_skip with label `no-changelog`_)
+- [ ] linked GitHub project? (_see the section on the right-hand side -->_)
```
paullatzelsperger, line 28: can non-committers do that? also, does that not happen automatically?

.github/workflows/scan-pull-request.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+  check-pull-request-title:
+    runs-on: ubuntu-latest
+    continue-on-error: false
+    steps:
+      - uses: actions/checkout@v2
+      - uses: deepakputhraya/action-pr-title@master
+        with:
+          # Match pull request titles conventional commit syntax
+          regex: '(?<type>\w+)(?<scope>(?:\([^()\r\n]*\)|\()?(?<breaking>!)?):(?<subject>.*)?'
```
paullatzelsperger, line 29: I cannot compile regex in my head - could you give an example or two as a comment in the yaml file?

.github/PULL_REQUEST_TEMPLATE.md
juliapampus: No, they can't. And same as for issues, as described [here](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/854#discussion_r822608819). The initial assignment could be easily automated if we had one project that was mandatory. But people were raising the eventuality of having multiple ones in GitHub's project tab. So we need to come up with an automatism based on labels or pr content or something like this. As a pr can only be merged by a committer, this can be done manually for now.

.github/workflows/scan-pull-request.yaml
juliapampus: Did, is it better now?

paullatzelsperger: yup, many thanks. that sounds reasonable.

paullatzelsperger (approved):

PR #862, opened 2022-03-09
cleanup(TransferProcessStore): add `findAll` and remove unneeded methods

### LINKED ISSUE #859: Cleanup `TransferProcessStore`: add paging, remove unneeded methods.
Add method `Stream<TransferProcess> findAll(QuerySpec querySpec)` to `TransferProcessManager` to support paging .

Implementations for all implementors must be provided as well. 

Also, the following methods are currently not implemented and should be removed:
- `createData()`
- `updateData()`
- `deleteData()`
- `findData()`

paullatzelsperger:
## What this PR changes/adds

1. Adds a `findAll(QuerySpec spec)` method, which returns all TPs covered by a specific query
2. Removes a couple of methods, that were not implemented anywhere. 

## Why it does that

ad 1. This is necessary because the Data Management API allows for querying TPs. This impl apparently slipped through the cracks in Milestone 2.
ad 2. The methods that were removed always threw an `UnsupportedOperationException`. Better not to have dead code.

## Further notes
n/a

## Linked Issue(s)

Closes #859 

## Checklist

### DIFF
--- a/extensions/in-memory/transfer-store-memory/src/main/java/org/eclipse/dataspaceconnector/transfer/store/memory/TransferProcessPredicateConverter.java
+++ b/extensions/in-memory/transfer-store-memory/src/main/java/org/eclipse/dataspaceconnector/transfer/store/memory/TransferProcessPredicateConverter.java
@@ -1,0 +1,34 @@
+/*
+ *  Copyright (c) 2020 - 2022 Microsoft Corporation
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.transfer.store.memory;
+
+import org.eclipse.dataspaceconnector.spi.query.BaseCriterionToPredicateConverter;
+import org.eclipse.dataspaceconnector.spi.query.Criterion;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcess;
+
+import java.util.function.Predicate;
+
+import static org.eclipse.dataspaceconnector.common.reflection.ReflectionUtil.getFieldValueSilent;
+
+/**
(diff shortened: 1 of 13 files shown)


extensions/in-memory/transfer-store-memory/src/main/java/org/eclipse/dataspaceconnector/transfer/store/memory/TransferProcessPredicateConverter.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.function.Predicate;
+
+import static org.eclipse.dataspaceconnector.common.reflection.ReflectionUtil.getFieldValueSilent;
+
+/**
+ * Converts a {@link Criterion} into a {@link Predicate} for use with an {@link InMemoryTransferProcessStore}.
+ * Currently, only the "=" and "in" operators are supported (cf. {@link Criterion#getOperator()} and {@link BaseCriterionToPredicateConverter}).
+ */
+class TransferProcessPredicateConverter extends BaseCriterionToPredicateConverter<TransferProcess> {
```
ndr-brt, line 29: This implementation is equal to other 3 implementations (the two `ContractDefinitionPredicateConverter`, `ContractNegotiationPredicateConverter`). Could be generalized in a single implementation called like `ReflectiveCriterionToPredicateConverter<T>`

paullatzelsperger: I had thought of that initially, but the problem is this:
- If the `ReflectiveCriterionToPredicateConverter` goes next to `BaseCriterionToPredicateConverter`, i.e. into `core-spi`, that would introduce a dependency onto `common:util`
- Conversely, putting it into `common:util` would introduce a dependency onto `core-spi`

So unless we find a suitable module to put this into, I think the better solution (or less bad?) is to leave it as is. Or did I miss anything?

ndr-brt: uhm... yes, this is a dog chasing its own tail. I guess that it's ok to leave it like this.

ronjaquensel (approved):

PR #870, opened 2022-03-09
core: apply 2-state transition pattern to ContractNegotiationManagers

### LINKED ISSUE #825: State machine two-state transitions.

ndr-brt:
## What this PR changes/adds

Migrate all the state transition to the 2-state pattern as described in https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/825 for `ContractNegotiationManager`s

## Why it does that

Use the 3-state pattern could lead to problems, described in https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/825

## Further notes

the `CONFIRMING_SENT` status has been removed since it's not needed anymore
I had to implement a lease mechanism into the `InMemoryContractNegotiationStore` (for the cosmos implementation was already done in #724)

I add the `-quiet` parameter to the gradle test task, to avoid printing the system tests logs on the pipeline

## Linked Issue(s)

Fix #825 (supposing that #831 would be merged before this)

## Checklist

### DIFF
--- a/extensions/in-memory/negotiation-store-memory/src/main/java/org/eclipse/dataspaceconnector/negotiation/store/memory/InMemoryContractNegotiationStore.java
+++ b/extensions/in-memory/negotiation-store-memory/src/main/java/org/eclipse/dataspaceconnector/negotiation/store/memory/InMemoryContractNegotiationStore.java
@@ -110,9 +110,20 @@ public class InMemoryContractNegotiationStore implements ContractNegotiationStor
     public @NotNull List<ContractNegotiation> nextForState(int state, int max) {
         return lockManager.readLock(() -> {
             var set = stateCache.get(state);
-            return set == null ? Collections.emptyList() : set.stream()
+
+            List<ContractNegotiation> toBeLeased = set == null ? Collections.emptyList() : set.stream()
                     .sorted(Comparator.comparingLong(ContractNegotiation::getStateTimestamp)) //order by state timestamp, oldest first
                     .limit(max)
+                    .collect(toList());
+
+            stateCache.compute(state, (key, value) -> {
+                if (value != null) {
+                    value.removeAll(toBeLeased);
+                }
+                return value;
+            });
+
+            return toBeLeased.stream()
                     .map(ContractNegotiation::copy)
                     .collect(toList());
         });
--- a/launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/EdcRuntimeExtension.java
+++ b/launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/EdcRuntimeExtension.java
@@ -14,6 +14,7 @@
 package org.eclipse.dataspaceconnector.junit.launcher;
 
 import org.eclipse.dataspaceconnector.spi.EdcException;
+import org.eclipse.dataspaceconnector.spi.monitor.ConsoleMonitor;
 import org.eclipse.dataspaceconnector.spi.monitor.Monitor;
 import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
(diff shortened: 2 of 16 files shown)


extensions/in-memory/negotiation-store-memory/src/main/java/org/eclipse/dataspaceconnector/negotiation/store/memory/InMemoryContractNegotiationStore.java
```diff
@@ -110,9 +110,20 @@ public void delete(String processId) {
...
+                    .collect(toList());
+
+            stateCache.compute(state, (key, value) -> {
+                if (value != null) {
+                    value.removeAll(toBeLeased);
+                }
+                return value;
+            });
+
+            return toBeLeased.stream()
```
ronjaquensel, line 126: Wouldn't a similiar type of lease also be needed for the `CosmosContractNegotiationStore`?

launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/EdcRuntimeExtension.java
```diff
@@ -172,8 +173,7 @@ public void afterTestExecution(ExtensionContext context) throws Exception {
 
     @Override
     protected @NotNull Monitor createMonitor() {
-        return new Monitor() {
-        };
+        return new ConsoleMonitor(logPrefix, ConsoleMonitor.Level.DEBUG);
```
ronjaquensel, line 176: While this is great for local testing, should the EDC log at debug level in the pipeline?

extensions/in-memory/negotiation-store-memory/src/main/java/org/eclipse/dataspaceconnector/negotiation/store/memory/InMemoryContractNegotiationStore.java
ndr-brt: Already done in #724

launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/EdcRuntimeExtension.java
ndr-brt: You're right, I tweaked a bit and probably the best solution would be to use the console monitor unless gradle is launched with the `-quiet` parameter. I modified the code accordingly. I also noticed that the `EdcRuntimeExtension` was shutting down extensions a second time in the `@AfterAll` methods, but the shutdowns were already triggered by an hook defined in the `boot` method.
(example output: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/runs/5551002718)

ronjaquensel: I really like that solution, as it also removes the need to exchange the monitor locally in order to see the test output. And good catch on the double shutdown, never noticed that.

ndr-brt: seems like the shutdown logs are printed with another monitor, will take a look

ndr-brt: @ronjaquensel I can't get why the shutdown logs are printed... however they are printed also currently on the pipeline so... if you are ok with this could you approve? thanks

ronjaquensel (approved):

ndr-brt:
@jimmarino can this be merged?

launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/EdcRuntimeExtension.java
ndr-brt: @ronjaquensel I found out that the "duplicated shoutdown" is needed by the `DataPlaneIntegrationTests`, so for the moment I left it as it is :(

PR #876, opened 2022-03-11
core: Call the listeners before the state transition is persisted

### LINKED ISSUE #434: Architecture - Essential - "at-least-once" TransferProcessListener
Triggering custom logic on TransferProcess state changes requires registering a [TransferProcessListener](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/spi/src/main/java/org/eclipse/dataspaceconnector/spi/transfer/TransferProcessListener.java). However listeners are not guaranteed to be called, e.g. if the instance crashes after persisting the state change but before invoking the listener. Custom logic in the listener can be crucial for the functioning of client applications, thus EDC needs to guarantee that these are called before advancing a transfer process to the next state.
 
Suggestion for improvements:
* Add additional TransferProcess states to indicate that a TP has advanced state but listeners have not yet been called.
* Add processing for these new states in the TransferProcessManager, ensuring that listeners are always called.
* (Optional) Consider introducing a retry strategy for failing listeners
* (Optional) Consider executing listeners in parallel to improve performance

Izzzu:
## What this PR changes/adds

- Call the listeners before the state transition is persisted.
- Add  _confirmingSent_ method to ContractNegotiationListener

## Why it does that

A custom extension may need to perform custom processing after a data transfer has reached a certain stage.

Triggering custom logic on TransferProcess state changes requires registering a TransferProcessListener. However listeners are not guaranteed to be called, e.g. if the instance crashes after persisting the state change but before invoking the listener. Custom logic in the listener can be crucial for the functioning of client applications, thus EDC needs to guarantee that these are called before advancing a transfer process to the next state is made durable.

## Further notes

- calling observable (_observable.invokeForEach(observe)_) and saving negotiation/transferProcess were grouped in one method to reduce code duplication

## Linked Issue(s)

Closes #434 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ the detailed section referring to by linking pull requests or issues.
 * Refactor (=generify) transformer subsystem (#779)
 * Extract interfaces for every api controller class to improve swagger documentation (#891)
 * Instrument executors with metrics (#912)
+* Call the listeners before the state transition is persisted. (#876)
 
 #### Removed
 
--- a/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
+++ b/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
@@ -21,6 +21,7 @@ import org.eclipse.dataspaceconnector.common.statemachine.StateMachine;
 import org.eclipse.dataspaceconnector.common.statemachine.StateProcessorImpl;
 import org.eclipse.dataspaceconnector.contract.common.ContractId;
 import org.eclipse.dataspaceconnector.spi.contract.negotiation.ConsumerContractNegotiationManager;
+import org.eclipse.dataspaceconnector.spi.contract.negotiation.observe.ContractNegotiationListener;
 import org.eclipse.dataspaceconnector.spi.contract.negotiation.response.NegotiationResult;
 import org.eclipse.dataspaceconnector.spi.iam.ClaimToken;
 import org.eclipse.dataspaceconnector.spi.result.Result;
@@ -40,6 +41,7 @@ import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import java.util.function.Function;
 
 import static java.lang.String.format;
@@ -108,8 +110,7 @@ public class ConsumerContractNegotiationManagerImpl extends AbstractContractNego
 
         negotiation.addContractOffer(contractOffer.getContractOffer());
         negotiation.transitionInitial();
-        negotiationStore.save(negotiation);
-        observable.invokeForEach(l -> l.requesting(negotiation));
+        update(negotiation, l -> l.requesting(negotiation));
 
         monitor.debug(String.format("[Consumer] ContractNegotiation initiated. %s is now in state %s.",
                 negotiation.getId(), ContractNegotiationStates.from(negotiation.getState())));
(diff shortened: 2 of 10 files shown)


ndr-brt:
I know that there was already a discussion about. My opinion is that a pattern based on events would fit better and it will give more flexibility and scalability. However that's not the right place to talk about that, will open a discussion about this.

Talking about this PR:

If a listener throws an exception, the state will never be changed, because nobody catches exceptions on listeners, and having a listener that could interfere with the state machine it's an antipattern.
On the same page, having a synchronous listener would permit a listener to block a processing, and I would avoid this too.

Izzzu:
> I know that there was already a discussion about. My opinion is that a pattern based on events would fit better and it will give more flexibility and scalability. However that's not the right place to talk about that, will open a discussion about this.
> 
> Talking about this PR:
> 
> If a listener throws an exception, the state will never change, because nobody catches exceptions on listeners, and having a listener that could interfere with the state machine it's an antipattern. On the same page, having a synchronous listener would permit a listener to block the entity processing, and I would avoid this too.

@ndr-brt Thanks for the comment. What do you mean by "nobody catches exceptions on listeners"?  When an exception is thrown from a listener after the state is saved it's also not a good practice. Well-behaved listener shouldn't let the exceptions escape to the outside. 

@algattik any thoughts on that?

ndr-brt:
@Izzzu a listener, as the name says, reacts on an action that happens and apply another logic on it, at this moment, if a listener throws an exception, the flow of the state machine is interrupted.

@algattik probably I'm not aware of the uses of this pattern, but eventually if this is what you want, I would rename all the "listener" things in "interceptor", to avoid confusion, because this is not a listener anymore.

algattik:
We can call them "Observer" then? This follows the design pattern

ndr-brt:
I'm not sure, I've never seen an observer pattern application that observes a state change that has not happened yet and that's able to avoid it at all (through throwing exceptions). 
To me it would be pretty confusing.

algattik:
The use case we had in Catena-X was for recursively fetching parts of a vehicle from multiple providers. Every time we received a result from one, we would use a listener to enqueue further transfers down the tree.

Izzzu:
> @Izzzu a listener, as the name says, reacts on an action that happens and apply another logic on it, at this moment, if a listener throws an exception, the flow of the state machine is interrupted.

@ndr-brt I don’t think it answers my question what do you mean by "nobody catches exceptions on listeners”. Could you elaborate if it’s still relevant?
btw listener name doesn’t indicate that it reacts on something. It can be used e.g. for audit purposes.

IMO Interceptor would suggest that we want to offer a way to change (eg. filter) the input but as I understand it’s not the case.

What about making the method names more descriptive and naming them instead of
_provisioning_, _completed_ -> _preProvisioning_, _preCompleted_? 
It would indicate explicitly that it’s called before the state is saved. If needed _postProvisioning_, _postComplete_ (or just _provisioning_, _completed_ to avoid redundant postfixes) can be introduced later if there’s use case for it?

ndr-brt:
> > @Izzzu a listener, as the name says, reacts on an action that happens and apply another logic on it, at this moment, if a listener throws an exception, the flow of the state machine is interrupted.
> 
> @ndr-brt I don’t think it answers my question what do you mean by "nobody catches exceptions on listeners”. Could you elaborate if it’s still relevant? btw listener name doesn’t indicate that it reacts on something. It can be used e.g. for audit purposes.
> 
> IMO Interceptor would suggest that we want to offer a way to change (eg. filter) the input but as I understand it’s not the case.
> 
> What about making the method names more descriptive and naming them instead of _provisioning_, _completed_ -> _preProvisioning_, _preCompleted_? It would indicate explicitly that it’s called before the state is saved. If needed _postProvisioning_, _postComplete_ (or just _provisioning_, _completed_ to avoid redundant postfixes) can be introduced later if there’s use case for it?

My concern is that in this case a listener could interrupt the flow of the state machine in case of exception throwing, don't think is something we really want.
By definition, a listener "listens and reacts" to an event, but it should not interfere on the system on which it's listening on (it could be a complete different sistem too). 
Audit is another example of reacting to events: something happens, I take count of it and that's it. And, to be precise, audit doesn't work well with the "at-least-one" pattern (counting things more than one could lead to vicious results)

I could accept the decision to make the listener act before an operation actually happens (the `pre`/`post` proposal seems reasonable), but giving to it the possibility to stop the flow of the state machine seems kinda weird to me..

ndr-brt (approved):
Ok, let's merge this and keep the discussion open on other places

PR #887, opened 2022-03-14
extensions: implement s3 protocol for data plane

### LINKED ISSUE #823: Provide an S3 implementation of DataSource/DataSink for the Data Plane
As discussed in #675 , the current `DataOperator` interface and implementations will be trashed, before that, we should adapt the AWS S3 implementation to the Data Plane Framework.

ndr-brt:
## What this PR changes/adds

Implements data-plane's `DataSource` and `DataSink` for AWS S3 service.

## Why it does that

Currently there are only an http and an azure blobstorage implementations. This will give the possibility to use the dataplane with the AWS widely used storage service.

## Further notes

I introduced a `ValidationRule` concept, is a function that takes an object and return a `Result<Void>`, every its implementation could verify a simple object but also a complex one through composition.
This concept could be extended to the other data plane extension modules to build a composable validator.

I noticed that the `validate` method on the `DataSourceFactory`/`DataSinkFactory` returns a `Result<Boolean>`, but the boolean inside there has no reason to exists, because the outcome of the operation has already been described by the `Result` class itself: 
* success -> validation passed
* failed -> validation failed
Would be better for clarity to define it as a `Result<Void>`. This could be addressed in another issue.

## Linked Issue(s)

Closes #823 

## Checklist


jimmarino (approved):

nmsadek:
Hi EDC Team, 

Thank you for integrating S3 as protocol in the data plane! I highly appreciate your efforts.. 

Could you please include an S3 example in the samples (and the documentation as well), so that we can exchange data between different buckets located in the same AWS Account... 


Thanks a lot! 
Best regards,
Nizar

PR #890, opened 2022-03-14
Added: DPF Selector module and clients to access it

### LINKED ISSUE #832: DataPlaneFramework: Create DPF Selector
## Description
We need a service which we'll call "DPF Selector" that contains a list of all the available DPF instances that a connector has access to.

## Requirements

- **Data Model:** every DPF is represented as a `DataPlaneInstance` with its name and metadata such as:
  - available `DataSource` implementations
  - available `DataSink` implementations
  - configuration (URL)

  As a starting point the API of that service should look something like this:
  ```java
  public interface DpfSelector {
     DataPlaneInstance select(DataRequest request, SelectionStrategy strategy);
  }
  ```

- **Storage:** those `DataPlaneInstances` are stored in a mutable `DpfSelectorStore` (provide in-mem implementation by default), so adding and removing is possible at runtime.

- **Selection algorithm:** the DPF Selector should also be able to use different _strategies_ for DPF selection, e.g. round-robin, select-first-matching, random, etc. Regardless of strategy the selector can only return DPFs that can handle the request.

## Deployment
The DPF Selector should be able to embedded in a connector runtime or standalone, e.g. as a K8S pod. This may influence how the control plane communicates with the selector (REST vs. Java SPI) and it may also influence the choice for a storage backend.

## Out of scope
In future developments we can collect dynamic metrics such as load average, number of transfers in progress, CPU utilization, etc. and report back to the selector who in turn updates the `DataPlaneInstance` in the store.

- [x] Create technical proposal #851
- [ ] Add minimal implementation #890 

[edit] added implementation PR

paullatzelsperger:
## What this PR changes/adds

This PR adds a "DPF Selector", essentially a router for acccessing DataPlaneFramework instances. Such a "DPF Selector" can be deployed as standalone runtime (cf. `launchers/dpf-selector`) and can run embedded in the EDC code base as extension.

For either of these use cases a client is implemented:
- `EmbeddedDataPlaneSelectorClient`: when embedded in control plane
- `RemoteDataPlaneSelectorClient`: when DPF selector runs standalone and exposes a REST API

## Why it does that

If there are more than one DPF instances available, for example, each with different capabilities, there needs to be a switching point to load-balance the requests between them.

## Further notes

As a follow-up we'll also need a way to provide idempotency, which means the control plane needs to make sure that the same data request isn't processed twice by different DPF instances, so it needs a way of asking "hey, is anyone already processing this?". This is also required for crash-recovery.

This feature will be added in a later PR.

## Linked Issue(s)

Closes #832 

## Checklist


extensions/data-plane-selector/selector-spi/src/main/java/org/eclipse/dataspaceconnector/dataplane/selector/store/DataPlaneInstanceStore.java
```diff
@@ -0,0 +1,36 @@
...
+package org.eclipse.dataspaceconnector.dataplane.selector.store;
+
+import org.eclipse.dataspaceconnector.dataplane.selector.instance.DataPlaneInstance;
+
+import java.util.Collection;
+import java.util.stream.Stream;
+
+/**
+ * Holds all {@link DataPlaneInstance} objects that are known to the DPF selector.
+ * The collection of those objects is not regarded immutable, so adding at runtime is possible, implementors need to be
```
jimmarino, line 24: Super nit, I think something like the following would be clearer:
The collection of {@link DataPlaneInstance} objects is mutable at runtime, so implementations must take that into account.

jimmarino (approved):

bscholtes1A (approved):

PR #892, opened 2022-03-15
doc(decision): Decision Record for Policy Scopes

jimmarino:
## What this PR changes/adds

Adds a decision record for policy scopes.

### DIFF
--- a/docs/developer/decision-records/2022-03-15-policy-scopes/README.md
+++ b/docs/developer/decision-records/2022-03-15-policy-scopes/README.md
@@ -1,0 +1,51 @@
+# Policy Scopes
+            
+## Decision
+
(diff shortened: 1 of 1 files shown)


MoritzKeppler:
I'm fine with the proposal, sounds reasonable - but still I'm far away from understanding all consequences. 
We should avoid limiting the solution space of discussions like https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/discussions/878
But having an implementation can also help us to move the discussion forward, realize what's solved and what's missing as running code is less ambiguous.

<sub> Moritz Keppler <moritz.keppler@mercedes-benz.com>, Daimler TSS GmbH, [legal info/Impressum](https://github.com/mercedes-benz/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

juliapampus (approved):

PR #897, opened 2022-03-15
core: increase Xmx in order to build "smoothly" with several gradle daemons

jbonofre:
## What this PR changes/adds

Increase -Xmx in gradle daemons

## Why it does that

Users may have OOM while building.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -36,3 +36,5 @@ httpMockServer=5.12.0
 micrometerVersion=1.8.2
 bouncycastleVersion=1.70
 gatlingVersion=3.7.5
+
+org.gradle.jvmargs=-Xmx512m


paullatzelsperger (approved):
no replacement for displacement :)

ndr-brt:
never run into this issue, feels like it's something that needs to be configured on the local gradle configuration, more than checked into the repository (the same, in my opinion, would be for the other `org.gradle` settings about parallelism into the gradle.properties file)

ndr-brt:
@jbonofre try to comment out the other `org.gradle` settings about parallelism and see if you still get that issue. I disabled parallelism locally because it slowed my machine, probably it can cause also OOM on yours.

jbonofre:
Good point. I remember I did it for Apache Beam as well. Let me try. Thanks !

MoritzKeppler:
+1 for not including the org.gradle.workers settings as part of the checked in gradle properties.

ndr-brt:
@jbonofre this problem probably has been resolved in #961, if so, can you close this? thanks

jbonofre:
@ndr-brt agree, it looks good to me. Thanks !

Issue #865, opened 2022-03-09
SQL: implementation of `TransferProcessStore`


paullatzelsperger:
An SQL-based implementation of the `TransferProcessStore` is needed. 

### Requirements
1. Exclusive Locking: the `nextForState` method must take into account the exclusive locking mechanism required by the asynchronous nature of the `TransferProcessManager`:
    - `nextForState()` locks the TP
    - `update()` unlocks the TP
    - `find()` ignores the lock
    - there should be a timeout after which the lock is automatically released.

2. Table Schema: the `DataRequest` should be in a separate table because `TransferProcess.processIdForTransferId` will access the `DataRequest.processId` field. The other fields can be stored as JSON objects for the sake of simplicity.
3. Querying: the query implementation must be flexible enough to accomodate  a `QuerySpec`.
3. Indexing: queries will be targeting `TransferProcess.id` and  TransferProcess.dataRequest.processId`. Indexes must be created accordingly. 

- [x] Design proposal: #899 
- [x] Implementation #950 

_Please provide a written proposal for requirements 1-3 before beginning the implementation._

paullatzelsperger:
assigning @algattik since he has worked in this area before

PR #899, opened 2022-03-15
Proposal: technical concept for a `TransferProcessStore` based on SQL

paullatzelsperger:
## What this PR changes/adds

This PR adds a document outlining some key aspects of the `TransferProcessStore` based on SQL such as table schema, SQL statements. 

## Why it does that

Before the implementation of #865 can start, there needs to be a vetted proposal in place.

## Further notes

I put the file in `extensions/sql/transfer-process-store`, but once the proposal is accepted it should go into the appropriate folder in `docs/developer`

## Linked Issue(s)

Contributes to #865

## Checklist


extensions/sql/transfer-process-store/sql_tps_proposal.md
```diff
@@ -0,0 +1,133 @@
...
+    managed_resources boolean default true,
+    properties        varchar,
+    transfer_type     varchar
+);
+
+comment on column edc.data_request.data_destination is 'DataAddress serialized as JSON';
+
+comment on column edc.data_request.properties is 'java Map serialized as JSON';
+
+comment on column edc.data_request.transfer_type is 'TransferType serialized as JSON';
```
DominikPinsel, line 25: As all three objects are pretty simple (just `Strings` in a `HashMap` or `String` / `Boolean` as field) I would do the extra effort create dedicated tables for them. This also might become handy when the EDC wants to query them or migrate the schema to another version.

extensions/sql/transfer-process-store/sql_tps_proposal.md
```diff
@@ -0,0 +1,133 @@
...
+        constraint data_request___fk
+            references edc.data_request
+            on update restrict on delete restrict
+);
+
+comment on column edc.transfer_process.trace_context is 'Java Map serialized as JSON';
+
+comment on column edc.transfer_process.resource_manifest is 'java ResourceManifest serialized as JSON';
+
+comment on column edc.transfer_process.provisioned_resource_set is 'ProvisionedResourceSet serialized as JSON';
```
DominikPinsel, line 53: As all three objects are pretty simple (just `Strings` in a `HashMap` or `String` / `Boolean` as field) I would do the extra effort create dedicated tables for them. This also might become handy when the EDC wants to query them or migrate the schema to another version.

extensions/sql/transfer-process-store/sql_tps_proposal.md
```diff
@@ -0,0 +1,133 @@
...
+```
+The `skip locked` statement will prevent any subsequent processes from blocking. It will most closely mimic the behaviour of the `lease`.
+
+_Nota bene: this locks a row until the transaction is completed, see [below](#open-questions) for details._ 
+
+
+#### `create`
+inserts a new TP into the database by first inserting the `DataRequest`, obtaining it's ID and then creating the `TransferProcess` entry.
+```sql
+with drq_id as
```
DominikPinsel, line 100: I assume `WITH` is also an extension to the ANSI/ISO SQL standard.

DominikPinsel: Thanks for pointing this out, I didn't know that. Maybe our committers can decide which SQL standard or databases the EDC needs to support. I think this was never specified.

I also found this in [Wikipedia](https://en.wikipedia.org/wiki/SQL:1999):

> Recursive CTEs from the standard were relatively close to the existing implementation in [IBM DB2](https://en.wikipedia.org/wiki/IBM_DB2) version 2.[[1]](https://en.wikipedia.org/wiki/Hierarchical_and_recursive_queries_in_SQL#cite_note-JimMelton1-1) Recursive CTEs are also supported by [Microsoft SQL Server](https://en.wikipedia.org/wiki/Microsoft_SQL_Server) (since SQL Server 2008 R2),[[2]](https://en.wikipedia.org/wiki/Hierarchical_and_recursive_queries_in_SQL#cite_note-msdnRCTEs-2) [Firebird 2.1](https://en.wikipedia.org/wiki/Firebird_(database_server)),[[3]](https://en.wikipedia.org/wiki/Hierarchical_and_recursive_queries_in_SQL#cite_note-3) [PostgreSQL 8.4+](https://en.wikipedia.org/wiki/PostgreSQL),[[4]](https://en.wikipedia.org/wiki/Hierarchical_and_recursive_queries_in_SQL#cite_note-4) [SQLite 3.8.3+](https://en.wikipedia.org/wiki/SQLite),[[5]](https://en.wikipedia.org/wiki/Hierarchical_and_recursive_queries_in_SQL#cite_note-5) [IBM Informix](https://en.wikipedia.org/wiki/IBM_Informix) version 11.50+, [CUBRID](https://en.wikipedia.org/wiki/CUBRID), [MariaDB 10.2+](https://en.wikipedia.org/wiki/MariaDB) and [MySQL 8.0.1+](https://en.wikipedia.org/wiki/MySQL).

And it looks like Derby DB doesn't support it ([source](https://issues.apache.org/jira/browse/DERBY-11)).

extensions/sql/transfer-process-store/sql_tps_proposal.md
paullatzelsperger: my reasoning behind serializing them was that to my knowledge there isn't any plans to query on those fields, so it seemed to be the lowest-hanging fruit, because it makes `insert` and `delete` statements simpler.

extensions/sql/transfer-process-store/sql_tps_proposal.md
paullatzelsperger: similar reasoning as below

extensions/sql/transfer-process-store/sql_tps_proposal.md
algattik: Recursive CTL is something very special - here we have non-recursive CTE

paullatzelsperger: Since we're using user-supplied IDs (as opposed to: DB-generated ones), we don't need the `returning id` part. Therefor I don't see a problem in replacing the `WITH` with two consecutive SQL statements as long as they are executed in the same transaction. @algattik do you agree, am I missing something?

extensions/sql/transfer-process-store/ddl.sql
```diff
@@ -0,0 +1,84 @@
+create table lease
+(
+    leased_by      varchar               not null,
+    leased_at      bigint,
+    lease_duration integer default 60000 not null,
```
bcronin90, line 5: I'm not sure this is the right time to set this default. If it's not already set in the lease, you're basically changing the object when persisting.

extensions/sql/transfer-process-store/ddl.sql
```diff
@@ -0,0 +1,84 @@
+create table lease
```
bcronin90, line 1: We've prefixed all table with "edc_", so this should probably be continued here for consistency.

denisneuling:
@paullatzelsperger Same as for the [ContractNegotiationStore proposal](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/938#pullrequestreview-918422353): Would you please be so kind and define a concrete value range on each of the columns?

extensions/sql/transfer-process-store/ddl.sql
paullatzelsperger: Shouldn't we have defaults at every level? There should be defaults in Java code, at the DML level and at the DDL level, no?

bcronin90: I'd prefer if they were maintained only at the level of the actual edc object. Otherwise a default might get added/changed/removed, while another default is hidden in the SQL implementation.

paullatzelsperger: fair enough. will remove.

paullatzelsperger:
> @paullatzelsperger Same as for the [ContractNegotiationStore proposal](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/938#pullrequestreview-918422353): Would you please be so kind and define a concrete value range on each of the columns?


I did the same modifications as for the other proposal (rename files, add comments). Ok for you?

bcronin90 (approved):

PR #910, opened 2022-03-16
DPF Azure extension: Azure Data factory and cloud testing with decision records

### LINKED ISSUE #773: DPF Azure extension: support Azure Data Factory (serverless)
Serverless data transfer infrastructure is provided by major cloud providers and integration software vendors. For example, Azure Data Factory (ADF) offers a serverless way to copy data at scale in an optimised manner. As a managed service, it offers increased scalability and reduced operational load on the application when copying data.

Serverless offerings cannot replace DPF stream-based copy, since EDC users may not want a vendor dependency, and each serverless product supports a limited range of integrations. Serverless transfer can, however, be an optional module in DPF. DataFlowRequests can be routed to either serverless transfer or DPF source/sink streams, depending on rules to be defined, such as whether the serverless platform supports the source and sink, as well as transfer size.

cpeeyush:
## What this PR changes/adds

Implement DPF copy with Azure Data Factory (ADF).

Also implements cloud testing.

Two decision records are created to explain the approaches to serverless transfers and cloud testing. (The cloud testing decision record is short, because details are given in a repository doc markdown file).

## Why it does that

- Allow serverless data transfer, with large scale and effective resource usage
- Enable full testing on CI, which requires cloud resources

## Further notes

Azure SDK classes are final, thus not mockable without workarounds. Wrapper classes and deep stubs are used to structure code for SRP and testability.

## Linked Issue(s)

Closes #773 

## Checklist


extensions/data-plane/data-plane-framework/src/main/java/org/eclipse/dataspaceconnector/dataplane/framework/DataPlaneFrameworkExtension.java
```diff
@@ -83,24 +80,22 @@ public void initialize(ServiceExtensionContext context) {
         var workers = context.getSetting(WORKERS, DEFAULT_WORKERS);
         var waitTimeout = context.getSetting(WAIT_TIMEOUT, DEFAULT_WAIT_TIMEOUT);
 
+        var store = new InMemoryDataPlaneStore(IN_MEMORY_STORE_CAPACITY);
```
jimmarino, line 83: The in-memory store should be pluggable, i.e. if a store implementation is not registered during initiaize, the in-memory one will be used in start()

jimmarino:
@cpeeyush hats off to an excellent and clear summary in the DRs included in this PR! That helps reviewers tremendously. 

Outside of the test issue, I had one comment with the change to the in-memory store.

docs/developer/decision-records/2022-03-01-serverless-transfers/README.md
```diff
@@ -0,0 +1,115 @@
+# Serverless transfers in DPF
+
+## Decision
+
+Serverless mechanisms for data transfers are supported in DPF. Example of serverless implementations include Azure Data Factory, a fully managed integration service that can be controlled through a REST API.
+
```
paullatzelsperger, line 6:
```suggestion
Serverless mechanisms for data transfers are supported in DPF. Examples of serverless implementations include Azure Data Factory, a fully managed integration service that can be controlled through a REST API.
```

docs/developer/decision-records/2022-03-01-serverless-transfers/README.md
```diff
@@ -0,0 +1,115 @@
...
+- One ADF pipeline, containing a single activity to copy data from source to destination. The activity is linked to both datasets.
+-  One ADF pipeline run.
+
+The pipeline run is asynchronous, and the DPF extension should use a state machine to regularly poll for completion. In the spike, for simplicity, we implemented busy polling to provide a synchronous implementation.
+
+When using Data Factory Studio, note that pipelines created through the REST API do not appear in the pipeline designer (for manual editing). However, the runs of those pipelines do appear in the Monitor tab.
+
+### Walkthrough
+
+The DPF `PipelineService` is responsible for looking up source and sink stream implementation and using them to effect a transfer. That service was extended to pass the request to a `AzureDataFactoryTransferService` if the source and destination are both blobs. This implementation violates SRP for the sake of simplicity in the spike.
```
paullatzelsperger, line 63: please write out "SRP"

docs/developer/cloud_testing.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Planning your deployment
+
+You will need:
+
+- An Azure subscription
+- At least one developer with the `Owner` role on the Azure subscription in order to deploy resources and assign roles
+- A service principal (instructions below)
+- For developers to be able to read configuration from Terraform outputs, grant them the `Storage Blob Data Reader` (or `Contributor`) role on the storage account holding the Terraform state file.
+- For developers to be able to use their own identity to run tests, grant the the following roles at the Subscription level, or on the Resource Group created to hold Terraform resources.
```
paullatzelsperger, line 34:
```suggestion
- For developers to be able to use their own identity to run tests, grant the  following roles at the Subscription level, or on the Resource Group created to hold Terraform resources.```

File: extensions/azure/azure-test/src/testFixtures/java/org/eclipse/dataspaceconnector/azure/testfixtures/TerraformOutputsExtension.java [hunk 5]
```diff
@@ -0,0 +1,59 @@
...
+
+import java.io.File;
+import java.util.Map;
+import java.util.Properties;
+
+/**
+ * A JUnit extension for setting system properties from a Terraform outputs JSON file.
+ * Used by cloud integration tests.
+ */
+public class TerraformOutputsExtension implements BeforeAllCallback, AfterAllCallback {
```
paullatzelsperger, line 32: Relying on a particular deployment mechanism (i.e. Terraform) and its artifacts (the json file) seems like an anti-pattern. Why can't we use Azure ResourceManager to deploy everything?   
Or, at least [run terraform from withing code](https://github.com/microsoft/terraform-spring-boot)

.github/workflows/integrationtests.yaml
```diff
@@ -100,3 +106,49 @@ jobs:
...
+
+      - name: 'Az CLI login'
+        uses: azure/login@v1
+        with:
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+
+      - name: "Fetch Terraform outputs"
+        run: printf "$TERRAFORM_OUTPUTS" > resources/azure/testing/terraform_outputs.json
```
paullatzelsperger, line 139: won't that leak potentially sensitive information, for example `tenant_id` or `client_id`?

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/AzureDataFactoryTransferManager.java
```diff
@@ -0,0 +1,117 @@
...
+
+        return awaitRunCompletion(runId);
+    }
+
+    @NotNull
+    private CompletableFuture<TransferResult> awaitRunCompletion(String runId) {
+        monitor.info("Awaiting ADF pipeline completion for run " + runId);
+
+        var timeout = clock.instant().plus(maxDuration);
+        while (clock.instant().isBefore(timeout)) {
```
paullatzelsperger, line 70: busy waiting without at least a `Thread.sleep()` seems wrong. Am I missing something?

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/DataPlaneAzureDataFactoryExtension.java
```diff
@@ -0,0 +1,106 @@
...
+import java.time.Duration;
+import java.util.Objects;
+
+/**
+ * Registers a {@link TransferService} for performing data transfers with Azure Data Factory.
+ */
+public class DataPlaneAzureDataFactoryExtension implements ServiceExtension {
+
+    @EdcSetting
+    private static final String KEY_VAULT_LINKED_SERVICE_NAME = "edc.data.factory.key.vault.linked.service.name";
```
paullatzelsperger, line 40: nit: is there a hierarchy to this? otherwise I'd prefer `...linked-service-name`

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/DataFactoryPipelineFactory.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Create a Data Factory pipeline for a transfer request.
+     *
+     * @param request the transfer request.
+     * @return the created pipeline resource.
+     */
+    PipelineResource createPipeline(DataFlowRequest request) {
+        var baseName = "EDC-DPF-" + UUID.randomUUID();
```
paullatzelsperger, line 56: magic strings should be constants IMO.

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/KeyVaultClient.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.eclipse.dataspaceconnector.azure.dataplane.azuredatafactory;
+
+import com.azure.security.keyvault.secrets.SecretClient;
+import com.azure.security.keyvault.secrets.models.KeyVaultSecret;
+
+/**
+ * Client for Azure Key Vault, wrapping the Azure SDK.
+ */
+class KeyVaultClient {
```
paullatzelsperger, line 22: Why not use `AzureVault`?

paullatzelsperger (commented):
Impressive PR, kudos @cpeeyush!
I had a couple of remarks, I'm esp. dubious about depending on terraform for test execution. I'd prefer passing the values in as environment or system props.

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/DataPlaneAzureDataFactoryExtension.java
algattik: Not sure this use of dashes would play out well with how we resolve system properties/env variables

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/AzureDataFactoryTransferManager.java
algattik: Right, API is rate limited at 12,500/hour

extensions/azure/azure-test/src/testFixtures/java/org/eclipse/dataspaceconnector/azure/testfixtures/TerraformOutputsExtension.java
algattik: That is explained in the design document, it is preferable to have a service identity with the minimal required privileges so that even if it leaks it can't be abused to create e.g. VMs.

We could have a script that converts the Terraform output json file into a key=value .env type file.

.github/workflows/integrationtests.yaml
algattik: These are not sensitive values. In any case, they are not more exposed because they are on the filesystem than because they are passed to test code.

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/KeyVaultClient.java
algattik: To reuse the resource-manager module consistently, and because `AzureVault` authenticates with client secret. We could have a follow up task to refactor AzureVault to use resource-manager (enforcing SRP)

docs/developer/decision-records/2022-03-01-serverless-transfers/README.md
cpeeyush: Updated. Thanks.

docs/developer/cloud_testing.md
cpeeyush: Updated. thanks.

cpeeyush: Updated. thanks.

.github/workflows/integrationtests.yaml
paullatzelsperger: if they are stored as repo secrects, they would be obscured in log output, right? Since the `client_id` is half of the credentials package, i'd very much regard it as sensitive.

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/DataPlaneAzureDataFactoryExtension.java
paullatzelsperger: underscores then

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/KeyVaultClient.java
paullatzelsperger: i'm good with it being a temporary solution, ultimately we should enhance the existing `AzureVault` implementation, and not duplicate functionality

extensions/azure/azure-test/src/testFixtures/java/org/eclipse/dataspaceconnector/azure/testfixtures/TerraformOutputsExtension.java
paullatzelsperger: I don't quite understand. Am I correct in assuming that the permissions/privileges would be the same regardless whether we provision resources with Terraform or the Azure ResourceManager SDK? So we need elevated privileges for resources deployment anyway. 

That would mean that during test setup (=deploying resources) a different identity should be used than during test execution, right?

However, I do understand that rebuilding the Terraform config in Java Code (Azure RM) is probably not worth the effort.

paullatzelsperger (commented):
could we move the contents of `resources/azure/tesing` into `extensions/azure/azure-test/src/main/resources`? Seems like a more appropriate place.

algattik:
> could we move the contents of `resources/azure/tesing` into `extensions/azure/azure-test/src/main/resources`? Seems like a more appropriate place.

Why? The directory `src/main/resources` is by default in the Gradle source set, i.e. it is in the runtime class path and gets packed in the JAR. This is not what we want for the shell scripts and especially terraform provider binaries and state files.

extensions/azure/azure-test/src/testFixtures/java/org/eclipse/dataspaceconnector/azure/testfixtures/TerraformOutputsExtension.java
algattik: > That would mean that during test setup (=deploying resources) a different identity should be used than during test execution, right?

Indeed. During deployment you use the developer's own identity (user), while during test you use also the developer's identity when running on your machine, and a service principal with limited permissions in CI.

> However, I do understand that rebuilding the Terraform config in Java Code (Azure RM) is probably not worth the effort.

That would not be much of an issue actually. We deploy very few resources and it could be ported to Java easily (assuming no bad surprises), if that's the way forward we choose.

Using scripts to deploy does make the dev setup more complex and potentially less portable (to Windows). On the other hand, having the resources already deployed and their configuration downloaded locally makes running the test many times for developers ("inner loop") much faster than having the overhead of running Terraform each time. We might have a "best of both worlds scenario" with a Java test framework that "caches" the configuration locally between runs.

.github/workflows/integrationtests.yaml
algattik: The only reason client_id is stored as a secret is that GitHub Actions does not propose a way to store repo action settings as "non-secrets". It is no more sensitive than your GitHub ID is.

In any case, it is not being exposed. It is written to the local file system that is wiped after execution. Just like the actual secrets are written to the local Terraform state file.

extensions/azure/azure-test/src/testFixtures/java/org/eclipse/dataspaceconnector/azure/testfixtures/TerraformOutputsExtension.java
cpeeyush: @paullatzelsperger we have updated it to use `.properties` file and removed use of terraform json artifact.

extensions/data-plane/data-plane-framework/src/main/java/org/eclipse/dataspaceconnector/dataplane/framework/DataPlaneFrameworkExtension.java
algattik: Implemented

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/AzureDataFactoryTransferManager.java
algattik: Implemented

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/DataPlaneAzureDataFactoryExtension.java
algattik: That would make it impossible to set it via env variables. Those underscores would be replaced with periods...

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/DataFactoryPipelineFactory.java
algattik: Implemented

extensions/azure/data-plane/data-factory/src/main/java/org/eclipse/dataspaceconnector/azure/dataplane/azuredatafactory/DataPlaneAzureDataFactoryExtension.java
paullatzelsperger: then lets just join them as "linkedservicename". not ideal, but (to my mind) better than the dots.

paullatzelsperger (approved):

paullatzelsperger (approved):

algattik:
@bscholtes1A would you please be able to review this PR today?

bscholtes1A (approved):

jimmarino:
@algattik are all the outstanding discussion points resolved? If so, could you fix the conflicts and let me know so I can merge?

PR #911, opened 2022-03-16
system-tests: avoid system-tests failure when runtimes are not fully built

ndr-brt:
## What this PR changes/adds

Makes the `EdcRuntimeExtension` load the resources of an "extension runtime" from the `main` folder.

## Why it does that

Currently, there are two way of handling extension runtime dependencies: 
- not define them into the "runner/tests" module as dependency (as it's done by the `system-tests:end-to-end-transfer:runner` module)
- define them as `testRuntimeOnly` (as it's done by the `system-tests:tests` module)

both of them are wrong, for these reasons:
- the first because it does not define a "build" order for the modules, so you must have every extension runtime module built by the test run, otherwise the extension runtimes will not be loaded.
- the second because it loads all the dependencies of the extension runtimes into the common class loader, and so every extension runtime will have the union of the dependencies of every runtime, and that is wrong, because every runtime should have its dependencies (e.g. control-plane dependencies are different from the data-plane one)

I found out that the best way is to use `testCompileOnly`, that takes the best from the used approaches.
The only issue is that the `testCompile` task happens before the `processResources`, so after that happens no resources are built into the module `build` directory. To resolve this, I made the `EdcRuntimeExtension.resolveClassPathEntry` to replace the "built" path with the "src" path on the resource entry. This works and saves lots of headaches.

## Further notes

-

## Linked Issue(s)

-

## Checklist

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -1,18 +1,19 @@
-# Writing Integration Tests
+# Writing Tests
 
 ## Definition and distinction
 
-While _unit tests_ test one single class by stubbing or mocking dependencies, and an end-2-end test relies on the _
-entire_
-system to be present, the _integration test_ tests one particular aspect of a software, which may involve external
-systems.
+* _unit tests_ test one single class by stubbing or mocking dependencies.
+* [_integration test_](#integration-tests) tests one particular aspect of a software, which may involve external systems.
+* [_system tests_](#system-tests) are end-2-end tests that rely on the _entire_ system to be present.
 
-## TL;DR
+## Integration Tests
+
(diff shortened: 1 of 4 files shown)


algattik:
Brilliant thinking!

Would we worth adding a couple of lines from the above description to the docs. In integration-testing.md?

bscholtes1A (approved):

ndr-brt:
@algattik good point, I generalized `integration-testing.md` to `testing.md` adding a `system tests` section

algattik (approved):

PR #925, opened 2022-03-18
Allow sub-paths to be propagated through http dataplane.

drcgjung:
## What this PR changes/adds

Currently, the inbound Filter forgets the actual path that is hit under the base-path/deployment-path.
This also leads to the "name" compartment of the HttpDataSource being left empty.
This fix computes the difference between basepath and requestpath and uses it as the "name" property of the DataAddress in the request --> HttpDataSource gets a (hierarchical) name that is appended to the target url.

## Why it does that

For wrapping arbitrary REST APIs using the http dataplane, it is necessary not only to take headers, params and body with the data request, but also the concrete URL that is hit (otherwise you would need indefinitely many data planes/addresses to be registered). 

## Further notes

Tests have been checked before and after (quite some were already red in the main branch, so this was rather a guess than a profound investigation), but no additional test was added (maybe you should think about it).

## Linked Issue(s)

## Checklist


drcgjung:
wrong source branch

PR #926, opened 2022-03-18
Feature/catenax wrapper path

drcgjung:
What this PR changes/adds
Currently, the inbound Filter forgets the actual path that is hit under the base-path/deployment-path.
This also leads to the "name" compartment of the HttpDataSource being left empty.
This fix computes the difference between basepath and requestpath and uses it as the "name" property of the DataAddress in the request --> HttpDataSource gets a (hierarchical) name that is appended to the target url.

Why it does that
For wrapping arbitrary REST APIs using the http dataplane, it is necessary not only to take headers, params and body with the data request, but also the concrete URL that is hit (otherwise you would need indefinitely many data planes/addresses to be registered).

Further notes
Tests have been checked before and after (quite some were already red in the main branch, so this was rather a guess than a profound investigation), but no additional test was added (maybe you should think about it).

## Linked Issue(s)

Closes # <-- _insert Issue number if one exists_

## Checklist

- [ ] added appropriate tests?
- [ ] performed checkstyle check locally?
- [ ] added/updated copyright headers?
- [ ] documented public classes/methods?
- [ ] added/updated relevant documentation?
- [ ] added relevant details to the changelog? (_skip with label `no-changelog`_)
- [ ] formatted title correctly? (_take a look at the [CONTRIBUTING](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/CONTRIBUTING.md#submit-a-pull-request) and [styleguide](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/styleguide.md) for details_)

PR #929, opened 2022-03-19
Fix DPF public API and add capability to map the path to the Http Dat…

### LINKED ISSUE #879: Data Plane Public API fails to map request HTTP method
# Bug Report

## Describe the Bug
Data Plane public API, which is modeled by the `DataPlanePublicApiRequestFilter` class does not map properly the request HTTP method into the `DataFlowRequest`. Every request is mapped as a _GET_ regardless of the actual HTTP method used to call the DPF public API. 

During testing we observed that proper behavior is obtained when a random path is appended at the end of the url. Thus it seems that `ContainerRequestFilter` only fails to map the HTTP method when targeting the root path. 

### Expected Behavior
The HTTP verb of the request should be mapped into the `DataFlowRequest`.

### Observed Behavior
The HTTP _GET_ verb is always mapped into the `DataFlowRequest`, regardless of the actual HTTP method used in the request. 

## Steps to Reproduce
1. Send _POST_ request to the DPF public API, i.e. POST http://my-dpf-instance.com:9191/public
2. Inspect which verbs is mapped into the `DataFlowRequest`

## Detailed Description

## Possible Implementation
A possible approach would be to try replacing the current ContainerRequestFilter by a Jetty [AbstractHandler](https://www.programcreek.com/java-api-examples/?api=org.eclipse.jetty.server.handler.AbstractHandler). Not sure it will fix the issue though, should be tested.

### LINKED ISSUE #881: Data Plane Framework - Pass Path Information
# Feature Request
The public dataplane api catches all incoming http calls but drops the path information before calling the actual service in the backend. Query params and http method get already passed properly.
Propose to append the path retrieved to the URL that is used to call the backend service.
We could also make this configurable (e.g. toggles for "pass http method", "pass query params", "pass path information", ...)

DPF and maybe Assets

Integration of REST APIs like the one described by AAS https://app.swaggerhub.com/apis/Plattform_i40/AssetAdministrationShell-REST-API/Final-Draft
It should be possible to grant access to multiple endpoints / pathes via one EDC contract, e.g. grant access to a complete AAS.

## Solution Proposal
pass path information similar to how we pass query params today, e.g. just add path information retrieved in `DataPlanePublicApiRequestFilter` to the end of the `DataAddress` that is called on provider side.

## Type of Issue
improvement

## Checklist

bscholtes1A:
## What this PR changes/adds

This PR repairs the public API of the DPF. Especially it now limit the usage of the `RequestFilter` to the token validation and creation of the `DataFlowRequest` based on the request properties, which is highly convenient as it completely generic (all verbs, any path...).

Then the request filter delegates to the newly introduced public API controllers which validate the request and then perform the data transfer. 

We added the capability to also map the path segments provided into the input request to the http data source (it was previously done for query parameters and for the request body).

Finally some integration tests are added in order to validate the behavior of the public API.

## Why it does that

Fix DPF public API and enable forwarding path segments to the Http data source.

## Linked Issue(s)

Closes #881, Closes #879.

## Checklist


extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/transfer/DataPlanePublicApiRequestFilter.java
```diff
@@ -90,36 +86,46 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
...
+        // create the data flow request and map it into the request body
+        var body = new BufferedReader(new InputStreamReader(requestContext.getEntityStream()))
+                .lines().collect(Collectors.joining("\n"));
+        var properties = createRequestProperties(requestContext.getRequest(), requestContext.getUriInfo(), requestContext.getMediaType(), body);
+        var dataFlowRequest = createDataFlowRequest(tokenValidationResult.getContent(), properties);
+        var serialized = typeManager.writeValueAsString(dataFlowRequest);
+        requestContext.setEntityStream(new ByteArrayInputStream(serialized.getBytes()));
+
+        // redirect to endpoints exposed by the controller
+        requestContext.setRequestUri(URI.create(requestContext.getUriInfo().getBaseUri() + publicApiPath));
```
drcgjung, line 98: Yeah that's base path + custom (data plane specific) parent segment.

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/transfer/DataPlanePublicApiRequestFilter.java
```diff
@@ -90,36 +86,46 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
...
+    /**
+     * Map relevant information from the container request context to the properties of the {@link DataFlowRequest}.
+     */
+    private Map<String, String> createRequestProperties(Request request, UriInfo uriInfo, @Nullable MediaType mediaType, String body) {
+        var props = new HashMap<String, String>();
+        props.put(METHOD, request.getMethod());
+        props.put(QUERY_PARAMS, convertQueryParamsToString(uriInfo));
+
+        if (!uriInfo.getPathSegments().isEmpty()) {
+            props.put(PATH, uriInfo.getPathSegments().stream()
```
drcgjung, line 111: very neat. However, the publicApiPath could me multi-segmented too? (stuff/moreStuff) So it could not be compared against a single segment of the request uri, doesn't it? Maybe just hypothetical but this should be documented somewhere.

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/dataspaceconnector/dataplane/http/pipeline/HttpDataSourceFactory.java
```diff
@@ -119,6 +120,9 @@ public DataSource createSource(DataFlowRequest request) {
             builder.header(authKey, secretResult.getContent());
         }
 
+        Optional.ofNullable(request.getProperties().get(PATH))
+                .ifPresent(builder::name);
```
drcgjung, line 124: And that was the final piece. Very well done, I'll take my PR back then ...

drcgjung (approved):
Checked it from the perspective of #926 ... will work. thx.

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/transfer/DataPlanePublicApiRequestFilter.java
bscholtes1A: Very good remark. 
I updated to something more resilient and add a unit-test for this scenario.

MoritzKeppler:
thanks, @bscholtes1A!
One general observation: can you please check the legal headers of the files and ensure "Amadeus ..." is present?
A detailed review will follow.

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/transfer/DataPlanePublicApiRequestFilter.java
```diff
@@ -90,36 +89,52 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
...
-        // create request and perform validation
-        var dataFlowRequest = createDataFlowRequest(tokenValidationResult.getContent(), requestContext);
-        var validationResult = dataPlaneManager.validate(dataFlowRequest);
-        if (validationResult.failed()) {
-            requestContext.abortWith(validationResult.getFailureMessages().isEmpty() ?
-                    validationError(String.format("Failed to validate request with id: %s", dataFlowRequest.getId())) :
-                    validationErrors(validationResult.getFailureMessages()));
-            return;
+        // create the data flow request and map it into the request body
+        var body = new BufferedReader(new InputStreamReader(requestContext.getEntityStream()))
```
ndr-brt, line 93: As github suggest, better close the Reader after extracting data, maybe with a `try` with resources statement

gradle.properties
```diff
@@ -1,15 +1,14 @@
...
-
 javaVersion=11
 infoModelVersion=4.2.7
 jetBrainsAnnotationsVersion=15.0
 jacksonVersion=2.13.1
 jettyVersion=11.0.6
 websocketVersion=2.0.0
 jerseyVersion=3.0.3
-servletApi=4.0.4
+servletApi=4.0.1
```
ndr-brt, line 11: Could you explain in the PR the reason for this downgrade?

extensions/data-plane/integration-tests/src/test/java/org/eclipse/dataspaceconnector/dataplane/http/DataPlaneHttpPullIntegrationTests.java
```diff
@@ -0,0 +1,268 @@
...
+    private static final Map<String, String> PROPS = Map.of(
+            "web.http.control.port", String.valueOf(DPF_CONTROL_API_PORT),
+            "web.http.public.port", String.valueOf(DPF_PUBLIC_API_PORT),
+            "web.http.public.path", DPF_PUBLIC_PATH,
+            "edc.controlplane.validation-endpoint", VALIDATION_API_HOST
+    );
+
+    @BeforeAll
+    public static void setUp() {
+        PROPS.forEach(System::setProperty);
```
ndr-brt, line 100: it's better to avoid system properties to inject configuration to the EdcExtension, it can be done with this call on a setup method:
```
@BeforeEach
void setup(EdcExtension extension) {
    extension.registerSystemExtension(ConfigurationExtension.class, (ConfigurationExtension) () -> ConfigFactory.fromMap(configuration));
}
```

extensions/data-plane/integration-tests/src/test/java/org/eclipse/dataspaceconnector/dataplane/http/DataPlaneHttpPullIntegrationTests.java
```diff
@@ -0,0 +1,268 @@
...
+    private static final int DPF_CONTROL_API_PORT = getFreePort();
+    private static final int DPF_PUBLIC_API_PORT = getFreePort();
+    private static final String DPF_PUBLIC_API_HOST = "http://localhost:" + DPF_PUBLIC_API_PORT;
+    private static final String DPF_PUBLIC_PATH = "/public";
+    private static final String DPF_DATA_PATH = String.format("%s/data", DPF_PUBLIC_PATH);
+
+    // Data source API
+    private static final int HTTP_SOURCE_API_PORT = getFreePort();
+    private static final String HTTP_SOURCE_API_HOST = "http://localhost:" + HTTP_SOURCE_API_PORT;
+    private static final String SOURCE_AUTH_VALUE = FAKER.lorem().word();
```
ndr-brt, line 79: unused

ndr-brt (approved):
looks good, left some comments on minor nits

MoritzKeppler (changes requested):
left some comments. One overall question: what is the main benefit from having a filter and a controller? Think it could be easier if it's just a controller (any maybe an authentication filter in addition)

extensions/data-plane/data-plane-api/src/test/java/org/eclipse/dataspaceconnector/dataplane/api/transfer/DataPlanePublicApiRequestFilterTest.java
```diff
@@ -260,6 +186,18 @@ private ContainerRequestContext createDefaultContext(MultivaluedMap<String, Stri
         return context;
     }
 
+    private static UriInfo createMockUriInfo(MultivaluedMap<String, String> queryParams) {
```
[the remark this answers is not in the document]
bscholtes1A: Yes but this is a simple approach to have a first layer of testing that goes very fast. 

More advanced test cases are tested in the newly introduced end-to-end integration tests.

gradle.properties
bscholtes1A: Restored as in main branch.

extensions/data-plane/data-plane-api/src/test/java/org/eclipse/dataspaceconnector/dataplane/api/response/ResponseFunctionsTest.java
```diff
@@ -11,16 +11,18 @@
  *       Microsoft Corporation - initial API and implementation
```
MoritzKeppler, line 11: you can add Amadeus here

extensions/data-plane/data-plane-http/src/test/java/org/eclipse/dataspaceconnector/dataplane/http/pipeline/HttpDataSourceFactoryTest.java
```diff
@@ -31,14 +32,14 @@
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.UUID;
 import java.util.stream.Stream;
```
MoritzKeppler, line 37: you could Amadeus to the list of contributors in the class header

extensions/data-plane/data-plane-http/src/test/java/org/eclipse/dataspaceconnector/dataplane/http/pipeline/HttpDataSourceTest.java
```diff
@@ -31,6 +31,10 @@
 class HttpDataSourceTest {
```
MoritzKeppler, line 31: Class does not have a legal header at all!?

extensions/data-plane/integration-tests/src/test/java/org/eclipse/dataspaceconnector/dataplane/http/DataPlaneHttpPullIntegrationTests.java
```diff
@@ -0,0 +1,273 @@
...
+ *  Copyright (c) 2022 Microsoft Corporation
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
```
MoritzKeppler, line 11: guess it should point to Amadeus

MoritzKeppler (changes requested):
thanks for the adjustments! looks almost good to me.
just some minor comments, mainly about the legal headers.

MoritzKeppler (approved):
lgtm

ndr-brt (approved):

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/controller/ContainerRequestContextApiImpl.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    /**
+     * Convert query parameters injected from the input into a string, e.g. foo=bar&hello=world
+     */
+    private static String convertQueryParamsToString(UriInfo uriInfo) {
+        return uriInfo.getQueryParameters().entrySet()
+                .stream()
+                .map(entry -> new QueryParam(entry.getKey(), entry.getValue()))
```
drcgjung, line 83: I am not sure how url encoded params (like json query params) will behave over this. My gut says this will work if not some magic "decode" happens in consumer-side uriInfo or somewhere else without the fitting "recode" in the provider part.

Cannot tell from the specs so easily ... so We will see in a few hours ;-)

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/response/ResponseFunctions.java
```diff
@@ -28,25 +28,28 @@
  */
 public final class ResponseFunctions {
 
+    private static final String DATA = "data";
```
drcgjung, line 31: I haven checked it, but in the previous (main) logic, data gets encoded twice (provider side and then again in the consumer side), so we did need to run "double decode" on the data response object. 

I could not find an evidence that this behaviour was improved ...

drcgjung (approved):
LGTM (with a few ideas, beautifications and maybe two remarks which may lead to follow-up issues).

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/response/ResponseFunctions.java
bscholtes1A: good catch indeed. 
I removed this data stuff so that only the raw data is now returned in output.

drcgjung (approved):
guess that has tested away my fear with the double-encoded data ...

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/response/ResponseFunctions.java
drcgjung: top notch.

jimmarino:
@bscholtes1A are you able to merge this (i.e. your permissions have been sorted)?

bscholtes1A:
> @bscholtes1A are you able to merge this (i.e. your permissions have been sorted)?

Not yet, ticket opened by @mspiekermann to EF

PR #930, opened 2022-03-21
DataManagementApi: In-memory policy store implementation

### LINKED ISSUE #803: Create `PolicyStore` and default impl
A `PolicyStore` interface must be implemented with the following API:
```java
public interface PolicyStore {
    // returns null if not found
    Policy findById(String policyId);
    Stream<Policy> findAll(QuerySpec spec);
    void save(Policy policy);
    // returns null if not found, returns the deleted item otherwise.
    Policy delete(String policyId);
}
```
If any exception occurs, a `EdcPersistenceException` should be thrown.

In addition, an in-memory implementation should be provided.

Izzzu:
## What this PR changes/adds

- Add PolicyStore interface.
- Implement in memory version of PolicyStore.
- Add InMemoryPolicyStoreExtension

## Why it does that

Implementation of PolicyStore according to definitions in upstream issue: [#803](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/803)

## Further notes

This PR covers also following points:

- Extracting common code from store implementations of findAll(QuerySpec spec) methods into a class ReflectionBasedQueryResolver.
- Removing not up to date javadoc from CosmosContractDefinitionStore for Stream<ContractDefinition> findAll(QuerySpec spec) method
- Adding test in CosmosContractNegotiationStoreIntegrationTest verifying that filtering by non existent sort field returns empty stream.
- Change in InMemoryContractNegotiationStore to be consistent with other stores to return empty stream in case of filtering by nonexistent sort field.

## Linked Issue(s)

Closes #803 

## Checklist


spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/ReflectionBasedQueryResolver.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.stream.Stream;
+
+import static org.eclipse.dataspaceconnector.common.reflection.ReflectionUtil.propertyComparator;
+
+/**
+ * Applies query on a stream. Uses reflection to fetch object fields. Used in stores implementations.
+ *
+ * @param <T> type of the stream elements.
+ */
+public class ReflectionBasedQueryResolver<T> extends QueryResolver<T> {
```
paullatzelsperger, line 28: Can we move this into `:core:core-base`? it would then have to be passed into constructors in the `*Stores`

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/QueryResolver.java
```diff
@@ -0,0 +1,33 @@
...
+package org.eclipse.dataspaceconnector.spi.query;
+
+import java.util.stream.Stream;
+
+/**
+ * Class responsible for applying queries on an input elements of type T. Currently supports querying a stream of elements.
+ *
+ * @param <T> type of the queried elements.
+ */
+public abstract class QueryResolver<T> {
```
paullatzelsperger, line 23: could this be an interface, even a `@FunctionalInterface`?

spi/core-spi/build.gradle.kts
```diff
@@ -30,6 +30,7 @@ dependencies {
 
     implementation("io.opentelemetry:opentelemetry-api:${openTelemetryVersion}")
 
+    api(project(":common:util"))
```
paullatzelsperger, line 33: unless you're exposing classes from `:common:util` in this module's public API, it's better to use `implementation` dependencies

extensions/azure/cosmos/contract-definition-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/definition/store/CosmosContractDefinitionStore.java
```diff
@@ -61,40 +62,17 @@ public CosmosContractDefinitionStore(CosmosDbApi cosmosDbApi, TypeManager typeMa
         this.retryPolicy = retryPolicy;
         this.partitionKey = partitionKey;
         lockManager = new LockManager(new ReentrantReadWriteLock(true));
+        queryResolver = new ReflectionBasedQueryResolver<>(ContractDefinition.class);
```
paullatzelsperger, line 65: this should get passed in through the CTor, so that the `ReflectionBasedQueryResolver` can be moved to core

paullatzelsperger (changes requested):

spi/contract-spi/src/main/java/org/eclipse/dataspaceconnector/spi/contract/policy/store/PolicyStore.java
```diff
@@ -0,0 +1,64 @@
...
+import org.eclipse.dataspaceconnector.spi.persistence.EdcPersistenceException;
+import org.eclipse.dataspaceconnector.spi.query.QuerySpec;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.stream.Stream;
+
+/**
+ * Persists {@link Policy}.
+ */
+public interface PolicyStore {
```
paullatzelsperger, line 26: I think this should be in `contract-spi`. We might need access to it in runtimes that don't need the contract stuff.

spi/contract-spi/src/main/java/org/eclipse/dataspaceconnector/spi/contract/policy/store/PolicyStore.java
```diff
@@ -0,0 +1,64 @@
...
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+package org.eclipse.dataspaceconnector.spi.contract.policy.store;
```
ndr-brt, line 14: Since the `Policy` class is in the `core-spi`, maybe also this store will fit better there? Maybe could be time for a `policy-spi` module?

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/ReflectionBasedQueryResolver.java
```diff
@@ -0,0 +1,85 @@
...
+        // filter
+        Stream<Predicate<T>> predicateStream = spec.getFilterExpression().stream().map(this::toPredicate);
+        var andPredicate = predicateStream.reduce(x -> true, Predicate::and);
+        Stream<T> filteredStream  = stream.filter(andPredicate);
+
+        // sort
+        var sortField = spec.getSortField();
+
+        if (sortField != null) {
+            // if the sort field doesn't exist on the object -> return empty
```
ndr-brt, line 63: This behaviour is kinda strange, I would prefer an error if the sort field does not exist, it's pretty common on query systems (like sql databases), otherwise it will be pretty hard to understand why the query returned an empty stream.

spi/contract-spi/src/main/java/org/eclipse/dataspaceconnector/spi/contract/policy/store/PolicyStore.java
ndr-brt: did you mean `:spi:policy-spi`?

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/ReflectionBasedQueryResolver.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.eclipse.dataspaceconnector.common.reflection.ReflectionUtil;
+
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+import static org.eclipse.dataspaceconnector.common.reflection.ReflectionUtil.propertyComparator;
+
+/**
+ * Applies query on a stream. Uses reflection to fetch object fields. Used in stores implementations.
```
paullatzelsperger, line 24: could you also add a note that this is the default implementation?

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/ReflectionBasedQueryResolver.java
Izzzu: Agree, but this code was extracted directly from query methods in *Store classes as a refactoring, not changed. I wouldn't increase the scope of this PR to change this logic, as it would impact several store implementations.

spi/contract-spi/src/main/java/org/eclipse/dataspaceconnector/spi/contract/policy/store/PolicyStore.java
Izzzu: Makes sense, I must have left it in this module by mistake. I created :spi:policy-spi module. Thanks

spi/contract-spi/src/main/java/org/eclipse/dataspaceconnector/spi/contract/policy/store/PolicyStore.java
Izzzu: I created policy-spi in spi and moved the PolicyStore there to be consistent with other store interfaces

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/ReflectionBasedQueryResolver.java
Izzzu: Done. Changed to 
> Default implementation of {@link QueryResolver} that applies query on a stream. Uses reflection to fetch object fields. Used in stores implementations.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/QueryResolver.java
Izzzu: fully agree, changed

extensions/azure/cosmos/contract-definition-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/definition/store/CosmosContractDefinitionStore.java
Izzzu: I think it's related to your previous comment: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/930#discussion_r833382897
Why we should move it to core-base? Are modules allowed to depend on core-base?

paullatzelsperger: what I meant was to _pass this as constructor argument_ rather than _instantiating directly_. That way, the `CosmosContractDefinitionStore` only has a dependency on `QueryResolve`.
Then, the `ReflectionBasedQueryResolver` instance could be `@Inject`-ed in the `CosmosContractDefinitionStoreExtension`. This would of course require some extension to have registered it previously. 

The more I think about this, the more I think the way it is now is actually better, because is much less hassle and it hides the implementation. So please feel free to disregard :)

paullatzelsperger (commented):
LGTM once checkstyle violations are resolved.

paullatzelsperger (approved):

jimmarino (approved):

jimmarino:
@Izzzu can you fix the conflicts and I'll merge?

Izzzu:
> @Izzzu can you fix the conflicts and I'll merge?

@jimmarino done, thanks!

PR #938, opened 2022-03-21
Proposal: technical concept for a `ContractNegotiationStore` based on SQL

### LINKED ISSUE #864: SQL: implementation of the `ContractNegotiationStore`
An SQL-based implementation of the `TransferProcessStore` is needed.

### Requirements
1. Exclusive Locking: the `nextForState` method must take into account the exclusive locking mechanism required by the asynchronous nature of the `ContractNegotiationManager`s:
    - `nextForState()` locks the TP
    - `update()` unlocks the TP
    - `find()` ignores the lock
    - there should be a timeout after which the lock is automatically released.
1. `save()`: implement upsert semantics
2. Querying: the query implementation must be flexible enough to accomodate  a `QuerySpec`.
3. Table Schema: `ContractAgreement` should be in a separate table (use FK) as `findContractAgreement()` will have to traverse that FK. 
4. Indexing: As a starting point there thould be indexes on `ContractNegotiation.id`, `ContractAgreement.id` and `ContractNegotiation.correlationId`.

_Please provide a written proposal for requirements 1, 3, 4 and 5 before beginning the implementation._

- [x] Design Proposal: #938 
- [ ] Implementation:

paullatzelsperger:
## What this PR changes/adds

Adds a technical concept for table schema and DML/DQL statements. No implementation.

## Why it does that

Before the implementation of #864 can start, there needs to be a vetted proposal in place.

## Further notes

Concept document currently resided in `extensions/sql/contract-negotiation/proposal.md` and will be renamed to `README.md` once the proposal is accepted.

## Linked Issue(s)

Closes #864 

## Checklist

### DIFF
--- a/extensions/sql/contract-negotiation/ddl.sql
+++ b/extensions/sql/contract-negotiation/ddl.sql
@@ -1,0 +1,60 @@
+create table if not exists contract_agreement
+(
+    id                varchar not null
+        constraint contract_agreement_pk
+            primary key,
+    provider_agent_id varchar,
+    consumer_agent_id varchar,
+    signing_date      bigint,
+    start_date        bigint,
+    end_date          integer,
+    asset_id          varchar not null,
+    policy_id         varchar
+);
+
+alter table contract_agreement
+    owner to "user";
+
+create table if not exists contract_negotiation
+(
+    id                    varchar                                            not null
+        constraint contract_negotiation_pk
+            primary key,
+    correlation_id        varchar                                            not null,
+    counterparty_id       varchar                                            not null,
+    counterparty_address  varchar                                            not null,
+    protocol              varchar default 'ids-multipart'::character varying not null,
+    type                  integer default 0                                  not null,
+    state                 integer default 0                                  not null,
+    state_count           integer default 0,
+    state_timestamp       bigint,
+    error_detail          varchar,
+    contract_agreement_id varchar
+        constraint contract_negotiation_contract_agreement_id_fk
+            references contract_agreement,
+    contract_offers       varchar,
+    trace_context         varchar,
+    lease_id              varchar
+        constraint contract_negotiation_lease_lease_id_fk
+            references lease
+            on delete set null
+);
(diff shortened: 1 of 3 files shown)


denisneuling (changes requested):
@paullatzelsperger Would you please define the length of all VARCHAR, INTEGER etc declarations? If omitted the DBMS will pick a default length, which could be really short (E.g. afaik MS SQL has a default varchar length of 30)

paullatzelsperger:
> @paullatzelsperger Would you please define the length of all VARCHAR, INTEGER etc declarations? If omitted the DBMS will pick a default length, which could be really short (E.g. afaik MS SQL has a default varchar length of 30)

that makes sense. will update.

denisneuling:
@paullatzelsperger  I am rather concerned about data integrity than about performance. 
If we omit declaration of the values length the DBMS will pick its default length for the columns data type. Then, if data to be persisted exceeds the (default) length, it will be truncated best case or raise an error. Either way is not wanted I suppose. So we should be explicit within our DDL and pick ranges applicable to DBMS we want to support. 
Moreover we should add checks/validations at the application layer prior to persisting data.

paullatzelsperger:
@denisneuling I guess the question then is what a reasonable size is. There is not really a reliable way to determine that, especially since we'll have columns containing JSON-serialized objects, so the safest way is to define `varchar(65535)` which is the max size for Postgres.

AFAIK Postgres doesn't have a `varchar(max)`, only an unbounded one (`varchar` or `text`). SQL Server on the other hand has both and as you said defaults to 30, but the syntax is different. 

The `*.sql` file we'll provide is for development/testing purposes only anyway, because in a real world scenario I suppose there would be a DB admin taking care of the _actual_ DDL, so about this: lets stick with `varchar` for now, and rename the file to `ddl_postgres.sql`, and add a comment that it is specific to Postgres. 

Providing more dialect-specific DDL could be done when #923 is implemented.

Issue #828, opened 2022-03-07
Refactor structure of business objects


paullatzelsperger:
## Problem Statement

Currently some business objects (BOs) have actual references to other objects (e.g. `ContractAgreement` -> `Asset`) instead of just business keys (i.e. `assetId`). This has two major implications:
- at the persistence level we may need to introduce another object, e.g. a `ContractAgreementEntity` would maintain a foreign-key to make sure the referenced object (here: the `Asset`) isn't duplicated. this would introduce another layer of transformation, testing, etc.
- since many of those BOs are exposed through an API like IDS or REST, they're exposing too much information. For example, there is no need for the entire `Policy` and `Asset` object to be contained inside the `ContractAgreement`. 
- [optiona] Can we replace `Asset` and `Policy` with their respective IDs in `ContractOffer` or should we add an `assetId` and `policyId` field to make a sparsely populated `ContractOffer` more transparent?

## Proposal
We therefore propose to refactor those BOs and replace object references with business keys. They will reduce the API surface, the size of the transmitted JSON and avoid duplication by design. They would also be more consistent towards the persistence layer, because we don't need to serialize anything to JSON, we could simply use a sub-query.

The following section lists all the proposed refactoring steps and the implications that each one has:
1. `ContractAgreement`: replace `asset (Asset)` with `assetId (String)`. Adapt all usages.
2. `ContractAgreement`: replace `policy (Policy)` with `policyId (String)`. Implications:
    - `ContractAgreementToIdsContractAgreementTransformer`: would need access to the `PolicyStore` (cf. Issue #803)
4. `ContractDefinition`: replace both policy fields with `[access|contract]PolicyId (String)`. Implications:
    - `ContractDefinitionServiceImpl` would need access to `PolicyStore`
    - `ContractValidationServiceImpl` needs access to `PolicyStore`
    - `ContractOfferServiceImpl` needs access to `PolicyStore`

_Note: `PolicyStore` should implement simple in-mem caching to speed up lookups during contract validation_

PR #940, opened 2022-03-22
extensions: implement ContractDefinition Data Management Api 

### LINKED ISSUE #806: DataManagementApi: add `ContractDefinitionService` and `ContractDefinitionTransformer`
For the implementation of the contract definition endpoint we need an `ContractDefinitionService` (and an `ContractDefinitionServiceImpl`) with roughly the following signature:
```java
public interface ContractDefinitionService {
  Asset findbyId(String contractDefinitionId);
  Collection<ContractDefinition> query(QuerySpec query);
  ContractDefinition delete(String contractDefinitionId);
  ContractDefinition create(ContractDefinition newDef);
}
```

In addition, there needs to be an `ContractDefinitionTransformer`, that converts an `ContractDefinition` into an `ContractDefinitionDto` and back. This depends on #779 

Important aspects:
- service must perform defensive checks before performing an operation, e.g. `delete`: must check first, whether a `ContractAgreement` already exists with the same ID, in which case it cannot be deleted.

ndr-brt:
## What this PR changes/adds

Implements ContractDefinition service for Data Management Api.
Depends on #931 

## Why it does that

To expose a clear and RESTful api to operate on ContractDefinition

## Further notes

* Implemented "like" operator in `BaseCriterionToPredicateConverter` to permit filter agreements by definition id.
* Inlined `DtoTransformer.canHandle` method, as its implementation was always the same for every implementor.
* Removed some tests on the controller that verified the `QuerySpec` object construction, but in my opinion those verifications should stay on the `QuerySpec` factory (that, in our case, is the builder object)
* The check for existing agreements on the definition before delete has been removed since it's impossible to do that currently (will be possible after #972)

## Linked Issue(s)

Closes #806

## Checklist

### DIFF
--- a/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
+++ b/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
@@ -44,7 +44,7 @@ public class AssetServiceImpl implements AssetService {
     }
 
     @Override
-    public Asset findbyId(String assetId) {
+    public Asset findById(String assetId) {
         return index.findById(assetId);
     }
 
@@ -84,7 +84,7 @@ public class AssetServiceImpl implements AssetService {
         var result = new AtomicReference<ServiceResult<Asset>>();
 
         transactionContext.execute(() -> {
-            if (findbyId(asset.getId()) == null) {
+            if (findById(asset.getId()) == null) {
                 loader.accept(asset, dataAddress);
                 result.set(ServiceResult.success(asset));
             } else {
--- a/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
+++ b/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
@@ -84,7 +84,7 @@ public class AssetServiceImpl implements AssetService {
         var result = new AtomicReference<ServiceResult<Asset>>();
 
         transactionContext.execute(() -> {
-            if (index.findById(asset.getId()) == null) {
+            if (findbyId(asset.getId()) == null) {
                 loader.accept(asset, dataAddress);
                 result.set(ServiceResult.success(asset));
             } else {
--- a/extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/ContractDefinitionApiController.java
+++ b/extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/ContractDefinitionApiController.java
@@ -35,7 +35,6 @@ import org.eclipse.dataspaceconnector.spi.monitor.Monitor;
 import org.eclipse.dataspaceconnector.spi.query.QuerySpec;
 import org.eclipse.dataspaceconnector.spi.query.SortOrder;
 import org.eclipse.dataspaceconnector.spi.result.Result;
-import org.eclipse.dataspaceconnector.spi.types.domain.asset.Asset;
 import org.eclipse.dataspaceconnector.spi.types.domain.contract.offer.ContractDefinition;
 
 import java.util.List;
@@ -87,7 +86,7 @@ public class ContractDefinitionApiController implements ContractDefinitionApi {
         monitor.debug(format("get contract definition with ID %s", id));
 
         return Optional.of(id)
-                .map(service::findbyId)
+                .map(service::findById)
                 .map(it -> transformerRegistry.transform(it, ContractDefinitionDto.class))
                 .filter(Result::succeeded)
                 .map(Result::getContent)
--- a/extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/ContractDefinitionApiController.java
+++ b/extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/ContractDefinitionApiController.java
@@ -25,12 +25,22 @@ import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.core.MediaType;
 import org.eclipse.dataspaceconnector.api.datamanagement.contractdefinition.model.ContractDefinitionDto;
+import org.eclipse.dataspaceconnector.api.datamanagement.contractdefinition.service.ContractDefinitionService;
+import org.eclipse.dataspaceconnector.api.exception.ObjectExistsException;
+import org.eclipse.dataspaceconnector.api.exception.ObjectNotFoundException;
+import org.eclipse.dataspaceconnector.api.result.ServiceResult;
+import org.eclipse.dataspaceconnector.api.transformer.DtoTransformerRegistry;
+import org.eclipse.dataspaceconnector.spi.EdcException;
 import org.eclipse.dataspaceconnector.spi.monitor.Monitor;
 import org.eclipse.dataspaceconnector.spi.query.QuerySpec;
 import org.eclipse.dataspaceconnector.spi.query.SortOrder;
+import org.eclipse.dataspaceconnector.spi.result.Result;
+import org.eclipse.dataspaceconnector.spi.types.domain.asset.Asset;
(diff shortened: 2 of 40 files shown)


extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
```diff
@@ -84,7 +84,7 @@ public Asset findbyId(String assetId) {
         var result = new AtomicReference<ServiceResult<Asset>>();
 
         transactionContext.execute(() -> {
-            if (index.findById(asset.getId()) == null) {
+            if (findbyId(asset.getId()) == null) {
```
paullatzelsperger, line 87: nit: `findById` (character-case)

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/service/ContractDefinitionService.java
```diff
@@ -0,0 +1,56 @@
...
+
+public interface ContractDefinitionService {
+
+    /**
+     * Returns a contract definition by its id
+     *
+     * @param contractDefinitionId id of the contract definition
+     * @return the contract definition, null if it's not found
+     */
+    ContractDefinition findbyId(String contractDefinitionId);
```
paullatzelsperger, line 31:
```suggestion
    ContractDefinition findById(String contractDefinitionId);
```

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/service/ContractDefinitionService.java
```diff
@@ -0,0 +1,56 @@
...
+
+package org.eclipse.dataspaceconnector.api.datamanagement.contractdefinition.service;
+
+import org.eclipse.dataspaceconnector.api.result.ServiceResult;
+import org.eclipse.dataspaceconnector.spi.query.QuerySpec;
+import org.eclipse.dataspaceconnector.spi.types.domain.contract.offer.ContractDefinition;
+
+import java.util.Collection;
+
+public interface ContractDefinitionService {
```
paullatzelsperger, line 23: please add a bit of javadoc

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/service/ContractDefinitionServiceImpl.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Override
+    public ServiceResult<ContractDefinition> delete(String contractDefinitionId) {
+        var result = new AtomicReference<ServiceResult<ContractDefinition>>();
+
+        transactionContext.execute(() -> {
+            var filter = format("contractAgreement.id like %s:%%", contractDefinitionId);
+            var query = QuerySpec.Builder.newInstance().filter(filter).build();
+
+            var negotiationsOnAsset = contractNegotiationStore.queryNegotiations(query);
```
paullatzelsperger, line 84: why not use `contractNegotiationStore.findContractAgreement(contractDefinitionId)`?

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/service/ContractDefinitionServiceImpl.java
```diff
@@ -0,0 +1,101 @@
...
+    public ContractDefinitionServiceImpl(ContractDefinitionStore store, ContractDefinitionLoader loader, TransactionContext transactionContext, ContractNegotiationStore contractNegotiationStore) {
+        this.store = store;
+        this.loader = loader;
+        this.transactionContext = transactionContext;
+        this.contractNegotiationStore = contractNegotiationStore;
+    }
+
+    @Override
+    public ContractDefinition findbyId(String contractDefinitionId) {
+        var querySpec = QuerySpec.Builder.newInstance()
```
paullatzelsperger, line 48: We might need to add a `findById` method to the `ContractDefinitionStore`, because SQL-based stores may not have the query feature fully implemented for M3, just paging. No action to be taken on your part, just an FYI.

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/transform/ContractDefinitionDtoToContractDefinitionTransformer.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public Class<ContractDefinition> getOutputType() {
+        return ContractDefinition.class;
+    }
+
+    @Override
+    public @Nullable ContractDefinition transform(@Nullable ContractDefinitionDto object, @NotNull TransformerContext context) {
+        return ContractDefinition.Builder.newInstance()
+                .id(object.getId())
+                .accessPolicy(Policy.Builder.newInstance().id(object.getAccessPolicyId()).build())
```
paullatzelsperger, line 42: #828 will get rid of having the full `Policy` objects there. could you add a todo to indicate that?

paullatzelsperger:
About read access: all read access in services should be wrapped in a transaction, even if there is just one store or one call involved. E.g. we may want to change the transaction isolation level later, to guard against race conditions in clustered environments (=DB write locks).

I know that currently it is a bit cumbersome, because the `TransactionBlock#execute` only returns `void`, so we have to use `AtomicReferences`, but in my PR #865 I added a [commit](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/950/commits/321cf3525917fe5c5fce2cc9ce4849975de902f2) that improves that.

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/service/ContractDefinitionServiceImpl.java
ndr-brt: Because the contract agreement id is composed by the contract definition id plus a random generated UUID, so using that method will not return the agreement

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/service/ContractDefinitionServiceImpl.java
ndr-brt: Yes, will open an issue for this

ndr-brt:
@paullatzelsperger will open an issue about wrapping also read accesses by a transaction context, because I miss that also on the `Asset` part

paullatzelsperger:
> @paullatzelsperger will open an issue about wrapping also read accesses by a transaction context, because I miss that also on the `Asset` part

I'll extract the refactoring of the TranactionContext in a separate pr

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/service/ContractDefinitionServiceImpl.java
ndr-brt: #967

paullatzelsperger (approved):

extensions/api/data-management/contractdefinition/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractdefinition/ContractDefinitionApiController.java
```diff
@@ -69,21 +85,52 @@ public ContractDefinitionApiController(Monitor monitor) {
...
     }
 
     @POST
     @Override
     public void createContractDefinition(ContractDefinitionDto dto) {
         monitor.debug("create new contract definition");
-    }
+        var transformResult = transformerRegistry.transform(dto, ContractDefinition.class);
+        if (transformResult.failed()) {
+            throw new IllegalArgumentException("Request is not well formatted");
```
marcgs, line 102: Consider covering this BAD_REQUEST case in `ContractDefinitionApiControllerIntegrationTest`

PR #949, opened 2022-03-23
extensions: introduce Hostname service

### LINKED ISSUE #948: EDC baseUri service
# Feature Request

Introduce a service that supplies the EDC baseUri (protocol://host:port)

Every part where a callback from an external service is required.

To avoid baseUri duplication in configuration

## Solution Proposal
Provide a service that loads the base uri from configuration. Probably it will be provided by the jetty extension

## Type of Issue
improvement

ndr-brt:
## What this PR changes/adds

Introduce `BaseUriService`, that provides the base URI of the connector, the only implementation loads it from the configuration.

## Why it does that

To have a single point where this information is available

## Further notes

-

## Linked Issue(s)

Closes #948

## Checklist


spi/web-spi/src/main/java/org/eclipse/dataspaceconnector/spi/BaseUriService.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.spi;
+
+import org.eclipse.dataspaceconnector.spi.system.Feature;
+
+@FunctionalInterface
+@Feature("edc:web:base-uri")
+public interface BaseUriService {
```
jimmarino, line 21: Should this go in web-spi?

extensions/http/jetty/src/main/java/org/eclipse/dataspaceconnector/extension/jetty/ConfigBaseUriService.java
```diff
@@ -0,0 +1,41 @@
...
+package org.eclipse.dataspaceconnector.extension.jetty;
+
+import org.eclipse.dataspaceconnector.spi.BaseUriService;
+import org.eclipse.dataspaceconnector.spi.monitor.Monitor;
+import org.eclipse.dataspaceconnector.spi.system.configuration.Config;
+
+import static org.eclipse.dataspaceconnector.extension.jetty.JettyExtension.WEB_HTTP_URI_BASE;
+
+public class ConfigBaseUriService implements BaseUriService {
+    protected static final String DEFAULT_BASE_URI = "http://localhost:8181";
```
jimmarino, line 24: How do we synchronize/align this with the port an API will use? The base URI may need to just be the domain

ndr-brt:
@jimmarino I wasn't considering the different port mappings, so probably this service should return only the hostname, without protocol nor port, will rename it accordingly

jimmarino (approved):

jimmarino (commented):
Is this the Http base address or host name? If it is the HTTP base address, should it go in web-spi?

ndr-brt:
At this point I think this should represent only the hostname, as the port, as you said depends on the context, and same is for protocol. 
This define an agnostic property that reports the hostname (or domain name) and can be used to construct callback addresses for external services. The protocol will be defined by the context.

core/base/README.md
```diff
@@ -0,0 +1,7 @@
+# core-base
+
+## Configuration settings
+
+| Parameter name | Description        | Mandatory | Default value |
+|----------------|--------------------|-----------|---------------|
+| `edc.hostname` | Connector hostname | false     | localhost     |
```
florianrusch-zf, line 7: Could we put a little more information here so that users who are not looking at the code understand what this configuration is used for?

Maybe something like this:

```suggestion
| `edc.hostname` | Connector hostname, which e.g. is used in referer urls | false     | localhost     |
```

florianrusch-zf (changes requested):

core/base/README.md
ndr-brt: seems fair, done

florianrusch-zf (approved):

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/Hostname.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.eclipse.dataspaceconnector.spi.system;
+
+@FunctionalInterface
+@Feature("edc:hostname")
+public interface Hostname {
+
+    /**
+     * Gives the hostname of the connector.
```
paullatzelsperger, line 22:
```suggestion
     * Provides the hostname of the connector.
```

ndr-brt:
@jimmarino I think can be merged

PR #950, opened 2022-03-23
Feature: Adds SQL-based implementation of the `TransferProcessStore`

### LINKED ISSUE #865: SQL: implementation of `TransferProcessStore`

paullatzelsperger:
## What this PR changes/adds

Adds an SQL-based implementation of the `TransferProcessStore`.

## Why it does that

Required for Milestone 3.

## Further notes

- A generic `Lease` mechanism has been introduced, which is required to exclusively lock `TransferProcess` entities. For this purpose another module called `:extensions:sql:lease` has been introduced. Subsequent implementations, e.g. for `ContractNegotiations` (#864) should make use of this.
- An override for the `TransactionContext#execute` was added to allow easily wrapping read calls into transactions. This feature has been backported to the `SqlContractDefinitionStore` as well.

## Linked Issue(s)

Closes #865 

## Checklist


extensions/sql/lease/src/main/java/org/eclipse/dataspaceconnector/sql/lease/SqlLeaseContextConfiguration.java
```diff
@@ -0,0 +1,68 @@
...
+import org.eclipse.dataspaceconnector.spi.transaction.TransactionContext;
+
+import java.sql.Connection;
+import java.time.Duration;
+import java.util.Objects;
+
+/**
+ * Utility class to dynamically assemble a {@link SqlLeaseContext} and execute operations on it using a {@link Connection}.
+ */
+public class SqlLeaseContextConfiguration {
```
bcronin90, line 26: Isn't this basically a builder? I think it would be cleaner to split the "withConnection()" up into a "connection()" that sets the connection and a "build()" that builds.

extensions/sql/transfer-process-store/doc/ddl_postgres.sql
```diff
@@ -14,15 +15,16 @@ COMMENT ON COLUMN edc_lease.leased_at IS 'posix timestamp of lease';
...
 COMMENT ON COLUMN edc_lease.lease_duration IS 'duration of lease in milliseconds';
 
+DROP TABLE IF EXISTS edc_transfer_process;
 CREATE TABLE IF NOT EXISTS edc_transfer_process
 (
     id                       VARCHAR           NOT NULL
         CONSTRAINT transfer_process_pk
             PRIMARY KEY,
-    type                     INTEGER DEFAULT 0 NOT NULL,
+    type                     INTEGER           NOT NULL,
```
bcronin90, line 24: These two enums should probably be persisted by their String value to avoid issues with changing enum definitions.

extensions/sql/transfer-process-store/src/test/java/org/eclipse/dataspaceconnector/sql/transferprocess/store/SqlTransferProcessStoreTest.java
```diff
@@ -0,0 +1,474 @@
...
+        IntStream.range(0, 10)
+                .mapToObj(i -> createTransferProcess(String.valueOf(i)))
+                .forEach(store::create);
+
+        var qs = QuerySpec.Builder.newInstance().limit(10).offset(12).build();
+        assertThat(store.findAll(qs)).isEmpty();
+
+    }
+
+    @Test
```
bcronin90, line 424: I personally don't like these empty tests. I know they represent functionality that will be added later, but I'd rather have them deleted or marked for skipping.

extensions/transaction/transaction-atomikos/src/main/java/org/eclipse/dataspaceconnector/transaction/atomikos/AtomikosTransactionContext.java
```diff
@@ -66,17 +66,49 @@ public void execute(TransactionBlock block) {
...
+        } finally {
+            commitOrRollback(startedTransaction);
+        }
+    }
+
+    private void commitOrRollback(boolean startedTransaction) {
+        if (startedTransaction) {
+            try {
+                var status = transactionManager.getStatus();
+                if (STATUS_ACTIVE == status) {
```
bcronin90, line 105: Implementing a case for other statuses would probably be good.

extensions/transaction/transaction-atomikos/src/main/java/org/eclipse/dataspaceconnector/transaction/atomikos/AtomikosTransactionContext.java
```diff
@@ -66,17 +66,49 @@ public void execute(TransactionBlock block) {
...
+                    transactionManager.commit();
+                } else if (STATUS_MARKED_ROLLBACK == status) {
+                    transactionManager.rollback();
                 }
+            } catch (HeuristicRollbackException | SystemException | HeuristicMixedException | RollbackException e) {
+                monitor.severe("Transaction error", e);
```
bcronin90, line 111: Exception should be rethrown.

extensions/transaction/transaction-local/src/main/java/org/eclipse/dataspaceconnector/transaction/local/LocalTransactionContext.java
```diff
@@ -87,6 +93,29 @@ public void registerResource(LocalTransactionResource resource) {
...
+                    } catch (Exception e) {
+                        monitor.severe("Error rolling back resource", e);
+                    }
+                });
+            } else {
+                resources.forEach(localTransactionResource -> {
+                    try {
+                        localTransactionResource.commit();
+                    } catch (Exception e) {
+                        monitor.severe("Error committing resource", e);
```
bcronin90, line 111: Exception should be rethrown.

extensions/sql/lease/src/main/java/org/eclipse/dataspaceconnector/sql/lease/SqlLeaseContext.java
```diff
@@ -0,0 +1,120 @@
...
+        this.connection = connection;
+    }
+
+    @Override
+    public void breakLease(String entityId) {
+        trxContext.execute(() -> {
+
+            var l = getLease(entityId);
+
+            if (l != null) {
```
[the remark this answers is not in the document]
paullatzelsperger: then there is no lease to break, so it's a NOOP.

extensions/sql/lease/src/main/java/org/eclipse/dataspaceconnector/sql/lease/SqlLeaseContextConfiguration.java
paullatzelsperger: it's very similar, but not quite. Users should not store references to the `SqlLeaseContext`, because connections should not be kept open. I know that it is still possible, but less obvious at least. The idea is that users configure the `SqlLeaseContextConfiguration` once, and then call `withConnection(...).[acquire|break]` several times. So the intent here was not to create a builder, that sets several potentially optional fields. I'll update the source comments.

extensions/sql/transfer-process-store/doc/ddl_postgres.sql
paullatzelsperger: good point. will update.

extensions/sql/transfer-process-store/src/test/java/org/eclipse/dataspaceconnector/sql/transferprocess/store/SqlTransferProcessStoreTest.java
paullatzelsperger: i don't either. i simply forgot to add them :/

extensions/transaction/transaction-atomikos/src/main/java/org/eclipse/dataspaceconnector/transaction/atomikos/AtomikosTransactionContext.java
paullatzelsperger: Can't really comment, that would probably be a topic for @jimmarino. In any event I think it's out of scope for this PR. It only turns up as changed because i added the execute-with-result temporarily.

bcronin90 (approved):

juliapampus (approved):

PR #957, opened 2022-03-24
extensions: implement ContractNegotiation Data Management Api

### LINKED ISSUE #807: DataManagementApi: add `ContractNegotiationService` and `ContractNegotiationTransformer`
For the implementation of the contract negotiation endpoint we need an `ContractNegotiationService` (and an `ContractNegotiationServiceImpl`) with roughly the following signature:
```java
public interface ContractNegotiationService {
  ContractNegotiation findbyId(String contractNegotiationId);
  Collection<ContractNegotiation> query(QuerySpec query);
  String getState(String negotiationId);
  Result cancel(String negotiationId);
  Result decline(String negotiationId, String reason);
  ContractAgreement getForNegotiation(String negotiationId);
  String initiateNegotiation(DataRequest request); // transformed from NegotiationInitiateRequestDto
}
```

In addition, there needs to be an `ContractNegotiationTransformer`, that converts an `ContractNegotiation` into a `ContractNegotiationDto` and back as well as a `ContractAgreementTransformer. This depends on #779 

Important aspects:
- service must perform defensive checks before performing an operation:
  - `cancel`: must check first, whether a `ContractNegotiation` can be cancelled according to valid state transitions
  - `decline`: check whether a `ContractNegotiation` can be declined according to valid state transitions
- state transitions must be done through the `CommandQueue` mechanism and their `Result` only reflects the successful submission of the command

ndr-brt:
## What this PR changes/adds

Implements ContractDefinition service for Data Management Api.
Depends on #940

## Why it does that

To expose a clear and RESTful api to operate on ContractContractNegotiation

## Further notes

* The `NegotiationInitiateRequestDto` contains an `offerId`, while the `ContractOfferRequest` in which it is transformed contain a full `ContractOffer` object. For the moment I created it with a `ContractOffer` that has only the id, but's obvious that we need a better process, probably we need to include a `ContractOfferDto` into the `NegotiationInitiateRequestDto`, or retrieve it from somewhere, I need some hints on this.
* Removed `negotiationId` from `ContractAgreementDto` as it seems to be unused (and since to get the agreement we need to pass from negotiation it would be needless)

## Linked Issue(s)

Closes #807

## Checklist

### DIFF
--- a/extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/service/ContractNegotiationServiceImpl.java
+++ b/extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/service/ContractNegotiationServiceImpl.java
@@ -17,7 +17,6 @@ package org.eclipse.dataspaceconnector.api.datamanagement.contractnegotiation.se
 import org.eclipse.dataspaceconnector.api.result.ServiceResult;
 import org.eclipse.dataspaceconnector.contract.negotiation.command.commands.CancelNegotiationCommand;
 import org.eclipse.dataspaceconnector.spi.contract.negotiation.ConsumerContractNegotiationManager;
-import org.eclipse.dataspaceconnector.spi.contract.negotiation.response.NegotiationResult;
 import org.eclipse.dataspaceconnector.spi.contract.negotiation.store.ContractNegotiationStore;
 import org.eclipse.dataspaceconnector.spi.iam.ClaimToken;
 import org.eclipse.dataspaceconnector.spi.query.QuerySpec;
@@ -26,11 +25,9 @@ import org.eclipse.dataspaceconnector.spi.types.domain.contract.agreement.Contra
 import org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiation;
 import org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractNegotiationStates;
 import org.eclipse.dataspaceconnector.spi.types.domain.contract.negotiation.ContractOfferRequest;
-import org.eclipse.dataspaceconnector.spi.types.domain.contract.offer.ContractDefinition;
 
 import java.util.Collection;
 import java.util.Optional;
-import java.util.concurrent.atomic.AtomicReference;
 
 import static java.lang.String.format;
 import static java.util.stream.Collectors.toList;
@@ -49,12 +46,12 @@ public class ContractNegotiationServiceImpl implements ContractNegotiationServic
 
     @Override
     public ContractNegotiation findbyId(String contractNegotiationId) {
-        return store.find(contractNegotiationId);
+        return transactionContext.execute(() -> store.find(contractNegotiationId));
     }
 
     @Override
     public Collection<ContractNegotiation> query(QuerySpec query) {
-        return store.queryNegotiations(query).collect(toList());
+        return transactionContext.execute(() -> store.queryNegotiations(query).collect(toList()));
     }
 
     @Override
@@ -69,49 +66,40 @@ public class ContractNegotiationServiceImpl implements ContractNegotiationServic
 
     @Override
     public ContractAgreement getForNegotiation(String negotiationId) {
-        return store.findContractAgreement(negotiationId);
+        return transactionContext.execute(() -> store.findContractAgreement(negotiationId));
     }
 
     @Override
     public ContractNegotiation initiateNegotiation(ContractOfferRequest request) {
-        var result = manager.initiate(request);
-        return result.getContent();
+        return transactionContext.execute(() -> manager.initiate(request).getContent());
     }
 
     @Override
     public ServiceResult<ContractNegotiation> cancel(String negotiationId) {
-        var result = new AtomicReference<ServiceResult<ContractNegotiation>>();
-
-        transactionContext.execute(() -> {
+        return transactionContext.execute(() -> {
             var negotiation = store.find(negotiationId);
             if (negotiation == null) {
-                result.set(ServiceResult.notFound(format("ContractNegotiation %s does not exist", negotiationId)));
-                return;
+                return ServiceResult.notFound(format("ContractNegotiation %s does not exist", negotiationId));
+            } else {
+                manager.enqueueCommand(new CancelNegotiationCommand(negotiationId));
+                return ServiceResult.success(negotiation);
             }
-            manager.enqueueCommand(new CancelNegotiationCommand(negotiationId));
-            result.set(ServiceResult.success(negotiation));
         });
-
-        return result.get();
     }
 
     @Override
     public ServiceResult<ContractNegotiation> decline(String negotiationId) {
-        var result = new AtomicReference<ServiceResult<ContractNegotiation>>();
-
-        transactionContext.execute(() -> {
+        return transactionContext.execute(() -> {
             try {
                 var declineResult = manager.declined(ClaimToken.Builder.newInstance().build(), negotiationId);
                 if (declineResult.succeeded()) {
-                    result.set(ServiceResult.success(declineResult.getContent()));
+                    return ServiceResult.success(declineResult.getContent());
                 } else {
-                    result.set(ServiceResult.conflict(format("Cannot decline ContractNegotiation %s", negotiationId)));
+                    return ServiceResult.conflict(format("Cannot decline ContractNegotiation %s", negotiationId));
                 }
             } catch (Exception e) {
-                result.set(ServiceResult.conflict(format("Cannot decline ContractNegotiation %s: %s", negotiationId, e.getLocalizedMessage())));
+                return ServiceResult.conflict(format("Cannot decline ContractNegotiation %s: %s", negotiationId, e.getLocalizedMessage()));
             }
         });
-
-        return result.get();
     }
 }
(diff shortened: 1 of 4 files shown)


algattik:
It can be challenging to write UTs for mapper code, that (1) do not simply repeat the code under test, (2) properly test invalid output without too much boilerplate, (3) is not fixed to specific input values.

Here is a structure (for another, but similar PR) that I think helps with those, building "correct" inputs and outputs in setup, with fake data, and then allowing methods to corrupt the data and verify expectations. It also avoids redundancy if testing roundtrip.


```
class TransferProcessToTransferProcessDtoTransformerTest {
    static Faker faker = new Faker();

    DtoTransformerRegistry registry = mock(DtoTransformerRegistry.class);
    TransformerContext context = new DtoTransformerRegistryImpl.DtoTransformerContext(registry);
    String id = faker.lorem().word();
    TransferProcess.Type type = faker.options().option(TransferProcess.Type.class);
    TransferProcessStates state = faker.options().option(TransferProcessStates.class);
    String errorDetail = faker.lorem().word();

    public TransferProcess.Builder entity = TransferProcess.Builder.newInstance()
            .id(id)
            .type(type)
            .state(state.code())
            .errorDetail(errorDetail);

    TransferProcessDto.Builder dto = TransferProcessDto.Builder.newInstance()
            .id(id)
            .type(type.name())
            .state(state.name())
            .errorDetail(errorDetail);

    TransferProcessToTransferProcessDtoTransformer transformer = new TransferProcessToTransferProcessDtoTransformer();
    List<String> problems = new ArrayList<>();
    
    @Test
    void transform() {
        assertThatEntityTransformsToDto();
    }

    @Test
    void transform_whenInvalidState() {
        entity.state(invalidStateCode);
        dto.state(null);
        problems.add("Invalid value for TransferProcess.state");

        assertThatEntityTransformsToDto();
    }

    void assertThatEntityTransformsToDto() {
        var result = transformer.transform(entity.build(), context);

        assertThat(result)
                .usingRecursiveComparison()
                .isEqualTo(dto.build());

        assertThat(context.getProblems()).containsExactlyElementsOf(problems);
    }
}
````

**Comment by ndr-brt** [comment]
@algattik yeah, that seems a good approach, but currently is not applicable, at least to contract negotiations, since not all the fields are equal (e.g. `ContractNegotiation` has a `ContractAgreement` instance, while the `Dto` has only the id).
Could be something to address to another issue for sure.

**Comment by algattik** [comment]
> @algattik yeah, that seems a good approach, but currently is not applicable, at least to contract negotiations, since not all the fields are equal (e.g. `ContractNegotiation` has a `ContractAgreement` instance, while the `Dto` has only the id).
> Could be something to address to another issue for sure.

From this answer it seems you might be thinking that the proposed code compares the mapping input and output directly by reflection? That is not what it does, maybe you should have another look. It does not assume any identity or similarity between input and output.

**Comment by ndr-brt** [comment]
> > @algattik yeah, that seems a good approach, but currently is not applicable, at least to contract negotiations, since not all the fields are equal (e.g. `ContractNegotiation` has a `ContractAgreement` instance, while the `Dto` has only the id).
> > Could be something to address to another issue for sure.
> 
> From this answer it seems you might be thinking that the proposed test code compares the mapping input and output directly by reflection? That is not what it does, maybe you should have another look. It does not assume any identity or similarity between input and output.

I will, my suggestion tho, is to address this optimization to another issue/pr

File: extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/service/ContractNegotiationServiceImpl.java [hunk 1]
```diff
@@ -0,0 +1,117 @@
...
+    public ContractNegotiation initiateNegotiation(ContractOfferRequest request) {
+        var result = manager.initiate(request);
+        return result.getContent();
+    }
+
+    @Override
+    public ServiceResult<ContractNegotiation> cancel(String negotiationId) {
+        var result = new AtomicReference<ServiceResult<ContractNegotiation>>();
+
+        transactionContext.execute(() -> {
```
marcgs, line 85: You can now directly return the result of the `execute` method without using an `AtomicReference`. See #968 .

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/service/ContractNegotiationServiceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public ContractNegotiationServiceImpl(ContractNegotiationStore store, ConsumerContractNegotiationManager manager, TransactionContext transactionContext) {
+        this.store = store;
+        this.manager = manager;
+        this.transactionContext = transactionContext;
+    }
+
+    @Override
+    public ContractNegotiation findbyId(String contractNegotiationId) {
+        return store.find(contractNegotiationId);
```
marcgs, line 52: Shouldn't all service methods (including read-only) be made transactional? This would allow for service composition with a common parent transaction. @paullatzelsperger

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/service/ContractNegotiationServiceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        return result.get();
+    }
+
+    @Override
+    public ServiceResult<ContractNegotiation> decline(String negotiationId) {
+        var result = new AtomicReference<ServiceResult<ContractNegotiation>>();
+
+        transactionContext.execute(() -> {
+            try {
+                var declineResult = manager.declined(ClaimToken.Builder.newInstance().build(), negotiationId);
```
marcgs, line 104: Not sure to understand why here we are not using a command to decline the negotiation like in the `cancel` method

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/ContractNegotiationApiController.java
```diff
@@ -63,63 +78,104 @@ public ContractNegotiationApiController(Monitor monitor) {
...
     public String initiateContractNegotiation(NegotiationInitiateRequestDto initiateDto) {
-
         if (!isValid(initiateDto)) {
             throw new IllegalArgumentException("Negotiation request is invalid");
         }
 
-        return "not-implemente"; //will be the negotiation-id
+        var transformResult = transformerRegistry.transform(initiateDto, ContractOfferRequest.class);
+        if (transformResult.failed()) {
+            throw new IllegalArgumentException("Negotiation request is invalid");
```
marcgs, line 135: I would suggest covering this case in `ContractNegotiationApiControllerIntegrationTest`

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/service/ContractNegotiationServiceImpl.java
ndr-brt: done

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/service/ContractNegotiationServiceImpl.java
ndr-brt: yes, we already discussed this, since it's something that came out after 2 api pr were already merged, I proposed to address this in another issue (#966),  but since it was pretty easy I did it here also on the other service also if it goes out of pr scope.

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/service/ContractNegotiationServiceImpl.java
ndr-brt: it was like this since the first version of the `ContractNegotiation` interface. Please address this to another issue

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/ContractNegotiationApiController.java
ndr-brt: It's not possible, since currently there's no possibility that the transformer returns a failure. it was did like this to respect the `Result` api and avoid possible unexpected exceptions in the future.

paullatzelsperger:
> The NegotiationInitiateRequestDto contains an offerId, while the ContractOfferRequest in which it is transformed contain a full ContractOffer object. For the moment I created it with a ContractOffer that has only the id, but's obvious that we need a better process, probably we need to include a ContractOfferDto into the NegotiationInitiateRequestDto, or retrieve it from somewhere, I need some hints on this.

Every connector should maintain in its catalog a list of `ContractOffers`, that other connectors have presented. Therefor the way to do this is to resolve the `ContractOffer` for that particular `offerId` using the `org.eclipse.dataspaceconnector.catalog.spi.QueryEngine`

paullatzelsperger (approved):

PR #970, opened 2022-03-25
sample 4: update permission

ronjaquensel:
## What this PR changes/adds

It updates the permission in the `transfer-file` module of sample 04.0. It changes the permission's action from `idsc:USE` to `USE` and adds a permission target.

## Why it does that

Currently, upon description request the transformer throws an error when transforming the permission from sample 4 due to a missing target and an unknown action type. While the description request is not required for sample 4 to run successfully, people might look at this sample and define their policies in the same way, which would then lead to these errors.


DominikPinsel (approved):
Good catch :)

ndr-brt (approved):

PR #976, opened 2022-03-25
chore: Add data address documentation

jimmarino:
## What this PR changes/adds

Nothing, it only documents the concept of a data address

## Checklist


docs/developer/data_address_types.md
```diff
@@ -0,0 +1,124 @@
...
+
+For pull scenarios, there is no Destination Data Address sent by the consumer.
+
+##### Asset Data Addresses (ADA)
+
+An Asset Data Address is loaded along with an `Asset` via the _Data Management API_. The `DataAddressResolver` is used to resolve the ADA when a provider processes a transfer request. The ADA always exists (e.g., for both pull and push scenarios).
+
+##### Content Data Address (CDA)
+
+A Content Data Address (CDA) is created from an ADA. In pull scenarios, the CDA is sent to the consumer. in push scenarios, the provider uses the CDA to resolve asset content and send it to the consumer.
```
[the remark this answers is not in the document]
jimmarino: The CDA is used to resolve the data content location. The CDA could be sent to the client directly (assuming it is public) or used by the DPF behind-the-scenes if the CDA is not a public address.

paullatzelsperger (approved):

docs/developer/data_address_types.md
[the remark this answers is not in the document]
MoritzKeppler: if it's just used behind-the-scenes, what type of data address is then being sent to the client?

docs/developer/data_address_types.md
```diff
@@ -0,0 +1,124 @@
+# A Note On Data Address Types
+
+A **_Data Address_** is a location reference used by EDC runtimes. This note outlines the different data address types and their use.
```
MoritzKeppler, line 3: one question from a discussion we have in parallel:
Is the data address just the **location** of the data or does it also include **"all knowledge to retrieve the data"**. E.g. in case of an HTTP typed address: is it just the URL or all values that are needed to perform an HTTP call to retrieve the data (potentially including http method, http headers or even a body that needs to be passed).

MoritzKeppler: Thanks, I agree, especially with the out-of-band storage of secrets. 
I thought about edge cases. What if data can just be retrieved if a special http header is set, for example `Accept` to define the expected mime type? 
That cannot be included in a URL. So it's no just a "location reference" but also any knowledge that is needed to retrieve data.

docs/developer/data_address_types.md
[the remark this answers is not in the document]
jimmarino: In the case of sync data transfers (e.g. proxied APIs), that' is decided by the data plane DataFlowController. In that case, the CDA would probably not be given out.

docs/developer/data_address_types.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### Asset Data Address (ADA)
+
+An **_Asset Data Address_** is associated with an asset on the provider and used by the latter to locate asset data. 
+
+#### Content Data Address (CDA)
+
+A **Content Data Address_** is used to resolve asset data. An CDA may be used by the provider to resolve data that is in turn pushed to the consumer as specified by a DDA. Alternatively, in pull transfer scenarios, an CDA may be sent to the consumer, which uses it to access the data.
+
+#### ADAs and SDAs in the Control Plane
```
bscholtes1A, line 21: SDA?

docs/developer/data_address_types.md
```diff
@@ -0,0 +1,124 @@
...
+
+In some cases, an ADA may be used as an CDA; in other cases, there is no direct mapping. Consider two scenarios.
+
+##### Scenario 1: Client API Access
+
+A consumer requests API access modeled as an asset from a provider. The provider resolves the ADA and passes it back to the consumer to pull the data (access the API). In this case, the ADA and CDA are the same.
```
bscholtes1A, line 27: This case can be more complex as the provide control plane can actually provide a CDA which is not the ADA. For example the CDA can be the public API of the Data Plane, which is used as proxy to access the ADA (the ADA is wrapped in the access token passed to the consumer for accessing the public API of the DPF)

docs/developer/data_address_types.md
```diff
@@ -0,0 +1,124 @@
...
+
+##### Scenario 2: Data Provisioned from a Backend System
+
+A consumer requests a big data set that must be prepared by a backend system. The provider resolves the ADA during the provision phase. Based on the ADA, a provisioner is engaged and communicates with the backend system to generate the big data. The backend system places the generated data in object storage. When the data is available, the backend system (asynchronously) calls back into the EDC, sending an CDA that resolves to the object storage. At that point, the data transfer proceeds, and the CDA can either be used by the provider to push the data to the client (via the DDA), or the CDA sent to the client to pull the data.
```
bscholtes1A, line 31: This scenario is a bit tricky to understand who provision what. Maybe a small draft would help?

docs/developer/data_address_types.md
```diff
@@ -0,0 +1,124 @@
...
+
+#### DDA, ADA, and CDA Propagation in the Control Plane
+
+##### Destination Data Addresses (DDA)
+
+For push scenarios, a Destination Data Address (DDA) is sent by the consumer to the provider as the `dataDestination` contained in the `DataRequest`. The `DataRequest` is associated with the provider `TransferProcess` and propagated through the runtime. The DDA can either be provided to the consumer EDC out-of-band (as part of the client operation that initiates a transfer request, e.g. when a destination is managed by an external system), or created by a `Provisioner` and associated with a transfer process.a
```
bscholtes1A, line 37: typo: process.a

docs/developer/data_address_types.md
bscholtes1A, line 47: typo: . **I**n push scenario

docs/developer/data_address_types.md
```diff
@@ -0,0 +1,124 @@
...
+
+A consumer requests data that it pulls from the provider over HTTP.
+
+
+###### Consumer Data Plane (Originator)
+
+| Control Plane  | DPF                |
+|:---------------|:-------------------|
+| CDA            | Source Address     |
+| Consumer Sink  | Destination Address|
```
bscholtes1A, line 85: Not sure here. 

In pull http scenario, the data is directly returned in output of the request (synchronous). There is no DDA, and thus no Consumer Sink.

docs/developer/data_address_types.md
```diff
@@ -0,0 +1,124 @@
...
+#### Pull Object Storage
+
+A consumer requests data that it pull from the provider's object storage.
+
+###### Consumer Data Plane (Originator)
+
+| Control Plane  | DPF                 |
+|:---------------|:--------------------|
+| CDA            | Source Address      |
+| Consumer Sink  | Destination Address |
```
bscholtes1A, line 120: Who triggers the actual data transfer here?
My understanding that only the provider executes the `DataFlowController`, but maybe I am wrong. 
If I am correct, then how consumer can be the originator of the data transfer?

Issue #902, opened 2022-03-15
DPF Test and Sample Updates


zeier:
## Description

Samples and tests currently use the `InlineDataFlowController`, which was a temporary solution until the Data Plane Framework was in place.

Issue #871 will use the `HttpDataFlowController` to perform communication between the Control Plane and the Data Plane.

The goal of this issue is to adapt samples and tests to use the Data Plane and validate that end to end observability works as expected.

## Acceptance Criteria

- [ ] Usage of `InlineDataflowController` is replaced by the `HTTPDataFlowController`
- [ ] Samples and system tests leverage the Data Plane for data transfer
- [ ] Existing samples can be used to show end to end observability (distributed tracing) for data transfers where both the Control Plane and the Data Plane are involved
- [ ] Test are available for both Azure Object Storage extensions (integration test for the simple data transfer and system test for Azure D) and can be used to show end to end observability (distributed tracing) for data transfers where both the Control Plane and the Data Plane are involved

zeier:
@bscholtes1A is my understanding correct that you will cover this issue as part of your ongoing work? 
If not, please let me know and we can coordinate accordingly.

bscholtes1A:
> @bscholtes1A is my understanding correct that you will cover this issue as part of your ongoing work? If not, please let me know and we can coordinate accordingly.

Hi @zeier 
I will tackle the creation of the Control Plane client that delegates the work to the DPF and the possibility to have the DPF embedded within the Control Plane runtime (which will be the config used in the samples...)
However I did not plan to update the samples by myself as there are still a lot to do on the DPF for this milestone.

zeier:
> 

Hi @bscholtes1A 
OK, could you please ping us when you have your changes in, so that we can update the samples?

bscholtes1A:
> > 
> 
> Hi @bscholtes1A OK, could you please ping us when you have your changes in, so that we can update the samples?

As soon as #988 is merged

PR #988, opened 2022-03-28
DataPlane: Introduces a client enabling to delegate data transfer to the DPF

### LINKED ISSUE #871: Data Plane Framework - Provider Data Push
This is a child issue of #463. 

## Overview

The Data Plane Framework (DPF) should support two transfer directions:
1. client pull (sync): transfer is triggered by the consumer backend applications which query actively the data using their DPF.
2. provider push (async): once agreement request is successfully processed, the provider control plane sends a request to the DPF that triggers the data transfer from the provider data source to the consumer data destination. Data transfer can be finite or non-finite (event streaming).

Two kinds of API are therefore required within the DPF (each API should be registered on a separated context):
- a Public API used for client pull, which can be directly targeted by the backend applications,
- a Control API used for provided push which is only hit by the Control Plane. 

**The present issue only covers the asynchronous provider data push and the associated DPF Control API**, as Synchronous Data Pull has already been covered extensively in #585. 

## Requirement

- success of the artifact request directly triggers the data transfer, 
- DPF can be part of the Control Plane runtime (testing, samples...) or it can run as a separated runtime, thus the DPF should provide:
  - a Java API enabling to run the DPF directly from the Control Plane,
  - a REST API that can be hit remotely by the Control Plane through a dedicated client. 

## Design

The transfer orchestration belongs to the **Control Plane** runtime. All classes related to this orchestration are therefore placed in a dedicated extension `data-plane-transfer` which is separated from the `data-plane` module. 

The Provider Data Push mode is active for any `DataRequest` whose destination type is different from `HttpProxy`, which is reserved for Synchronous Data Pull (see #585 for more details). When the `DataRequest` is first received by the consumer Control Plane, it will first provision the data destination if needed (e.g. S3 bucket, Azure storage...), put the corresponding `DataAddress` in the `DataRequest` and finally send it to the provider Control Plane. 

The provider Control Plane will then map the `DataRequest` into `DataFlowRequest` (counterpart of the `DataRequest` in the DPF data model) and send this `DataFlowRequest` to the Control API exposed its DPF. The selection of the DPF instance is delegated to the DPF selector which is treated in a separated issue #832. We propose to rely on the existing `transfer-functions` extension for sending the `DataFlowRequest` to the DPF. 

The Control API of the DPF will then simply validate the received `DataFlowRequest` and, if valid, enqueue it in the `DataPlaneManager`. Note depending on the data source/destination combination some optimization of the data transfer are possible that avoids streaming the data through the DPF, which becomes crucial when dealing with massive data transfers. 

The whole sequence is depicted below:

provider_push drawio

bscholtes1A:
## What this PR changes/adds

This PR introduces a client enabling to delegate data transfers to the Data Plane (DPF) through its Control API. 

## Why it does that

As of current implementation, data transfers are today performed directly performed by the Control Plane through dedicated `DataFlowController`. 
This approach does not provide the ability to scale differently the Control Plane and the DPF as in real production environment there will potentially be many DPF instances controlled by the same Control Plane. 

In #585, we previously tackled synchronous data transfers, wherein a dedicated `DataFlowController` was introduced to orchestrate the "data pull" transfer mode by leveraging the Public API exposed by the DPF. 

In the present issue we add another `DataFlowController` implementation which is basically the asynchronous counterpart of the previous issue and enables to delegate data transfer request to one of the Data Plane instance. It supports the two possible topologies:
 - DPF embedded within the Control Plane, which is the exact counterpart of what we are doing with current implementation. This deployment strategy is essentially used for testing/demo purposes. Note that this will enable to remove `DataOperator` and `InlineDataFlowController` from the code base as it will be removed by the "embedded DPF". This replacement is tackled by #902. 
 - DPF runs in a separated runtime and should be accessed remotely through its Control API. The choice of the DPF instance that should run the transfer is dictated by the Data Plane Selector (see #832).

The flow diagram is depicted below:

provider_push drawio

## Linked Issue(s)

Closes #871.

## Checklist


CHANGELOG.md
```diff
@@ -39,6 +38,7 @@ the detailed section referring to by linking pull requests or issues.
 * Implement Asset service for Data Management API (#931)
 * Implement ContractDefinition service for Data Management API (#940)
 * Implement ContractNegotiation service for Data Management API (#957)
+* Add client to delete data transfer to Data Plane (#988)
```
MoritzKeppler, line 41:
```suggestion
* Let consumer delegate data transfer to Data Plane (#988)
```

MoritzKeppler:
I'm still failing to understand the exact differences of this mode and the sync mode... Need to get that sorted in my head :-)
My view:
* every data transfer will go through the data plane.
* every data transfer can include provisioning steps (e.g. provision data and get a CDA for it)
* the asset of a contract offer limits the way it can get transferred: e.g. "push via http", "pull via http", ... 
* this is independent from how the producer data plane will retrieve the asset (once again pull vs push possible) and how the consumer data plane will pass the asset to the backend application (again pull vs push etc.)
E.g. a consumer backend application may tell the control plane "please start transfer for asset X from contract Y. Give me the endpoint information so I can query it." => 'proxy mode'
or it is: "please start transfer for asset X and store it under that address" => depending if it's pull or push the consumer data plane would than pull the data or wait for data coming in and finally store it at the given address (i.e. push it) 

Does that make sense?

bscholtes1A:
> 

If the Asset you want to consume is a Rest API, how would you consume it without going through the sync mode? Especially:

-  how would pass the request parameters (method, path, query param...).
- performance-wise, how would you handle it? Going through the state machine (i.e. performing an Artifact Request) for every call to the Rest API is not an option (remember that a data consumer wants to poll the Rest API at regular interval to check for changes on the data...)

Also the DPF does not know about the Assets and Contracts. Creating the proxy which wraps the actual data address is only feasible from the Control Plane which is the only one to be able to resolve the data address associated with each asset. The same is for the Contracts, which are only known by the Control Plane.  

More info: #585 and #871

CHANGELOG.md
bscholtes1A: This is not related to the consumer. This PR is about offering the possibility to the Control Plane to delegate the data transfers to its Data Plane(s).

bscholtes1A:
@zeier 
Follow-up of our recent meeting, this is the one needed to get rid of the `ÌnlineDataFlowController` and related stuff, so that we can update the samples #902

zeier:
> @zeier Follow-up of our recent meeting, this is the one needed to get rid of the `ÌnlineDataFlowController` and related stuff, so that we can update the samples #902

@bscholtes1A, @marcgs will have a look at the PR this afternoon.

CHANGELOG.md
```diff
@@ -41,6 +40,7 @@ the detailed section referring to by linking pull requests or issues.
 * Implement ContractDefinition service for Data Management API (#940)
 * Implement ContractNegotiation service for Data Management API (#957)
 * In-memory implementation of PolicyStore (#930)
+* Let Control Plane delete data transfer to Data Plane (#988)
```
marcgs, line 43: Do you mean `delegate` instead of `delete`?

extensions/data-plane-transfer/data-plane-transfer-client/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+There are two possible configurations:
+
+- Data Plane is embedded in the Control Plane runtime: this is mostly a configuration used for testing, demo...
+- Data Plane "lives" in a separated runtime (generally in the same cluster): this is the typical production
+  configuration.
+
+## Data Plane is embedded in Control Plane runtime
+
+The cornerstone of the Data Plane is the `DataPlaneManager`. Thus, we rely on the presence/absence of this service in
```
marcgs, line 21: Could you specify what needs to be done to enable embedded VS remote transfer clients from an EDC client perspective? Which extensions need to be in place or absent in each case?

extensions/data-plane-transfer/data-plane-transfer-client/README.md
```diff
@@ -0,0 +1,37 @@
+# Data Plane Transfer Client extension
```
marcgs, line 1: I wonder if this documentation belongs rather in `/docs/developer' @paullatzelsperger

system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/dataspaceconnector/test/e2e/EndToEndTransferTest.java
```diff
@@ -45,6 +45,7 @@
 import static org.awaitility.Awaitility.await;
 import static org.eclipse.dataspaceconnector.common.testfixtures.TestUtils.getFreePort;
 import static org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcessStates.COMPLETED;
+import static org.eclipse.dataspaceconnector.transfer.dataplane.spi.DataPlaneTransferType.SYNC;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.notNullValue;
```
marcgs, line 51: This E2E test verifies the "HTTP pull" flow. I think we should write another `EndToEndTransferTest` for the "HTTP pull" flow using the `DataPlaneTransferFlowController` and `EmbeddedDataPlaneSelectorClient`.

jimmarino: +1 but we can also add this in another PR

extensions/data-plane-transfer/data-plane-transfer-client/README.md
jimmarino: Originally we had the idea of also including doc in extensions. We need to go through and rationalize the doc as a lot of it has become outdated.

jimmarino:
@MoritzKeppler Here are a few scenarios:

## Client pull, a CDA is generated by a backend system and is private

Some proxy modes are examples of this case (if they need to generate a CDA). Here are the steps as I see it:

1. Client Control Plane (CCP) issues a request.
2. Provider Control Plane (PCP) receives the request and issues a provision request to a backend system.
3. The backend system calls back the PCP at some later point and passes a CDA to resolve provisioned data.  
4. A token is generated by the PCP and sent to the CCP. A URL to access the Provider DPF is also sent.
5. The CCP passes the token and Provider DPF URL to its DPF.
6. The Client DPF invokes the Provider DPF passing the token.
7. The Provider DPF resolves the CDA and returns the content.


## Client pull, the CDA is generated by a backend system and is public.

1. Client Control Plane (CCP) issues a request.
2. Provider Control Plane (PCP) receives the request and issues a provision request to a backend system.
3. The backend system calls back the PCP at some later point and passes a CDA to resolve provisioned data.  
4. The PCP sends the CDA to the CCP.
5. The CCP passes the CDA to its DPF. 
6. The Client DPF uses the CDA to pull data. 

## Provider push, the CDA is generated by a backend system and is private

1. Client Control Plane (CCP) issues a request. In this request is a DDA where the data is to be pushed.
2. Provider Control Plane (PCP) receives the request and issues a provision request to a backend system.
3. The backend system calls back the PCP at some later point and passes a CDA to resolve provisioned data.  
4. The PCP passes the DDA and CDA to its DPF.
5. The Provider DPF uses the CDA to resolve the data and the DDA to push it to the destination. The destination may be the Client DPF or some other endpoint (e.g. object storage access).

jimmarino (approved):
I'm fine if we merge once the conflicts are resolved and make any required changes in subsequent PRs

system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/dataspaceconnector/test/e2e/EndToEndTransferTest.java
marcgs: I am ok with doing this in a separate PR as well. It still probably makes sense to do this work before starting #902 to prevent duplicating efforts with integration issues. WDYT @bscholtes1A ?

bscholtes1A: There should be no duplication. Samples will rely on the embedded client, while the future e2e test will rely on separated runtimes for the Control Plane, DPF, and possibility the Selector. So for #902 can be started once the present PR is merged.

marcgs: After a discussion with the team we agreed we will start with a simpler sample with the embedded client (#902), and probably add a second one for the advanced setup with the remote DPF in a follow-up story. We can take over writing the e2e as part of this follow-up. Does this work for you @bscholtes1A ?

marcgs:
Although the PR has been merged I think it is worth clarifying the following point mentioned by @bscholtes1A :
`DPF embedded within the Control Plane, which is the exact counterpart of what we are doing with current implementation. This deployment strategy is essentially used for testing/demo purposes.`
From my understanding the embedded DPF is a valid setup that can be used in simpler production setups as well. Is this correct @jimmarino ?

jimmarino:
Hi @marcgs I would not recommend embedding it for production cases but someone could do that.

Issue #176, opened 2021-10-21
gradle check is not failing if checkstyle violations are found


ndr-brt:
I noticed that my local execution of `./gradlew clean check` went well, but then I got an error from the github action.

I investigated a little and seems like the checkstyle gradle plugin mark the problems as warnings, but I think that this should change.

The solution is to set [`maxWarnings`](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.Checkstyle.html#org.gradle.api.plugins.quality.Checkstyle:maxWarnings) to `0`

PR #1000, opened 2022-03-29
CI: Dependency checks (#176)

### LINKED ISSUE #827: Code Quality: introduce dependency checks
## Requirements
In order to avoid "invalid" modules dependencies we'll introduce a possibility to check dependencies during build time.

There should be a tool with some sort of rule engine that evaluates every module based on a set of global rules and fails the build if one of them is violated. This tool should be executable as gradle task.

_Selecting an appropriate tool is part of this issue._

## Examples for rules would be:
- modules may only depend on `-spi` modules, with the following exceptions:
  - launchers and tests
  - `common`
  - `extensions:http:jetty` (this exception should be removed once there is an SPI for jetty)
- there should not be "cross-module" dependencies at the same level, for example `core:boot` should not depend on `core:contract`. Exceptions 
  - `core:policy` since there is no SPI for it.
  - technology libs, i.e. common modules like `azure-commons`
- `core:spi` cannot depend on any other module
- no module may depend directly on any of the `core` modules
- no module may depend on a `launcher`, `samples` or `system-tests` module
- no cyclic dependencies, direct or indirect

## Further validations:
- launchers should not reference two modules providing impls for the same interface
- there can not be two modules with identical names (that would confuse Gradle)
- there should not be unused dependencies

marcgs:
## What this PR changes/adds

 The CI verify workflow is extended with two Gradle tasks for analyzing EDC module dependencies:

 - A custom gradle plugin defines rules on which module dependencies are allowed within EDC (e.g. no module may depend directly on any of the core modules).
 - The [Dependency Analysis Gradle Plugin](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin) used for detecting unused dependencies and dependencies declared on the wrong configuration (`api` vs `implementation` vs `compileOnly`, etc.).

 The tasks are initially set to only emit warnings, but will be configured to fail builds once all warnings have been resolved.

Outputs can be seen in two tasks in the `Test Code (Style, Tests)` workflow -> `Dependency-analysis`:
- Dependency rules report (408 violations, see "Dependency Rules" in [CI run](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/runs/5805278157?check_suite_focus=true))
- Dependency analysis report (~500 lines, see "Dependecy Analysis" in [CI run](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/runs/5805278157?check_suite_focus=true))

## Why it does that

The custom Gradle plugin contains logic implementing the rules described in #827 . These rules can be executed locally using:

```
./gradlew applyDependencyRules -Pdependency.analysis=warn -Pdependency.analysis.clear.artifacts=false
```

 The [Gradle Java library plugin](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_recognizing_dependencies) defines a clear separation between dependency configurations. This allows controlling and limiting the use of transitive dependencies.

 The Java library plugin introduces the concept of ABI (Application Binary Interface) which includes (among others) types used in public methods. Dependencies that define those types should be defined in the `api` configuration. Other dependencies should be defined in the `implementation` configuration so that they are not transitively exposed, and therefore do not leak into the consumers' compile classpath.

 The Dependency Analysis Gradle Plugin analyzes the abstract syntax tree to determine the ABI. On that basis, it can recommend which import configurations should be changed. The dependency analysis can be executed locally using:

```
./gradlew buildHealth -Pdependency.analysis=warn -Pdependency.analysis.clear.artifacts=false
```

## Further notes

Modified some of the project files to implement the recommendations from the tool for the `core` module. This validated that the recommendations were reasonable:
* `:core:contract`:  scope of `:common:state-machine-lib` and `:common:util` changed from api to implementation. 
* `:core:contract`:  scope of `:common:state-machine-lib` changed from api to implementation. Removed unused dependency to `:common:util`

## Linked Issue(s)

Closes #827 

## Checklist


gradle.properties
```diff
@@ -1,4 +1,5 @@
 javaVersion=11
+org.gradle.jvmargs=-Xmx4096m
```
ndr-brt, line 2: Is this really necessary? 
This is equal to default value on my machine (this can be obtained by `java -XX:+PrintFlagsFinal -version | grep MaxHeapSize`), so, in my case this does nothing, aside that better not to have gradle specific arguments configuration checked in (we already did, and then we came back)

marcgs: Removed.

ndr-brt: It that causes problems in future only in the CI, a cli parameter on the CI script parameter could be added

.github/workflows/verify.yaml
```diff
@@ -32,6 +32,34 @@ jobs:
...
+  Unit-Test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-java@v2
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+
+      - name: Setup Gradle cache
```
ndr-brt, line 60: What's the difference of using `gradle-build-action` instead of the `cache: 'gradle'` attribute on the `setup-java`?
If this is better could be used in every job/workflow?

marcgs: The gradle build action has [some benefits](https://github.com/gradle/gradle-build-action#why-use-the-gradle-build-action) over the `setup-java` approach. Additionally `gradle.properties` are not included in the cache checksum when using `setup-java`. But it is true that it is unrelated to this PR, and could be done in a separate cleanup PR. I can revert this here and open a separate PR 👍🏻

ndr-brt (approved):

jimmarino (approved):

PR #1001, opened 2022-03-29
DPF: share thread pool

### LINKED ISSUE #618: DPF - Azure Object Storage Extension

algattik:
## What this PR changes/adds

Share a thread pool across all DPF transfers. Made number of threads configurable.
Instrument the thread pool with metrics.

## Why it does that

 This allows better control of resources.

## Further notes

Had to introduce `ExecutorInstrumentationImplementation` class for better control of optional dependencies.

## Linked Issue(s)

Closes #618 

## Checklist

### DIFF
--- a/core/base/src/main/java/org/eclipse/dataspaceconnector/core/CoreServicesExtension.java
+++ b/core/base/src/main/java/org/eclipse/dataspaceconnector/core/CoreServicesExtension.java
@@ -36,6 +36,7 @@ import org.eclipse.dataspaceconnector.spi.policy.RuleBindingRegistry;
 import org.eclipse.dataspaceconnector.spi.security.PrivateKeyResolver;
 import org.eclipse.dataspaceconnector.spi.system.BaseExtension;
 import org.eclipse.dataspaceconnector.spi.system.ExecutorInstrumentation;
+import org.eclipse.dataspaceconnector.spi.system.ExecutorInstrumentationImplementation;
 import org.eclipse.dataspaceconnector.spi.system.Hostname;
 import org.eclipse.dataspaceconnector.spi.system.Inject;
 import org.eclipse.dataspaceconnector.spi.system.Provides;
@@ -67,6 +68,7 @@ import static java.util.Optional.ofNullable;
         RemoteMessageDispatcherRegistry.class,
         RetryPolicy.class,
         RuleBindingRegistry.class,
+        ExecutorInstrumentation.class,
 })
 public class CoreServicesExtension implements ServiceExtension {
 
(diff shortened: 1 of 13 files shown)


jimmarino (commented):
Hi @algattik I'm not sure I follow why ExecutorInstrumentationImplementation is needed. 

I think we would only need to topologically sort the base extensions in `DefaultServiceExtensionContext.addDefaultExtensions` based on @Inject before they are added to the front of the load list. The core extension can then check if an `ExecutorInstrumentation` is already injected, and if not, add the default one.

That would avoid the issue of having to create an additional interface.

What do you think?

paullatzelsperger:
@algattik since @jimmarino reviews this, I took the liberty of removing myself. however, if there is anything you indeed want me to look at, feel free to re-request.

algattik:
> Hi @algattik I'm not sure I follow why ExecutorInstrumentationImplementation is needed.
> 
> I think we would only need to topologically sort the base extensions in `DefaultServiceExtensionContext.addDefaultExtensions` based on @Inject before they are added to the front of the load list. The core extension can then check if an `ExecutorInstrumentation` is already injected, and if not, add the default one.
> 
> That would avoid the issue of having to create an additional interface.
> 
> What do you think?

I've tried it, but it doesn't work. `CoreServicesExtension` needs to inject an (optional) `ExecutorInstrumentation`, and outjects one if none is provided. If we make it `@Provides(ExecutorInstrumentation.class)`, it creates a cyclic dependency error. If we don't, we get `EdcInjectionException: The following injected fields were not provided` in other classes.

algattik:
@jimmarino could we go ahead, and plan for the next milestone an improvement in IoC container to support things like default and named dependencies?

bscholtes1A (approved):

jimmarino:
> named

Let's discuss. The style of "named" dependencies I prefer is annotation-based like Guice, however, we should look if it is really needed.

PR #1007, opened 2022-03-30
Transfers: reliably send transfers from consumer to provider

### LINKED ISSUE #336: Consumer does not reattempt transfers when Provider not available at initiation time
If the consumer calls `org.eclipse.dataspaceconnector.spi.transfer.TransferProcessManager#initiateConsumerRequest` but the Provider is not available, the consumer does not receive an error. If the Provider later becomes available, the request does not get processed.

algattik:
## What this PR changes/adds

If the Consumer tries to initiate a transfer but the Provider is not available, the consumer does not manage the error. If the Provider later becomes available, the request did not get processed. This PR solves this issue.

If the request is unsuccessful, we keep the entity in the same state (with a state update, to increment its internal state count) and save it to break the lease and send the TP again potentially on the next iteration.

This "long retry" mechanism implements an exponential wait mechanism so as not to overflow the provider when it becomes available again. There might be many consumers queuing up requests to one provider.

After a given number of retries is exhausted, the transfer state moves to ERROR.

An additional "short retry" mechanism could also be implemented later in the IdsMultipartSender side, as a complement to this mechanism. The low level retry should be very short so as not to hog consumer resources and hammer the provider. 

## Why it does that

Reliable delivery

## Linked Issue(s)

Closes #336 

## Checklist

### DIFF
--- a/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
+++ b/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
@@ -27,6 +27,7 @@ import org.eclipse.dataspaceconnector.spi.message.RemoteMessageDispatcherRegistr
 import org.eclipse.dataspaceconnector.spi.monitor.Monitor;
 import org.eclipse.dataspaceconnector.spi.policy.store.PolicyArchive;
 import org.eclipse.dataspaceconnector.spi.response.ResponseStatus;
+import org.eclipse.dataspaceconnector.spi.retry.ExponentialWaitStrategy;
 import org.eclipse.dataspaceconnector.spi.retry.WaitStrategy;
 import org.eclipse.dataspaceconnector.spi.security.Vault;
 import org.eclipse.dataspaceconnector.spi.system.ExecutorInstrumentation;
@@ -53,6 +54,7 @@ import org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcess;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcessStates;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.command.TransferProcessCommand;
 
+import java.time.Clock;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
@@ -114,6 +116,7 @@ public class TransferProcessManagerImpl implements TransferProcessManager, Provi
     private StateMachine stateMachine;
     private DataAddressResolver addressResolver;
     private PolicyArchive policyArchive;
+    private SendRetryManager<TransferProcess> sendRetryManager;
 
     private TransferProcessManagerImpl() {
     }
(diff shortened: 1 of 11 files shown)


core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
```diff
@@ -570,42 +574,77 @@ private void processProviderRequest(TransferProcess process, DataRequest dataReq
         }
     }
 
+    private boolean processConsumerRequest(TransferProcess process, DataRequest dataRequest) {
+        int retryCount = process.getStateCount() - 1;
+        if (retryCount > 0) {
```
ndr-brt, line 579: This logic seems too complicated to handle it here, because it make also testing more difficult.
I would advice to introduce an interface like (names are just the first proposal, could be better for sure :) )
```java
private interface SendEntityRetryManager<T> {
    boolean shouldWaitNextTick(T entity);
}
```
(generic because this will likely be implemented for `ContractNegotiation` and probably others).

doing so the `processConsumerRequest` body will become only
```java
if (sendEntityRetry.shouldWaitNextTick(process)) {
    // Break lease
    transferProcessStore.update(process);

    return false;
} else {
    sendConsumerRequest(process, dataRequest);
    return true;
}
```
that's more easily testable, and then all the logic and the clock will be moved in a `TransferProcess` implementation of the interface, and one instance of that will become a collaborator of `TransferProcessManager` and it would become usable also in other parts of the manager if needed.

algattik: Implemented

ndr-brt (approved):
Tight! :+1:

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessSendRetryManager.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public boolean shouldDelay(TransferProcess process) {
+        int retryCount = process.getStateCount() - 1;
+        if (retryCount <= 0) {
+            return false;
+        }
+        var delayStrategy = delayStrategySupplier.get();
+        delayStrategy.failures(retryCount);
```
marcgs, line 36: The amount of failures are added to `errorCount` within the `failures` method here, but as far as I can tell the state count is increased as well every time a request fails. Are we counting things twice?

marcgs (approved):

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessSendRetryManager.java
algattik: Added comments to make the use of `Supplier` more visible

juliapampus:
@algattik Pls resolve the conflicts, I can merge afterwards

algattik:
@juliapampus I've resolved the conflicts, thanks in advance :)

jimmarino (approved):

PR #1015, opened 2022-03-30
Feature: Add SQL-based PolicyStore

### LINKED ISSUE #866: SQL: implement `PolicyStore`
An SQL-based implementation of the `PolicyStore` is needed. 

### Requirements
1. `save()`: implement upsert semantics
2. Querying: the query implementation must be flexible enough to accomodate  a `QuerySpec`.
2. Table Schema: serialize all related objects (Duty, Permission, ...) to JSON.
3. Indexing: queries will be targeting `Policy.id`.

_Please provide a written proposal for requirements 2 and 3 before beginning the implementation._

**This issue depends on #803**

- [x] Design Proposal: #920 
- [ ] Implementation (PR reference goes here)

tuncaytunc-zf:
## What this PR changes/adds

Add SQL PolicyStore implementation.

## Why it does that

To allow for persistence of PolicyStore

## Further notes
Documentatoin is not done yet.

## Linked Issue(s)

Closes #866 

## Checklist


extensions/sql/policy/store/src/main/java/org/eclipse/dataspaceconnector/sql/policy/store/SqlPolicyStore.java
```diff
@@ -0,0 +1,176 @@
...
+        });
+    }
+
+    @Override
+    public void save(Policy policy) {
+        Objects.requireNonNull(policy);
+
+        transactionContext.execute(() -> {
+            try (var connection = getConnection()) {
+                executeQuery(connection, sqlPolicyStoreStatements.getSqlSaveClauseTemplate(),
```
paullatzelsperger, line 96: I'd rather have an explicit "if-exists" check here (either using `SELECT` or `SELECT COUNT`), even if that's another SQL statement

extensions/sql/policy/store/src/test/java/org/eclipse/dataspaceconnector/slq/policy/store/SqlPolicyStoreTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+    @Test
+    @DisplayName("Find policy by ID that exists")
+    void findById_whenPresent() {
+        Policy policy =  getDummyPolicy(getRandomId());
+        sqlPolicyStore.save(policy);
+
+        var policyFromDb = sqlPolicyStore.findById(policy.getUid());
+
+        assertThat(policy.getUid()).isEqualTo(policyFromDb.getUid());
```
paullatzelsperger, line 127: this performs a stronger assertion:
```suggestion
assertThat(policyFromDb).usingRecursiveComparison().isEqualTo(policy);
```

also you got `actual` and `expected` mixed up

extensions/sql/policy/store/src/test/java/org/eclipse/dataspaceconnector/slq/policy/store/SqlPolicyStoreTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+    @Test
+    @DisplayName("Find policy by ID when not exists")
+    void findById_whenNonexistent() {
+        assertThat(sqlPolicyStore.findById("nonexistent")).isNull();
+    }
+
+    @Test
+    @DisplayName("Find all policies with limit and offset")
+    void findAll_withSpec() {
```
paullatzelsperger, line 138: tests for the following cases should be added:
- page size larger than DB collection
- page size oversteps DB collection size (e.g. 10 elements, offset=7, limit=5)

extensions/sql/policy/store/src/test/java/org/eclipse/dataspaceconnector/slq/policy/store/SqlPolicyStoreTest.java
```diff
@@ -0,0 +1,212 @@
...
+        assertThat(policiesFromDB).hasSize(limit);
+    }
+
+    @Test
+    @DisplayName("Delete existing policy")
+    void deleteById_whenExists() {
+        var policy = getDummyPolicy(getRandomId());
+
+        sqlPolicyStore.save(policy);
+        assertThat(sqlPolicyStore.findById(policy.getUid()).getUid()).isEqualTo(policy.getUid());
```
paullatzelsperger, line 160: not really necessary, as it would verify the `findById` rather than the `delete`

paullatzelsperger (changes requested):
About testing: please do not use the SUT (system-under-test) to verify itself. Meaning, when you `save` something, don't use the `findById` method to verify `save`

extensions/sql/policy/store/src/test/resources/schema.sql
```diff
@@ -0,0 +1,28 @@
...
+--  SPDX-License-Identifier: Apache-2.0
+--
+--  Contributors:
+--       ZF Friedrichshafen AG - Initial API and Implementation
+--
+
+-- table: edc_policies
+CREATE TABLE IF NOT EXISTS edc_policies
+(
+    policy_id VARCHAR(255) NOT NULL,
```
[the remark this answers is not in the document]
tuncaytunc-zf: The tests are meant to run with H2 DB, I think it is better to use VARCHAR(MAX) as it was defined it proposal. I just changed it to a certain int value to be able to run it with Postgres DB since Postgress doesn't have MAX and forget to change it back.

extensions/sql/policy/store/src/main/java/org/eclipse/dataspaceconnector/sql/policy/store/SqlPolicyStore.java
```diff
@@ -0,0 +1,176 @@
...
+            return single(executeQuery(getConnection(), this::mapResultSet, sqlPolicyStoreStatements.getSqlFindByClauseTemplate(), id));
+        } catch (SQLException exception) {
+            throw new EdcPersistenceException(exception);
+        }
+    }
+
+    @Override
+    public Stream<Policy> findAll(QuerySpec spec) {
+        return transactionContext.execute(() -> {
+            Objects.requireNonNull(spec);
```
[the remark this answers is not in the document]
tuncaytunc-zf: You are right, since it is just select could be of course done outside the transaction, will do so.

extensions/sql/policy/store/src/main/java/org/eclipse/dataspaceconnector/sql/policy/store/SqlPolicyStore.java
```diff
@@ -0,0 +1,176 @@
...
+
+            var limit = Limit.Builder.newInstance()
```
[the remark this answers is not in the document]
tuncaytunc-zf: will do so, see my comment above.

extensions/sql/policy/store/src/test/java/org/eclipse/dataspaceconnector/slq/policy/store/SqlPolicyStoreTest.java
tuncaytunc-zf: Will do so.

extensions/sql/policy/store/src/test/java/org/eclipse/dataspaceconnector/slq/policy/store/SqlPolicyStoreTest.java
tuncaytunc-zf: I have removed it

extensions/sql/policy/store/src/test/java/org/eclipse/dataspaceconnector/slq/policy/store/SqlPolicyStoreTest.java
tuncaytunc-zf: Will implement the additional tests.

extensions/sql/policy/store/src/main/java/org/eclipse/dataspaceconnector/sql/policy/store/SqlPolicyStoreStatements.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    String getSqlFindByClauseTemplate();
+
+    /**
+     * DELETE statement for policies
+     */
+    String getSqlDeleteClauseTemplate();
+
+
+    default String getPolicyTableName() { return "edc_policies"; }
```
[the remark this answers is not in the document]
paullatzelsperger: please apply _our_ checkstyle config, as stated in the documentation.

extensions/sql/policy/store/src/main/java/org/eclipse/dataspaceconnector/sql/policy/store/SqlPolicyStore.java
paullatzelsperger: yeah, but relying on an exception for condition checking is bad style

extensions/sql/policy/store/src/test/resources/schema.sql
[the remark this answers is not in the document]
paullatzelsperger: why would we artificially constrain ourselves? H2 supports a `VARCHAR` datatype, same as Postgres. Seems a bit like having fixed-length `char[255]` in java code instead of `String`, no?
Now on some platforms (eg. MS SQL) `VARCHAR` has different semantics, so _there_ a length constraint makes sense, but as you said, we're targeting H2 and Postgres.

extensions/sql/policy/store/src/main/java/org/eclipse/dataspaceconnector/sql/policy/store/SqlPolicyStore.java
tuncaytunc-zf: I have implemented two additional private methods update and insert. When a policy with same ID exists I do an update and when not I insert it.

extensions/sql/policy/store/src/main/java/org/eclipse/dataspaceconnector/sql/policy/store/SqlPolicyStore.java
```diff
@@ -0,0 +1,190 @@
...
+                }
+                return entity;
+            } catch (Exception e) {
+                throw new EdcPersistenceException(e.getMessage(), e);
+            }
+        });
+    }
+
+    private void insert(Policy policy) {
+        try (var connection = getConnection()) {
```
paullatzelsperger, line 108: `insert`/`update`: they also should execute their code in a transaction. I know they are private, and that the `save` method already creates one, but that would just "attach" `save`/`update` to the existing transaction and would increase robustness, should we ever call those two from another method.

extensions/sql/policy/store/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Provides SQL persistence for policies.
+
+## Prerequisites
+
+Please apply this [schema](docs/schema.sql) to your SQL database.
+
+## Entity Diagram
+
+![ER Diagram](https://plantuml.com/plantuml/png/VT11QWCn38NX_Pn2wq0kK4B8lflUm3YUrucAHpAI1ZAKtht5RQ6XcIpz_WyCEdbaYsMk0oGuLi9OKjFAXU7qFX3jg3_NnECJUHZBH8V3o_Fn-1Nt-ovoAN1Ft_2FUdELxdPKLC9oYOV8KL52BU7Q3EwiSXyssCoz-mOmlV2POrkCIp0s6gu1SJvNf6RQODkKvGBZyFHgEl-INh8xJmAxD-9cZ6mJucRwkNci3V___HXEa76PwoS0)
```
paullatzelsperger, line 11: it's better to generate the image, add it to the project and reference it directly

extensions/sql/policy/store/README.md
```diff
@@ -0,0 +1,39 @@
+# SQL Policy Store
+
+Provides SQL persistence for policies.
+
+## Prerequisites
+
+Please apply this [schema](docs/schema.sql) to your SQL database.
```
paullatzelsperger, line 7: The link is incorrect, should be `../docs/schema.sql`. Also, that schema is not valid for Postgres, please adapt accordingly (remove size constraint, as commented before) and add comment "this is only tested for postgres" etc. for comparison check [here](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/3b7d17938562948e34522d4c919e725269ad0b8b/extensions/sql/transfer-process-store/doc/ddl_postgres.sql)

extensions/sql/policy/store/src/main/java/org/eclipse/dataspaceconnector/sql/policy/store/SqlPolicyStoreServiceExtension.java
```diff
@@ -0,0 +1,41 @@
...
+package org.eclipse.dataspaceconnector.sql.policy.store;
+
+import org.eclipse.dataspaceconnector.spi.policy.store.PolicyStore;
+import org.eclipse.dataspaceconnector.spi.system.Inject;
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtension;
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtensionContext;
+import org.eclipse.dataspaceconnector.spi.transaction.TransactionContext;
+import org.eclipse.dataspaceconnector.spi.transaction.datasource.DataSourceRegistry;
+
+public class SqlPolicyStoreServiceExtension implements ServiceExtension {
```
paullatzelsperger, line 24: please add the `@Provides(PolicyStore.class)` annotation, to inform the dependency injection mechanism

extensions/sql/policy/store/src/main/java/org/eclipse/dataspaceconnector/sql/policy/store/ConfigurationKeys.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Defines configuration keys used by the SqlPolicyStoreServiceExtension.
+ */
+public interface ConfigurationKeys {
+
+    /**
+     * Name of the datasource to use for accessing policies.
+     */
+    @EdcSetting(required = true)
+    String DATASOURCE_SETTING_NAME = "edc.datasource.policy.name";
```
paullatzelsperger, line 28: nit: at this point, moving this directly to the `SqlPolicyStoreServiceExtension` would probably be even better

extensions/sql/policy/store/src/main/java/org/eclipse/dataspaceconnector/sql/policy/store/SqlPolicyStoreServiceExtension.java
paullatzelsperger, line 24: nit: rename to `SqlPolicyStoreExtensions`

extensions/sql/policy/store/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
...
+
+dependencies {
+    implementation(project(":spi:core-spi"))
+    implementation(project(":spi:contract-spi"))
+    implementation(project(":extensions:dataloading"))
+    implementation(project(":extensions:transaction:transaction-spi"))
+    implementation(project(":extensions:transaction:transaction-datasource-spi"))
+    implementation(project(":extensions:sql:common"))
+
+    testImplementation(testFixtures(project(":launchers:junit")))
```
paullatzelsperger, line 31: the following dependencies are not needed and should be removed:
- `:launchers:junit`
- `:core:base`
- `:org.assertj...` -> already added by the root buildfile

paullatzelsperger (changes requested):
we're getting there.

paullatzelsperger (approved):

PR #1018, opened 2022-03-30
CI: control test verbosity

### LINKED ISSUE #994: Tests: improve experience
# Feature Request

The experience of running tests locally on developer machines should be improved in the following aspects:
- the amount of log output should be significantly reduced. No test should output anything to the console by default. [Optional]: create a commandline parameter that controls the log output, something like a "verbose" flag.
- Failing tests should be clearly identifiable. Currently it is very cumbersome to scroll hundreds of lines up until the failing test comes into view. Once the first aspect is done, this should already be improved significantly, but further improvements may be in order, such as a test summary.
- Reduce the set of tests that are run by `./gradlew clean test` (see issue #819): by default only unit tests should be run. 

build, unit tests

Building the entire code base (and thus running tests) is now really convoluted, and it needs to be cleaned up.

## Solution Proposal
Mitigations to this problem may include (but are not limited to):
- prohibit using the `ConsoleMonitor` in tests. We may even be able to verify this with a [checkstyle rule](https://checkstyle.sourceforge.io/config_regexp.html)
- in integration tests, that spin up EDC extensions, we need to also replace the `ConsoleMonitor` with a dummy one, or at the very least set its level to `SEVERE`.
- avoid `System.out.println` (obviously)

## Type of Issue
improvement

## Checklist

algattik:
## What this PR changes/adds

- `./gradlew clean test` does not log anything from (successful) test runs.
- A runtime flag is available (and documented) to make test runs verbose.
- With `./gradlew clean test`, if a test fails, the test name and failure cause are printed

## Why it does that

Improve developer experience

## Further notes

Fixed BaseRuntime so as not to apply a shutdown hook to tests, that was causing extra logging at test completion.

Sample run showing clear message when a test fails:
https://github.com/agera-edc/DataSpaceConnector/runs/5749889045?check_suite_focus=true
```
2022-03-30T07:23:21.2000714Z CoreServicesExtensionIntegrationTest > shouldProvideHostnameExtension(Hostname) FAILED
2022-03-30T07:23:21.2001408Z     java.lang.AssertionError: Something strange happened
2022-03-30T07:23:21.2002146Z         at org.eclipse.dataspaceconnector.core.CoreServicesExtensionIntegrationTest.shouldProvideHostnameExtension(CoreServicesExtensionIntegrationTest.java:40)
2022-03-30T07:23:27.6996011Z 
```

## Linked Issue(s)

Closes #994

## Checklist

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -1,5 +1,13 @@
 # Writing Tests
 
+## Controlling test verbosity
+
+To run tests verbosely (displaying test events and output and error streams to the console), use the following system property:
+
+```shell
+./gradlew test -PverboseTest
+```
+
 ## Definition and distinction
 
 * _unit tests_ test one single class by stubbing or mocking dependencies.
--- a/launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/EdcExtension.java
+++ b/launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/EdcExtension.java
@@ -42,7 +42,6 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 import static org.eclipse.dataspaceconnector.common.types.Cast.cast;
 
@@ -79,23 +78,18 @@ public class EdcExtension extends BaseRuntime implements BeforeTestExecutionCall
 
     @Override
     public void beforeTestExecution(ExtensionContext extensionContext) throws Exception {
-        boot();
+        bootWithoutShutdownHook();
     }
 
     @Override
     public void afterTestExecution(ExtensionContext context) throws Exception {
-        // TODO: this shutdown is duplicated but it's necessary to DataPlaneHttpIntegrationTests, needs to be fixed
-        if (runningServiceExtensions != null) {
-            shutdown(runningServiceExtensions, getMonitor());
-        }
-
+        shutdown();
         // clear the systemExtensions map to prevent it from piling up between subsequent runs
         ((MultiSourceServiceLocator) serviceLocator).clearSystemExtensions();
     }
 
     @Override
     protected void bootExtensions(ServiceExtensionContext context, List<InjectionContainer<ServiceExtension>> serviceExtensions) {
-        this.runningServiceExtensions = serviceExtensions.stream().map(InjectionContainer::getInjectionTarget).collect(Collectors.toList());
         super.bootExtensions(context, serviceExtensions);
     }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -179,7 +179,11 @@ allprojects {
         }
 
         testLogging {
-            events("failed")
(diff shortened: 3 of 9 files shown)


docs/testing.md
```diff
@@ -1,5 +1,13 @@
 # Writing Tests
 
+## Controlling test verbosity
+
+To run tests verbosely (displaying test events and output and error streams to the console), use the following system property:
+
+```shell
+./gradlew test -PverboseTest
```
ndr-brt, line 8: As I said in #994 I think it would be better to have verbose tests by default and then provide a flag to silence them, especially because otherwise it will be tricky on the IDE part having to set every time this flag.

launchers/junit/src/testFixtures/java/org/eclipse/dataspaceconnector/junit/launcher/EdcExtension.java
```diff
@@ -76,23 +74,18 @@
 
     @Override
     public void beforeTestExecution(ExtensionContext extensionContext) throws Exception {
-        boot();
+        bootWithoutShutdownHook();
     }
 
     @Override
     public void afterTestExecution(ExtensionContext context) throws Exception {
-        // TODO: this shutdown is duplicated but it's necessary to DataPlaneHttpIntegrationTests, needs to be fixed
```
ndr-brt, line 84: thank you for doing this :+1:

algattik:
@ndr-brt thanks for starting a review - I just force pushed to fix conflicts

docs/testing.md
algattik: @paullatzelsperger I don't have an opinion on this - since you wrote the story, what's your take?

paullatzelsperger: @ndr-brt I am of the opposite opinion: make them as sparse as possible by default, so that they don't needlessly spam the terminal window, and on CI builds we can enable some "verbosity" flag

algattik: Can we go ahead and merge it as is, since it is already the current behavior? We can create an issue to change the setting to verbose and collect opinions

ndr-brt: @paullatzelsperger there are two way of using tests (on my side):
- from IDE: probably launching one test or one test class most of the time, if the test is broken, some logs would be appreciated for sure, if everytime I launch a different test from the IDE I have to modify the "run configuration" and add the flag, it could become annoying.
- from command line: at the end before the commit to verify that everything works correctly, probably here logs are not wanted, and would be straightforward to add a flag to shut them down, same on the CI.

When we introduced the system tests some other devs were asking me why there weren't logs printed by default.

algattik: Created #1049 to capture this proposal. We should please go forward with this PR that solidifies the current behavior and ensures we have proper reporting of failing tests in CI. We can take the discussion to #1049 and reach an agreement for another PR.

algattik: @ndr-brt ok to approve?

ndr-brt (approved):

paullatzelsperger (approved):

Issue #882, opened 2022-03-11
Express Access to REST API as Asset


MoritzKeppler:
# Feature Request
Data is often shared via REST APIs that exist independent from a connector/IDS, an AAS e.g. https://app.swaggerhub.com/apis/Plattform_i40/AssetAdministrationShell-REST-API/Final-Draft
A connector in front of such an API can ensure, that any call is covered by a contract that describes the interactions allowed (http methods, endpoints, parameters).
We need a way to express that as a asset / contract (or a policy) and enforce the rules during data transfer.

## Which Areas Would Be Affected?
DPF, data management API, maybe policy handling

## Why Is the Feature Desired?
Integration of AAS, e.g. a asset / contract for access to "all submodels of type x from all shells"

## Solution Proposal
maybe a regex match to URLs?

## Type of Issue
new feature

## Checklist

- [x] assigned appropriate label?
- [x] **Do NOT select a milestone or an assignee!**

related to #881

jimmarino:
relates to #858

MoritzKeppler:
`HttpDataSourceFactory` is implemented in a way that it partially overwrites values given in the `DataAddress` with values from `DataFlowRequest` properties (a list filled by the http call coming in `DataPlanePublicApiController`), especially the http path. But also the actual http method is taken from that list.

Propose to change it in a way, that the DataAddress always wins and just if a value is not defined there it is taken from the `DataFlowRequest` properties.
Advantages:
* improved security: caller cannot access any address accidentally
* covers some basic scenarios in http proxy behavior of DPF, e.g. "one Asset = one URL" and "Asset = all subpathes" etc. (improved solution will be available as soon as the DPF can evaluate policies on the call)

Propose to add the HTTP method to the `DataAddress` (optional) with Fallback to `GET`

What do you think @bscholtes1A @paullatzelsperger  @stefan-ettl ?

bscholtes1A:
Hi @MoritzKeppler 

for the path I get it, and that's quite easy to put precedence in the code to the path coming from the `DataAddress`. 

However, for the method I am not sure what you mean. What if I call with a POST an API for which the `DataAddress` does not contain any verb?

MoritzKeppler:
> 
> However, for the method I am not sure what you mean. What if I call with a POST an API for which the `DataAddress` does not contain any verb?

If DataAddress does not contain any Verb, POST would be used.
[...]

PR #1019, opened 2022-03-30
dpf: http proxy: don't overwrite values set in DataAddress for security reasons

MoritzKeppler:
## What this PR changes/adds
The DPF can act as http proxy. For security reasons it's important to not overwrite values set in the source dataaddress defined at the asset but just take those aspects (path, query params, http method, ...) that are not predefined.

## Why it does that
security reasons

## Linked Issue(s)
see #882

## Checklist


extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/controller/DataPlanePublicApiController.java
```diff
@@ -158,17 +160,49 @@ private Response handle(ContainerRequestContext requestContext) {
     /**
      * Create a {@link DataFlowRequest} based on the decoded claim token and the request content.
      */
-    private DataFlowRequest createDataFlowRequest(ClaimToken claims, Map<String, String> properties) {
-        var dataAddress = typeManager.readValue(claims.getClaims().get(DataPlaneConstants.DATA_ADDRESS), DataAddress.class);
+    private DataFlowRequest createDataFlowRequest(ClaimToken claims, ContainerRequestContext requestContext) {
+        var origDataAddress = typeManager.readValue(claims.getClaims().get(DataPlaneConstants.DATA_ADDRESS), DataAddress.class);
+        if (!HttpDataAddressSchema.TYPE.equals(origDataAddress.getType())) {
```
bscholtes1A, line 165: Seems there is a misunderstanding here. Public DPF API is not restricted to accessing Rest API. 
It can be used to interrogate a database, a store...

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/controller/DataPlanePublicApiController.java
```diff
@@ -158,17 +160,49 @@ private Response handle(ContainerRequestContext requestContext) {
...
+        if (StringUtils.isNullOrBlank(origDataAddress.getProperty(HttpDataAddressSchema.BODY))) {
+            dataAddress.property(HttpDataAddressSchema.BODY, requestContextApi.body(requestContext));
+        }
+
+        if (StringUtils.isNullOrBlank(origDataAddress.getProperty(HttpDataAddressSchema.MEDIA_TYPE))) {
+            dataAddress.property(HttpDataAddressSchema.MEDIA_TYPE, requestContextApi.mediaType(requestContext));
+        }
+
+        if (StringUtils.isNullOrBlank(origDataAddress.getProperty(HttpDataAddressSchema.QUERY_PARAMS))) {
+            dataAddress.property(HttpDataAddressSchema.QUERY_PARAMS, requestContextApi.queryParams(requestContext));
```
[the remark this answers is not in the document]
MoritzKeppler: What about cases where an internal data source needs query params independent from the proxy behavior?
If a path and a domain is part of a DataAddress, then query params, method and potentially also a body that needs to be passed to retrieve the data could be part of the address. Or?

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/dataspaceconnector/dataplane/api/controller/DataPlanePublicApiController.java
MoritzKeppler: ok, so I need to make it conditional, i.e. just update the Data Address if it's of type `HttpDataAddressSchema.TYPE`
Does not make sense to pass path and query params if it's pointing to a SQL database.

bscholtes1A: Some can still be useful, for example the query parameters if you want to target a DB.
This is why I would be in favor of puting there info anyway in the DataFlowRequest properties, and then in the factory (e.g. HttpDataSourceFactory) that is called, you choise whether you want to use these properties to complete the data address of not.

Issue #323, opened 2021-11-30
Create Fluent Query API for AssetIndex etc.


paullatzelsperger:
This is an improvement of the current `Criterion` that is used to submit queries. 

## Problem statement
The main problem with the Criterion API is that the user has to supply valid operators and operands. Valid as in "valid in the database's dialect". So the query syntax is dependent on the DB implementation.

For example, submitting an `"IN"`-query to an SQL-style DB would require the right-hand operand to be a valid list of values complete with brackets and commas, resulting in `someValue IN ('value1', 'value2', ...'valueN')`. 
In addition, the user needs to know the data type of the left-hand operand, because that influences, whether the single quotation marks are necessary or not.

## Proposition
In order to improve on that I propose an explicit query API with the following properties:
- offers a small but specific set of operands (e.g. `equals`, `in`, `gt`, `lt`, `geq`, `leq`)
- type safety: the type of the left-hand operand is known either through a schema or through a parameter, thus we can validate the operator and the right-hand operator
- fluent
- extensible: there is a possibility to supply "raw" queries, that are not checked and that is tunneled through to the data store more or less directly.
- serializable: makes it possible to transmit a query easily over the wire

### String Equals
```java
var query = QueryBuilder.newQuery()
                         .where("someProp").havingType(String.class)
                         .eq("someValue").build();
```

### String in 
```java
var query = QueryBuilder.newQuery()
                         .where("someProp").havingType(String.class)
                         .in("value1", "value2", "value3")
                         .build();
```
### Greater-Than with Schema
```java
org.eclipse.dataspaceconnector.spi.types.domain.schema.Schema schema = ...; //obtain Schema, e.g. from JSON Schema
var query = QueryBuilder.newQuery()
                         .where("someProp").havingSchema(schema)
                         .gt(420)
                         .build();
```

jimmarino:
Nice and succinct +1

PR #1021, opened 2022-03-30
AssetIndex: add support for `QuerySpec` filtering

### LINKED ISSUE #1014: AssetIndex: add support for `QuerySpec` filtering for the SQL-impl.
# Feature Request

For the AssetIndex we do in fact need the full `QuerySpec` filtering support, because here that's not used for paging etc. but for the Offer creation process. Currently, the SQL-based implementation only uses the `limit` and `offset` parameters, which must be improved.

AssetIndex, ContractOffer creation

When creating offers, the provider must be able to fetch all `Assets` for a `ContractDefinition` using the `AssetSelectorExpression`.

## Type of Issue
(Urgent) Enhancement to #984 

## Checklist

paullatzelsperger:
## What this PR changes/adds

Adds support for `QuerySpec` filtering and `AssetSelectorExpression`, i.e. filtering on Asset properties.

## Why it does that

This is necessary for generating and validating `ContractOffer`s.

## Further notes

- Currently not all operators support prepared statements, e.g., the `IN` operator.
- sorting is not yet implemented

## Linked Issue(s)

Closes #1014 

## Checklist

### DIFF
--- a/extensions/sql/asset/index/src/main/java/org/eclipse/dataspaceconnector/sql/asset/index/SqlAssetIndex.java
+++ b/extensions/sql/asset/index/src/main/java/org/eclipse/dataspaceconnector/sql/asset/index/SqlAssetIndex.java
@@ -9,6 +9,7 @@
  *
  *  Contributors:
  *       Daimler TSS GmbH - Initial API and Implementation
+ *       Microsoft Corporation - added full QuerySpec support
  *
  */
 
@@ -22,6 +23,7 @@ import org.eclipse.dataspaceconnector.spi.asset.AssetIndex;
 import org.eclipse.dataspaceconnector.spi.asset.AssetSelectorExpression;
 import org.eclipse.dataspaceconnector.spi.asset.DataAddressResolver;
 import org.eclipse.dataspaceconnector.spi.persistence.EdcPersistenceException;
+import org.eclipse.dataspaceconnector.spi.query.Criterion;
 import org.eclipse.dataspaceconnector.spi.query.QuerySpec;
 import org.eclipse.dataspaceconnector.spi.transaction.TransactionContext;
 import org.eclipse.dataspaceconnector.spi.transaction.datasource.DataSourceRegistry;
@@ -33,16 +35,20 @@ import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.AbstractMap;
+import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.LinkedList;
+import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.sql.DataSource;
 
+import static java.lang.String.format;
 import static org.eclipse.dataspaceconnector.sql.SqlQueryExecutor.executeQuery;
 
 public class SqlAssetIndex implements AssetLoader, AssetIndex, DataAddressResolver {
+
+    private static final List<String> SUPPORTED_PREPARED_STATEMENT_OPERATORS = List.of("=", "like");
 
     private final ObjectMapper objectMapper;
     private final DataSourceRegistry dataSourceRegistry;
@@ -67,7 +73,7 @@ public class SqlAssetIndex implements AssetLoader, AssetIndex, DataAddressResolv
             transactionContext.execute(() -> {
                 try {
                     if (existsById(asset.getId(), connection)) {
-                        throw new EdcPersistenceException(String.format("Cannot persist. Asset with ID '%s' already exists.", asset.getId()));
+                        throw new EdcPersistenceException(format("Cannot persist. Asset with ID '%s' already exists.", asset.getId()));
                     }
 
                     executeQuery(connection, sqlAssetQueries.getSqlAssetInsertClause(),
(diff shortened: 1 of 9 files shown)


DominikPinsel (changes requested):

extensions/sql/asset/index/src/main/java/org/eclipse/dataspaceconnector/sql/asset/index/SqlAssetIndex.java
```diff
@@ -193,34 +215,67 @@ public DataAddress resolveForAsset(String assetId) {
...
+        return format(" WHERE %s", String.join(" AND ", subSelects));
+    }
+
+    /**
+     * Converts a {@linkplain Criterion} into a dynamically assembled SELECT statement.
+     */
+    private String toSubSelect(Criterion c) {
+        return format("%s %s %s)", sqlAssetQueries.getQuerySubSelectClause(), c.getOperator(),
+                (supportsPreparedStatement(c) ? "?" : c.getOperandRight()));
+    }
```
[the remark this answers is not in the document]
paullatzelsperger: this is clearly stated in the PR description as a known issue.

ndr-brt:
Why don't force a `Criterion` object to have a `Collection` as the right operand instead of a string and then build the `subSelect` for the "in" criterion as "leftoperand IN (?, ?, ? ..." with a question mark for every collection item. 
This will probably avoid the sql injection issue from the bottom. (because, doing so, every operator will support prepared statement operators)

paullatzelsperger:
@ndr-brt there is already an issue to improve the `Criterion` API (#323), which could be used to tackle that. 
As an intermediate fix, we could use a `List` as `rightOperand` - the `Criterion` would support that already, since it uses `Object`s. However, the store would then explicitly have to check for that, which seems like a fair trade-off. WDYT?

ndr-brt:
@paullatzelsperger the issue is known so no hurry with this, we can address it to #323

ndr-brt (approved):

paullatzelsperger:
@ndr-brt I saw you already approved, but in the meantime I added handling for `IN` operators in [this commit](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/1021/commits/210575801785f029cfe73c936f3895d4f4cadd4d). Care to take a look?

ndr-brt (approved):
looks good!

extensions/sql/asset/index/src/main/java/org/eclipse/dataspaceconnector/sql/asset/index/SqlAssetIndex.java
```diff
@@ -127,23 +125,53 @@ public Asset deleteById(String assetId) {
...
+
     @Override
     public Stream<Asset> queryAssets(AssetSelectorExpression expression) {
         Objects.requireNonNull(expression);
 
-        return getAssetStream(sqlAssetQueries.getSqlAssetListClause());
+        var criteria = expression.getCriteria();
+        var querySpec = QuerySpec.Builder.newInstance().filter(criteria)
+                .offset(0)
+                .limit(Integer.MAX_VALUE) // means effectively no limit
```
Izzzu, line 142: Would it make sense to have limit in QuerySpec initialised by default with  Integer.MAX_VALUE (effectively no limit)? I see now it's 50

paullatzelsperger: i don't think so, because this is really a specific case: we want the `LIMIT ` clause there so as not to complicate the statement generation even more, but effectively there shouldn't be a limit. 
The `QuerySpec` should definitely have sane defaults that don't bork the database :)

extensions/sql/asset/index/src/main/java/org/eclipse/dataspaceconnector/sql/asset/index/SqlAssetIndex.java
```diff
@@ -127,23 +125,53 @@ public Asset deleteById(String assetId) {
...
     public Stream<Asset> queryAssets(QuerySpec querySpec) {
         Objects.requireNonNull(querySpec);
 
-        var query = String.format("%s LIMIT %s OFFSET %s",
-                sqlAssetQueries.getSqlAssetListClause(),
-                querySpec.getLimit(),
-                querySpec.getOffset());
 
-        return getAssetStream(query);
+        var conditions = querySpec.getFilterExpression().stream().map(SqlConditionExpression::new).collect(Collectors.toList());
```
Izzzu, line 152: Is it possible that this piece of code can be reused in different stores? 
Maybe it would be worth to extract it to separate class (called eg. SQLQueryResolver) that would implement QueryResolver interface. See eg. ReflectionBasedQueryResolver.
_[I see the PR was just merged but will leave the comment for future reference]_

paullatzelsperger: again, this is really specific to the `AssetIndex`, where all the properties are located in a separate table. However, I'm sure we can extract some useful code when we need it, especially the `SqlConditionExpression`, which "parses" the `Criterion`

extensions/sql/asset/index/src/main/java/org/eclipse/dataspaceconnector/sql/asset/index/SqlAssetIndex.java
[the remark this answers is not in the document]
paullatzelsperger: ...and on top of it I put a temp fix in.

PR #1039, opened 2022-03-31
Feature: adds HTTP provisioner webhook endpoint

paullatzelsperger:
## What this PR changes/adds

Adds a webhook for the external system to call back into during an HTTP provisioning sequence.

## Why it does that

The HTTP provisioner calls out to an external system via HTTP and includes the callback URL to that request. After the external system has finished, it needs a webhook endpoint with which the EDC runtime can be informed about the completion.

## Further notes

- the `apiToken` property on the webhook's request body is opaque
- the webhook's URL is provided by the `HttpProvisionerWebhookUrl` service, that can be `@Inject`-ed
- the `SimpleSecretToken` is now located directly in the HTTP provisioner package, but since that is a pretty generic object, we may want to move it into a more general package if needed.
- Much of the changeset is caused by refactoring the package structure.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ the detailed section referring to by linking pull requests or issues.
 * Support for HTTP-based provisioning (#963)
 * Let Control Plane delegate data transfer to Data Plane (#988)
 * CosmosDb based `PolicyStore` (#826)
+* Http Provisioner Webhook endpoint (#1039)
 
 #### Changed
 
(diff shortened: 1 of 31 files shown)


paullatzelsperger:
@MoritzKeppler adding you because C-X backend services will have to use this webhook.

paullatzelsperger:
Seems like the CI build fails with a compile error. I don't have this error locally, will need to investigate.

jimmarino (approved):

PR #1040, opened 2022-03-31
chore: Provisioning types cleanup

jimmarino:
## What this PR changes/adds

This PR adds a `ProvisionedDataAddressResource` and creates a class hierarchy for data address resources. This consolidates common properties and simplifies the `TPM`.

## Why it does that

It simplifies the codebase.

## Checklist

### DIFF
--- a/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
+++ b/core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
@@ -43,7 +43,9 @@ import org.eclipse.dataspaceconnector.spi.transfer.provision.ResourceManifestGen
 import org.eclipse.dataspaceconnector.spi.transfer.store.TransferProcessStore;
 import org.eclipse.dataspaceconnector.spi.types.TypeManager;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataRequest;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.ProvisionResponse;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.ProvisionedContentResource;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.ProvisionedDataAddressResource;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.ProvisionedDataDestinationResource;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.ProvisionedResource;
 import org.eclipse.dataspaceconnector.spi.types.domain.transfer.ResourceManifest;
@@ -59,6 +61,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 
 import static java.lang.String.format;
+import static java.lang.String.join;
 import static java.util.Collections.emptyList;
 import static java.util.UUID.randomUUID;
 import static org.eclipse.dataspaceconnector.spi.types.domain.transfer.TransferProcess.Type.CONSUMER;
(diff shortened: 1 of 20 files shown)


core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
```diff
@@ -457,6 +444,31 @@ private void handleProvisionResult(TransferProcess transferProcess, List<Provisi
         }
     }
 
+    private boolean handleProvisionDataAddressResource(ProvisionedDataAddressResource resource, ProvisionResponse response, TransferProcess transferProcess) {
```
ndr-brt, line 447: This method provides side effects, probably I would avoid to have something that returns `boolean` (and looks like a "query" method) and that changes the state at the same time.
Since a `false` result will cause `transitionError` I would put that method call right there (avoiding `store`, because it will be called anyway in the block at line 431), and making the method be a `void` one

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
```diff
@@ -517,6 +518,18 @@ private void handleDeprovisionResult(TransferProcess transferProcess, List<Depro
         }
     }
 
+    private void removeDeprovisionedSecrets(ProvisionedDataAddressResource provisionedResource, TransferProcess transferProcess) {
+        if (!provisionedResource.hasToken()) {
```
ndr-brt, line 522: this guard condition could be moved to the condition at line 499 to improve clarity.

core/transfer/src/test/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImplTest.java
```diff
@@ -613,4 +628,25 @@ public static Builder newInstance() {
...
     }
+
+    private static class TokenTestProvisionResource extends TestProvisionedDataDestinationResource {
+        public TokenTestProvisionResource(String resourceName, String id) {
+            super(resourceName, id);
+            this.hasToken = true;
+        }
+    }
+
+    private class TestToken implements SecretToken {
```
ndr-brt, line 639: a `mock` instance would replace well this dummy class

jimmarino: The problem is it needs to be serialized to the vault, so a mock will throw an exception when the ObjectMapper attempts the write.

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
jimmarino: +1

core/transfer/src/test/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImplTest.java
ndr-brt: got it, thanks

core/transfer/src/main/java/org/eclipse/dataspaceconnector/transfer/core/transfer/TransferProcessManagerImpl.java
jimmarino: Looking at this, I think the issue was the method was doing two separate things. I pulled the secret storing up and integrated that with fatal errors, which allowed error transitioning to be handled by a single code path.

Let me know what you think.

ndr-brt (approved):
looks good

PR #1051, opened 2022-04-01
Cosmos DB integration tests: pick up TransferProcess after the lease expired

### LINKED ISSUE #423: High Availability - Critical - Provider High Availability
If a provider connector fails after receiving a request, a transfer process needs to be picked up by another provider connector instance. This is specially important when deploying to Kubernetes, since pods can be restarted at any time (e.g. VM autoscale).

A `TransferProcessStore` backed by an external storage is required to achieve this. In addition a transfer process leasing mechanism to orchestrate the assignment of transfer processes to provider connector instances needs to be in place as well. 

A CosmosDB-based solution is [partially implemented](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/extensions/azure/transfer-process-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/transfer/store/cosmos/CosmosTransferProcessStore.java) in EDC. This solution requires two stored procedures which available in test resources but not packaged for runtime ([#1](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/extensions/azure/transfer-process-store-cosmos/src/test/resources/lease.js), [#2](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/extensions/azure/transfer-process-store-cosmos/src/test/resources/nextForState.js)). Finally this implementation is missing lease expiration logic to allow a leased transfer process to be taken by another provider connector instance after a timeout.

Suggested fixes and improvements:
* Provide working implementation of the Cosmos stored procedures in a source folder together with terraform files for deployment.
* Implement lease expiration logic to allow a second provider connector instance to take over a transfer process on failure of a first provider connector instance.
* Provide a working sample for this solution.

Izzzu:
## What this PR changes/adds

- Added test to CosmosTransferProcessStoreIntegrationTest to cover scenario when item is processed after the lease expired.
- ADR for Cosmos DB lease mechanism with stored procedures. 
- Small clean ups in CosmosTransferProcessStoreIntegrationTest and CosmosContractNegotiationStoreIntegrationTest

## Why it does that

Ensure that lease mechanism works correctly in a situation when the lease of the transfer process expired.
Example use case: if the processing fails in one instance of the connector then another instance should be able to picks up the processing after the lease of the item expired.

ADR is added retroactively to record decision about the behaviour implemented in the Cosmos DB stored procedures.

## Further notes

- Method uploadStoredProcedure moved into newly created StoredProcedureTestUtils class in cosmos-commons test fixtures to avoid duplicated code.

## Linked Issue(s)

Closes #423

## Checklist


docs/developer/decision-records/2022-03-30-cosmosdb-lease-mechanism/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+1. [`lease.js`](/extensions/azure/cosmos/cosmos-common/src/main/resources/lease.js) - queries the database to look for an item and if an item is found applies a lease on it. It's used to acquire the lock on single item in the database before an update and to remove the lock after.
+
+```java
+   leaseContext.acquireLease(process.getId());
+   failsafeExecutor.run(() -> cosmosDbApi.saveItem(document));
+   leaseContext.breakLease(process.getId());
+```
+
+1. [`nextForState.js`](/extensions/azure/cosmos/cosmos-common/src/main/resources/nextForState.js) - queries the database to look for all items that has no lease or expired lease. It applies a new lease on all returned items. This stored procedure is used to get a batch of elements that are next to be processed for given state.
```
paullatzelsperger, line 27: I would emphasize the reason this ist done in an SPROC is, because the "SELECT" and the "write-lease" operations must be ACID and SPROCs are the only way to achieve that.

paullatzelsperger (approved):

docs/developer/decision-records/2022-03-30-cosmosdb-lease-mechanism/README.md
Izzzu: added

PR #1060, opened 2022-04-04
DPF: Adapt system-test to use Embedded DPF

### LINKED ISSUE #902: DPF Test and Sample Updates

cpeeyush:
## What this PR changes/adds

Adapt system-test to use Embedded DPF to perform file copy, rather than the legacy `DataStreamPublisher` mechanism.

## Why it does that

Include DPF in system test scope.

## Linked Issue(s)

Closes #902 

## Checklist


jimmarino (approved):
Hey @cpeeyush This looks good. One comment, it looks like there is a mix of var and typed declaration. Can you switch it to var for consistency?

ndr-brt (commented):
I have a question that does not strictly relate with the modifications: why `FileTransferIntegrationTest` does not load the `sample-04.0` runtimes but it loads other two runtimes (`file-transfer-consumer`, `file-transfer-provider`) that are the pretty much same?
I think would be better to avoid this duplication.

cpeeyush:
> `FileTransferIntegrationTest`

@ndr-brt you are right, initially when `FileTransferIntegrationTest` was added it was to ensure that sample04 is working as expected or not, and it was using sample04 runtimes. 
But later I think it some point it was decided that these system tests should not cover samples and have their own runtimes that's why we see this duplication here.

cpeeyush:
> Hey @cpeeyush This looks good. One comment, it looks like there is a mix of var and typed declaration. Can you switch it to var for consistency?

Hi @jimmarino, Thanks, I have updated to use `var` consistently.

jimmarino:
> > @ndr-brt you are right, initially when `FileTransferIntegrationTest` was added it was to ensure that sample04 is working as expected, and it was using sample04 runtimes. But later I think at some point it was decided that these system tests should not cover samples and have their own runtimes that's why we see this duplication here.
> 
> I don't see downsides in having the system test covering the sample, also because it does not represent a real use case, but only a sample specific one. It will mash two potatoes with one fork.

@ndr-brt Would it make sense to come up with common runtime BOMs and still keep the test decoupling? I think our needs for integration testing will wind up diverging a bit from samples as we fill those out.

ndr-brt:
> @ndr-brt Would it make sense to come up with common runtime BOMs and still keep the test decoupling? I think our needs for integration testing will wind up diverging a bit from samples as we fill those out.

I think `FileTransferIntegrationTest` should not be considered an EDC end-to-end test but a test that avoids regressions on the samples, since it does not represent any real case that we want to verify or to run performance tests against.

As emerged in the last contributors meeting we really need to keep samples in good health, so I'm in favor of everything that guarantees it. If this test is not to be considered a sample test, then we should have one: having two tests that verify two runtime sets could be a good solution too.

cpeeyush:
@jimmarino @ndr-brt PR #1015 removed `include(":extensions:data-plane-transfer:data-plane-transfer-client")` from root `settings.gradle.kts` and seems like that is causing build failure for this PR. Was that removed on purpose ? or can be add it back ?

ndr-brt:
> @jimmarino @ndr-brt PR #1015 removed `include(":extensions:data-plane-transfer:data-plane-transfer-client")` from root `settings.gradle.kts` and seems like that is causing build failure for this PR. Was that removed on purpose ? or can be added back ?

Don't know, maybe a rebase issue? @tuncaytunc-zf what was the reason?

juliapampus:
@cpeeyush Can you take a look why the tests are failing, or should the pipeline be that red?

cpeeyush:
Hi @juliapampus  it was due to removal of `include(":extensions:data-plane-transfer:data-plane-transfer-client")` as discussed above with @tuncaytunc-zf. I have restored it, pipeline are green now.

PR #1062, opened 2022-04-04
Cleanup: Adds smaller refactorings in the `SqlContractDefinitionStore` 

paullatzelsperger:
## What this PR changes/adds

Mainly the following items have been refactored:
- SQL statements are now in an overridable interface/implementor structure
- all methods in the store use transactions
- removed `MAX` length constraints on `VARCHAR` as Postgres does not support this.

## Why it does that

This was done mainly to maintain consistency with the other SQL-based implementations.

## Further notes
n/a

## Checklist


extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
```diff
@@ -133,11 +93,11 @@ public void save(Collection<ContractDefinition> definitions) {
         transactionContext.execute(() -> {
             try (var connection = getConnection()) {
                 for (var definition : definitions) {
-                    executeQuery(connection, SQL_SAVE_CLAUSE_TEMPLATE,
+                    executeQuery(connection, statements.getInsertTemplate(),
```
tuncaytunc-zf, line 96: You should check beforehand with "existById(definition.getId, connection)" if a ContractDefinition with same ID exist before you insert it.

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
```diff
@@ -196,19 +152,37 @@ public ContractDefinition deleteById(String id) {
...
+        return transactionContext.execute(() -> {
+            var contractDefinitions = executeQuery(connection, this::mapResultSet, statements.getFindByTemplate(), id);
+            if (contractDefinitions.isEmpty()) {
+                return null;
+            }
+            if (contractDefinitions.size() == 1) {
+                return contractDefinitions.get(0);
+            }
+            throw new IllegalStateException("Expected result set size of 1 but got " + contractDefinitions.size());
+        });
```
tuncaytunc-zf, line 185: This can be executed outside from transaction and with just single line "return single(executeQuery(connection, this::mapResultSet, statement.getFindByTemplate(), id)" and of course by catching the SQLException.

tuncaytunc-zf:
The README should be moved to docs folder and the png diagram inside should be stored and referenced locally.

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
paullatzelsperger: The `ContractDefinitionStore` has an explicit `update` method, so we don't need `upsert` semantics here.
Removing the `update` method altogether, and implementing `upsert` semantics is possible, but something for another PR, as it will affect a rather wide swath of files.

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
paullatzelsperger: I agree on adding a `single` method, but the exception must be thrown inside the transaction in order to trigger a rollback.

extensions/sql/contract-definition/store/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
paullatzelsperger: on closer inspection, I noticed that the `Cosmos` impl also has upsert semantics. Will make it consistent then.

tuncaytunc-zf (approved):

PR #1079, opened 2022-04-05
DPF: control proxy behavior 

MoritzKeppler:
## What this PR changes/adds
The DPF can act as http proxy. For security reasons it's important to not overwrite values set in the source dataaddress defined at the asset but just take those aspects the data address allows to be taken from the current request.

## Why it does that

security reasons

## Linked Issue(s)

relates to #882

## Checklist


bscholtes1A:
Ok for me as this is something that needs to be delivered soon.

bscholtes1A:
Ok for me as this is to be delivered soon, however this kind of logic should be later on handled through Policies, no ? @jimmarino  @paullatzelsperger WDYT?

MoritzKeppler:
The actual rules what is allowed to be called (e.g. path patterns) should be part of the policies.
But nevertheless it could be of value to configure if a DataAddress is allowed to be used in a proxy behavior or describes a finally defined endpoint. This is typically something that happens during creation of an asset - independent from contract definitions / offers.

bcronin90 (approved):

bscholtes1A:
> The actual rules what is allowed to be called (e.g. path patterns) should be part of the policies. But nevertheless it could be of value to configure if a DataAddress is allowed to be used in a proxy behavior or describes a finally defined endpoint. This is typically something that happens during creation of

MoritzKeppler:
@bscholtes1A 
We (@paullatzelsperger, @jimmarino, @stefan-ettl, @MoritzKeppler) decided to merge it as it is and check if we need to refactor or improve it later.

PR #1104, opened 2022-04-08
extensions: implement TransferProcess Data Management Api

### LINKED ISSUE #809: DataManagementApi: add `TransferProcessService` and `TransferProcessTransformer`
For the implementation of the transfer process endpoint we need an `TransferProcessService` (and an `TransferProcessServiceImpl`) with roughly the following signature:
```java
public interface TransferProcess {
  TransferProcess findbyId(String transferProcessId);
  Collection<TransferProcess> query(QuerySpec query);
  String getState(String transferProcessId);
  Result cancel(String transferProcessId);
  Result deprovision(String transferProcessId, String reason);
}
```

In addition, there needs to be an `TransferProcessTransformer`, that converts an `TransferProcess` into a `TransferProcessDto` and back. This depends on #779 

Important aspects:
- service must perform defensive checks before performing an operation:
  - `cancel`: must check first, whether a `TransferProcess` can be cancelled according to valid state transitions
  - `deprovision`: check whether a `TransferProcess` can be declined according to valid state transitions
- state transitions must be done through the `CommandQueue` mechanism and their `Result` only reflects the successful submission of the command

Depends on #859

### LINKED ISSUE #811: DataManagementApi: add services and transformers
The actual implementation of the DataManagementApi endpoints will be done in services and transformers.

**Services** perform database operations, handle defensive checking and transactional consistency
**Transformers** convert DTOs into business objects and back
**Controllers** have fields for both and invoke them, handle errors, etc.

Those are the relevant issues:
- [x] #805 
- [x] #806 
- [x] #807 
- [x] #808
- [x] #809
- [x] #969 

Depends on #919

### LINKED ISSUE #812: DataManagementApi: controller integration tests
For all the data management api controllers there should be (currently) empty integration tests, which should now be implemented. It's highly likely that test methods need to be added to achieve acceptable coverage.

ndr-brt:
## What this PR changes/adds

Implements controller and add integration test to the `TransferProcess` data management API

## Why it does that

To expose a clear and RESTful api to operate on `TransferProcess`s

## Further notes

-

## Linked Issue(s)

Closes #809
Closes #811 
Closes #812 

## Checklist


extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/model/TransferRequestDto.java
```diff
@@ -34,6 +34,7 @@
     private TransferType transferType;
     private String protocol = "ids-multipart";
     private String connectorId;
+    private String assetId;
```
paullatzelsperger, line 37: Just FYI: I think eventually the `DataRequest` should only carry the `contractId`, and no `assetId`, because provider connectors should be to resolve the asset through the contract. Not for this PR though, just an FYI.

extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/TransferProcessApiController.java
```diff
@@ -58,63 +73,103 @@ public TransferProcessApiController(Monitor monitor) {
...
 
     @GET
-    @Path("{id}")
+    @Path("/{id}")
     @Override
     public TransferProcessDto getTransferProcess(@PathParam("id") String id) {
-        monitor.debug(format("get TransferProcess with ID %s", id));
-
-        return null;
+        monitor.debug(format("Get TransferProcess with ID %s", id));
```
paullatzelsperger, line 89: could you get rid of all the log spam? no real value in that IMO...

extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/service/TransferProcessServiceImpl.java
```diff
@@ -72,6 +75,18 @@ public TransferProcessServiceImpl(TransferProcessStore transferProcessStore, Tra
...
+                    .filter(AbstractResult::succeeded)
+                    .map(AbstractResult::getContent)
+                    .map(ServiceResult::success)
+                    .orElse(ServiceResult.conflict("Request couldn't be initialised."));
+        });
+    }
+
     private ServiceResult<TransferProcess> apply(String transferProcessId, Function<TransferProcess, ServiceResult<TransferProcess>> function) {
         return transactionContext.execute(() -> {
             var transferProcess = transferProcessStore.find(transferProcessId);
```
paullatzelsperger, line 92: modifying transfer processes should NOT be done directly through the store, but using the `CommandQueue`, to avoid race conditions. We should already have commands to cancel and deprovision. Could you do this directly in this PR? otherwise lets open another one.

ndr-brt: There are no modifications done through the `store`, it's used only to validate the "cancellation" and the "deprovisioning", but at the end the actions are executed through the commands

extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/TransferProcessApiController.java
ndr-brt: removed the log in the "query" methods, left in the "command" ones. Also in the other data management api controllers there are a lot of logs, probably we could cleanup them in another PR.

extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/model/TransferRequestDto.java
ndr-brt: I agree, let's do that in another PR.

Izzzu (approved):
Looks like it needs rebasing to fix the dependency check. It was recently merged to main

ndr-brt:
> Looks like it needs rebasing to fix the dependency check. It was recently merged to main

@Izzzu is already rebased against main. The problem seems to be `Java heap space`, will try pushing it again

extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/service/TransferProcessServiceImpl.java
paullatzelsperger: ah you're right.

paullatzelsperger (approved):

PR #1114, opened 2022-04-11
ids: IDS policy fixes

bscholtes1A:
## What this PR changes/adds

This PR brings two fixes for IDS policies:
 - add binding for _USE_ action to _ALL_SCOPE_ (otherwise, policy constraints are never evaluated)
 - improve resiliency of CustomIdsConstraintDeserializer as it was not properly checking potentially null values before using them.

## Checklist


juliapampus (approved):

juliapampus:
@bscholtes1A I am okay with the deserializer fix/improvement, I am not sure about the scope thing. @ndr-brt Could you tell me what this ids-policy extension is used for?

ndr-brt:
@juliapampus it was something added to show how to register functions on the policy engine, it hasn't other uses.
In my opinion it can be deleted.

juliapampus:
> @juliapampus it was something added to show how to register functions on the policy engine, it hasn't other uses. In my opinion it can be deleted.

@ndr-brt Is there any documentation for this? We could add the files inline to a step-for-step guide. I would support deleting unnecessary modules to avoid confusions.

ndr-brt:
Makes sense, opened an issue #1156 1156

ndr-brt:
Could be, yes, but we should find a way to make this more explicit. 
E.g. probably the name `ids-policy` and its position (extensions) do not fit well. If they should come out of the box probably the `core/policy` module would be a better choice.
Probably @jimmarino has an opinion on this

PR #1141, opened 2022-04-13
Fix: corrects content types in DataManagementApi controllers

### LINKED ISSUE #1126: DataManagementAPI: fix content-types
# Feature Request

Currently all API methods define 
```java
@Consumes({ MediaType.APPLICATION_JSON })
@Produces({ MediaType.APPLICATION_JSON })
```
this is not correct in many cases, because plain `String`s are returned, therefor the content type should be `MediaType.TEXT_PLAIN`. 

Public APIs (DataManagement)

Incorrect content types can cause problems down the road, especially when using client libs generated by the [Swagger Codegen](https://swagger.io/tools/swagger-codegen/), because clients may attempt to interpret a `String` as JSON, which would fail.

## Solution Proposal
Technically there are two possibilities: 
1. put the correct (and possibly different) content types on every controller method
2. have everything produce `application/json`, and enforce JSON-conformant return objects. We could achieve that in a generic way using a [`ContainerResponseFilter`](https://www.baeldung.com/jersey-filters-interceptors#2-implementing-a-response-server-filter) that checks if the response is convertible to JSON, and wraps it in JSON if that's not the case.

_Personally I'm leaning slightly toward the first solution as it avoids useless wrapping and unwrapping. Having different content types do exist for a reason._

## Type of Issue
Cleanup

## Checklist

paullatzelsperger:
## What this PR changes/adds

Adapts the `@Produces` and `@Consumes` annotations on all DataManagementAPI controller such that they reflect what the methods actually consume/produce.

## Why it does that

Wrong content types may cause problems when generating client SDKs from the OpenApi spec.

## Further notes

- fixed a wrong validity check in the `ContractNegotiationApiController`
- added default `provider` and `consumer` fields in the transformer.
- removed a logline from `StateMachine` that spammed the log every couple of seconds.

## Linked Issue(s)

Closes #1126 

## Checklist

### DIFF
--- a/common/state-machine-lib/src/main/java/org/eclipse/dataspaceconnector/common/statemachine/StateMachine.java
+++ b/common/state-machine-lib/src/main/java/org/eclipse/dataspaceconnector/common/statemachine/StateMachine.java
@@ -50,7 +50,7 @@ public class StateMachine {
         this.name = name;
         this.monitor = monitor;
         this.waitStrategy = waitStrategy;
-        this.executor = instrumentation.instrument(
+        executor = instrumentation.instrument(
                 Executors.newSingleThreadScheduledExecutor(r -> {
                     var thread = Executors.defaultThreadFactory().newThread(r);
                     thread.setName("StateMachine-" + name);
@@ -68,17 +68,11 @@ public class StateMachine {
         return submit(0L);
     }
 
-    @NotNull
-    private Future<?> submit(long delayMillis) {
-        monitor.debug(format("StateMachine [%s] delaying for %d ms", name, delayMillis));
-        return executor.schedule(loop(), delayMillis, MILLISECONDS);
-    }
-
     /**
      * Stop the loop gracefully
      *
      * @return a future that will complete when the loop is fully stopped.
-     *         The content of the future will be true if stop happened before the timeout, false elsewhere.
+     * The content of the future will be true if stop happened before the timeout, false elsewhere.
      */
     public CompletableFuture<Boolean> stop() {
         active.set(false);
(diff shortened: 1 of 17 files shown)


common/state-machine-lib/src/main/java/org/eclipse/dataspaceconnector/common/statemachine/StateMachine.java
```diff
@@ -50,7 +50,7 @@ private StateMachine(String name, Monitor monitor, ExecutorInstrumentation instr
         this.name = name;
```
ndr-brt, line 50: Since it's not scope of this PR, I would exclude these modification on this `StateMachine.java` file

ndr-brt (approved):

common/state-machine-lib/src/main/java/org/eclipse/dataspaceconnector/common/statemachine/StateMachine.java
paullatzelsperger: it's just one line... do you really want to open a separate PR for that? and it annoyed me very much :)

paullatzelsperger:
@ndr-brt: talked to @jimmarino and he is in favor of having every method produce APPLICATION_JSON, which would mean that API controller methods have to return some sort of `ApiResult` object, into which all actual return types get wrapped. This is more involved than meets the eye, because we'd then also have to improve the OpenAPI doc, adapt all controllers + tests, etc. so I'd suggest doing this in a separate PR. What's your take?

ndr-brt:
@paullatzelsperger yeah, having all the API return json it will make things simpler. Ok for doing that in another issue 👍

paullatzelsperger:
> @paullatzelsperger yeah, having all the API return json it will make things simpler. Ok for doing that in another issue +1

FYI I created #1175 as followup

PR #1144, opened 2022-04-13
core: replace policies with ids in ContractDefinition

### LINKED ISSUE #1137: `ContractDefinitionDtoToContractDefinitionTransformer` does not transform `Policies`
# Bug Report

## Describe the Bug

The `ContractDefinitionDtoToContractDefinitionTransformer` does not transform `Policies`. Therefore, all contract definitions that are created using the DataMgmt API do not contain any policy information besides the ID.

### Expected Behavior

`ContractDefinitionDtoToContractDefinitionTransformer` does not leave out information when creating `ContractDefinition` objects.

### Observed Behavior

`ContractDefinitions` that are created using the DataMgmtAPI carry the expected policy information. The `ContractDefinition` is takes later from the `ContractDefinitionStore` and causes unexpected behavior (e.g. `Offers` are created when the `Permission` wouldn't allow it).

## Steps to Reproduce

- Obvious in the code...

## Context Information

```java
  @Nullable
    public ContractDefinition transform(@Nullable ContractDefinitionDto object, @NotNull TransformerContext context) {
        return Builder.newInstance().id(object.getId()).accessPolicy(org.eclipse.dataspaceconnector.policy.model.Policy.Builder.newInstance().id(object.getAccessPolicyId()).build()).contractPolicy(org.eclipse.dataspaceconnector.policy.model.Policy.Builder.newInstance().id(object.getContractPolicyId()).build()).selectorExpression(org.eclipse.dataspaceconnector.spi.asset.AssetSelectorExpression.Builder.newInstance().criteria(object.getCriteria()).build()).build();
    }
```

## Detailed Description

-

## Possible Implementation

I assume there are two ways to fix this.
1. The `ContractDefinitionDtoToContractDefinitionTransformer` creates the `ContractDefinition` with the correct policy from the policy store.
2. The `ContractDefinition` should only contain the ID of the `Policy` and not the policy objects itself. This would force other components to always get the lates policy from the corresponding store.

I'd go with solution 2. What do you think?

Edit: Just saw that there is a ToDo for solution #2 in the source code. When I was debugging I only had the class files and didn't see it.

ndr-brt:
## What this PR changes/adds

Replace the fields `accessPolicy` and `contractPolicy` with `accessPolicyId` and `contractPolicyId` on `ContractDefinition`, and inject `PolicyStore` to `ContractDefinitionServiceImpl`, `ContractValidationServiceImpl` and `ContractOfferServiceImpl`

## Why it does that

Because policies are stored separately from contract definitions and they can be referenced by their id

## Further notes

- removed `NullAssetIndex` as its use was impossible to reach.
- probably `data-loader-cli` needs further work but seemed too much for this PR. Since data-management api is in place is it still useful?

## Linked Issue(s)

Closes #1137 
Part 3. of #828 

## Checklist


extensions/sql/contract-definition/store/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStoreTest.java
```diff
@@ -158,17 +159,11 @@ void saveMany_noneExist() {
...
-        sqlContractDefinitionStore.save(newDefs);
-
-        //verify that all have the properties
+        sqlContractDefinitionStore.save(getContractDefinitions(10));
 
         var definitionsRetrieved = sqlContractDefinitionStore.findAll();
 
-        assertThat(definitionsRetrieved).isNotNull();
-        assertThat(definitionsRetrieved).allSatisfy(cd -> assertThat(cd.getContractPolicy().getExtensibleProperties()).containsEntry("somekey", "someval"));
+        assertThat(definitionsRetrieved).hasSize(10);
```
paullatzelsperger, line 166: the `allSatisfy` assertion is actually important, as it makes sure, that all elements have been updated, even the pre-existing ones.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
```diff
@@ -37,41 +42,39 @@ public class ContractOfferServiceImpl implements ContractOfferService {
...
     private final ContractDefinitionService definitionService;
     private final AssetIndex assetIndex;
+    private final PolicyStore policyStore;
 
-    public ContractOfferServiceImpl(ParticipantAgentService agentService, ContractDefinitionService definitionService, AssetIndex assetIndex) {
+    public ContractOfferServiceImpl(ParticipantAgentService agentService, ContractDefinitionService definitionService, AssetIndex assetIndex, PolicyStore policyStore) {
         this.agentService = Objects.requireNonNull(agentService, "ParticipantAgentService must not be null");
         this.definitionService = Objects.requireNonNull(definitionService, "ContractDefinitionService must not be null");
         this.assetIndex = Objects.requireNonNull(assetIndex, "AssetIndex must not be null");
+        this.policyStore = policyStore;
```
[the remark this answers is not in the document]
ndr-brt: removed them all, they could not be null since the inject mechanism already assure that they cannot be null

extensions/sql/contract-definition/store/src/test/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStoreTest.java
ndr-brt: I had to remove it since now it's not possible to modify the policy object anymore (the policy object has been replaced by an id). The current assertion verifies that, when you save 3 definitions with id `0, 1, 2` and then you save 10 other definitions `0, 1, 2 ... 9`, the `0, 1, 2` actually got update, so at the end in the store there will be 10 entries, I think it's enough to verify the behavior.

paullatzelsperger: verifying that only makes sure that no more than 10 entities exist in total after the second `save`, but it does not assert whether the pre-existing ones have been _skipped_ or _updated_. 
So should the behavior of the store change at some point from upserting to skipping existing ones, the test would still succeed, which seems odd. Personally I generally want stuff to break when I change something.
what do you think about this:
```java
    @Test
    @DisplayName("Save multiple Contract Definitions with some preexisting Definitions")
    void saveMany_someExist() {
        var preexisting = getContractDefinitions(3);
        sqlContractDefinitionStore.save(preexisting);

        var newDefs = getContractDefinitions(10);
        newDefs.forEach(cd -> cd.getSelectorExpression().getCriteria().add(new Criterion("foo", "=", "bar")));
        sqlContractDefinitionStore.save(newDefs);

        var definitionsRetrieved = sqlContractDefinitionStore.findAll();

        assertThat(definitionsRetrieved).hasSize(10);
        var expectedCriterion = new Criterion("foo", "=", "bar");
        assertThat(definitionsRetrieved).allSatisfy(cd -> assertThat(cd.getSelectorExpression().getCriteria()).hasSize(1)
                .allSatisfy(c -> assertThat(c).isEqualTo(expectedCriterion)));
    }
```
[gist is here](https://gist.github.com/paullatzelsperger/426bdf961cd1e5cc52a0b240b55253aa)

ndr-brt: @paullatzelsperger In my opinion this is over-complicated. I would prefer to have a dedicated test to verify that.
For example, this assertion should be done by `saveOne_alreadyExist_shouldUpdate`:
```
    @Test
    @DisplayName("Save a single Contract Definition that already exists")
    void saveOne_alreadyExist_shouldUpdate() {
        sqlContractDefinitionStore.save(getContractDefinition("id", "policy", "contract"));
        sqlContractDefinitionStore.save(getContractDefinition("id", "updatedAccess", "updatedContract"));

        var result = sqlContractDefinitionStore.findAll();
        
        assertThat(result).hasSize(1).containsExactly(getContractDefinition("id", "updatedAccess", "updatedContract"));
    }
```

PR updated

paullatzelsperger: that's fine too, thanks.

bscholtes1A (approved):

Issue #1130, opened 2022-04-12
Story: use in-memory implementations by default


paullatzelsperger:
# Feature Request

All the `in-memory` implementations of various stores, registries, etc. that we currently have should be made the default, "if nothing else is on the classpath". This means no explicit gradle dependency on the `in-memory` modules is then necessary anymore.

Once #1056 is ready, we can utilize the default dependency provider mechanism to realize that. 
Also, we could then remove the `:extensions:in-memory` module.

## Which Areas Would Be Affected?
- [x] assetindex-memory
- [ ] contractdefinition-store-memory
- [ ] did-document-store-inmem
- [ ] fcc-node-directory-memory
- [ ] fcc-store-memory
- [ ] identity-hub-memory
- [ ] negotiation-store-memory
- [ ] policy-store-memory
- [ ] transfer-store-memory

_we may want to create separate issues for all those items listed here_

## Why Is the Feature Desired?
Reduce the error probability for new devs, reduce dependency complexity, always provide sane defaults.

## Solution Proposal
Use the default dependency provider mechanism (#1056) and move all the in-mem implementations into core.

## Type of Issue
Improvement, story

## Checklist

- [x] assigned appropriate label?
- [ ] **Do NOT select a milestone or an assignee!**

PR #1150, opened 2022-04-13
Feature: enhance dependency injection with factory methods 

paullatzelsperger:
## What this PR changes/adds

Adds functionality to annotate methods in extensions as `@Provider`, effectively marking them as factory methods.
This also enables us to provide dependencies "by default"

## Why it does that

to improve the DI mechanism and enable default dependencies.

## Further notes

- removes the `@Feature` annotation, as it is not needed anymore. This is also where many of the changes come from.
- many test objects, which previously were private inner classes, are now standalone classes. This also causes a few changed files.
- the `InjectorImpl` can now have resolve services from the `ServiceExtensionContext` and from an optional list of default services
- the `ExtensionLoader` class has been cleaned and stripped, now we have a `DependencyGraph` (takes care of sorting) and a `ExtensionLifecycleManager` (takes care of constructing the extension)
- developer documentation was updated

## Linked Issue(s)

Prerequisite for #1130 

## Checklist


algattik:
Given you've added classes, it would be advisable to rebase to apply the new checkstyle rule for headers (https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/1136)

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/ExtensionLoader.java
```diff
@@ -145,8 +141,7 @@ public static void loadVault(ServiceExtensionContext context, ExtensionLoader lo
     public List<InjectionContainer<ServiceExtension>> loadServiceExtensions() {
         List<ServiceExtension> serviceExtensions = loadExtensions(ServiceExtension.class, true);
 
-        //the first sort is only to verify that there are no "upward" dependencies from PRIMORDIAL -> DEFAULT
-        var sips = sortExtensions(serviceExtensions);
+        var sips = new DependencyGraph().of(serviceExtensions);
 
         return Collections.unmodifiableList(sips);
```
ndr-brt, line 146: Not done in this PR, but this `unmodifiableList` wrapper is not needed, as the list returned by `of` is already immutable

core/base/src/test/java/org/eclipse/dataspaceconnector/core/CoreServicesExtensionTest.java
```diff
@@ -16,18 +16,18 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
 
-import static org.mockito.ArgumentMatchers.anyInt;
-import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
 
+@ExtendWith(TypeManagerDependencyInjectionExtension.class)
```
ndr-brt, line 30: This starts to be overcomplicated for a test that verifies an initialization and not a behavior. I would remove the test (`verifyPolicyTypesAreRegistered`) as it doesn't give significant value

launchers/data-plane-server/config.properties
```diff
@@ -0,0 +1,17 @@
...
+#  https://www.apache.org/licenses/LICENSE-2.0
+#
+#  SPDX-License-Identifier: Apache-2.0
+#
+#  Contributors:
+#       Microsoft Corporation - initial API and implementation
+#
+#
+web.http.public.port=1234
+web.http.public.path=/yomama
```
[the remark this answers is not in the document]
paullatzelsperger: oops, this should never have gotten committed. thanks for noticing.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/injection/ProviderMethod.java
```diff
@@ -0,0 +1,68 @@
...
+ * Represents a {@link Method} that is annotated with the {@link Provider} annotation.
+ */
+public class ProviderMethod {
+    private final Method method;
+    private final boolean isDefault;
+
+    public ProviderMethod(Method method) {
+        this.method = method;
+        isDefault = ofNullable(method.getAnnotation(Provider.class)).map(Provider::isDefault)
+                .orElseThrow(() -> new IllegalArgumentException(format("Method %s is not annotated with @Provider!", method)));
```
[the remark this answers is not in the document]
paullatzelsperger: I guess it's a matter of preference, but strictly speaking this deals with validating _input_, rather than actual dependency injection, therefore I went with the `IllegalArgumentException`. Theoretically `ProviderMethod`s could be used for something else too

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/ExtensionLoader.java
paullatzelsperger: `DependencyGraph.of` returns a `List`, the fact that it happens to be immutable is an implementation detail that could change at any time, so we cannot rely on it.

ndr-brt: @paullatzelsperger same is for `loadServiceExtensions` method and the `createExtensions` that calls it, since it's internal code (not to be extended or so) I wouldn't be so defensive

paullatzelsperger: mh, you're probably right. will ax it.

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java
```diff
@@ -98,6 +98,11 @@ public <T> T getService(Class<T> type, boolean isOptional) {
 
     @Override
     public <T> void registerService(Class<T> type, T service) {
+        registerServiceRaw(type, service);
```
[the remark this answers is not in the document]
paullatzelsperger: typically I think people should use the typed one to increase type safety and avoid runtime crashes.
I had to introduce the `*raw` one because `Method#getReturnType()` returns an unbound wildcard `Class<?>`. 
I'll just cast that to a `T`, then the `*raw` method can be removed.

core/base/src/test/java/org/eclipse/dataspaceconnector/core/CoreServicesExtensionTest.java
paullatzelsperger: we do test serialization (`PolicyTest`), but it's not just about making sure that Policies and all the other types can be deserialized, it's also about making sure that we ship code that has those types _already registered_. 

Otherwise we could run into a situation where serialization per se works fine, but the API still crashes, because we forgot to register some `Constraint` implementor with the `TypeManager`.

ndr-brt (approved):

jimmarino (approved):

PR #1196, opened 2022-04-21
fix: Introduce registry for EndpointDataReferenceTransformer

### LINKED ISSUE #1181: `EndpointDataReferenceTransformer` might be injected before it is overwritten
# Bug Report

## Describe the Bug
In current implementation, a default `EndpointDataReferenceTransformer` is register within `CoreServicesExtension`. This default implementation returned the `EndpointDataReference` without any modification. This default transformer can then be overwritten by another extension (see `data-plane-transfer` example for an example). 

However, depending on the order in which Extensions are initialized, it is possible that the default `EndpointDataReferenceTransformer` is injected before it is overridden by a dedicated extension. This might leads to situation in which the `EndpointDataReference` is returned as it went while we expect a transformation.

### Expected Behavior
If a `EndpointDataReferenceTransformer` implementation is registered, then it should be apply regardless of the extension initialization order. 

### Observed Behavior
Depending on the extension initialization order, expected transformation might not be applied.

## Possible Implementation
A registry pattern should be preferred for the `EndpointDataReferenceTransformer`, in the same way it is done for the `EndpointDataReferenceReceiver`. This approach offers the advantage that the transformation logic is dynamically determined (i.e. not at extension initialization time), which makes it independent from the extension initialization order.

bscholtes1A:
## What this PR changes/adds

This PR introduces the registry pattern for `EndpointDataReferenceTransformer`, in the same it is done for `EndpointDataReferenceReceiverRegistry`.

## Why it does that

When performing end-to-end testing it was observed that, depending on the injection order of the extensions, the `DefaultEndpointDataReferenceTransformer` (which does nothing) instantiated in `CoreServicesExtension` can be injected before it is actually overriden. 

This might lead to some situation wherein we expect a transformation operation to be performed on the `EndpointDataReference`, while actually nothin is done.

## Linked Issue(s)

Closes #1181.

## Checklist


data-protocols/ids/ids-api-multipart-endpoint-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/handler/EndpointDataReferenceHandler.java
```diff
@@ -70,24 +69,15 @@ public boolean canHandle(@NotNull MultipartRequest multipartRequest) {
...
-                .collect(asyncAllOf())
-                .thenApply(results -> results.stream()
-                        .filter(Result::failed)
-                        .findFirst()
-                        .map(failed -> Result.failure(failed.getFailureMessages()))
-                        .orElse(Result.success("Successful operation")))
-                .exceptionally(throwable -> Result.failure("Unhandled exception raised when transferring data: " + throwable.getMessage()))
-                .join();
 
+        var receiveResult = receiverRegistry.receiveAll(transformedEdr).join();
```
ndr-brt, line 80: Not for this PR but we should make this thing async to avoid blocking threads

ndr-brt (approved):
neat

PR #1200, opened 2022-04-22
Improvement: REST API strict body checking

### LINKED ISSUE #1128: Story: add strict body validation for REST APIs
# Feature Request

This story deals with two things actually:
- [ ] validate the request body using Jax-RS `@Valid` annotations and constraints
- [ ] disallow unrecognized JSON properties.

REST APIs/endpoints

While the first item serves mostly serves to fail fast when invalid request bodies are supplied, the second one could actually be a security feature (overloading an endpoint with very large JSON).

## Solution Proposal
ad 1) use the Jax-RS `@Valid` annotation plus the corresponding constraint annotations (e.g. `@NotNull`,...)
ad 2) use an `ObjectMapper` that sets the `FAIL_ON_UNKNOWN_PROPERTIES` flag to true: 
```java
objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
_Note that this MUST ONLY be set for REST APIs, possibly that means generating/duplicating the `ObjectMapper` specifically for that purpose_

## Type of Issue
Improvement

relates to #1111 

## Checklist

paullatzelsperger:
## What this PR changes/adds

This PR introduces DTO/body validation for public facing REST APIs using `jakarta.validation` annotations.

## Why it does that

To provide a transparent and easy way to validate request bodies.

## Further notes

- n/a
 
## Linked Issue(s)

Closes #1128 

## Checklist


extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/AssetApi.java
```diff
@@ -26,7 +27,7 @@
 @Tag(name = "Asset")
 public interface AssetApi {
 
-    void createAsset(AssetEntryDto assetEntryDto);
+    void createAsset(@Valid AssetEntryDto assetEntryDto);
```
ndr-brt, line 30: jakarta annotations should be defined only on implementations, leaving only swagger ones on the interfaces.

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/model/NegotiationInitiateRequestDto.java
```diff
@@ -70,7 +73,7 @@ public Builder connectorId(String connectorId) {
             return this;
         }
 
-        public Builder offerId(ContractOfferDescription offerId) {
+        public Builder offer(ContractOfferDescription offerId) {
```
ndr-brt, line 76: Also the parameter could be called `offer`

extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/TransferProcessApiController.java
```diff
@@ -133,7 +134,7 @@ public void deprovisionTransferProcess(@PathParam("id") String id) {
     @Consumes({ MediaType.APPLICATION_JSON })
     @Produces({ MediaType.TEXT_PLAIN })
     @Override
-    public String initiateTransfer(TransferRequestDto transferRequest) {
+    public String initiateTransfer(@Valid TransferRequestDto transferRequest) {
         if (!isValid(transferRequest)) {
```
ndr-brt, line 138: this validation could be done with the jakarta validation as well

ndr-brt (commented):
What about introducting also an object that contains all the `QuerySpec` values and add validations on that too? (through `@BeanParam`)

extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/AssetApi.java
paullatzelsperger: unfortunately that won't work, as the validator will then raise an exception: 
```java
HV000151: A method overriding another method must not redefine the parameter constraint configuration, but method <XYZ> redefines the configuration of <XYZ>
```
so we're stuck with leaving it there, i'm afraid.

extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/TransferProcessApiController.java
paullatzelsperger: should be implicitly done, removed `isValid`

extensions/api/data-management/contractnegotiation/src/test/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/model/ContractOfferDescriptionValidationTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @BeforeEach
+    void setUp() {
+        try (ValidatorFactory factory = Validation.buildDefaultValidatorFactory()) {
+            validator = factory.getValidator();
+        }
+    }
+
+    @ParameterizedTest
+    @MethodSource(value = "provideInvalidProperties")
```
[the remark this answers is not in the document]
paullatzelsperger: > What about introducing also an object that contains all the `QuerySpec` values and add validations on that too? (through `@BeanParam`)

didn't know about this - seems like a nice and elegant way to group request parameters +`

paullatzelsperger:
@ndr-brt 
> What about introducing also an object that contains all the QuerySpec values and add validations on that too? (through @BeanParam)

didn't know about this TBH, seems like a clean way to supply request params. will adapt +1

ndr-brt:
> > What about introducing also an object that contains all the QuerySpec values and add validations on that too? (through @BeanParam)
> 
> didn't know about this TBH, seems like a clean way to supply request params. ~will adapt +1~
> 
> [edit] On closer inspection this would require us to put JAX-RS annotations directly on the `QuerySpec` object so that it gets picked up by the Swagger YAML generator, which would introduce a dependency not only onto `jakarta.validation` (which is expected) but also onto `jakarta.ws.rs:jakarta.ws.rs-api` in `core-spi`. @ndr-brt do we really want that?

Probably we need to introduce another object representing the query that will be mapped to `QuerySpec` after validation. For sure something for another PR. In the meantime LGTM

ndr-brt (approved):

PR #1213, opened 2022-04-25
DataManagementApi: all API endpoints produce and consume APPLICATION_JSON

### LINKED ISSUE #1175: Chore: use `application/json` as content type in all public API methods
# Feature Request

In order to have consistency across all public APIs, every controller method must `@Produce` and `@Consume` `application/json`. This means that all plaintext responses, for example the response to a contract-negotiation-initiation would have to be wrapped in a JSON object.

We could even go so far as to devise a global `ApiResult` object, that wraps _every_ response object. API methods would still return the proper HTTP status code (200 and others), and carry the `ApiResult` in the response body. 

We have to check how the OpenApi generator plugin behaves in this situation, it would be bad to lose information about the returned JSON object. Maybe we can counter this with custom annotations somehow.

Data Management API

Having a consistent API scheme is desirable and will make code-generated clients easier to handle.

## Solution Proposal
The `ApiResult` and could look like the this:
```java
public class ApiResult<T> {
   private T content; //only non-null if no error
   private Integer errorCode; //null if no error, maybe leave out?
   private String errorMessage; //null if no error
}
```

the `errorCode` could contain either the HTTP status code, which seems redundant, or it could contain application error codes, should we ever support those, or we could leave it out altogether. (I'm in favor of leaving it out)

## Type of Issue
Improvement

## Tasks
- [ ] Investigate how OpenApi behaves with those generic objects
- [ ] Update all `@Produces` and `@Consumes` annotations, move to the class level
- [ ] wrap every controller response in an `ApiResult`

## Checklist

paullatzelsperger:
## What this PR changes/adds

All API endpoints of the DataManagementApi now `@Produce` and `@Consume` `APPLICATION_JSON`. Those controller methods that previously returned a simple `String` now return a complex object that can be serialized to JSON in order to make things consistent.

## Why it does that

To allow easier consumption in (generated) API clients.

## Further notes

n/a

## Linked Issue(s)

Closes #1175 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ in the detailed section referring to by linking pull requests or issues.
 * Restructure sql extension folder tree (#1154)
 * Extract single `PolicyArchive` implementation (#1158)
 * Replace `accessPolicy` and `contractPolicy` with `accessPolicyId` and `contractPolicyId` on `ContractDefinition` (#1144)
+* All DMgmt Api methods now produce and consume `APPLICATION_JSON` (#1175)
 
 #### Removed
 
--- a/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/AssetApiController.java
+++ b/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/AssetApiController.java
@@ -47,6 +47,8 @@ import java.util.stream.Collectors;
 
 import static java.lang.String.format;
 
+@Consumes({ MediaType.APPLICATION_JSON })
+@Produces({ MediaType.APPLICATION_JSON })
 @Path("/assets")
 public class AssetApiController implements AssetApi {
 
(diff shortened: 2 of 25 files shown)


bscholtes1A (approved):

ndr-brt (commented):
Probably also the samples need to be aligned

paullatzelsperger:
> Probably also the samples need to be aligned

which samples did you have in mind? They don't use the DM API afaik. The system tests have already been adapted.

ndr-brt:
> which samples did you have in mind? They don't use the DM API afaik. The system tests have already been adapted.

sample 04.0 for sure, probably also 04.1 04.2 04.3 and 05 as well

paullatzelsperger:
> > which samples did you have in mind? They don't use the DM API afaik. The system tests have already been adapted.
> 
> sample 04.0 for sure, probably also 04.1 04.2 04.3 and 05 as well

I only found occurrences of the DM API in the README.md (`curl`) and those are written such that no adaptation seems necessary.

ndr-brt (approved):

PR #1220, opened 2022-04-26
Refactor: replace policy with policyId on ContractAgreement

### LINKED ISSUE #828: Refactor structure of business objects

ndr-brt:
## What this PR changes/adds

Replace `Policy` instance with `policyIs` on `ContractAgreement`

## Why it does that

To avoid data duplication

## Further notes

- It's the second point of #828 and the final one
- `PolicyArchive` now relies on `PolicyStore` to get the `Policy` instance

## Linked Issue(s)

Closes #828

## Checklist


data-protocols/ids/ids-api-multipart-endpoint-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/handler/ContractAgreementHandler.java
```diff
@@ -105,19 +104,18 @@ public boolean canHandle(@NotNull MultipartRequest multipartRequest) {
...
             return createBadParametersErrorMultipartResponse(message);
         }
 
         // TODO get hash from message
-        var agreement = result.getContent();
+        var output = result.getContent();
         var processId = message.getTransferContract();
         var negotiationResponse = negotiationManager.confirmed(claimToken,
-                String.valueOf(processId), agreement, null);
+                String.valueOf(processId), output.getContractAgreement(), null);
```
ronjaquensel, line 118: The policy from the output should be persisted. Just the agreement is handed over to the negotiation manager, so the policy is lost at that point. Therefore, the agreement then references a policy id that does not exist in the store.

ndr-brt: @ronjaquensel thanks, yes, this made total sense, I was just wondering why nobody was using that policy object but it's because it needs to be stored locally so then after it can be used later on. I will introduce policy persistence in the `confirmed` method of the negotiation manager.

ronjaquensel (approved):

paullatzelsperger (approved):

Issue #1183, opened 2022-04-21
End-to-end blob storage transfer is not working


algattik:
# Bug Report

## Describe the Bug

End-to-end blob storage transfer is not working

## Steps to Reproduce
- Run consumer and provider EDC
- Perform contract negotiation
- Issue a transfer request such as
```
{
    "managedResources": false,
    "protocol": "ids-multipart",
    "connectorAddress": "http://localhost:7171/api/v1/ids/data",
    "dataDestination": {
        "properties": {
            "container": "testdocument",
            "type": "AzureStorageBlobData",
            "account": "testdocument",
            "blobname": "5a69f2f5-86b0-4244-8d17-1b786889f153"
        }
    },
    "id": "f5266fba-bf2c-4947-bfca-07c60f41c49b",
    "connectorId": "consumer",
    "contractId": "4a75736e-001d-4364-8bd4-9888490edb56:559bed97-0873-4b36-af4c-da953e4591f8",
    "assetId": "text-document",
    "transferType": {
        "contentType": "application/octet-stream",
        "isFinite": true
    }
}
```
On the provider side message, no blob account key is received

<img width="469" alt="image" src="https://user-images.githubusercontent.com/22099722/164403250-2efd0939-1dc6-40bb-bd8b-2d66486f5999.png">

## Detailed Description
_If applicable, add screenshots and logs to help explain your problem._

PR #1221, opened 2022-04-26
TransferProcess API: provide provisioned resource information

algattik:
## What this PR changes/adds

Provide DataManagement API client information about data destination properties, such as the automatically provisioned Azure storage container name when `managedResources=true`.

## Why it does that

End-to-end flow for blob storage transfers across EDC and DPF is not working (#1183). This is one of multiple PRs to solve several issues causing the problem; this PR is required to build a system test to validate the flow end to end.

When performing a transfer to Azure blob with `managedResources=true`, the consumer EDC provisions a storage container with a randomly generated name. The container is used as destination of the data copy on the provider DPF.

The client polls the TransferProcess API for completion. When the transfer is completed, the client can access the data in the storage container. However, the client currently had no way of knowing what storage container was created by the consumer provisioner. 

When using a blob storage provisioned account, the `dataDestination` will look like

```
{HashMap@15791}  size = 4
 "container" -> "7162210a-5ee9-4f43-8443-7a104c4d15fc"
 "keyName" -> "resource"
 "type" -> "AzureStorage"
 "account" -> "account2"
```

## Further notes

This is required for building system tests that verify actual data has been written. It can also be used e.g. by web app clients to generate download links.

Added a `transform_whenMinimalData` UT case to ensure no NPE when missing custom DataDestination properties and the like.

## Linked Issue(s)

Closes #1221

## Checklist

### DIFF
--- a/extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/model/TransferProcessDto.java
+++ b/extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/model/TransferProcessDto.java
@@ -18,6 +18,8 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 
+import java.util.Map;
+
 @JsonDeserialize(builder = TransferProcessDto.Builder.class)
 public class TransferProcessDto {
     private String id;
@@ -25,6 +27,7 @@ public class TransferProcessDto {
     private String state;
     private String errorDetail;
     private DataRequestDto dataRequest;
+    private Map<String, String> dataDestination = Map.of();
 
     private TransferProcessDto() {
     }
@@ -47,6 +50,10 @@ public class TransferProcessDto {
 
     public DataRequestDto getDataRequest() {
         return dataRequest;
+    }
+
+    public Map<String, String> getDataDestination() {
+        return dataDestination;
     }
 
     @JsonPOJOBuilder(withPrefix = "")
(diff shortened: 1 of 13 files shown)


extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/model/TransferProcessDto.java
```diff
@@ -18,13 +18,16 @@
...
+import java.util.Map;
+
 @JsonDeserialize(builder = TransferProcessDto.Builder.class)
 public class TransferProcessDto {
     private String id;
     private String type;
     private String state;
     private String errorDetail;
     private DataRequestDto dataRequest;
+    private Map<String, String> dataDestination = Map.of();
```
ndr-brt, line 30: For coherence with the `asset` data management api, a dedicated `DataAddressDto` object could be used instead

algattik: @ndr-brt would that object only contain a map? Since the values (keys) are dependent on the address type, not sure what that would bring?

ndr-brt: Yes, it does not bring much, it's just for coherence, if here the `DataAddress` is represented as a `Map`, the same should be on the `asset` api.

algattik: It turns out the `DataAddressDto` already exists :) will reuse

algattik: Added `DataAddressInformationDto` (to avoid name conflict with `DataAddressDto` used in `asset`)

ndr-brt: @algattik I don't think they have to be called differently, as they are in different modules and have different packages, also `ContractAgreementDto` has two different definitions

paullatzelsperger (approved):
I'm not too crazy about the `TransferProcessTransformerTestData`, mostly because prefer a `Testfunctions` class with static methods. Thus I find it a bit convoluted and difficult to read the accompanying test, but that might just be my personal preference.

PR #1223, opened 2022-04-27
Feature: provide in-mem implementations by default

### LINKED ISSUE #1130: Story: use in-memory implementations by default

paullatzelsperger:
## What this PR changes/adds

All in-memory implementations of various stores are now provided as default (using the `@Provider` annotation). This means that it is not necessary anymore to add dependencies onto `:extensions:in-memory:xyzstore` anymore.

## Why it does that

Having all those in-memory extensions had a couple of major downsides:
- no defaults were provided
- it was easy to forget to add them to the build file
- there were lots of pretty thin `in-memory:`-modules

*With this PR, sane defaults are provided at all times, reducing user error and erroneous build configurations.*

## Further notes

- all the "core" stores, such as `AssetIndex`, `ContractDefinitionStore` etc. are now provided through the new `:core:defaults` module, which `:core:base` depends on. It is therefore sufficient to add `:core:base`, and a default (=in-mem) impl will be registered if no other impl is found on the classpath. 
- in-mem implementations for catalog-related stores was moved into `:extensions:catalog:federated-catalog-cache`
- in-mem implementations for identityhub store and did document store were moved into `:extensions:iam:decentralized-identity:identity-did-core`

## Linked Issue(s)

Closes #1130

## Checklist


samples/other/run-from-junit/src/test/java/org/eclipse/dataspaceconnector/junit/EndToEndTest.java
```diff
@@ -161,6 +162,9 @@ private void loadNegotiation(ContractNegotiationStore negotiationStore) {
 
     @Provides(IdentityService.class)
     private static class TestServiceExtension implements ServiceExtension {
+        @Inject
+        private AssetLoader loader;
```
bscholtes1A, line 166: needed?

bscholtes1A (approved):

samples/other/run-from-junit/src/test/java/org/eclipse/dataspaceconnector/junit/EndToEndTest.java
paullatzelsperger: unfortunately yes, because the way default provides work is that they only get instantiated "on demand", and if there is no "demand" (which is what `@Inject` is), it won't get instantiated, thus not being provided by the `EdcExtension` resulting in an exception when running the test.

core/base/src/main/java/org/eclipse/dataspaceconnector/core/defaults/DefaultServicesExtension.java
```diff
@@ -0,0 +1,89 @@
...
+ * Provides (in-mem) defaults for various stores, registries etc.
+ * Provider methods are only invoked if no other implementation was found on the classpath.
+ */
+public class DefaultServicesExtension implements ServiceExtension {
+
+    private final InMemoryAssetIndex assetIndex;
+    private final InMemoryContractDefinitionStore contractDefinitionStore;
+
+    public DefaultServicesExtension() {
+        assetIndex = new InMemoryAssetIndex();
```
ndr-brt, line 46: Not a big issue, but doing like this, there will be an `InMemoryAssetIndex` and an `InMemoryContractDefinitionStore` instance loaded in every connector, also if the store configured is `sql`. We could not care about that, but I would put a lazy loaded singleton there to make sure that the instance is created only if it's needed.

samples/other/run-from-junit/src/test/java/org/eclipse/dataspaceconnector/junit/EndToEndTest.java
ndr-brt: @paullatzelsperger I think this test class is somewhat obsolete since now `system-tests` exist, I would delete it completely

core/base/src/main/java/org/eclipse/dataspaceconnector/core/defaults/DefaultServicesExtension.java
paullatzelsperger: that makes sense.

launchers/ids-connector/README.md
```diff
@@ -29,7 +29,7 @@ The following modules are used for this launcher:
 | Name                                                 | Description                                                                        |
 |------------------------------------------------------|------------------------------------------------------------------------------------|
 | extensions:in-memory:transfer-store-memory           | store for persisting `TransferProcesses`                                           |
-| extensions:in-memory:assetindex-memory               | store for persisting `Assets`                                                      | 
+| core:defaults               | store for persisting `Assets`                                                      | 
 | extensions:in-memory:negotiation-store-memory        | store for persisting `ContractNegotiations`                                        | 
 | extensions:in-memory:contractdefinition-store-memory | store for persisting `ContractDefinitions`                                         | 
 | extensions:in-memory:policy-store-memory             | store for persisting `Policies`                                         |
```
ronjaquensel, line 35: The other in-memory stores should be removed from the table, as they're not in the build file anymore.

ronjaquensel (approved):

core/base/src/main/java/org/eclipse/dataspaceconnector/core/defaults/DefaultServicesExtension.java
```diff
@@ -0,0 +1,96 @@
...
+    public DataAddressResolver defaultDataAddressResolver() {
+        return getAssetIndex();
+    }
+
+    @Provider(isDefault = true)
+    public AssetLoader defaultAssetLoader() {
+        return getAssetIndex();
+    }
+
+    private static InMemoryAssetIndex getAssetIndex() {
```
ndr-brt, line 65: could be not static

core/base/src/main/java/org/eclipse/dataspaceconnector/core/defaults/DefaultServicesExtension.java
```diff
@@ -0,0 +1,96 @@
...
+ * Provider methods are only invoked if no other implementation was found on the classpath.
+ */
+public class DefaultServicesExtension implements ServiceExtension {
+
+    private static InMemoryAssetIndex assetIndex;
+    private final InMemoryContractDefinitionStore contractDefinitionStore;
+
+    public DefaultServicesExtension() {
+        assetIndex = new InMemoryAssetIndex();
+        contractDefinitionStore = new InMemoryContractDefinitionStore();
```
ndr-brt, line 47: also the contract definition store should be loaded lazily, and the initialization in the constructor avoided

ndr-brt (approved):

PR #1231, opened 2022-04-28
Chore: Fixed some minor snags with OpenApi

paullatzelsperger:
## What this PR changes/adds

Makes operation IDs unique (`CatalogApiController` and `FederatedCatalogApiController` both had a method with the same name), adds description to `CatalogApiController#getCatalog`.

## Why it does that

The OpenAPI spec was invalid - the OpenAPI Generator validates the spec and reported those errors.

## Further notes

- removed an old file (`catalog.yaml`) as it is replaced by `federated-catalog-cache.yaml` due to a previous renaming.

## Checklist


bscholtes1A (approved):

extensions/catalog/federated-catalog-cache/src/main/java/org/eclipse/dataspaceconnector/catalog/cache/controller/FederatedCatalogApiController.java
```diff
@@ -29,8 +29,8 @@
 
 import java.util.Collection;
 
-@Consumes({MediaType.APPLICATION_JSON})
-@Produces({MediaType.APPLICATION_JSON})
+@Consumes({ MediaType.APPLICATION_JSON })
```
ndr-brt, line 32: I noticed that everyone has their taste in fact of whitespaces on arrays, why don't avoid curly braces since we only produce/consume only json?
Like
```
@Consumes(MediaType.APPLICATION_JSON)
@Produces(MediaType.APPLICATION_JSON)
```

ndr-brt (approved):

extensions/catalog/federated-catalog-cache/src/main/java/org/eclipse/dataspaceconnector/catalog/cache/controller/FederatedCatalogApiController.java
paullatzelsperger: Those sorts of things should be handled by the IntelliJ code formatter. The one i'm using is located at `resources/edc-codestyle.xml`.

ndr-brt: yeah, I don't care that much too, but looking at the history seems like it's not the first time that this changes, just to avoid ping-pong changelogs :)

Issue #1163, opened 2022-04-14
DPF: public API input validation


MoritzKeppler:
# Feature Request
`DataPlanePublicApiController` provides an endpoint that is accessible from 'the world outside'.
The data plane should provide at least some basic mechanisms to protect itself against fraudulent or accidental broken calls, e.g. limit the size of the body, path, query params etc.
Some basic checks are sufficient. More sophisticated checks can always be done as part of an ingress controller, WAF, proxy, .... placed before the endpoint - if needed.

## Which Areas Would Be Affected?
DPF

## Why Is the Feature Desired?
protect service

## Solution Proposal
define max body size (config value with sensible default)
max size for URL parts

## Type of Issue
improvement

ndr-brt:
Jetty by default already provides limit for them: 

- "request header size", that's the limit for the header size of a request, it's set to 8KB by default:
https://github.com/eclipse/jetty.project/blob/cb127793e5d8b5c5730b964392a9a905ba49191d/jetty-server/src/main/java/org/eclipse/jetty/server/HttpConfiguration.java#L57

- "form content size", that's the limit for the body, it's set to 200000 by default (less than 200KB):
https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-server/src/main/java/org/eclipse/jetty/server/handler/ContextHandler.java#L146

So this issue can be closed in my opinion, unless we want those parameters configurable, but that would be a different issue.

MoritzKeppler:
thanks for the investigation @ndr-brt!

Definitely good to know that Jetty comes with some useful defaults (even if one could argue that it would be better to set values explicitly and not trust other libs to be configured properly)

If I got it right the max form content size is just used when parsing params from a http body that contains form values and has no influence on the number of bytes we read in  `ContainerRequestContextApiImpl`.

Suggest to at least add some lines to the README of the extension describing what can be configured to protect the dataplane / what needs to be done in the infrastructure added in front of the dataplane.

<sub> Moritz Keppler <moritz.keppler@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

ndr-brt:
@MoritzKeppler I dived deep into this, apparently, the `request header size` is working fine, while the `max form content size`, as the name speaks, limits only the form requests.

[...]

PR #1258, opened 2022-05-04
data-management-api: validate query parameters

### LINKED ISSUE #1235: DataManagement API: query parameters validation
# Feature Request

Currently, the query params used to query the data management services are not validated.
Some fields need to be validated, e.g. `limit` should have a maximum value, cannot be lower than 0, and so on

Data management API

To block invalid requests as soon as possible.

## Solution Proposal
An object that represent the query should be introduced and validated through the jakarta validation as done in #1200

## Type of Issue
improvement

## Checklist

ndr-brt:
## What this PR changes/adds

Adds validation on query params before building the `QuerySpec` object.
To do that introduces a `QuerySpecDto` object that is valued through the `@BeanParam` annotation.

## Why it does that

To better validate input on query call

## Further notes

- Bump some dependencies

## Linked Issue(s)

Closes #1235 

## Checklist


extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/query/QuerySpecDto.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @QueryParam("limit")
+    @Positive
+    private Integer limit = 50;
+
+    @QueryParam("filter")
+    private String filter;
+
+    @QueryParam("sort")
+    private String sortOrder = "ASC";
```
ronjaquensel, line 37: Is it possible to change the type to `SortOrder` instead of String? This seems to cause the allowed enum values to be removed from the openAPI yaml files.

extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/query/QuerySpecDto.java
```diff
@@ -0,0 +1,128 @@
...
+
+    public String getSortOrder() {
+        return sortOrder;
+    }
+
+    public String getSortField() {
+        return sortField;
+    }
+
+    @AssertTrue
```
paullatzelsperger, line 66: This needs to have the `@JsonIgnore` annotation, otherwise Jackson will look for a `valid` field. A serialization test would have surfaced that. Mind adding one?

paullatzelsperger (changes requested):

extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/query/QuerySpecDto.java
ndr-brt: I used a string to decouple the `dto` model from the `domain` model. 
But yes, introducing a `SortOrderDto` seems like to much, I will set it back to `SortOrder`

extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/query/QuerySpecDto.java
ndr-brt: This class is never serialized to/deserialized from json, so it's not needed

ronjaquensel (approved):

extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/query/QuerySpecDto.java
paullatzelsperger: I haven't checked, but what about tests (integration, e2e)?

ndr-brt: test are working just fine without touching them, I just add a "failing query" test for every controller

bscholtes1A:
Hi @ndr-brt 
looks good!
Not related to this PR but we have in the backlog the input validation for Data Plane API #1163 
The different is that the public API of the DPF takes random path/query params and body (they are forwarded to the backend data source). 

Will the approach you introduce in this PR also applicable in the context of the Data Plane API?

ndr-brt:
hey @bscholtes1A I don't think the same approach could be used there because, as you said, there isn't the possibility to know what parameters will be passed upfront.
Probably the simplest (and good enough?) validation there could be just a size limit for body and query

extensions/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/query/QuerySpecDto.java
paullatzelsperger: sure, that'll work, but as far as I'm aware @algattik has plans to generate a java API client from our own OpenAPI spec and use that in tests. If that is the case we'd indeed have to serialize the `QuerySpecDto`.

ndr-brt: @paullatzelsperger I'm not sure about that, however I would leave this for that issue, if it's needed it will be done

paullatzelsperger: ok, fine for me.

paullatzelsperger (approved):

bscholtes1A (approved):

PR #1263, opened 2022-05-05
refactor: move vault accessors to the ServiceExtensionContext class

### LINKED ISSUE #1237: Give to non-injectable services a dedicated method on the `ServiceExtensionContext`
# Feature Request

Currently, services like `Vault`, `PrivateKeyResolver` and `CertificateResolver` that are registered by default are obtained throught the `ServiceExtensionContext.getService`, this is unclear because that method is used to get services that are injected through `@Inject` or required by `@Requires` but these services are not provided by anyone.

core

Having dedicated methods on the context will avoid ambiguity.

## Solution Proposal
-

## Type of Issue
improvement

## Checklist

ndr-brt:
## What this PR changes/adds

Move `Vault`, `PrivateKeyResolver` and `CertificateResolver` to `ServiceExtensionContext`

## Why it does that

Currently the `Vault`/`PrivateKeyResolver`/`CertificateResolver` are not "injectable" (with the `@Inject` annotation) by anyone because they aren't "provided" (with the `@Provides` annotations), because they are mandatory and loaded at the "runtime" stage.
This is pretty confusing, so I moved their instance to the context class as the other mandatory services (like monitor, type manager, telemetry and so on...)

## Further notes

- Inlined some code from the `ExtensionLoader.loadVault` to the `BaseRuntime` as there was too much indirection (and static methods are not good for keeping such behavior)
- switching the vault-related `registerServiceMock` calls to the `registerSystemExtension` as a vault extension should be registered otherwise the `Null` one will be loaded

## Linked Issue(s)

Closes #1237 

## Checklist


jimmarino:
@ndr-brt @paullatzelsperger  Can't we just makes these injectable? It seems like we are starting to add too much to the ServiceExtensionContext.

ndr-brt:
@jimmarino what should differentiate a service that needs to be in the context and one that needs to be injected? Should also `Monitor`, `TypeManager` and `Telemetry` being "injectable"?

paullatzelsperger:
> @jimmarino what should differentiate a service that needs to be in the context and one that needs to be injected? Should also `Monitor`, `TypeManager` and `Telemetry` being "injectable"?

@ndr-brt I like the idea of having the `Vault` etc. as "normal" services, thus being registered into the context, just like normal services. We could still have dedicated methods `getVault()` or `getTypemanager()` on the service context, if we need them. WDYT?

jimmarino:
> @jimmarino what should differentiate a service that needs to be in the context and one that needs to be injected? Should also `Monitor`, `TypeManager` and `Telemetry` being "injectable"?

+1 on making those injectable

ndr-brt:
Fine for me, same for `Monitor`, `TypeManager` and `Telemetry` I imagine. Will close this and open an issue about.

PR #1273, opened 2022-05-10
Bugfix: clear FCC store before updating

paullatzelsperger:
## What this PR changes/adds

This PR cleans out the FCC cache store before re-populating it again with the results of a crawl run.

## Why it does that

To avoid having stale entries: when a connector removes a contract definition, all other connectors would still have corresponding (but invalid) offers.

## Further notes
n/a

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ in the detailed section referring to by linking pull requests or issues.
 * Make data-plane public api controller asynchronous (#1228)
 * Provide In-mem implementations by default (#1130)
 * Changed Catalog config keys and switched from minutes to seconds
+* Clean FCC store before updating
 
 #### Removed
 
--- a/extensions/catalog/federated-catalog-cache/src/test/java/org/eclipse/dataspaceconnector/catalog/defaults/store/InMemoryFederatedCacheStoreTest.java
+++ b/extensions/catalog/federated-catalog-cache/src/test/java/org/eclipse/dataspaceconnector/catalog/defaults/store/InMemoryFederatedCacheStoreTest.java
@@ -17,6 +17,7 @@ package org.eclipse.dataspaceconnector.catalog.defaults.store;
 
 import org.eclipse.dataspaceconnector.catalog.spi.FederatedCacheStore;
 import org.eclipse.dataspaceconnector.catalog.store.InMemoryFederatedCacheStore;
+import org.eclipse.dataspaceconnector.common.concurrency.LockManager;
 import org.eclipse.dataspaceconnector.policy.model.Policy;
 import org.eclipse.dataspaceconnector.spi.query.CriterionConverter;
 import org.eclipse.dataspaceconnector.spi.types.domain.asset.Asset;
@@ -26,7 +27,9 @@ import org.junit.jupiter.api.Test;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Predicate;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -35,12 +35,15 @@ in the detailed section referring to by linking pull requests or issues.
...
 * Extract single `PolicyArchive` implementation (#1158)
-* Replace `accessPolicy` and `contractPolicy` with `accessPolicyId` and `contractPolicyId` on `ContractDefinition` (#1144)
+* Replace `accessPolicy` and `contractPolicy` with `accessPolicyId` and `contractPolicyId` on `ContractDefinition` (
+  # 1144)
 * Replace `policy` with `policyId` on `ContractAgreement` (#1220)
 * All DMgmt Api methods now produce and consume `APPLICATION_JSON` (#1175)
 * Make data-plane public api controller asynchronous (#1228)
 * Provide In-mem implementations by default (#1130)
 * Changed Catalog config keys and switched from minutes to seconds
+* Usage of `NooTransactionContext` in (SQL-)Tests (#1119)
```
marcgs, line 45:
```suggestion
* Usage of `NoopTransactionContext` in (SQL-)Tests (#1119)
```

extensions/catalog/federated-catalog-cache/src/test/java/org/eclipse/dataspaceconnector/catalog/defaults/store/InMemoryFederatedCacheStoreTest.java
```diff
@@ -110,4 +113,23 @@ void queryCacheContainingTwoDistinctAssets_shouldReturnBothContractOffers() {
                 .anySatisfy(co -> assertThat(co.getAsset().getId()).isEqualTo(assetId1))
                 .anySatisfy(co -> assertThat(co.getAsset().getId()).isEqualTo(assetId2));
     }
+
+    @Test
+    void deleteAll() {
```
marcgs, line 118: @paullatzelsperger do you think it would be worth writing integration tests for the federated catalog logic? Probably in a separate story as there are none at the moment AFAIS.

marcgs (approved):

extensions/catalog/federated-catalog-cache/src/test/java/org/eclipse/dataspaceconnector/catalog/defaults/store/InMemoryFederatedCacheStoreTest.java
paullatzelsperger: @marcgs yes that is likely a good idea :)
however, I wanted to get this in quickly, and i-tests or e2e-tests for the catalog are no small feat.

bscholtes1A:
Hi @paullatzelsperger 
what happens if one node is not available in the current crawl run? (temporary unavailability)
If my understanding is correct, then it means we will lose the assets proposed by this node, thus we lose a bit of the "caching" aspect.  WDYT?

paullatzelsperger:
> Hi @paullatzelsperger what happens if one node is not available in the current crawl run? (temporary unavailability) If my understanding is correct, then it means we will lose the assets proposed by this node, thus we lose a bit of the "caching" aspect. WDYT?

@bscholtes1A If the request comes back unsuccessful, the loader would not neither clean nor update the cache, so the dirty items remain. even if we chose to clear the cache before receiving the response, those assets would "re-appear" on the next run. as we now do not have a standalone catalog runtime yet, the catalog being down would also mean the connector is down, so no negotiation or transfer would be possible anyway. 

TL;DR: I don't think it's a problem :)

extensions/catalog/federated-catalog-cache/src/test/java/org/eclipse/dataspaceconnector/catalog/defaults/store/InMemoryFederatedCacheStoreTest.java
ndr-brt: Created an issue about: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/1279

bscholtes1A (approved):

PR #1277, opened 2022-05-11
Bugfix: several small fixes in Cosmos' `WhereClause`

### LINKED ISSUE #1276: Bug: several problems in `WhereClause`
# Bug Report

## Describe the Bug
The implementation of `WhereClause` has several small problems:
- check for valid operators is case-sensitive (e.g. `IN` != `in`)
- chained `"WHERE"` clauses must use the `"AND"` keyword, even if there are no statement parameters

_Note:
Another issue will address the problematic specifcation of lists for the `IN` operator, which now must be `('val1', 'val2',...)`. We should use prepared statements for that._

paullatzelsperger:
## What this PR changes/adds

Fixes several smaller bugs in the `WhereClause` implementation

## Why it does that

So that where clauses get constructed correctly.

## Further notes

Another issue will address the problematic specifcation of lists for the IN operator, which now must be ('val1', 'val2',...). We should use prepared statements for that.

This issue will also contain the CHANGELOG entry then.

## Linked Issue(s)

Closes #1276 

## Checklist

### DIFF
--- a/extensions/azure/cosmos/cosmos-common/src/main/java/org/eclipse/dataspaceconnector/azure/cosmos/dialect/WhereClause.java
+++ b/extensions/azure/cosmos/cosmos-common/src/main/java/org/eclipse/dataspaceconnector/azure/cosmos/dialect/WhereClause.java
@@ -39,7 +39,7 @@ import static org.eclipse.dataspaceconnector.azure.cosmos.CosmosDocument.sanitiz
  */
 class WhereClause implements Clause {
     public static final String EQUALS_OPERATOR = "=";
-    public static final String IN_OPERATOR = "IN";
+    public static final String IN_OPERATOR = "in";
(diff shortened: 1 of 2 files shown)


ndr-brt (approved):
I'm ok with the fixes themselves, but I think this could be refactored in a different way, like extracting a `WhereCondition` class where to put the logic to create the condition as string and the list of parameter, then the `WhereClause` will just keep a list of `WhereCondition` and the `getWhere` would be only a joining of string representation of the conditions, and the `getParameters` a flatMap of the conditions parameters. Not to be done right now, just a thought.

paullatzelsperger:
@ndr-brt I agree - there is a similar thing for the SQL implementations, it's called `SqlConditionExpression`. I'm working on replicating the same functionality for Cosmos.

paullatzelsperger:
closing this because #1283 exists.

PR #1285, opened 2022-05-11
core: make all the services injectable

### LINKED ISSUE #1272: CosmosPolicyStoreExtension: Unusable because vault is not injectable
# Bug Report

## Describe the Bug

The `@Inject` and `@Requires` cannot be used together with the `Vault`. All but the CosmosPolicyStoreExtension resolve the vault by calling `context.getService(Vault.class);` without the usage of any annotations mechanism. As the CosmosPolicyStoreExtension does this different it became unusable.

### Expected Behavior

- `Vault` becomes injectable or
- CosmosPolicyStoreExtension should resolve the vault like all the other extensions

### Observed Behavior

Exception when running a connector with the `CosmosPolicyStoreExtension`

```bash
[2022-05-10 13:32:24] [SEVERE ] Error booting runtime
org.eclipse.dataspaceconnector.spi.EdcException: The following @Require'd features were not provided: [org.eclipse.dataspaceconnector.spi.security.Vault]
	at org.eclipse.dataspaceconnector.boot.system.DependencyGraph.of(DependencyGraph.java:109)
	at org.eclipse.dataspaceconnector.boot.system.ExtensionLoader.loadServiceExtensions(ExtensionLoader.java:142)
	at org.eclipse.dataspaceconnector.boot.system.runtime.BaseRuntime.createExtensions(BaseRuntime.java:155)
	at org.eclipse.dataspaceconnector.boot.system.runtime.BaseRuntime.boot(BaseRuntime.java:229)
	at org.eclipse.dataspaceconnector.boot.system.runtime.BaseRuntime.boot(BaseRuntime.java:89)
	at org.eclipse.dataspaceconnector.boot.system.runtime.BaseRuntime.main(BaseRuntime.java:77)

```

## Steps to Reproduce
1. Start connector with CosmosPolicyStoreExtension

## Context Information
This bug is currently blocking some CX Teams from deploying their connector. It would be great if we could get a quick fix on the main branch, for example by having the CosmosPolicyStoreExtension resolving the vault like all the other extensions.

## Detailed Description
-

## Possible Implementation
-

### LINKED ISSUE #1275: Make all the services injectable
# Feature Request

Current there are some "special" services as: 
- `Vault`
- `PrivateKeyResolver`
- `CertificateResolver`
- `Monitor`
- `TypeManager`
- `Telemetry`

that aren't injectable with the `@Inject` annotation. Would be good to have them injectable.

core

Will fix #1272 

## Solution Proposal
-

## Type of Issue
improvement

## Checklist

ndr-brt:
## What this PR changes/adds

Makes these services injectable by adding a `@Provider` method on `CoreServicesExtension`

- `Vault`
- `PrivateKeyResolver`
- `CertificateResolver`
- `Monitor`
- `TypeManager`
- `Telemetry`

## Why it does that

To make sure that those services can be injected through `@Inject`

## Further notes

- format some code on `DependencyGraph` to make the core more readable
- extract a `ServiceExtensionComparator` to being able to sort extensions by their type (`Base`, `Core` or generic) and make the code more readable
- improve the signature of the `ExtensionLoader.loadSingletonExtension` as it was used only to get the `VaultExtension` to better separate the responsabilities
- avoid the need to call both `initialize` and `initializeVault` on `VaultExtension`, renaming the latter to `initialize` and make calling the `initialize(Monitor)` by default.

## Linked Issue(s)

Closes #1275

## Checklist


paullatzelsperger (commented):
Making the services that you mention in the issue available through `@Provider`-methods and simply accessing them through the context seems sort of ... hacky, because it doesn't really instantiate those services.

How about this:
Since we now have dependency injection, I don't see the need for a separate `BootExtension` interface anymore. We could ax it, or at least we could let `BootExtension extends ServiceExtension`, so it gets picked up by the extension loader. Once that's done, implementors such as the `FsVaultExtension` could directly have a `@Provider` methods, which would further decouple the extensions. I'll add the link to a gist shortly.
WDYT?

ndr-brt:
@paullatzelsperger that was my first approach, but then I found out that, making `VaultExtension` extending `ServiceExtension` doesn't make it automatically be loaded by the service loader, because the manifest file in the resources is needed, that wouldn't be a big issue, just adding the two `...ServiceExtension` files, the problem is on the `NullVaultExtension` that's not a separate extension but it's in the `core-spi` module, and having an extension loaded on an `spi` module seems not that good.
Will investigate it further

ndr-brt:
What about getting rid of the `VaultExtension` interface and make their implementors implementing `ServiceExtension`?
I'll work on that, let's see how it will come out.

paullatzelsperger (approved):

PR #1308, opened 2022-05-16
docs: add logging guide

### LINKED ISSUE #1124: docs: add logging guide
# Feature Request
document logging rules

all modules

logging is missing at a lot of places or implemented inconsistently

## Solution Proposal
document rules like:
* when to use which log level
* when to log, e.g.
  * log every incoming call
  * log every action that changes the state
  * log every every exception
* what must never be part of the log file (e.g. secrets)
* maybe also patterns how messages should look like

## Type of Issue
cleanup

ndr-brt:
## What this PR changes/adds

Add a first version of the logging guide.

## Why it does that

To have a unique and agreed approach on how to log.

## Further notes

- this is a proposal, please add comments where things are missing or wrong
- after this PR will be merged, another one that adapt logs to the rules provided here should be done

## Linked Issue(s)

Closes #1124

## Checklist


paullatzelsperger (commented):
I would amend the following:
- exceptions should be thrown or logged, not both
- "clogging" is to be avoided: we shouldn't log stuff like "entering method X", "leaving block Y", "returning HTTP 200"

paullatzelsperger (approved):

ronjaquensel (approved):

docs/developer/logging.md
```diff
@@ -0,0 +1,40 @@
...
+Used to describe the normal flow of the application.
+ 
+### `debug` 
+> Fine-grained informational events that are most useful to debug an application.
+ 
+Used to describe details of the normal flow that are not interesting for a production environment.
+
+## What should be logged
+
+- every incoming call with `debug`
```
jimmarino, line 32: Is this every call into the runtime or every call into an internal service? I don't think the latter provides much value.

docs/developer/logging.md
```diff
@@ -0,0 +1,40 @@
...
+- every state change with `info`
```
jimmarino, line 33: Shouldn't state changes be debug? Otherwise, these messages will likely drown out other important (non-warning) information.

docs/developer/logging.md
ndr-brt: The idea was every call coming from an external service, e.g. an HTTP call

docs/developer/logging.md
ndr-brt: @jimmarino I can agree with this, but what have to be logged with `info` level?

docs/developer/logging.md
jimmarino: May say "incoming call at the API layer" or something like that?

docs/developer/logging.md
jimmarino: For info, I think any important message that is not an error. For example, an extension is installed and booted. Or, a maintenance process is run. 

IMO granular runtime operations in production are better handled by observability/telemetry.  For development, I generally never use logs, just a debugger.

jimmarino (approved):

MoritzKeppler (commented):
What about adding some notes on what kind of information of a caller can be included (e.g. are we allowed to log the connectorId?)

Suggest to move the definition of the loglevel to the `Monitor.java` file. Needs to be present during development. Also a link to this document.

docs/developer/logging.md
florianrusch-zf: > IMO granular runtime operations in production are better handled by observability/telemetry. For development, I generally never use logs, just a debugger

@jimmarino I agree with you that for us developers the debugger is the better tooling. But in our recent rollout of the connector to the first companies of catena-x, we identified that most of the people trying to setup the connector don't have deep insights in devops approaches like observability/telemetry/tracability and therefor weren't able to debug into their connectors which were deployed with k8s or docker-compose (mostly they just opened the port and we edc developer then debugged into the connector). They just looked at the logs and were confused when they don't see an error message (e.g. when they got a 400 for their request).

So I totally agree with @MoritzKeppler that we need to **log as much useful informations as possible**. Therefor maybe we can introduce also a new log level like `audit`, which then is just for state changes and creations, updates, deletion of domain entities (e.g. via management api).

docs/developer/logging.md
```diff
@@ -0,0 +1,39 @@
...
+ 
+### `debug` 
+> Fine-grained informational events that are most useful to debug an application.
+ 
+Used to describe details of the normal flow that are not interesting for a production environment.
+
+## What should be logged
+- every exception with `severe` or `warning`
+- every important message that's not an error with `info`
+- other informative events like incoming calls at the API layer or state changes with `debug`
```
DominikPinsel, line 33: Maybe we can also add a sentence how these result objects are handled. Currently they are also used to manage unforeseen issues and the result is then handled by a parent method. 
We should decide whether the result receiving or result creating component must log these things (or both).

ndr-brt:
Thank you all for the exhaustive reviews.

@MoritzKeppler 
> What about adding some notes on what kind of information of a caller can be included (e.g. are we allowed to log the connectorId?)
I think this is described by the "potentially sensitive data" in the `what should be not logged` section

> Suggest to move the definition of the loglevel to the `Monitor.java` file. Needs to be present during development. Also a link to this document.
The `LogLevel` wouldn't fit well in `Monitor` there as it is only used by `ConsoleMonitor`, probably every `Monitor` implementation would use its own `Level` class (like it's done in the `LoggerMonitor` implementation

@MoritzKeppler @jimmarino 
regarding the `info` logs probably I would prefer to have state changes logged so I put them as example in "what should be logged with info" section. I propose to get this merged and address further discussions about this somewhere in the future

docs/developer/logging.md
jimmarino: I don't think state changes should be logged as info. That will pollute the logs with useless information in production deployments.

+1 for auditing, but that should be done via architected events. @ndr-brt is working on an eventing subsystem that can be used for that.

@florianrusch-zf The issue you identify appears to be twofold:

1. Developers should know how to use a debugger. Relying on logs as a substitute isn't a maintainable or effective solution.
2. The API should send back reasonable error information, preferrably formatted as JSON. It looks like that did not happen and would be worth filing a separate issue.

MoritzKeppler: There are great tools to filter and aggregate logs, so I wouldn't fear the amount of logs lines. 
Analysis later in time cannot be done by using a debugger. And the connector will not only be run by developers.
Log pollution will be caused by inconsistent or chatty loglines, this is something we need to avoid.

jimmarino: @MoritzKeppler introducing a filtering tool IMO just papers over the problem.  I don't think it makes sense to log debug information under "INFO". If we log in under DEBUG, people can always choose to enable that level and use filtering tools to sift through logs.

jimmarino:
> Thank you all for the exhaustive reviews.
> 
> @MoritzKeppler
> 
> > What about adding some notes on what kind of information of a caller can be included (e.g. are we allowed to log the connectorId?)
> > I think this is described by the "potentially sensitive data" in the `what should be not logged` section
> 
> > Suggest to move the definition of the loglevel to the `Monitor.java` file. Needs to be present during development. Also a link to this document.
> > The `LogLevel` wouldn't fit well in `Monitor` there as it is only used by `ConsoleMonitor`, probably every `Monitor` implementation would use its own `Level` class (like it's done in the `LoggerMonitor` implementation
> 
> @MoritzKeppler @jimmarino regarding the `info` logs probably I would prefer to have state changes logged so I put them as example in "what should be logged with info" section. I propose to get this merged and address further discussions somewhere in the future

I definitely think state changes should be logged but at debug, not info level.

docs/developer/logging.md
```diff
@@ -0,0 +1,41 @@
...
+Used to describe the normal flow of the application.
+ 
+### `debug` 
+> Fine-grained informational events that are most useful to debug an application.
+ 
+Used to describe details of the normal flow that are not interesting for a production environment.
+
+## What should be logged
+- every exception with `severe` or `warning`
+- every `Result` object evaluated as `failed` with `severe` if this is something that someone should take care of, `warning` otherwise
```
ndr-brt, line 33: That's not correct, since there are failed `Result` that are not so interesting for the logs, e.g. an entity not found by the api that would end up returning a 404.
```suggestion
- every `Result` object evaluated as `failed`: 
  - with `severe` if this is something that blocks the flow and someone should take care of immediately
  - with `warning` if this is something that doesn't block the flow and someone should take care of (because it could give worse results in the future)
```

docs/developer/logging.md
```diff
@@ -0,0 +1,43 @@
...
+### `debug` 
+> Fine-grained informational events that are most useful to debug an application.
+ 
+Used to describe details of the normal flow that are not interesting for a production environment.
+
+## What should be logged
+- every exception with `severe` or `warning`
+- every `Result` object evaluated as `failed`: 
+  - with `severe` if this is something that interrupts the flow and someone should take care of immediately
+  - with `warning` if this is something that doesn't interrupts the flow but someone should take care of, because it could give worse results in the future
```
ndr-brt, line 35:
```suggestion
- every `Result` object evaluated as `failed`: 
  - with `severe` if this is something that interrupts the flow and someone should take care of immediately
  - with `warning` if this is something that doesn't interrupt the flow but someone should take care of, because it could give worse results in the future
```

ndr-brt:
@MoritzKeppler seems like the upcoming event framework will be accomplish the needs for audit and keep on track with entity transitions, so seems like logging state changes with `debug` is the most sensible decision, as this will be interesting for some developers that are trying to debug the connector trying to fixing bugs on a test environment, but not so interesting for ops people in a production environment that needs to get from logs useful information about the connector state (blocking exceptions, warnings and other cruise info)

ndr-brt:
@MoritzKeppler seems like the upcoming event framework will be accomplish the needs for audit and keep on track with entity transitions, so seems like logging state changes with `debug` is the most sensible decision, as this will be interesting for some developers that are trying to debug the connector trying to fixing bugs on a test environment, but not so interesting for ops people in a production environment that needs to get from logs useful information about the connector state (blocking exceptions, warnings and other cruise info)

PR #1318, opened 2022-05-17
bugfix: move provide service extension phase before initialization

### LINKED ISSUE #1317: Warning about missing service registrations
# Feature Request

When a connector is starting, e.g. the `sample-01` one, it prints this warning:
```
WARNING 2022-05-17T11:01:16.883365294 There were missing service registrations in extension class org.eclipse.dataspaceconnector.core.CoreServicesExtension: org.eclipse.dataspaceconnector.spi.system.Hostname, okhttp3.OkHttpClient, org.eclipse.dataspaceconnector.spi.agent.ParticipantAgentService, org.eclipse.dataspaceconnector.spi.policy.PolicyEngine, org.eclipse.dataspaceconnector.spi.message.RemoteMessageDispatcherRegistry, org.eclipse.dataspaceconnector.spi.policy.RuleBindingRegistry
```

This should not be done as those services are provided by the `@Provider` annotation.

startup

To avoid misleading logs.

## Solution Proposal
Maybe the simplest thing could be to anticipate the `RegistrationPhase` before the `InitializePhase`

## Type of Issue
cleanup

## Checklist

ndr-brt:
## What this PR changes/adds

Moves the `provide` phase of services before the initialize phase

## Why it does that

Otherwise the validation of the `ServiceExtensionContext` by the `InjectionContainer` fails and log a warning message as pointed out by #1317 

## Further notes

- I had to remove all the `Phase` concept because it was too convoluted and it was pretty much impossible to switch the order of the phases without breaking down everything. I moved all the knowledge about phases inside the `ExtensionLifecycleManager` component since it was not too much to be handled by one single component, and moved all the tests to its test class.

## Linked Issue(s)

Closes #1317 

## Checklist


paullatzelsperger (commented):
There is two things to mention:
- The validation can only ever validate, that whatever was declared in `@Provides`, was indeed registered in the `initialize()` method. Now, the problem here is that there are some dependencies declared in `@Provides`, but in fact they are provided through a factory method (= `@Provider`), which triggered the validation warning. Simply removing the entries in `@Provides` should do the trick.
- Phases: they were introduced as a safeguard against accidentally (or erroneously) changing the order of inject-initialize-provide, so they are rigid _by design_. @ndr-brt points out that this could also be achieved through a test, which is true. So replacing the `Phase` concept with a very particular unit test would be a viable alternative. Doing this is another PR would probably be a good idea.

All in all, it seems like the only thing that's necessary to avoid the warning message is to remove the redundant (and thus offending) entries from `@Provides`.

ndr-brt:
superseded by #1336

Issue #1344, opened 2022-05-23
Bug: `CosmosTranferProcessStore` attempts to transition to `INITIAL`


paullatzelsperger:
# Bug Report

## Describe the Bug
The `CosmosTransferProcessStore` attempts to transition the TP to `INITIAL` upon creation, which is already done by the `TransferProcessManagerImpl`.

### Expected Behavior
The store does not transition anything.

Remove `CosmosTransferProcessStore#L116`

paullatzelsperger:
Will be fixed by #1346

PR #1346, opened 2022-05-23
system-test: add cosmosdb e2e test

ndr-brt:
## What this PR changes/adds

Add the CosmosDB version of the `end-to-end-transfer-test`

## Why it does that

To give an end to end coverage using cosmos db as persistence store.

## Further notes

- Introduce `CosmosClientProvider` to permit end-to-end test to provide the test connector
- Remove an unnecessary `transitionInitial` on the `CosmosTransferProcessStore.save` method 
- Verified locally all the `Cosmos*IntegrationTest` as they won't be verified until this will be merged.

## Linked Issue(s)

Closes: #1344 

## Checklist


extensions/azure/cosmos/cosmos-common/src/main/java/org/eclipse/dataspaceconnector/azure/cosmos/CosmosClientProvider.java
```diff
@@ -0,0 +1,22 @@
...
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.azure.cosmos;
+
+import com.azure.cosmos.CosmosClient;
+import org.eclipse.dataspaceconnector.spi.security.Vault;
+
+public interface CosmosClientProvider {
```
paullatzelsperger, line 20: pls document

system-tests/e2e-transfer-test/control-plane-cosmosdb/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
...
+
+plugins {
+    `java-library`
+}
+
+dependencies {
+    implementation(project(":system-tests:e2e-transfer-test:control-plane"))
+    implementation(project(":extensions:azure:cosmos"))
+    implementation(testFixtures(project(":extensions:azure:azure-test")))
+    implementation("com.azure:azure-cosmos:${cosmosSdkVersion}")
```
paullatzelsperger, line 25: could we make the `cosmos-common` extension expose `"com.azure:azure-cosmos` as `api` configuration? then we'd not have to state it here again

extensions/azure/cosmos/cosmos-common/src/main/java/org/eclipse/dataspaceconnector/azure/cosmos/CosmosDbApiImpl.java
```diff
@@ -31,26 +29,21 @@
...
 
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Scanner;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class CosmosDbApiImpl implements CosmosDbApi {
```
paullatzelsperger, line 45: Wouldn't it also have been possible to create a subclass for the `CosmosDbApiImpl`, e.g. named `TestCosmosDbApiImpl` that simply overrides the `createClient` method (which doesn't need to be static anyways) and uses the `CosmosTestClient`? That way we would be able to contain the `CosmosClient` (which is a third-party class) and not expose it.

system-tests/e2e-transfer-test/control-plane-cosmosdb/build.gradle.kts
ndr-brt: this was a spiking leftover, not needed, removed

extensions/azure/cosmos/cosmos-common/src/main/java/org/eclipse/dataspaceconnector/azure/cosmos/CosmosDbApiImpl.java
ndr-brt: To me creating a client is a thing that goes over the 'CosmosDbApi' component responsibilities. 
`CosmosClient` is a third party class indeed, but a `CosmosDbApi` class could not exist without it, so exposing it is not a big issue.

paullatzelsperger: Well, in this case it would have saved the creation of another extension, plus another "factory" interface + impl.

ndr-brt: the extension model is a base concept of the EDC, I don't see nothing wrong in using it, in this case it will permit users to customize the `CosmosClient` at extension level, like, for example, to accomplish specific needs like using a different authentication type or adding more than one preferred regions.

Making the `Api` instantiate the client, other than the responsibility issue I mentioned in the previous comment,  would need to expose a `protected getClient` method that would be needed only for testing purposes, that's an habit to be avoided if possible.

paullatzelsperger: The missing piece in my mind was the fact that when using the `CosmosClient` for testing with the CosmosDB emulator, it requires some special setup (downloading and storing a certificate), in connection with the fact that that we're dealing with and e2e test, which uses EDC's dependency injection capabilities. That means, that the "local" client must be supplied through DI.

Just one caveat: the CosmosDB emulator is not suitable for usage with CI/CD pipelines, it simply is not stable enough. Thus, E2E tests should run against a real cosmos instance, or not run at all, if no `COSMOS_KEY` is present. Same way we do for the other cosmos integration tests.

extensions/azure/cosmos/cosmos-common/src/main/java/org/eclipse/dataspaceconnector/azure/cosmos/CosmosClientProvider.java
```diff
@@ -17,6 +17,16 @@
 import com.azure.cosmos.CosmosClient;
 import org.eclipse.dataspaceconnector.spi.security.Vault;
 
+/**
+ * Provides a `CosmosClient` instance
```
paullatzelsperger, line 21: minor nit: 

```suggestion
 * Provides a `CosmosClient` instance. 
 * This is mainly intended for usage in test scenarios, where a {@code CosmosClient} is used, that requires special setup (for example with the emulator).
```

paullatzelsperger (approved):
LGTM

PR #1363, opened 2022-05-27
helm: Make Helm charts more generic

bscholtes1A:
## What this PR changes/adds

This PR brings some improvement to the existing Helm charts, which was tainted for only being used to deploy the Control Plane of the EDC. 

In order to avoid code duplication for deploying other components of the EDC (data plane, data plane selector, federated catalog...), we make the existing Helm charts more generic. Especially,:

- We remove the hardcoded environment variables within the Deployment objects, as this is specific to the EDC component which is being deployed (for example, Data Plane does not have an IDS port, so existing charts could not be used to deploy the DPF). 
- Ports numbering is now up to the client which is using the helm charts, and is no longer hardcoded in the helm chart (as this is something that is defined at image build-time).
- Unlimited number of ports can now be registered (http, http-ids, http-mgnt was hardcoded before).
- We offer the possibility to mount file in volume within the POD (this unlock usage of some filesystem extensions.
- Ingress can now expose more than one port, which was not possible with the current implementation. 

## Why it does that

Avoid duplication of Helm charts for the other EDC components.

## Checklist

### DIFF
--- a/resources/charts/dataspace-connector/values.yaml
+++ b/resources/charts/dataspace-connector/values.yaml
@@ -27,25 +27,36 @@ securityContext: {}
   # runAsNonRoot: true
   # runAsUser: 1000
 
+ports: {}
+  # Example:
+  #  http:
+  #    containerPort: 8181
+  #    servicePort: 80 (Optional: if not provided, the value of containerPort will be used)
+  #    path: /api
+
+volumes: []
+  # Define a volume that can then be mounted into a container. Please refer to https://kubernetes.io/docs/concepts/storage/volumes/
+  # for more details. Below example will create a volume from the content of the `log-config` ConfigMap keyed with `log_level`:
+  # - name: config-vol
+  #   configMap:
+  #     name: log-config
+  #     items:
+  #     - key: log_level
+  #       path: log_level
+
 edc:
-  port: 8181
-  path: /api
-  managementApi:
-    enabled: true
-    port: 9191
-    path: /api/v1/data
-  idsApi:
-    port: 8282
-    path: /api/v1/ids
   env: {}
   envSecrets: {}
+  volumeMounts: []
+  # Mount the provided volume in the POD filesystem at the specific path. Please refer to
+  # https://kubernetes.io/fr/docs/concepts/storage/volumes/ for more details.
+  # Example:
+  # - name: config-vol
+  #   mountPath: /etc/config
+  # will mount the `config-vol` volume at /etc/config
+
 service:
   type: ClusterIP
-  port: 80
-  managementApi:
-    port: 91
-  idsApi:
-    port: 8282
 
 ingress:
   enabled: false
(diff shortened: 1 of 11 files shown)


resources/charts/dataspace-connector/values.yaml
```diff
@@ -27,25 +27,17 @@ securityContext: {}
   # runAsNonRoot: true
   # runAsUser: 1000
 
+ports: {}
```
florianrusch-zf, line 30: It would be cool to have here an example how the configuration could be look like

florianrusch-zf: Maybe same for volumes & volumeMounts

system-tests/minikube/values-controlplane.yaml
```diff
@@ -0,0 +1,34 @@
...
+    WEB_HTTP_DATA_PATH: "{{ .Values.ports.apimgnt.path }}"
+    WEB_HTTP_IDS_PORT: "{{ .Values.ports.ids.containerPort }}"
+    WEB_HTTP_IDS_PATH: "{{ .Values.ports.ids.path }}"
+
+ports:
+  http:
+    containerPort: 8181
+    servicePort: 80
+    path: /api
+    type: NodePort
```
florianrusch-zf, line 15: Why do you output the ports as "NodePort" even though you have configured the ingress?

florianrusch-zf (changes requested):

resources/charts/dataspace-connector/values.yaml
florianrusch-zf: @bscholtes1A You did it for volumes and VolumeMounts. Did you explicitly omit the ports or was that a mistake?

resources/charts/dataspace-connector/values.yaml
```diff
@@ -27,25 +27,29 @@ securityContext: {}
...
-  idsApi:
-    port: 8282
-    path: /api/v1/ids
   env: {}
   envSecrets: {}
+  volumeMounts: []
+  # Mount the provided volume in the POD filesystem at the specific path, for example:
+  # - name: config-vol
+  #   mountPath: /etc/config
+  # will mount the `config-vol` volume at /etc/config
```
florianrusch-zf, line 49: Why do we need to have `env`, `envSecrets`, `volumeMounts` under `edc`? Could be on the root level or not?

florianrusch-zf (changes requested):

resources/charts/dataspace-connector/values.yaml
bscholtes1A: > @bscholtes1A You did it for volumes and VolumeMounts. Did you explicitly omit the ports or was that a mistake?

My bad, I missed this one. 
Corrected!

resources/charts/dataspace-connector/values.yaml
bscholtes1A: We could, but for now all POD settings are placed under `edc:`. `volumeMounts` refers to the way a volume is mounted into the container so for me this hierarchy makes sense.

PR #1372, opened 2022-05-30
Add validation for contract definition id

### LINKED ISSUE #1347: Add validation to avoid `ContractDefinition` id to contain `:`
# Feature Request

Creating a `ContractDefinition` with a `:` contained in the string would cause this error:
```
ContractRejectionHandler: Received contract rejection to message null. Negotiation process: 67abd883-b5b5-4e57-bf71-f5844f012a29. Rejection Reason: "Invalid id: urn:1edcc7c0-4796-46bc-bafd-87bd440cc68a:8d9e7ac4-770c-431c-9e2f-2e575c870e9a"
```

Would be good to have a validation on the `contractdefinition` api, to avoid this possibility.
Another way could be to have the "contract id parser" being able to permit the use of `:` inside the contract definition id, but seems to be more convoluted.

data management api

To avoid exceptions

## Solution Proposal
-

## Type of Issue
improvement

## Checklist

tuncaytunc-zf:
## What this PR changes/adds

Add validation for contract definition id

## Why it does that

Validation of contract definition id by creation solves the problem in negotiation process when the id is not valid.

## Further notes

Marking jersey-bean-validation helps for the correct validation of parameters.

## Linked Issue(s)

Closes #1347

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ in the detailed section referring to by linking pull requests or issues.
 * Postgresql end to end test (#1278)
 * Add signing/publishing config (#1147)
 * Verify OpenAPI definitions (#1312)
+* Add validation to contract definition id (#1347)
 
 #### Changed
 
--- a/common/util/build.gradle.kts
+++ b/common/util/build.gradle.kts
@@ -21,8 +21,11 @@ plugins {
 val jupiterVersion: String by project
 val mockitoVersion: String by project
 val okHttpVersion: String by project
+val jakartaValidationApi: String by project
 
 dependencies {
+    implementation("jakarta.validation:jakarta.validation-api:${jakartaValidationApi}")
+
     testFixturesImplementation("org.mockito:mockito-core:${mockitoVersion}")
     testFixturesImplementation("org.junit.jupiter:junit-jupiter-api:${jupiterVersion}")
     testFixturesRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:${jupiterVersion}")
--- a/common/util/src/main/java/org/eclipse/dataspaceconnector/common/annotations/Uuid.java
+++ b/common/util/src/main/java/org/eclipse/dataspaceconnector/common/annotations/Uuid.java
@@ -1,0 +1,35 @@
+/*
+ *  Copyright (c) 2022 ZF Friedrichshafen AG
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       ZF Friedrichshafen AG - Initial Implementation
+ *
+ */
+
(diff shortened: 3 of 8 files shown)


paullatzelsperger (changes requested):
We cannot and will not enforce the usage of the UUID format for `ContractDefinition`s.

tuncaytunc-zf:
Hi @paullatzelsperger the idea to validate the contractDefinitionId against the UUID was coming from @jimmarino please see the comments in issue #1347, maybe he has a reason why to use UUID. Otherwise I can revert the PR and check only if the ID contains any ":".

paullatzelsperger:
> Hi @paullatzelsperger the idea to validate the contractDefinitionId against the UUID was coming from @jimmarino please see the comments in issue https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/1347, maybe he has a reason why to use UUID. Otherwise I can revert the PR and check only if the ID contains any ":".

@jimmarino never said to validate against a UUID, or a GUID.

tuncaytunc-zf:
> > Hi @paullatzelsperger the idea to validate the contractDefinitionId against the UUID was coming from @jimmarino please see the comments in issue #1347, maybe he has a reason why to use UUID. Otherwise I can revert the PR and check only if the ID contains any ":".
> 
> @jimmarino never said to validate against a UUID, or a GUID. Using a UUID is what we recommend, to ensure a high level of uniqueness, but APIs should never be opinionated like that unless there is a technical reason.

My understanding from his comment _"The id should be a GUID (not URN) so +1 on adding validation in the API. ..."_ was that the id should be a GUID and should be validated in the API :-) But as I said if there are no special reason to use UUID of course I will remove the UUID validation and only check if id contains a ":"

florianrusch-zf:
> But as I said if there are no special reason to use UUID of course I will remove the UUID validation and only check if id contains a ":"

@paullatzelsperger @jimmarino @tuncaytunc-zf Do we have to look also for other chars which should't be part of the ID?

paullatzelsperger:
> > But as I said if there are no special reason to use UUID of course I will remove the UUID validation and only check if id contains a ":"
> 
> @paullatzelsperger @jimmarino @tuncaytunc-zf Do we have to look also for other chars which should't be part of the ID?

none that I can think of, from a technical perspective.

jimmarino:
> > > Hi @paullatzelsperger the idea to validate the contractDefinitionId against the UUID was coming from @jimmarino please see the comments in issue #1347, maybe he has a reason why to use UUID. Otherwise I can revert the PR and check only if the ID contains any ":".
> > 
> > 
> > @jimmarino never said to validate against a UUID, or a GUID. Using a UUID is what we recommend, to ensure a high level of uniqueness, but APIs should never be opinionated like that unless there is a technical reason.
> 
> My understanding from his comment _"The id should be a GUID (not URN) so +1 on adding validation in the API. ..."_ was that the id should be a GUID and should be validated in the API :-) But as I said if there are no special reason to use UUID of course I will remove the UUID validation and only check if id contains a ":"

I think this PR should be about 1-3 lines of code. We don't need to drag in validation dependencies into the core.

tuncaytunc-zf:
Will close this PR and implement just to check if the id contains any ":".

PR #1383, opened 2022-06-01
Bugfix: right-hand operand may contain equals signs

### LINKED ISSUE #1380: Filter asset not working properly for certain values
# Bug Report

## Describe the Bug
If the value provided to filter an asset has "=" inside, EDC splits it based on the equals sign and takes only the first half.

When trying to filter assets based on their properties, the following occurs:
- let's consider having the folllowing Asset:

```
[
    {
        "properties": {
            "additionalProp1": "a/b=c/d",
        }
    }
]
```

- when filtering by `additionalProp1` having the value  `a/b=c/d` using the request` GET assets?filter=additionalProp1=a/b=c/d` EDC takes as an input only `a/b` ignoring everything that comes after the sign "="

### Expected Behavior
EDC should take as an input the entire value and filter properly based on it.

### Observed Behavior
The call described above returns `[]` and 200 status code.
I've tried encoding the equals sign as well as the entire input, but nothing worked.

## Steps to Reproduce
1. Create an asset and populate a property with a value containing "="
2. Filter the assets based on the property and using the value having "=" inside it

## Context Information
- OS: [Windows]

## Logs
DEBUG 2022-05-31T13:41:10.1428741 get all Assets from QuerySpec{offset=0, pageSize=50, filterExpression=[asset:prop:p1 = a/b], sortOrder=null, sortField=null}

paullatzelsperger:
## What this PR changes/adds

This PR adds the `limit` parameter to the `String#split` method, causing the right-hand operand to be used "as-is", and not chop off parts after the "=" sign.

## Why it does that

When building `QuerySpec`s out of a filter string, there was a bug when the right-hand operand contained a `"="` sign: the portion right of it would get chopped off because `String#split` would tokenize into too many elements.

## Further notes

The Regex to detect EQUALS expressions and the generic parse step (the `else` path) was adapted, so that the right-hand operand may also contain spaces.

**Please comment in case this is not the desired behaviour**

## Linked Issue(s)

Closes #1380 

## Checklist

### DIFF
--- a/spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/QuerySpec.java
+++ b/spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/QuerySpec.java
@@ -67,7 +67,7 @@ public class QuerySpec {
     }
 
     public static final class Builder {
-        private static final String EQUALS_EXPRESSION_PATTERN = "[^\\s\\\\]*(\\s*)=(\\s*)[^\\s\\\\]*";
+        private static final String EQUALS_EXPRESSION_PATTERN = "[^\\s\\\\]*(\\s*)=(\\s*)[^\\\\]*";
         private final QuerySpec querySpec;
         private boolean equalsAsContains = false;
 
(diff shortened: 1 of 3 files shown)


spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/QuerySpec.java
```diff
@@ -67,7 +67,7 @@ public SortOrder getSortOrder() {
     }
 
     public static final class Builder {
-        private static final String EQUALS_EXPRESSION_PATTERN = "[^\\s\\\\]*(\\s*)=(\\s*)[^\\s\\\\]*";
+        private static final String EQUALS_EXPRESSION_PATTERN = "[^\\s\\\\]*(\\s*)=(\\s*)[^\\\\]*";
```
ndr-brt, line 70: this regex matches the same expressions that a `contains("=")` will do, I will suggest to avoid the regex then

paullatzelsperger: not quite. "some key=some value" (left hand contains whitespace) would match `contains("=")` but not the regex

ndr-brt (approved):

PR #1407, opened 2022-06-02
Feature: add validation for contract definition id #1372

### LINKED ISSUE #1347: Add validation to avoid `ContractDefinition` id to contain `:`

tuncaytunc-zf:
## What this PR changes/adds

Add validation for contract definition id

## Why it does that

Validation of contract definition id by creation solves the problem in negotiation process when the id is not valid.

## Further notes

Marking jersey-bean-validation helps for the correct validation of parameters.

## Linked Issue(s)

Closes #1347 

## Checklist

### DIFF
--- a/extensions/api/api-core/build.gradle.kts
+++ b/extensions/api/api-core/build.gradle.kts
@@ -28,12 +28,12 @@ dependencies {
     implementation(project(":common:util"))
     implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
     implementation("jakarta.validation:jakarta.validation-api:${jakartaValidationApi}")
+    implementation("org.glassfish.jersey.ext:jersey-bean-validation:${jerseyVersion}") //for validation
 
     testImplementation("org.glassfish.jersey.core:jersey-common:${jerseyVersion}")
     testImplementation("org.glassfish.jersey.core:jersey-server:${jerseyVersion}")
 
     testImplementation(testFixtures(project(":launchers:junit")))
-    testRuntimeOnly("org.glassfish.jersey.ext:jersey-bean-validation:${jerseyVersion}") //for validation
 }
 
 publishing {
--- a/extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/model/TransferRequestDto.java
+++ b/extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/model/TransferRequestDto.java
@@ -29,7 +29,6 @@ public class TransferRequestDto {
 
     @NotNull
     private String connectorAddress;
-    @NotNull
     private String id;
     @NotNull
     private String contractId;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ in the detailed section referring to by linking pull requests or issues.
 * Add signing/publishing config (#1147)
 * Verify OpenAPI definitions (#1312)
 * Documentation for CosmosDB (#1334)
+* Add validation to contract definition id (#1347)
 
 #### Changed
 
--- a/extensions/api/data-management/asset/build.gradle.kts
+++ b/extensions/api/data-management/asset/build.gradle.kts
@@ -38,7 +38,6 @@ dependencies {
     testImplementation(testFixtures(project(":common:util")))
     testImplementation(testFixtures(project(":launchers:junit")))
     testImplementation("io.rest-assured:rest-assured:${restAssured}")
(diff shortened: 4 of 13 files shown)


florianrusch-zf:
If we now would specify an incorrect ID, what would the error message look like?

tuncaytunc-zf:
> If we now would specify an incorrect ID, what would the error message look like?

You would get "400 Bad Request"

florianrusch-zf:
> You would get "400 Bad Request" like other validation fails.

Without any kind of error message?

florianrusch-zf:
> You would get "400 Bad Request" like other validation fails.

Without any kind of error message?

tuncaytunc-zf:
> > You would get "400 Bad Request" like other validation fails.
> 
> Without any kind of error message?

Unfortunately not like all other fields but as I can remember there were some discussions ongoing to improve the error handling or logging, maybe it exists already an issue for it. I think to improve the error handling was not part of this issue.

florianrusch-zf:
@ndr-brt what do you think about this? In my opinion we should always consider to improve the error handling - also if we just fix a bug.

@tuncaytunc-zf we have since a few days the [Logging Guide](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/docs/developer/logging.md). But I think that this special case is currently not be considered within the logging guide.

In addition we also discussed this kind of error handling here: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/1332

extensions/api/api-core/build.gradle.kts
```diff
@@ -28,12 +28,12 @@ dependencies {
     implementation(project(":common:util"))
     implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
     implementation("jakarta.validation:jakarta.validation-api:${jakartaValidationApi}")
+    implementation("org.glassfish.jersey.ext:jersey-bean-validation:${jerseyVersion}") //for validation
```
ndr-brt, line 31: This could be removed from all the extension where it's defined as `testRuntimeOnly`

ndr-brt (commented):
I think this goes beyond the logging guide, as this information (failed input validation) is interesting for the client, but not for the server.
We should add some details in the response body when validation fails to make the client aware of what's needed to fix. 

Talking about the PR, this detail (the interdiction of `:` in contract definition id) should be described in the openapi specification

ndr-brt (approved):

extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/model/TransferRequestDto.java
```diff
@@ -29,7 +29,6 @@ public class TransferRequestDto {
 
     @NotNull
     private String connectorAddress;
-    @NotNull
     private String id;
```
paullatzelsperger, line 32: Why was this removed? Transfer requests should/must always have an ID

paullatzelsperger (changes requested):

extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/model/TransferRequestDto.java
ndr-brt: @paullatzelsperger I think this was a typo, as every system-test doesn't pass an id in the request as the EDC will generate one by itself in the transformer:

https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/60da937f68c72f471419c2c779ac1d499bc0517b/extensions/api/data-management/transferprocess/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/transferprocess/transform/TransferRequestDtoToDataRequestTransformer.java#L46-L47

paullatzelsperger (approved):

PR #1413, opened 2022-06-03
docs: add event framework decision record

### LINKED ISSUE #1412: Event Framework Decision Record
# Feature Request

We need to carve a decision record about Event Framework in EDC.

docs

-

## Solution Proposal
-

## Type of Issue
documentation

## Checklist

ndr-brt:
## What this PR changes/adds

Adds a decision records about Event Framework

## Why it does that

To define some basic concepts for this feature

## Further notes

-

## Linked Issue(s)

Closes #1412

## Checklist

### DIFF
--- a/docs/developer/decision-records/2022-06-03-event-framework/README.md
+++ b/docs/developer/decision-records/2022-06-03-event-framework/README.md
@@ -92,3 +92,7 @@ public class EventRouterImpl implements EventRouter {
 ```
 
 At this point, at the extension level, a `EventListener` can be registered to the `EventRouter` and it will be able to get all the EDC events, filter them, publish them to some external event infrastructure.
+
+Note that this implementation does not respect the "at-least-one" rule, as something can happen between event creation and its dispatch. 
+To achieve such a behavior events should be persisted. With this implementation we would have events that will be delivered "at most once", they
+should not be used to trigger important business logic but only to give insights on what's happening in the EDC.
--- a/docs/developer/decision-records/2022-06-03-event-framework/README.md
+++ b/docs/developer/decision-records/2022-06-03-event-framework/README.md
@@ -22,9 +22,10 @@ An event is a representation of something that happened in the past, and has ver
 - it describes which layer emitted the event: domain, system, security, and so on...
 
 #### Main differences between Events and Callbacks
-- Callbacks are always correlated with a request, events don't
+- Callbacks are always correlated with a request, events are not
 - Callbacks are unicast, events are multicast
 - Callbacks are at the protocol level, events can be at a different level
+- Callbacks are critical for the correct operation of a connector, where events are optional
 
 ## Approach
 
@@ -49,9 +50,24 @@ public interface TransferProcessEventListener {
 }
 ```
 
-The `Listener` would have a single implementation that will take care of generate the event class and pass it to the `EventRouter` component.
-`Event` will be the abstract class that represents all the events, and every event class will extend it.
+`Event` will be the superclass that represents all the events, and every event class will extend it.
+```java
+public abstract class Event {
+    private final long createdAt;
 
+    ...
+}
+```
+
+Event example implementation:
+```java
+public class TransferProcessCreatedEvent extends Event {
+    private String id;
+    ...
+}
+```
+
+The `Listener` would have a single implementation that will take care of generate the event class and pass it to the `EventRouter` component.
 
 ```java
 
--- a/docs/developer/decision-records/2022-06-03-event-framework/README.md
+++ b/docs/developer/decision-records/2022-06-03-event-framework/README.md
@@ -1,0 +1,94 @@
+# Event framework
+
+## Decision
+
+This decision record covers concepts and semantics about asynchronous communication in the EDC.
+This is needed to define what's an `Event` and how this concept should be introduced into EDC.
+
+## Rationale
+
+There are two asynchronous communication levels in EDC:
+
+#### Protocol level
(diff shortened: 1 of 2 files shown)


docs/developer/decision-records/2022-06-03-event-framework/README.md
```diff
@@ -0,0 +1,94 @@
...
+public interface EventSubscriber {
+    void on(Event event);
+}
+
+public class EventRouterImpl implements EventRouter {
+
+    private List<EventSubscriber> subscribers;
+
+    public void publish(Event event) {
+        subscribers.forEach(subscriber -> {
```
jimmarino, line 87: Do you think this should be an async future? That will mean events may be lost. We could leave the dispatch choice to the listener implementation, which would provide more flexibility.

ndr-brt: @jimmarino I think the event framework shouldn't block the normal flow of the EDC, since the event is generated it should be sent to the subscribers without having one blocking the others. 
The event could be lost also if handled in a sync manner

jimmarino: With the sync behavior, we could have listeners abort an operation but that may not be desirable. 

I do like the idea of not blocking the normal flow of execution. I think we should document that a side effect of this design is that there is no guarantee events will be delivered (e.g. at least once).

Also. should we use a thread pool for the dispatch? That may be getting too much into implementation detail, but the code sample brought the issue up.

ndr-brt: I documented this detail in the DR, I think we'd like to reach "at least once" adding persistence in the future, but for the moment this seems a bit out of the scope (and would complicate the feature too much in my opinion).

I think the "thread pool" topic is absolutely an implementation detail, since it's only a matter on how the `runAsync` is called. Currently we have different thread pools spawned in the codebase, I guess we will do pretty much the same (making the pool size configurable)

bscholtes1A: Is there any plan to implement a DLQ in case a event dispatch failed?

bscholtes1A:
Looks good to me. 
Thanks @ndr-brt  for the nice write-up!

bscholtes1A (approved):

docs/developer/decision-records/2022-06-03-event-framework/README.md
```diff
@@ -0,0 +1,98 @@
...
+#### Instance level
+The EDC should have the possibility to communicate with third party systems (known as subscribers) notifying them about what's happening in the connector. This communication needs to be made through **events**.
+An event is a representation of something that happened in the past, and has very specific characteristics, like:
+- it's named with a short sentence in the past tense (e.g. "contract negotiation confirmed")
+- it's serializable
+- it contains the issue timestamp
+- it describes which layer emitted the event: domain, system, security, and so on...
+
+#### Main differences between Events and Callbacks
+- Callbacks are always correlated with a request, events don't
```
paullatzelsperger, line 25:
```suggestion
- Callbacks are always correlated with a request, events are not
```

docs/developer/decision-records/2022-06-03-event-framework/README.md
```diff
@@ -0,0 +1,98 @@
...
+- Callbacks are unicast, events are multicast
+- Callbacks are at the protocol level, events can be at a different level
+
```
paullatzelsperger, line 28: I would also add that:
"Callbacks are critical for the correct operation of a connector, where events are optional"

paullatzelsperger (commented):
Do I understand correctly, that the `EventRouter` receives events from all `*listenerImpl`s? It would therefore act as the central event "hub" (even if that's a somewhat worn-out term).

If so, it would be beneficial to include 
- a clear statement that all events in EDC have a common base class/interface
- a code snippet outlining that `Event` class (even if the final impl looks slightly different)

docs/developer/decision-records/2022-06-03-event-framework/README.md
ndr-brt: @bscholtes1A I think an eventual DLQ should be in charge of the dispatcher extension.

ndr-brt:
@paullatzelsperger, yes, the `EventRouter` (that could be call also `Hub` or `SubscriberRegistry`, let's see) will handle all the subscribers and will dispatch every event to every one of them.

The statement about the `Event` superclass was already in place, I tried to highlight it in a better way.

docs/developer/decision-records/2022-06-03-event-framework/README.md
janpmeyer: If I understand it correctly, at the beginning of its existence the Event Framework will not be able to guarantee that all events can really be read from the EventRouter, for example due to an overflow.

Is then the use of this framework for an audit logging usable or are first less critical tasks to be served with it? I don't know which tasks you currently hope to solve with this framework. Of course, this framework offers many application possibilities in general.

ndr-brt: @janpmeyer every subscriber will receive all the events but, in the case the EDC is stopped after an event emission but before its dispatchment, that event will be lost.
the first use-case we want to accomplish is asynchronous notification. 
For auditing an event persistence will be necessary, but's not part of this proposal, as it is big enough. Will be implemented in the future.

paullatzelsperger (approved):

jimmarino (approved):

Issue #341, opened 2021-12-02
Change contract dates to UTC epoch date


juliapampus:
Dates should be UTC epoch date. At least in the following classes: `ContractNegotiationManagerImpl`, `ProviderContractNegotiationManagerImpl`, `ContractValidationServiceImpl`, `ContractValidationServiceImplTest`, `ArtifactRequestHandler`

jimmarino:
- ContractOffer ZonedDateTime fields need to be converted to UTC timestamps

- IdsContractRequestToContractOfferTransformer and IdsResourceCatalogToDataCatalogTransformer need to convert received XMLGregorianCalendar to UTC, not zoned date times

- Cleanup dead ZonedDateTime imports (e.g. ContractAgreement)

- Seeing that IDS uses XMLGregorianCalendar, ContractOffer and ContractAgreement should have millisecond resolution, not day resolution. The corresponding fields need to be updated and renamed; 

- Update ContractValidationServiceImplTest, TestFunctions, InMemoryContractNegotiationStoreTest for consistency

@juliapampus do you have this underway or can I submit a PR?

juliapampus:
You can open a PR if you want @jimmarino

juliapampus:
ContractOfferFromIdsContractOfferOrRequestTransformer

Issue #1364, opened 2022-05-30
Core: Add timestamp to entities and standardize the use of `Clock`


algattik:
Standardize the use of `Clock` in the codebase, rather than constructs like `Instant.now()` which cannot be mocked easily. The `Clock` could be provided centrally.

See also #1370

A decision record should be provided.

_Originally posted by @ndr-brt in https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/1350#pullrequestreview-988906511_

PR #1416, opened 2022-06-03
Core: shared clock

algattik:
## What this PR changes/adds

Use a mockable `Clock` to get the time.

Removes calls to `Instant.now()`, `new Date()`, `System.currentTimeMillis()` from production code.

## Why it does that

- Make access to system time consistent in code base.
- Make time mockable for testing.
- Make tests simple and reliable (avoid approximate time comparisons, etc.).

## Further notes

- Moved convenience method implementations from `DefaultServiceExtensionContext` to the interface
- Changed `DataPlaneInstance` contract to return null instead of 0 for empty `lastActive` time.
- `nbf`: in discussion with Ben
- Removed `rollbackState` method in ContractNegotiation and TransferProcess, as it's unused and violates state machine contract.
- Consistently inject services in `ContractServiceExtension`

## Linked Issue(s)

Contributes to #1364 (a decision record will also be provided)

## Checklist


core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java
```diff
@@ -47,28 +48,14 @@ public DefaultServiceExtensionContext(TypeManager typeManager, Monitor monitor,
         registerService(TypeManager.class, typeManager);
         registerService(Monitor.class, monitor);
         registerService(Telemetry.class, telemetry);
+        registerService(Clock.class, Clock.systemUTC());
```
paullatzelsperger, line 51: Unless there is a good reason to the contrary we should provide the clock with a `@Provider(isDefault=true)`-annotated factory method in the `DefaultServicesExtension`.

juliapampus:
Does this also resolve issue #341? @algattik

jimmarino (approved):
Great stuff @algattik. I'd like to have a Decision Record since it is important to have this approach documented and everyone aware across the codebase.

Given the conflict issue, that could be done in a subsequent PR, even though it would have been better to commit the DR first.

algattik:
> Does this also resolve issue #341? @algattik

@juliapampus no, this is a pure "internal utils" change and does not affect any contracts or behaviours

algattik:
> Great stuff @algattik. I'd like to have a Decision Record since it is important to have this approach documented and everyone aware across the codebase.
> 
> Given the conflict issue, that could be done in a subsequent PR, even though it would have been better to commit the DR first.

Great point, I've updated the parent issue to add the need for a DR, and removed the "Closes" comment from this PR. I'll do a DR next.

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java
algattik: Great input, done :)

paullatzelsperger (approved):

core/boot/src/main/java/org/eclipse/dataspaceconnector/boot/system/DefaultServiceExtensionContext.java
algattik: The reasons for this are:
- to provide a utility method `ServiceExtensionContext#getClock` so that extensions don't need to inject a clock service. If we move the default clock to `DefaultServicesExtension`, we can't have such a utility method, since we need the dependency resolver to start extensions in the right order.
- that the clock is a common service just like `TypeManager`, `Monitor` and `Telemetry`, and there's no obvious reason to manage it differently

Issue #1357, opened 2022-05-24
Refactor multiple modules handling JWT generation and validation


algattik:
# Feature Request

- Redundant JWT generation and validation logic in common/token-generation-lib (used by OAuth2 module) and identity-did-service/identity-did-crypto
- ids-token-validation extension is not actually working (META-INF folder is misnamed)

## Which Areas Would Be Affected?
Several

## Why Is the Feature Desired?
Allow cross-cutting changes, such as #1176 

## Solution Proposal
- standardise on token-generation-lib / token-validation-lib modules, remove similar logic from identity-did-crypto
- fix ids-token-validation META-INF folder name
- write system test covering ids-token-validation and identity-did-service

## Type of Issue
cleanup, refactoring

## Checklist

- [ ] assigned appropriate label?
- [x] **Do NOT select a milestone or an assignee!**

paullatzelsperger:
I am in favor of cleaning up and making stuff more unified.

juliapampus:
> ids-token-validation extension is not actually working 

Relates to #1274

florianrusch-zf:
@algattik @paullatzelsperger Isn't this still a valid issue?

paullatzelsperger:
@florianrusch-zf it sure is. i'll reopen.

PR #1418, opened 2022-06-05
Fix: remove `token-validation` dependency from BOM

### LINKED ISSUE #1387: IDS: mandatory dependency to OAuth2
# Bug Report

## Describe the Bug
Importing `:data-protocols:ids` imports `:data-protocols:ids:token-validation` which requires a `Oauth2ValidationRulesRegistry`.

### Expected Behavior
Importing `:data-protocols:ids` should not require `OAuth2` module. If it really is required, it should be included in `:data-protocols:ids` BOM.

paullatzelsperger:
## What this PR changes/adds

Removes the `ids:token-validation` dependency from the IDS BOM.

## Why it does that

To avoid transitive dependency onto `Oauth2ValidationRulesRegistry`

## Further notes

n/a

## Linked Issue(s)

Closes #1387 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ in the detailed section referring to by linking pull requests or issues.
 * Deprecated Control API (#1310)
 * Remove sample module `:extensions:policy:ids-policy` (#1348)
 * Unused `:launchers:basic` (#1360)
+* Dependency onto `token-validation` from IDS BOM (#1387)
 
 #### Fixed
 
--- a/data-protocols/ids/build.gradle.kts
+++ b/data-protocols/ids/build.gradle.kts
@@ -25,7 +25,6 @@ dependencies {
     api(project(":data-protocols:ids:ids-transform-v1"))
     api(project(":data-protocols:ids:ids-api-multipart-endpoint-v1"))
     api(project(":data-protocols:ids:ids-api-multipart-dispatcher-v1"))
-    api(project(":data-protocols:ids:ids-token-validation"))
     api(project(":data-protocols:ids:ids-api-configuration"))
 }
 


bscholtes1A:
Hi @paullatzelsperger 
wouldnt it a most proper fix to put `ids-token-validation` into an extension, i.e. outside of the `ids` module?

Because for me you may want to rely on ids for the communication between EDC components, while still using the identity service of your choice. If this is correct, then `ids-token-validation` should then be seen at something you can plug on top, but not mandatory --> which is typically the concept of an extension.

paullatzelsperger:
> Hi @paullatzelsperger
wouldnt it a most proper fix to put ids-token-validation into an extension, i.e. outside of the ids module?

> Because for me you may want to rely on ids for the communication between EDC components, while still using the identity service of your choice. If this is correct, then ids-token-validation should then be seen at something you can plug on top, but not mandatory --> which is typically the concept of an extension.

we may even be able to consolidate several JWT validation code fragments as suggested by #1357 . However, this issue solely deals with removing the dependency from the BOM, because the way it was set up, it would have introduced a transitive dependency on a `Oauth2ValidationRulesRegistry`, which was *not* in the BOM.

bscholtes1A:
> > Hi @paullatzelsperger
> > wouldnt it a most proper fix to put ids-token-validation into an extension, i.e. outside of the ids module?
> 
> > Because for me you may want to rely on ids for the communication between EDC components, while still using the identity service of your choice. If this is correct, then ids-token-validation should then be seen at something you can plug on top, but not mandatory --> which is typically the concept of an extension.
> 
> we may even be able to consolidate several JWT validation code fragments as suggested by #1357 . However, this issue solely deals with removing the dependency from the BOM, because the way it was set up, it would have introduced a transitive dependency on a `Oauth2ValidationRulesRegistry`, which was _not_ in the BOM.

Yeap understood, but my suggestion was to remove it from the BOM **and** to move the ids token validation to a dedicated extension. But this can be done later on.

bscholtes1A (approved):

PR #1442, opened 2022-06-08
Api: Create integration test for basic authentication service

### LINKED ISSUE #1107: Implement integration test for auth-basic module
# Feature Request

`:extensions:api:auth-basic`

Regarding the finalization of Release 3 we decided to postpone the implementation of the integration test for the `auth-basic` module.

## Type of Issue
improvement

## Checklist

diegogomez-zf:
## What this PR changes/adds

Adds an integration test from the class BasicAuthenticationService. 

## Why it does that

This Integration test checks that the authentication service while accessing to a connector is correctly performed under 3 different scenarios: No header, incorrect credentials, and correct credentials.

## Linked Issue(s)

Closes #1107

## Checklist

- [x] formatted title correctly?


extensions/api/auth-basic/build.gradle.kts
```diff
@@ -25,6 +25,15 @@ dependencies {
     api(project(":extensions:api:auth-spi"))
     api(project(":spi:core-spi"))
     implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
+    implementation("org.junit.jupiter:junit-jupiter:5.8.1")
+    testImplementation("org.assertj:assertj-core:3.21.0")
+    testImplementation("org.mockito:mockito-inline:3.6.28")
+    testImplementation(project(":extensions:api:data-management:asset"))
```
paullatzelsperger, line 31: I understand that it might seem convenient to just use the Asset API as test dummy, but that's a bad idea. If the Asset API gets refactored, this would suddenly break tests in this module. We're violating the separation of concerns. Apart from that, it introduces a dependency across package boundaries, even if just for testing.

Although it might be more effort now, creating a simple test rest controller for testing is a lot cleaner, and will save us pain down the road, e.g. when new auth methods are added.

paullatzelsperger (changes requested):

extensions/api/auth-basic/build.gradle.kts
diegogomez-zf: The idea was to test if the AuthenticationRequestFilter works correctly together with the BasicAuthenticationService and the API would response with an appropriately status code, this is the reason why it was necessary to call a EDC runtime.

Regarding the other points, I rewrote the integration test to get rid of the Asset dependency and now there is a dummy API controller registered on the data management context.

extensions/api/auth-basic/build.gradle.kts
```diff
@@ -25,6 +25,14 @@ dependencies {
     api(project(":extensions:api:auth-spi"))
     api(project(":spi:core-spi"))
     implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
+    implementation("org.junit.jupiter:junit-jupiter:5.8.1")
+    testImplementation("org.assertj:assertj-core:3.21.0")
+    testImplementation("org.mockito:mockito-inline:3.6.28")
+    testImplementation(testFixtures(project(":launchers:junit")))
+    testImplementation(testFixtures(project(":common:util")))
+    testImplementation("io.rest-assured:rest-assured:${restAssured}")
+    testRuntimeOnly("org.glassfish.jersey.ext:jersey-bean-validation:${jerseyVersion}")
```
ndr-brt, line 34: not necessary

extensions/api/auth-basic/build.gradle.kts
```diff
@@ -25,6 +25,14 @@ dependencies {
     api(project(":extensions:api:auth-spi"))
     api(project(":spi:core-spi"))
     implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
+    implementation("org.junit.jupiter:junit-jupiter:5.8.1")
+    testImplementation("org.assertj:assertj-core:3.21.0")
+    testImplementation("org.mockito:mockito-inline:3.6.28")
```
ndr-brt, line 30: not needed (mockito-core should be added by default, and we don't use any feature from the `inline` package)

extensions/api/auth-basic/build.gradle.kts
```diff
@@ -25,6 +25,14 @@ dependencies {
     api(project(":extensions:api:auth-spi"))
     api(project(":spi:core-spi"))
     implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
+    implementation("org.junit.jupiter:junit-jupiter:5.8.1")
```
ndr-brt, line 28: junit is already imported by default, it can be removed from here

samples/02-health-endpoint/src/main/java/org/eclipse/dataspaceconnector/extensions/health/HealthEndpointExtension.java
```diff
@@ -28,4 +28,6 @@ public class HealthEndpointExtension implements ServiceExtension {
     public void initialize(ServiceExtensionContext context) {
         webService.registerResource(new HealthApiController(context.getMonitor()));
     }
+
```
ndr-brt, line 31: out of the scope of the PR

extensions/api/auth-basic/src/test/java/org/eclipse/dataspaceconnector/api/auth/BasicAuthenticationServiceIntegrationTest.java
```diff
@@ -0,0 +1,86 @@
...
+        when(vault.resolveSecret("api-basic-auth-hello")).thenReturn("myPassword");
+
+        baseRequest()
+                .header(new Header("Authorization", "Basic aGVsbG86bXlQYXNzd29yZA=="))
+                .get("/dummy")
+                .then()
+                .statusCode(200);
+    }
+
+    private RequestSpecification baseRequest() {
```
ndr-brt, line 80: At this point this method could be called `request`, get the header as parameter and do the effective call, leaving to the callers only to call `.then().statusCode(...)` on it

extensions/api/auth-basic/build.gradle.kts
```diff
@@ -25,6 +25,14 @@ dependencies {
     api(project(":extensions:api:auth-spi"))
     api(project(":spi:core-spi"))
     implementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
```
ndr-brt, line 27:
```suggestion
    testImplementation("jakarta.ws.rs:jakarta.ws.rs-api:${rsApi}")
```

extensions/api/auth-basic/build.gradle.kts
```diff
@@ -25,6 +25,14 @@ dependencies {
...
+    implementation("org.junit.jupiter:junit-jupiter:5.8.1")
+    testImplementation("org.assertj:assertj-core:3.21.0")
```
ndr-brt, line 29: assertj is already implemented by default

extensions/api/auth-basic/src/test/java/org/eclipse/dataspaceconnector/api/auth/DummyApiExtension.java
```diff
@@ -0,0 +1,47 @@
...
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtensionContext;
+
+public class DummyApiExtension implements ServiceExtension {
+
+    @Inject
+    WebService webService;
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        webService.registerResource("data", new DummyApiController());
```
ndr-brt, line 33: tests are broken because you need to register also an `AuthenticationRequestFilter` with the injected `AuthenticationService` and also the `EdcApiExceptionMapper`.
This could be made more simple, but let's do in another PR

extensions/api/auth-basic/build.gradle.kts
paullatzelsperger: Ok, the `AuthenticationRequestFilter` is already covered in a test, and anything beyond that would effectively be testing Jetty (which I think we can assume to be working).
As for the `BasicAuthenticationService`, there is also a test for that.

In summary, I think this integration test adds little value compared to the impact on overall test runtime it imposes.

ndr-brt: This integration test shown some things that probably are not standardized/documented enough.
E.g., to have a fully working `AuthenticationService` an `ExceptionMapper` has to be explicitly defined, and that's not documented anywhere. 

Until those things are not documented/refactored I see the value brought by this IT

paullatzelsperger: Then this PR should have added documentation about the `ExceptionMapper`, etc. It would have been more valuable, and less effort at that.
Apart from that, the entire issue circumvented our planning process.

ndr-brt (approved):

extensions/api/auth-basic/src/test/java/org/eclipse/dataspaceconnector/api/auth/DummyApiExtension.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Consumes({ MediaType.APPLICATION_JSON })
+    @Produces({ MediaType.APPLICATION_JSON })
+    @Path("/")
+    public class DummyApiController {
+
+        @GET
+        @Path("dummy")
+        public String checkHealth() {
```
florianrusch-zf, line 50:
```suggestion
        public String dumy() {
```

florianrusch-zf (approved):

PR #1443, opened 2022-06-08
Bugfix: brings back the edcVersion variable in the build file

paullatzelsperger:
## What this PR changes/adds

This PR reinstates the previously discarded `edcVersion` variable to hold the default value for the version.

## Why it does that

Due to some gradle weirdness it was possible that the version string was `"unspecified"` in subprojects.

**We won't be able to produce SNAPSHOT versions until this is merged**

## Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -46,23 +46,15 @@ val edcDeveloperEmail: String by project
 val edcScmConnection: String by project
 val edcWebsiteUrl: String by project
 val edcScmUrl: String by project
-val groupId: String = "org.eclipse.dataspaceconnector"
-var edcVersion: String = "0.0.1-SNAPSHOT"
-
-if (project.version == "unspecified") {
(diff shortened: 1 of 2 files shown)


ndr-brt (approved):
We are handling it in the hard way, better would be to have the current version in the `gradle.properties` file as:
```
version=0.0.1-SNAPSHOT
```
and then it could be overrided with a `-Pversion=<new version>` option at runtime.
This could be done in another PR in the next days, now lgtm

paullatzelsperger:
> We are handling it in the hard way, better would be to have the current version in the `gradle.properties` file as:
> 
> ```
> version=0.0.1-SNAPSHOT
> ```
> 
> and then it could be overrided with a `-Pversion=<new version>` option at runtime. This could be done in another PR in the next days, now lgtm

I updated my code a bit. apparently calling the thing `version` is dangerous, because all projects have a `project.version` that cannot be overwritten.

PR #1453, opened 2022-06-09
core: dispatch Asset events

### LINKED ISSUE #1435: Event Framework: dispatch Asset events
# Feature Request

The EDC should dispatch events regarding `Asset` lifecycle as:
- `created`
- `deleted`

The feature must follow the guidelines described in the [event-framework decision record](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/tree/6678501fe0dca78d5727adfbe0ac1cad9310c9a1/docs/developer/decision-records/2022-06-03-event-framework)

core

to permit asynchronous notifications

## Solution Proposal
-

## Type of Issue
new feature

## Checklist

ndr-brt:
## What this PR changes/adds

Introduce events for domain class `Asset` and publish them through the `EventRouter`

## Why it does that

To permit asynchronous notifications

## Further notes

- Add a first documentation markdown about how to subscribe to and how to publish events
- Implement a default behavior for `AssetLoader.accept(Asset asset, DataAddress dataAddress)` to remove some duplication
- Add the event publishing in the `AssetService` class. Currently not every component that act on Assets pass through the service (e.g. `DataLoader`). So here I have a question: should `AssetService` moved to an upstream module (like the core module or a dedicated core-asset module?)
- Add serialization capabilities to all the events. The deserialization could be achived registering the event types on the `TypeManager`. This process is described in the doc markdown but since it's not needed (the EDC doesn't need to deserialize events for the moment) I left this out.
- The `EventRouter` publishes events in an asynchronous manner to avoid blocking the main thread and catches exceptions that could be thrown by the subscribers. It doesn't implement any retry behavior nor it uses a configurable `ExecutorService`. Those two functionalities could be addressed to further PRs.

## Linked Issue(s)

Closes #1435

## Checklist

### DIFF
--- a/core/base/src/main/java/org/eclipse/dataspaceconnector/core/event/EventRouterImpl.java
+++ b/core/base/src/main/java/org/eclipse/dataspaceconnector/core/event/EventRouterImpl.java
@@ -1,0 +1,54 @@
+/*
+ *  Copyright (c) 2022 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.core.event;
+
+import org.eclipse.dataspaceconnector.spi.event.Event;
+import org.eclipse.dataspaceconnector.spi.event.EventRouter;
+import org.eclipse.dataspaceconnector.spi.event.EventSubscriber;
+import org.eclipse.dataspaceconnector.spi.monitor.Monitor;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.lang.String.format;
+import static java.util.concurrent.CompletableFuture.runAsync;
+
+public class EventRouterImpl implements EventRouter {
+
+    private final List<EventSubscriber> subscribers = new ArrayList<>();
+    private final Monitor monitor;
+
+    public EventRouterImpl(Monitor monitor) {
+        this.monitor = monitor;
+    }
+
+    @Override
+    public void register(EventSubscriber subscriber) {
+        subscribers.add(subscriber);
+    }
+
+    @Override
+    public void publish(Event event) {
+        subscribers.stream()
+                .map(subscriber -> runAsync(() -> subscriber.on(event)).thenApply(v -> subscriber))
+                .forEach(future -> future.whenComplete((subscriber, throwable) -> {
+                    if (throwable != null) {
+                        var subscriberName = subscriber.getClass().getSimpleName();
+                        var eventName = event.getClass().getSimpleName();
+                        monitor.severe(format("Subscriber %s failed to handle event %s", subscriberName, eventName), throwable);
+                    }
+                }));
+    }
+}
--- a/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetListener.java
+++ b/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetListener.java
@@ -24,7 +24,7 @@ import org.eclipse.dataspaceconnector.spi.types.domain.asset.Asset;
 public interface AssetListener {
 
     /**
-     * Called afer a {@link Asset} has created.
+     * Called after a {@link Asset} was created.
      *
      * @param asset the asset that has been created.
      */
@@ -33,7 +33,7 @@ public interface AssetListener {
     }
 
     /**
-     * Called afer a {@link Asset} has deleted.
+     * Called after a {@link Asset} was deleted.
      *
      * @param asset the asset that has been deleted.
      */
--- a/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetListener.java
+++ b/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetListener.java
@@ -1,0 +1,44 @@
+/*
+ *  Copyright (c) 2022 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.api.datamanagement.asset.service;
+
+import org.eclipse.dataspaceconnector.spi.observe.Observable;
+import org.eclipse.dataspaceconnector.spi.types.domain.asset.Asset;
+
+/**
+ * Interface implemented by listeners registered to observe asset state changes via {@link Observable#registerListener}.
+ * The listener must be called after the state changes are persisted.
+ */
+public interface AssetListener {
+
+    /**
+     * Called afer a {@link Asset} has created.
+     *
+     * @param asset the asset that has been created.
+     */
+    default void created(Asset asset) {
+
+    }
+
+    /**
+     * Called afer a {@link Asset} has deleted.
+     *
+     * @param asset the asset that has been deleted.
+     */
+    default void deleted(Asset asset) {
+
+    }
+
+}
--- a/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
+++ b/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
@@ -33,12 +33,14 @@ public class AssetServiceImpl implements AssetService {
     private final AssetLoader loader;
     private final ContractNegotiationStore contractNegotiationStore;
     private final TransactionContext transactionContext;
+    private final AssetObservable observable;
 
-    public AssetServiceImpl(AssetIndex index, AssetLoader loader, ContractNegotiationStore contractNegotiationStore, TransactionContext transactionContext) {
+    public AssetServiceImpl(AssetIndex index, AssetLoader loader, ContractNegotiationStore contractNegotiationStore, TransactionContext transactionContext, AssetObservable observable) {
         this.index = index;
         this.loader = loader;
         this.contractNegotiationStore = contractNegotiationStore;
         this.transactionContext = transactionContext;
+        this.observable = observable;
     }
 
     @Override
@@ -56,6 +58,7 @@ public class AssetServiceImpl implements AssetService {
         return transactionContext.execute(() -> {
             if (findById(asset.getId()) == null) {
                 loader.accept(asset, dataAddress);
+                observable.invokeForEach(l -> l.created(asset));
                 return ServiceResult.success(asset);
             } else {
                 return ServiceResult.conflict(format("Asset %s cannot be created because it already exist", asset.getId()));
@@ -77,6 +80,7 @@ public class AssetServiceImpl implements AssetService {
                 return ServiceResult.notFound(format("Asset %s does not exist", assetId));
             }
 
+            observable.invokeForEach(l -> l.deleted(deleted));
             return ServiceResult.success(deleted);
         });
     }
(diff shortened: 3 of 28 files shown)


extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
```diff
@@ -56,6 +58,7 @@ public ServiceResult<Asset> create(Asset asset, DataAddress dataAddress) {
         return transactionContext.execute(() -> {
             if (findById(asset.getId()) == null) {
                 loader.accept(asset, dataAddress);
+                observable.invokeForEach(l -> l.created(asset));
```
paullatzelsperger, line 61: maybe add a test to verify this. same for line 83

extensions/api/data-management/asset/src/test/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetEventDispatchTest.java
```diff
@@ -0,0 +1,72 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.api.datamanagement.asset.service;
+
+import org.eclipse.dataspaceconnector.junit.launcher.EdcExtension;
```
paullatzelsperger, line 17: I assume this was a rebase/merge artifact:
```suggestion
import org.eclipse.dataspaceconnector.junit.extensions.EdcExtension;
```

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/event/AssetDeleted.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ */
+
+package org.eclipse.dataspaceconnector.spi.event;
+
+import java.util.Objects;
+
+public class AssetDeleted extends Event {
+
+    private String id;
```
paullatzelsperger, line 21: is this supposed to be the `assetId`? if so, i'd rename it to avoid ambivalence: one could be lead to believe this were the event's ID.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/event/AssetDeleted.java
```diff
@@ -0,0 +1,52 @@
...
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.spi.event;
+
+import java.util.Objects;
+
+public class AssetDeleted extends Event {
```
paullatzelsperger, line 19: maybe a line of documentation?

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/event/AssetCreated.java
```diff
@@ -0,0 +1,52 @@
...
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.spi.event;
+
+import java.util.Objects;
+
+public class AssetCreated extends Event {
```
paullatzelsperger, line 19: maybe a line of documentation?

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/event/Event.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+/**
+ * The Event base class.
+ * It provides a default field "at" that gives the timestamp when the event was created.
+ * When serialized to JSON, will add a "type" field with the name of the runtime class name.
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type", include = JsonTypeInfo.As.EXTERNAL_PROPERTY)
+public abstract class Event {
```
paullatzelsperger, line 25: Do you think it would be worthwhile to have the `at` field initialized with the current time by default? That would reduce the amount of duplicated code greatly, because then event sources wouldn't have to call `at(clock.millis())`.  
The general use of `Clock` makes this more cumbersome, but we could create a factory method of sorts, just to make it a bit sleeker to use, wdyt?

spi/core-spi/src/test/java/org/eclipse/dataspaceconnector/spi/event/AssetCreatedTest.java
```diff
@@ -0,0 +1,45 @@
...
+        var typeManager = new TypeManager();
+        typeManager.registerTypes(new NamedType(AssetCreated.class, AssetCreated.class.getSimpleName()));
+
+        var event = AssetCreated.Builder.newInstance()
+                .id("id")
+                .at(Clock.systemUTC().millis())
+                .build();
+
+        var json = typeManager.writeValueAsString(event);
+        System.out.println(json);
```
paullatzelsperger, line 38: not std_out in tests pls (i assume you just forgot to remove it).

extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetListener.java
```diff
@@ -0,0 +1,44 @@
...
+ * The listener must be called after the state changes are persisted.
+ */
+public interface AssetListener {
+
+    /**
+     * Called afer a {@link Asset} has created.
+     *
+     * @param asset the asset that has been created.
+     */
+    default void created(Asset asset) {
```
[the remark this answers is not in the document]
ndr-brt: @paullatzelsperger not a big fan too. Will open a separate issue to avoid this also in other places

core/base/src/main/java/org/eclipse/dataspaceconnector/core/event/EventRouterImpl.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public void publish(Event event) {
+        subscribers.stream()
+                .map(subscriber -> runAsync(() -> subscriber.on(event)).thenApply(v -> subscriber))
+                .forEach(future -> future.whenComplete((subscriber, throwable) -> {
+                    if (throwable != null) {
+                        var subscriberName = subscriber.getClass().getSimpleName();
+                        var eventName = event.getClass().getSimpleName();
+                        monitor.severe(format("Subscriber %s failed to handle event %s", subscriberName, eventName), throwable);
```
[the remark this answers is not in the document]
ndr-brt: @paullatzelsperger `RetryPolicy` and the `Result` are already on my list, will open an issue later on.
About the DLQ I'm not sure should be an edc-core responsability, I would leave this up to the extensions but we can talk about this.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/event/AssetCreated.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private String id;
```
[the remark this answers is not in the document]
ndr-brt: @paullatzelsperger You're right, I was thinking about the `Event` structure, maybe we could keep an `EventEnvelope` that contains an Event field so that we won't mind about field name conflicts. wdyt about this?

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/event/AssetDeleted.java
ndr-brt: @ndr-brt the name of the class is pretty self-explanatory, I find it a bit difficult to write something that doesn't sound totally useless :)

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/event/Event.java
ndr-brt: @paullatzelsperger this could go in the discussion about the event envelope mentioned in a comment above, if we do that we could create the event instance separately and then having someone (also the event router`) that would wrap the event with an envelope with other informations like `at`, an id, the type, and so on.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/event/AssetDeleted.java
jimmarino: How about what "deleted" means and what it entails, e.g. is "deletion" a marker and the Asset entity still in the backing store?

ndr-brt: @jimmarino ok, got it, I would avoid the detail about the logical/phisical deletion, as it is an implementation delay of the store layer (I guess). 
I reported that, after the event emission, the asset won't be available anymore.

extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/EventAssetListener.java
```diff
@@ -0,0 +1,52 @@
...
+
+package org.eclipse.dataspaceconnector.api.datamanagement.asset.service;
+
+import org.eclipse.dataspaceconnector.spi.event.AssetCreated;
+import org.eclipse.dataspaceconnector.spi.event.AssetDeleted;
+import org.eclipse.dataspaceconnector.spi.event.EventRouter;
+import org.eclipse.dataspaceconnector.spi.types.domain.asset.Asset;
+
+import java.time.Clock;
+
```
jimmarino, line 23: How about some doc here since it is in api?

ndr-brt:
@jimmarino @paullatzelsperger , I added a commit where I introduced the `EventEnvelope` class, that contains the event and the metadata about the event (for the moment only the timestamp, but other information could be included in the future, as an id, the connector id and so on).
I'm in doubt about the naming, because I'm not sure that the `Event` class now represent an event, since it misses the timestamp, an idea could be to rename `EventEnvelope` to `Event` and `Event` to `EventData` or something like this. wdyt?

core/base/src/main/java/org/eclipse/dataspaceconnector/core/event/EventRouterImpl.java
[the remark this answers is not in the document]
jimmarino: I think events should be fire-and-forget and non-reliable (i.e., no delivery guarantees). Otherwise, we will wind up building a reliable messaging platform. Another approach would be to architect reliability points like we do with the state machine implementations for specific cases. 

I don't think the events in this PR (e.g., asset created, etc.) need to be reliable. If someone needs this behavior, they should account for it in the client that issues the command that triggered the event.

 Let's discuss this further after this PR.

extensions/dataloading/src/main/java/org/eclipse/dataspaceconnector/dataloading/AssetLoader.java
```diff
@@ -20,7 +20,9 @@
 
 public interface AssetLoader extends DataSink<AssetEntry> {
```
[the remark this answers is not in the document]
ndr-brt: Is out of the scope of the PR, since I just refactored it. Will open a dedicated issue

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/event/Event.java
jimmarino: > @jimmarino @paullatzelsperger , I added a commit where I introduced the `EventEnvelope` class, that contains the event and the metadata about the event (for the moment only the timestamp, but other information could be included in the future, as an id, the connector id and so on). I'm in doubt about the naming, because I'm not sure that the `Event` class now represent an event, since it misses the timestamp, an idea could be to rename `EventEnvelope` to `Event` and `Event` to `EventData` or something like this. wdyt?

Agreed the "Event" class should be the top-level name. Why not collapse the two classes into a single Event class? Generally, envelopes are introduced so that code can either process parts of a message without having to access the body (e.g. encrypted data) and/or to "unwrap" a message so that header information does not pollute non-transport areas of a code base.

BTW, this is a great PR! :-)

extensions/dataloading/src/main/java/org/eclipse/dataspaceconnector/dataloading/AssetLoader.java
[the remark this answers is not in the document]
ndr-brt: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/1467

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/event/Event.java
ndr-brt: @jimmarino so your advice would be to have everything (data and metadata) inside the `Event` class as it was before my last commit?

ndr-brt:
@jimmarino @paullatzelsperger as we agreed I came back to the `Event` version without envelope, let's see how this will evolve with the next PRs

paullatzelsperger (approved):

extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetListener.java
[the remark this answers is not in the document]
jimmarino: FWIW, I like them when used to avoid boilerplate code :-)

jimmarino (approved):

PR #1470, opened 2022-06-13
refactor: remove ConnectorVersionProvider

### LINKED ISSUE #1466: Move `ConnectorVersion` interface from ids to edc core
This is not an IDS specific property. Should be located in the core spi.

juliapampus:
## What this PR changes/adds

Renames and moves `ConnectorVersionProvider` interface and impl from ids spi to core spi.

## Why it does that

Resolve TODO. The software version is not an IDS specific property.

## Further notes

Is `spi.system.version` the right location, or is there a better one?

## Linked Issue(s)

Closes #1466

## Checklist


ndr-brt (approved):

jimmarino:
Can we rename this to EdcVersion since this can be used by all EDC runtimes and not just the connector?

juliapampus:
> Can we rename this to EdcVersion since this can be used by all EDC runtimes and not just the connector?

Sure

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/system/version/EdcVersion.java
```diff
@@ -12,10 +12,9 @@
  *
  */
 
-package org.eclipse.dataspaceconnector.ids.spi.version;
+package org.eclipse.dataspaceconnector.spi.system.version;
 
-// TODO move to edc-spi
-public interface ConnectorVersionProvider {
+public interface EdcVersion {
```
paullatzelsperger, line 17: Why do we need all these interfaces and impls for a simple version string?

How about we let the `ConnectorServiceImpl` either use a hard-coded version string (`0.0.1-SNAPSHOT`) instead of the `ConnectorVersionProvider`, and we delete the `ConnectorVersionProvider` and the `EdcVersion`?
In the future we could improve this:
- create a gradle task that updates this hard-coded value before compile/build
- create a gradle task that generates a separate version file during build, which is then parsed by the `ConnectorServiceImpl`. That file would have to be shipped together with EDC code.

juliapampus: @paullatzelsperger Changed as proposed

paullatzelsperger (approved):
LGTM

PR #1479, opened 2022-06-14
api: permit verbose body error message

### LINKED ISSUE #1411: api: verbose response body error messages
# Feature Request

Sometimes it's hard to understand why an API request fails, as in the body there are no details about what happened.
Would be useful to have a configuration setting to enable a verbose request body in these cases.
By default this should be off as it is not recommended in a production environment to give such details to the outside.

api

To help debug

## Solution Proposal
`EdcApiExceptionMapper` is a place where all the exceptions thrown in the api context are handled. 
Should be enough to put the `Result` object into the `EdcApiExtension` and if a particular setting is enabled, the error detail could be included in the body.
Should worth how to make also `jakarta-validation` exceptions pass through this, as I don't think they currently do this (needs further investigation).
If the "verbose response body" setting is enabled, a warning at startup should be printed.

## Type of Issue
improvement

## Checklist

ndr-brt:
## What this PR changes/adds

Permits verbose response message in case of failure. This will be configurable with a setting that's `false` by default

## Why it does that

To give a detailed reason to the client, that will permit a better understanding of the failure.

## Further notes

- Moved `EdcApiException` and subclasses to the `web-spi` module
- Moved `EdcApiExceptionMapper` to the `jersey` module and loaded by default on every context.
- Renamed `CorsFilterConfiguration` to `JerseyConfiguration` with the additional setting about verbose response.
- Added README for the jersey module

## Linked Issue(s)

Closes #1411 

## Checklist


paullatzelsperger (approved):

jimmarino:
I'm good with this PR but think we need to improve error reporting further. The default runtime behavior should be to return actionable error messages when invalid data is sent. For example, a client not under the control of the EDC operator may send a message containing invalid data. The response returned by the EDC should assist the client in diagnosing why the data is invalid.

This can be done by:

1. Modifying the `EdcApiExceptionMapper` to handle messages returned by `ConstraintViolationException`. These messages should be returned as a payload containing a defined JSON error object.
2. All constraint annotations used by EDC should contain a message. Eventually, the message should be i18n-enabled via configuration but maybe not in the first go-around.
3. This behavior should be enabled by default. If we want extended debug information, that could be optionally enabled.

@ndr-brt what do you think?

jimmarino (approved):

ndr-brt:
@jimmarino I agree this could improve, in particular:
1. Yes, currently the json payload structure for the `ConstraintViolationException` is defined by the `ValidationErrorData` class located in the validation library, we need to define a common structure for all the errors.
2. I wasn't aware of the possibility to have a custom message on the annotations, totally agree on this.
3. Makes sense. What could be an "extended debug information"? Do you mean attaching the stack trace to the message for example? (that's something that needs to be avoided in production but could be useful for debug.

will create the issues accordingly

Issue #1463, opened 2022-06-11
Core: base classes for entities and in-memory store


algattik:
# Feature Request

Developing an extension that introduces a custom entity with `state-machine-lib` requires writing boilerplate code that is already duplicated between `TransferProcess` and `ContractNegotiation`, as well as their store implementations (presumably this will apply to `DataFlowRequest` as well, when persistence for it is developed). There is an opportunity to introduce extensibility and reduce technical debt instead.

The code in question is primarily associated with state code and associated timestamps.

To reduce code duplication, we could develop:
- Generic base class for entities with associated state.
- Generic base class for in-memory entity store.
- Adapt `TransferProcess` and `ContractNegotiation` and their in-memory stores to use the base classes.

As follow-up work, generic base classes for Cosmos DB and PostgreSQL entity stores could be developed.

The interfaces and behavior would not change.

Note: `InMemoryTransferProcessStore` uses an additional caching layer and custom locking mechanism. Given this class is only meant to use for testing, it seems reasonable to remove this additional layer and use a simple `ConcurrentHashMap` for storage in the new generic in-memory store to reduce complexity.

## Which Areas Would Be Affected?

Core

## Solution Proposal

```java

public abstract class StateMachineEntity<T> implements TraceCarrier {

    protected String id;
    protected long createdTimestamp;
    protected int state;
    protected int stateCount;
    protected long stateTimestamp;
    protected Map<String, String> traceContext = new HashMap<>();
    protected String errorDetail;
    protected Clock clock;

   protected abstract static class Builder<T extends StateMachineEntity, B extends Builder<T, B>> {
   /...
}

public class TransferProcess extends StateMachineEntity<TransferProcess> {
}
```

```java
public class InMemoryEntityStore<T extends StateMachineEntity<T>> implements StateEntityStore<T> {
    T find(String id);
    void create(T entity);
    void update(T entity);
    void delete(String processId);
    Stream<T> findAll(QuerySpec querySpec);
    List<T> nextForState(int state, int max);
}
```

## Type of Issue

Technical debt refactoring

## Checklist

- [ ] assigned appropriate label?
- [x] **Do NOT select a milestone or an assignee!**

jimmarino:
This would be a great improvement to have for M5!

PR #1489, opened 2022-06-16
bugfix: avoid endless loops on contract negotiation sending failure

### LINKED ISSUE #1487: core: if negotiation sending fails start loops endlessly
# Bug Report

## Describe the Bug
If for any reason the provider connector respond with an error to the negotiation sending, the consumer will start try to send it again endlessly

### Expected Behavior
There should be a limit of retries, after which the contract should transit to ERROR.

### Observed Behavior
In the `ContractNegotiationManager`s instances there is no retry handling 

## Steps to Reproduce
-

## Context Information
-

## Detailed Description
-

## Possible Implementation
A `SendRetryManager` should be introduced and consulted after every sending failure, if the number of retries exceeds the limit, the negotiation will transit to `ERROR` (as done in the `TransferProcessManagerImpl`)

ndr-brt:
## What this PR changes/adds

Avoid endless loops on contract negotiation sending failures.

## Why it does that

To avoid log flooding.

## Further notes

- used the `SendRetryManager` component, generalized to work also with negotiations and moved to the `state-machine-lib` module
- introduced `Entity` interface, that could be a starting point for #1463 (used an interface because it was the easiest thing to do, probably a class would fit better)
- introduced an internal contract negotiation manager component called `AsyncSendResultHandler` that takes care of handling the async results, this was to remove a lot of duplication, and actually makes the code more readable.
- I really think the `SendRetryManager` is pretty trivial to understand, I have in mind a refactor that could make everything more understandable, will propose it in another PR as this brought already enough modifications.

## Linked Issue(s)

Closes #1487 

## Checklist


ronjaquensel (approved):

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/AbstractContractNegotiationManager.java
```diff
@@ -136,10 +160,68 @@ public T build() {
...
+            return this;
+        }
+
+        public AsyncSendResultHandler onFailure(Consumer<ContractNegotiation> onFailureHandler) {
+            this.onFailureHandler = onFailureHandler;
+            return this;
+        }
+
+        public BiConsumer<Object, Throwable> build() {
+            return (response, throwable) -> {
```
paullatzelsperger, line 201: `onFailureHandler` and `onSuccessHandler` both can be `null` here. Either defensively add a null-check to the `if`-statements in L209 and L218, or add non-null requirements in the `build` method.
As a third option, both handlers could have default noop consumers.

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -246,6 +244,11 @@ public void enqueueCommand(ContractNegotiationCommand command) {
         commandQueue.enqueue(command);
     }
 
+    @Override
+    protected String getName() {
+        return "Consumer";
```
paullatzelsperger, line 249: nit: could be a constant

core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/negotiation/ProviderContractNegotiationManagerImpl.java
```diff
@@ -196,6 +193,11 @@ public StatusResult<ContractNegotiation> consumerApproved(ClaimToken token, Stri
         return StatusResult.success(negotiation);
     }
 
+    @Override
+    protected String getName() {
+        return "Provider";
```
paullatzelsperger, line 198: nit: could be a constant

paullatzelsperger (approved):

PR #1510, opened 2022-06-22
feat: Extra configuration for HttpDataSink #1480

### LINKED ISSUE #1480: Extra configuration for HttpDataSink
# Feature Request

HttpDataSink supports at the moment only HTTP POST and it changes the initial endpoint by appending the part name.

In order to make it compatible with upload AWS S3 presigned URLs it needs extra adjustments, which are part of this feature request.

Dataplane HTTP extension

We need this feature for small files uploads in cloud (< 50 MB). 
The flow is:
- create upload presign URL
- use http data source dataplane to get the data 
- use http data sink dataplane to push the data to the created presign URL 

## Solution Proposal

In case we want to use it for uploading data to an upload presign URL it needs to be changed to support HTTP PUT and not to modify the endpoint (unfortunately a single presign URL can be used for the all content).

We propose adding 3 new asset properties:
- httpVerb- possible values:
      - POST (default) - use HTTP POST to send data
      - PUT - use HTTP PUT to send data
- additionalHeaders - make it possible to add extra headers like Azure x-ms-blob-type:BlockBlob
      - added not as String bu as Json key-value object like:
      ```"dataDestination": {
    "properties": {
      "endpoint": "https://qwerty.blob.core.windows.net/test/test.dat?sv=2021-06-08&ss=qwerty&srt=qwerty&sp=qaz&se=2022-06-20T17:19:22Z&st=2022-06-20T09:19:22Z&spr=https&sig=qwerty",
      "name": "",
      "type": "HttpData",
     "additionalHeaders" : {"Content-Type" : "application/octet-stream","x-ms-blob-type": "BlockBlob"} }
      ```
and a new config property:
- edc.dataplane.http.sink.partition.size - used to setup the partition size (5 by default as hardcoded now) - see HttpDataSinkFactory

### Limitations
In order to enable multipart cloud data upload (for big files > 50MB) a different solution should be used instead - AWS/Azure extension (which require cloud credentials) or a solution similar to - see [AWS doc](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html) and [Boto example](https://dev.to/traindex/multipart-upload-for-large-files-using-pre-signed-urls-aws-4hg4).

image

see [CloudHttpDataSink](https://github.com/mindsphere/DataSpaceConnector-Fork/blob/fac2993f1efdc6e718405195be4e91e0c865bfba/extensions/data-plane/data-plane-cloud-http/src/main/java/org/eclipse/dataspaceconnector/dataplane/cloud/http/pipeline/CloudHttpDataSink.java)

### When to use
This solution is usable in case of:
- no cloud direct access (wrapped access like when using [Integrated Data Lake service offered by Mindsphere](https://documentation.mindsphere.io/MindSphere/apis/iot-integrated-data-lake/api-integrated-data-lake-overview.html))
- data small in size (< 50 MB) 

## Type of Issue
new feature

## Checklist

lucian-torje-siemens:
## What this PR changes/adds

Adds support for changing:
* http verb used
* additional headers
* number of sink partitions

## Why it does that

In order to achieve compatibility with AWS S3, Azure Blob and ALI Object Storage presigned upload URLs

## Further notes

Dataplane Http Extension

## Linked Issue(s)

Closes #1480

## Checklist

## Related

[catenax-ng pull request](https://github.com/catenax-ng/product-edc/pull/215)

## Notify list

@sn0wcat
@kbData
@ndr-brt


extensions/data-plane/data-plane-http/src/main/java/org/eclipse/dataspaceconnector/dataplane/http/pipeline/HttpDataSink.java
```diff
@@ -65,6 +83,42 @@ protected StatusResult<Void> transferParts(List<DataSource.Part> parts) {
...
+
+            return Optional.of(
+                    requestBuilder
+                            .url(endpoint)
+                            .put(body)
+                            .build());
+        }
+    }
+
+    private class HttpDataSinkRequestPost implements HttpDataSinkRequest {
```
ndr-brt, line 110: This and `HttpDataSinkRequestPut` could be the same thing, as the method can be passed to the request builder with the `method` method that would need the `ContentType` (that could be a parameter of `HttpDataAddress` that's  `OCTET_STREAM` by default)

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/dataspaceconnector/dataplane/http/DataPlaneHttpExtension.java
```diff
@@ -54,11 +61,21 @@ public String name() {
...
-        var sinkFactory = new HttpDataSinkFactory(httpClient, executorContainer.getExecutorService(), 5, monitor);
+        var sinkFactory = new HttpDataSinkFactory(httpClient, executorContainer.getExecutorService(), sinkPartitionSize, monitor);
         pipelineService.registerFactory(sinkFactory);
     }
+
+    @NotNull
+    private Integer getSinkPartitionSize(ServiceExtensionContext context) {
+        try {
+            return context.getSetting(EDC_DATAPLANE_HTTP_SINK_PARTITION_SIZE, DEFAULT_PART_SIZE);
+        } catch (NumberFormatException e) {
```
ndr-brt, line 77: `NumberFormatException` is already handled by the `Config` service implementation, so it's not needed to be caught

ndr-brt (changes requested):

ndr-brt:
You need to sign the Eclipse Contributor Agreement as well

bscholtes1A (changes requested):

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/dataspaceconnector/dataplane/http/pipeline/HttpDataSinkFactory.java
```diff
@@ -75,6 +75,9 @@ private Result<DataSink> createDataSink(DataFlowRequest request) {
         }
         var authKey = dataAddress.getAuthKey();
         var authCode = dataAddress.getAuthCode();
+        var httpVerb = dataAddress.getHttpVerb();
```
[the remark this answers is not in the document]
lucian-torje-siemens: @bscholtes1A  When could we expect to have the new code with HTTP method as part of the input?

spi/core-spi/src/test/java/org/eclipse/dataspaceconnector/spi/types/domain/DataAddressTest.java
```diff
@@ -60,4 +60,16 @@ void verifyNullKeyThrowsException() {
                 .hasMessageContaining("Property key null.");
     }
 
+    @Test
+    void verifyGetDefaultPropertyValue() {
+        assertEquals("defaultValue", DataAddress.Builder.newInstance().type("sometype").build()
```
ndr-brt, line 65: We use `assertj` for assertion, I suggest to stick with it instead of using basic junit assertion methods

extensions/data-plane/data-plane-http/src/test/java/org/eclipse/dataspaceconnector/dataplane/http/pipeline/HttpDataSinkFactoryTest.java
```diff
@@ -110,6 +110,34 @@ void verifyCreateAuthenticatingSource() throws InterruptedException, ExecutionEx
         verify(call).execute();
     }
 
+    @Test
+    void verifyCreateAdditionalHeaders() throws InterruptedException, ExecutionException, IOException {
+        var dataAddress = HttpDataAddress.Builder.newInstance()
+                .baseUrl("http://example.com")
+                .addAdditionalHeader("Content-Type", "application/test-octet-stream")
```
ndr-brt, line 117: `Content-Type` would fit better as an `HttpDataAddress` field, and then be used into the `StreamingRequestBody`, as it's not an additional header, more a mandatory one, and on okhttp the content type is attached to the request body directly.
It should have a default value of `application/octet-stream` to avoid breaking changes.

ndr-brt (changes requested):
great, from my side we are almost ready

extensions/data-plane/data-plane-http/src/test/java/org/eclipse/dataspaceconnector/dataplane/http/pipeline/HttpDataSinkFactoryTest.java
lucian-torje-siemens: Content-Type has it's own field now - in case it is added as additional header, it will not be taken into consideration

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/dataspaceconnector/dataplane/http/pipeline/HttpDataSink.java
```diff
@@ -91,6 +103,16 @@ public Builder httpClient(OkHttpClient httpClient) {
             return this;
         }
 
+        public Builder additionalHeaders(Map<String, String> additionalHeaders) {
+            sink.additionalHeaders = additionalHeaders;
```
jimmarino, line 107: This should addAll() to an existing member map

spi/core-spi/src/test/java/org/eclipse/dataspaceconnector/spi/types/domain/HttpDataAddressTest.java
```diff
@@ -0,0 +1,66 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+class HttpDataAddressTest {
+
+    @Test
+    void verifyGetProperties() {
+        HttpDataAddress dataAddress = HttpDataAddress.Builder.newInstance()
```
bscholtes1A, line 26:
~~~
var dataAddress = ...
~~~

bscholtes1A (approved):

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/dataspaceconnector/dataplane/http/pipeline/HttpDataSinkFactory.java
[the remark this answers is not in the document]
bscholtes1A: During the week

PR #1511, opened 2022-06-22
Core: state machine entity base class

### LINKED ISSUE #1463: Core: base classes for entities and in-memory store

algattik:
## What this PR changes/adds

- Extracted `StateMachine` utility class for state machine entities
- Renamed previous `StateMachine` class to `StateMachineManager` for clarity and uniformity
- Developer documentation on how to use/develop persistent state machine
- Removed caching layer in in-memory store (as it added complexity with unclear benefits for code meant to be used for testing/at small scale)

## Why it does that

- Use the concept of [state machine](https://en.wikipedia.org/wiki/Finite-state_machine) in a way that follows its definition
- Reduce duplication in TransferProcess and ContractNegotiation and their in-memory stores
- Make state machine easier to use together with state-machine-lib in extension projects.

## Further notes

Also fixed race condition in ContractNegotiation (`transition(DECLINED, DECLINING, PROVIDER_OFFERED, CONFIRMING, CONFIRMED, REQUESTED);`), that manifested itself during a test run

## Linked Issue(s)

Closes #1463 

## Checklist


jimmarino (approved):

ndr-brt (commented):
What's the motivation behind modeling `StatefulEntity` as an interface and `StateMachine` as a superclass of `TransferProcess` and `ContractNegotiation`?
In my opinion they are not a `StateMachine`, they are `StatefulEntity` which state transitions are handled through a `StateMachine`, so the latter shouldn't be a superclass but something that interacts with the entities into the manager.

I would keep the `StatefulEntity` as an abstract class that would end up having an `Entity` superclass in turn being superclass of `TransferProcess` and `ContractNegotiation`.
The `Entity` class in the future would become the superclass of other non-stateful entities as `Asset` and `PolicyDefinition`

algattik:
> What's the motivation behind modeling `StatefulEntity` as an interface and `StateMachine` as a superclass of `TransferProcess` and `ContractNegotiation`? In my opinion they are not a `StateMachine`, they are `StatefulEntity` which state transitions are handled through a `StateMachine`, so the latter shouldn't be a superclass but something that interacts with the entities into the manager.

Formally, "A state machine is an abstract machine that can be in exactly one of a finite number of states at any given time" (https://en.wikipedia.org/wiki/Finite-state_machine) so the state machine is the entity, not the manager.

Nevertheless, I have renamed `StateMachine` to `StatefulEntity`

> I would keep the `StatefulEntity` as an abstract class that would end up having an `Entity` superclass in turn being superclass of `TransferProcess` and `ContractNegotiation`. The `Entity` class in the future would become the superclass of other non-stateful entities as `Asset` and `PolicyDefinition`

As I understand this will be done by @diegogomez-zf as part of https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/1370. Is it ok to do things in two steps like this?

ndr-brt (approved):

docs/developer/state-machine.md
```diff
@@ -0,0 +1,68 @@
+# State machine
+
+EDC provides a lightweight framework to facilitate the development of persistent state machines.
+
+The framework currently manage a single execution thread.
```
paullatzelsperger, line 5:
```suggestion
The framework currently manages a single execution thread.
```

docs/developer/state-machine.md
```diff
@@ -0,0 +1,68 @@
...
+      return true;
+    }
+   return false;
+}
+...
+```
+
+## State-machine store
+
+The Store which manage `StatefulEntity` persistence must persist entities in a storage system. In-memory implementations are provided for testing.
```
paullatzelsperger, line 55: nit: manages

paullatzelsperger (approved):
LGTM apart from minor nits.

We could look into introducing another base class `Entity` as superclass of `StatefulEntity`. Analogously we could devise an `InMemoryEntityStore` that in turn is a superclass of `InMemoryEntityStore` and takes those `Entity` objects.
There, we could store things like `Policy`, `ContractDefnition`, `Asset`, etc. That need not be done in this PR though.

algattik:
> LGTM apart from minor nits.
> 
> We could look into introducing another base class `Entity` as superclass of `StatefulEntity`. Analogously we could devise an `InMemoryEntityStore` that in turn is a superclass of `InMemoryEntityStore` and takes those `Entity` objects. There, we could store things like `Policy`, `ContractDefnition`, `Asset`, etc. That need not be done in this PR though.

Yes indeed as I understand that is (at least in part) in scope for @diegogomez-zf in https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/1370.

PR #1534, opened 2022-06-23
feat: add Hashicorp Vault secrets backend

### LINKED ISSUE #1026: HashiCorp Vault extension
# Feature Request

It's time to support an OSS `Vault` implementation.

`vault`

To being able to deploy the application on a K8s cluster without relying on external any (cloud) service.

## Solution Proposal
An extension module that provides an implementation for the `Vault.class` service interface.

## Type of Issue
improvement

## Checklist

bcronin90:
## What this PR changes/adds

Add Hashicorp Vault backend for storing secrets.

## Why it does that

To be able to deploy the EDC without any cloud dependencies.

## Further notes

## Linked Issue(s)

Closes #1026 

## Checklist


bcronin90:
@ndr-brt @paullatzelsperger @juliapampus Hey, I'd appreciate some input in the following matter:

Downstream, we used Testcontainers for the integration test of this component. I've mostly adapted that to use services in github actions, as per your standards upstream. However, I am having some issues debugging this. I can run the actions locally with ACT, but I'm unable to find the issue there, as the test output is very sparse, even on -v. Do you have an idea how best to handle this?

paullatzelsperger:
@bcronin90 not entirely sure what ACT refers to, but I usually run the tests locally in IntelliJ using docker containers. 
Moving to services from there is normally pretty straight-forward.

bcronin90:
> @bcronin90 not entirely sure what ACT refers to, but I usually run the tests locally in IntelliJ using docker containers. Moving to services from there is normally pretty straight-forward.

It's a similar approach: https://github.com/nektos/act
Basically runs the action locally. I can set up the necessary containers manually, but I was worried that might not be representative.

ndr-brt:
@bcronin90 to run a local instance of vault to be used for integration tests you can do (taken from the [documentation](https://hub.docker.com/_/vault))
```bash
 docker run  -e 'VAULT_DEV_ROOT_TOKEN_ID=myroot' -p "8200:8200" vault
```
(on the docs there are some details about a `--cap-add=IPC_LOCK` parameter, but on my machine it worked also without)

`http://localhost:8200` would be the url and `myroot` the token to be configured to connect

theoretically, the github service would need to be configured like this:
```yaml
    services:
      hashicorp-vault:
        image: vault
        ports:
          - 8200:8200
        env:
          VAULT_DEV_ROOT_TOKEN_ID: myroot
```

There is another problem, I noticed that, the `OkHttpClient` injection in the vault creates a cyclic dependency between the `CoreServicesExtension`, this could be a bug inside the injection loading, as in fact there would be no cyclic dependency, so my advice would be, momentarily, to have everything running, to avoid injecting the `OkHttpClient` and creating one from scratch leaving a TODO to be fixed later on.

extensions/hashicorp-vault/src/test/resources/logback.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jimmarino, line 1: Looks like this file was inadvertently added

gradle.properties
```diff
@@ -46,3 +46,4 @@ azureIdentityVersion=1.4.6
 azureResourceManagerVersion=2.12.0
 azureResourceManagerDataFactory=1.0.0-beta.12
 azureKeyVaultVersion=4.2.3
+testContainersVersion=1.17.2
```
jimmarino, line 49: Can this be removed?

extensions/hashicorp-vault/src/test/resources/logback.xml
bcronin90: Yes, this was needed downstream. Removed it.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpCertificateResolver.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+
+/**
+ * Resolves an X.509 certificate in Hashicorp vault.
+ */
+public class HashicorpCertificateResolver implements CertificateResolver {
```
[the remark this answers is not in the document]
bcronin90: It would probably be best to move the PEMUtil upwards and use it in both modules. Alternatively I can use the Azure-based method here too. Which would you prefer?

bcronin90:
> @bcronin90 to run a local instance of vault to be used for integration tests you can do (taken from the [documentation](https://hub.docker.com/_/vault))
> 
> ```shell
>  docker run  -e 'VAULT_DEV_ROOT_TOKEN_ID=myroot' -p "8200:8200" vault
> ```
> 
> (on the docs there are some details about a `--cap-add=IPC_LOCK` parameter, but on my machine it worked also without)
> 
> `http://localhost:8200` would be the url and `myroot` the token to be configured to connect
> 
> theoretically, the github service would need to be configured like this:
> 
> ```yaml
>     services:
>       hashicorp-vault:
>         image: vault
>         ports:
>           - 8200:8200
>         env:
>           VAULT_DEV_ROOT_TOKEN_ID: myroot
> ```
> 
> There is another problem, I noticed that, the `OkHttpClient` injection in the vault creates a cyclic dependency between the `CoreServicesExtension`, this could be a bug inside the injection loading, as in fact there would be no cyclic dependency, so my advice would be, momentarily, to have everything running, to avoid injecting the `OkHttpClient` and creating one from scratch leaving a TODO to be fixed later on.

Running a vault and interfacing it with the tests works fine. And the command I put into the github action definition is identical to that even. But doing it by hand works, while the github action fails. I'm not quite sure why. Currently it seems there is some dependency problem, but I'm not 100% sure yet.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpCertificateResolver.java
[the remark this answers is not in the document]
jimmarino: Hi @bcronin90 I would put PEMUtils in a library or abstract out the Certificate Resolver. @ndr-brt and @paullatzelsperger what do you think?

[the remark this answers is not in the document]
ndr-brt: I'd abstract `CertificateResolver` to have an unique instance like the `AzureCertificateResolver` that relies only on base java libraries (does not require bouncycastle), could be called `DefaultCertificateResolver` and be registered as default implementation for `CertificateResolver` and avoid `NoopCertificateResolver` at all.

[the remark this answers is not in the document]
bcronin90: Is the idea to have the bouncycastle dependency isolated to PEMUtil or to remove it entirely?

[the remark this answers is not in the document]
ndr-brt: @bcronin90 in my opinion for this task is unnecessary. The `AzureCertificateResolver` is not using it, it uses only base java classes.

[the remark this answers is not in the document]
bcronin90: I can rewrite the `HashicorpCertificateResolver` to use `CertificateFactory` like `AzureCertificateResolver` does. But I'd really prefer it the other way around. I think the BouncyCastle implementation is a bit less clunky, especially because it doesn't rely on manually replacing parts of a string. I'm open to either direction though.

extensions/hashicorp-vault/README.md
```diff
@@ -0,0 +1,9 @@
+# [HashiCorp Vault](https://www.vaultproject.io/) Extension
+
+## Configuration
```
ndr-brt, line 3: Please add a section that explains how to run a local vault instance to run the integration tests

extensions/hashicorp-vault/src/test/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultTest.java
```diff
@@ -0,0 +1,87 @@
...
+        vaultClient = Mockito.mock(HashicorpVaultClient.class);
+        final Monitor monitor = Mockito.mock(Monitor.class);
+        vault = new HashicorpVault(vaultClient, monitor);
+    }
+
+    @Test
+    void getSecret() {
+        // prepare
+        var value = UUID.randomUUID().toString();
+        var result = Mockito.mock(Result.class);
```
ndr-brt, line 44: `Result` is a data class, please don't mock it, it's easy to instantiate

ndr-brt (changes requested):
I don't agree with the bouncycastle over java plain certificate parser, as every dependency adds weight to the engine, potentially also transitive dependencies, since the plain java logic is really straightforward I would stick to that, let's see what the others think.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpCertificateResolver.java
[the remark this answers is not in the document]
bcronin90: I would prefer keeping the bouncycastle dependency. It appears, other packages already use it, so it's not a new dependency.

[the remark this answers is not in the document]
bcronin90: Could you have a look at this @juliapampus ? I completely agree with the rest of the review by @ndr-brt , but this detail is a bit problematic. I tried to sum it up in the above comment, but if you need any further info, please ask.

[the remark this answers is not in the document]
jimmarino: I agree with @ndr-brt - we should not use Bouncycastle and in general, should avoid unnecessary dependencies. I don't think the Azure approach is kludgey since the certificate format is standard.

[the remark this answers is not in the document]
jimmarino: Also, we should support a common format across Azure Vault and Hashicorp

jimmarino (changes requested):
Please see comments about removing Bouncycastle and aligning with Azure vault

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpCertificateResolver.java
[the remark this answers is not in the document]
bcronin90: Thanks for the input @jimmarino . I've removed the bouncycastle dependency and consolidated the certificate validation. I'll push it with the rest of the review implementation. I do however have some difficulty with removing bouncycastle as a test dependency. We rely on generated certificates for our tests and creating those with plain java is an extreme pain. So my proposition would be to keep the test dependency, as it isn't deployed with the project and thus doesn't bloat it.

extensions/hashicorp-vault/src/test/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpCertificateResolverTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @BeforeEach
+    void setup() {
+        vault = Mockito.mock(HashicorpVault.class);
+        final Monitor monitor = Mockito.mock(Monitor.class);
+        certificateResolver = new HashicorpCertificateResolver(vault, monitor);
+    }
+
+    @Test
+    void resolveCertificate() throws RuntimeException, IOException {
```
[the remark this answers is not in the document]
bcronin90: That would mean we have no unit tests for the certificate resolver at all. Is that okay?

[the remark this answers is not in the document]
ndr-brt: @bcronin90 in my opinion you could choose to test this in two ways:
- with an Integration Test, that will cover the client part to resolve a secret and parsing it to a certificate
- with an unit test, that mocks the vault response and also mocks a "pem parser" component that transforms a string in a certificate (and that's tested separately), and verifies eventual edge cases.

I would keep either one or the other, not both (probably I would go with the latter, since the Vault client it's already integration-tested by the `HashicorpVaultIntegrationTest`)

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpCertificateResolver.java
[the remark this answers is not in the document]
jimmarino: I don't think we should have the dependency. Please remove it. The creation of test certificates can be encapsulated with a test utility method that is shared across modules if needed.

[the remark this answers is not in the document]
bcronin90: I've already removed it.

common/util/src/main/java/org/eclipse/dataspaceconnector/common/security/PemUtil.java
```diff
@@ -0,0 +1,48 @@
...
+import java.security.cert.CertificateEncodingException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.Base64;
+
+public final class PemUtil {
+    private static final String HEADER = "-----BEGIN CERTIFICATE-----";
+    private static final String FOOTER = "-----END CERTIFICATE-----";
+
+    public static X509Certificate readX509Certificate(@NotNull String encoded) throws GeneralSecurityException {
```
ndr-brt, line 31: We usually avoid to use exceptions, better to catch it and throw an `EdcException` (that's unchecked).

ndr-brt (commented):
We are close to the top :)

common/util/src/main/java/org/eclipse/dataspaceconnector/common/security/PemUtil.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
+import java.security.GeneralSecurityException;
+import java.security.cert.CertificateEncodingException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.Base64;
+
+public final class PemUtil {
```
ndr-brt, line 27: Please document

extensions/hashicorp-vault/build.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+dependencies {
+    api(project(":spi:core-spi"))
+
+    implementation("com.squareup.okhttp3:okhttp:${okHttpVersion}")
+    implementation(project(":common:util"))
+    implementation("dev.failsafe:failsafe:${failsafeVersion}")
+
+    testImplementation(project(":extensions:junit"))
+    testImplementation(testFixtures(project(":common:util")))
+    testImplementation("org.mockito:mockito-inline:${mockitoVersion}")
```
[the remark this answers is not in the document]
bcronin90: I'm pretty sure we need mockito inline to mock static methods. Or what do you mean?

extensions/hashicorp-vault/src/test/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpCertificateResolverTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @BeforeEach
+    void setup() {
+        vault = Mockito.mock(HashicorpVault.class);
+        final Monitor monitor = Mockito.mock(Monitor.class);
+        certificateResolver = new HashicorpCertificateResolver(vault, monitor);
+    }
+
+    @Test
+    void resolveCertificate() throws RuntimeException, IOException, CertificateEncodingException {
```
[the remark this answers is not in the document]
ndr-brt: @bcronin90 The null check is covered by the test, but not the exception handling. Anyway I'm ok to leave it as it is.

extensions/hashicorp-vault/build.gradle.kts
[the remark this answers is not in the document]
ndr-brt: I removed it and everything worked.

ndr-brt (approved):

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultExtension.java
```diff
@@ -0,0 +1,96 @@
...
+                .withBackoff(minBackoff, maxBackoff, ChronoUnit.MILLIS)
+                .build();
+
+        var okHttpClient = new OkHttpClient.Builder().build();
+        var client = new HashicorpVaultClient(config, okHttpClient, context.getTypeManager(), retryPolicy);
+
+        var vault = new HashicorpVault(client, context.getMonitor());
+        var privateKeyResolver = new VaultPrivateKeyResolver(vault);
+
+        context.registerService(Vault.class, vault);
```
paullatzelsperger, line 72: registering services could be done using the `@Provider` method functionality, which is the preferred and recommended way. In that case, the `@Provides` annotation on the class can go away.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultExtension.java
```diff
@@ -0,0 +1,96 @@
...
+
+import java.time.temporal.ChronoUnit;
+
+@Provides({ Vault.class, PrivateKeyResolver.class })
+public class HashicorpVaultExtension implements ServiceExtension {
+
+    @EdcSetting(required = true)
+    public static final String VAULT_URL = "edc.vault.hashicorp.url";
+
+    @EdcSetting(required = true)
```
paullatzelsperger, line 36: `@EdcSetting` now has more attributes, such as descriptions, etc. They are used to auto-generate documentation - please use them.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultClient.java
```diff
@@ -0,0 +1,166 @@
...
+
+    Result<String> getSecretValue(@NotNull String key) {
+        var requestUri = getSecretUrl(key, VAULT_SECRET_DATA_PATH);
+        var headers = getHeaders();
+        var request = new Request.Builder().url(requestUri).headers(headers).get().build();
+
+        try (var response = with(retryPolicy).get(() -> okHttpClient.newCall(request).execute())) {
+
+            if (response.isSuccessful()) {
+                if (response.code() == 404) {
```
paullatzelsperger, line 70: use a constant please

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultClient.java
```diff
@@ -0,0 +1,166 @@
...
+            } else {
+                return Result.failure(String.format(CALL_UNSUCCESSFUL_ERROR_TEMPLATE, response.code()));
+            }
+
+        } catch (IOException e) {
+            return Result.failure(e.getMessage());
+        }
+    }
+
+    Result<HashicorpVaultCreateEntryResponsePayload> setSecret(
```
paullatzelsperger, line 92: I cannot see a reason for this to be package-private. Even if the class is package-private, the methods should either be `public` and be tested, or be `private` to indicate clearly what the intention was.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultClient.java
```diff
@@ -0,0 +1,166 @@
...
+                return Result.success(responsePayload);
+            } else {
+                return Result.failure(String.format(CALL_UNSUCCESSFUL_ERROR_TEMPLATE, response.code()));
+            }
+        } catch (IOException e) {
+            return Result.failure(e.getMessage());
+        }
+    }
+
+    Result<Void> destroySecret(@NotNull String key) {
```
paullatzelsperger, line 121: this definitely should be private

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultCreateEntryRequestPayload.java
```diff
@@ -0,0 +1,77 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+import java.util.Map;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonDeserialize(builder = HashicorpVaultCreateEntryRequestPayload.Builder.class)
+class HashicorpVaultCreateEntryRequestPayload {
+
+    @JsonProperty("options")
```
paullatzelsperger, line 29: nit: this is superfluous

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultCreateEntryRequestPayload.java
```diff
@@ -0,0 +1,77 @@
...
+    public Map<String, String> getData() {
+        return this.data;
+    }
+
+    @JsonPOJOBuilder(withPrefix = "")
+    public static class Builder {
+        private HashicorpVaultCreateEntryRequestPayloadOptions options;
+        private Map<String, String> data;
+
+        Builder() {
```
paullatzelsperger, line 53: private

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultCreateEntryRequestPayload.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        @JsonProperty("data")
+        public Builder data(Map<String, String> data) {
+            this.data = data;
+            return this;
+        }
+
+        public HashicorpVaultCreateEntryRequestPayload build() {
+            return new HashicorpVaultCreateEntryRequestPayload(options, data);
```
paullatzelsperger, line 74: throughout the entire codebase we use builders that use a single field. please adapt that style

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultCreateEntryRequestPayloadOptions.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonDeserialize(builder = HashicorpVaultCreateEntryRequestPayloadOptions.Builder.class)
+class HashicorpVaultCreateEntryRequestPayloadOptions {
```
paullatzelsperger, line 25: see my comments in `...RequestPayload`

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultCreateEntryResponsePayload.java
```diff
@@ -0,0 +1,63 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonDeserialize(builder = HashicorpVaultCreateEntryResponsePayload.Builder.class)
+class HashicorpVaultCreateEntryResponsePayload {
```
paullatzelsperger, line 25: see my comments in `...RequestPayload`

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultGetEntryResponsePayload.java
```diff
@@ -0,0 +1,63 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonDeserialize(builder = HashicorpVaultGetEntryResponsePayload.Builder.class)
+class HashicorpVaultGetEntryResponsePayload {
```
paullatzelsperger, line 25: see my comments in ...RequestPayload

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultCreateEntryRequestPayload.java
```diff
@@ -0,0 +1,77 @@
...
+
+        Builder() {
+        }
+
+        @JsonCreator
+        public static Builder newInstance() {
+            return new Builder();
+        }
+
+        @JsonProperty("options")
```
paullatzelsperger, line 61: why is this needed?

extensions/hashicorp-vault/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+
+plugins {
+    `java-library`
+}
+
+dependencies {
+    api(project(":spi:core-spi"))
+
+    implementation("com.squareup.okhttp3:okhttp:${okHttpVersion}")
+    implementation(project(":common:util"))
```
paullatzelsperger, line 26: nit: internal dependencies first

extensions/hashicorp-vault/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+    implementation("dev.failsafe:failsafe:${failsafeVersion}")
```
paullatzelsperger, line 27: not needed, as `RetryPolicy` should be provided by one of the core modules

paullatzelsperger (changes requested):
Comments inline. 

Also, test coverage is lacking: for example there is no test verifying the correct serialization/deserializetion of all the Hashicorp API objects.

Also, I would avoid prefixing all the classnames with `Hashicorp`, since they are internal classes only and it's clear by their package declaration what module they belong to.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultExtension.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        var config = loadHashicorpVaultClientConfig(context);
+
+        var maxRetries = context.getSetting(MAX_RETRIES, 5);
+        var minBackoff = context.getSetting(BACKOFF_MIN_MILLIS, 500);
+        var maxBackoff = context.getSetting(BACKOFF_MAX_MILLIS, 10_000);
+
+        var retryPolicy = RetryPolicy.builder()
```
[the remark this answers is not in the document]
bcronin90: That would lead to a cyclic dependency: `Cyclic extension dependency for [org.eclipse.dataspaceconnector.core.CoreServicesExtension@7e7b159b]`

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultExtension.java
```diff
@@ -0,0 +1,96 @@
...
+                .withMaxRetries(maxRetries)
+                .withBackoff(minBackoff, maxBackoff, ChronoUnit.MILLIS)
+                .build();
+
+        var okHttpClient = new OkHttpClient.Builder().build();
```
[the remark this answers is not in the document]
bcronin90: Same here.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultClient.java
bcronin90: These methods were tested in previous versions of these had unit tests. But it was decided that testing them via integration tests was enough. If this stance has changed, we can bring those tests back, but it would be great if the code owners could align their requirements here.

As to them not being private: If they were, we couldn't use them.

extensions/hashicorp-vault/build.gradle.kts
bcronin90: That doesn't seem to be the case.

paullatzelsperger: Just verified - it works. pick either one of these variants:
```java
@Inject private RetryPolicy retrypolicy;
@Inject private RetryPolicy<?> retryPolicy;
@Inject private RetryPolicy<Object> retryPolicy;
```
and add it as a class member to the extension.

paullatzelsperger: mmh, you're right. that needs to be fixed, since the `CoreServicesExtension` provides the `RetryPolicy`.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultClient.java
paullatzelsperger: I understand that the intention was to make methods package-private so that other classes have access to them, but my point was not to make them _private_, but make them _public_. 

Simply to make the intent clear to any future reader, that these methods should be called by collaborators.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultExtension.java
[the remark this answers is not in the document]
paullatzelsperger: unfortunately, you're right. #1765 should fix this: the `vault` could just be instantiated in the `prepare()` phase then. 
Could you add a comment mentioning the cyclic dependency?

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultExtension.java
paullatzelsperger: almost there. the `@Provides` annotation on the extension class should be removed as well.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultExtension.java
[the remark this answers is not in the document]
paullatzelsperger: I created #1775 to combat the cyclic dependency issue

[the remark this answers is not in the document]
bcronin90: Maybe I'm misunderstanding something here. I rebased onto main and as I understand it, the `okHttpClient` is supposed to be provided by `core:base`, but it just isn't. If I add the functionality there to provide it, I get another cyclical dependency. I assume that's now between HashicorpVault -> core:base -> common:util -> HashicorpVault.

But I might just be misunderstanding the design here. If so, please help me understand what is still to do here.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultExtension.java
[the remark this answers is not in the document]
bcronin90: Maybe I'm misunderstanding something here. I rebased onto main and as I understand it, the `okHttpClient` is supposed to be provided by `core:base`, but it just isn't. If I add the functionality there to provide it, I get another cyclical dependency. I assume that's now between HashicorpVault -> core:base -> common:util -> HashicorpVault.

But I might just be misunderstanding the design here. If so, please help me understand what is still to do here.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultExtension.java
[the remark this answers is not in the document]
bcronin90: Injecting `RetryPolicy` works now, `OkHttpClient` does not. See below for further comment.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultExtension.java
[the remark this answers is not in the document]
bcronin90: Since this is resolved, I assume it's okay to use our own OkHttpClient for now?

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/CreateEntryRequestPayloadOptions.java
```diff
@@ -0,0 +1,58 @@
...
+
+package org.eclipse.dataspaceconnector.core.security.hashicorpvault;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
paullatzelsperger, line 23: If the `ObjectMapper` instantiated by the `TypeManager` is used (which it should), then this annotation is superfluous. cf. `Typemanager.java#L50`
Same goes for all the other payload classes.

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/CreateEntryRequestPayloadOptions.java
```diff
@@ -0,0 +1,58 @@
...
+@JsonDeserialize(builder = CreateEntryRequestPayloadOptions.Builder.class)
+class CreateEntryRequestPayloadOptions {
+    @JsonProperty()
```
paullatzelsperger, line 26: needed? Same goes for all the other payload classes.

extensions/hashicorp-vault/src/test/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultTest.java
```diff
@@ -0,0 +1,77 @@
...
+    @Test
+    void getSecret() {
+        var value = UUID.randomUUID().toString();
+        var result = Result.success(value);
+        when(vaultClient.getSecretValue(KEY)).thenReturn(result);
+
+        var returnValue = vault.resolveSecret(KEY);
+
+        verify(vaultClient, times(1)).getSecretValue(KEY);
+        Assertions.assertEquals(value, returnValue);
```
paullatzelsperger, line 52: nit: static imports, and use AssertJ (`assertThat(xyz).isNull()`) instead for consistency

extensions/hashicorp-vault/src/test/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultIntegrationTest.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final String VAULT_TEST_TOKEN = "test-token";
+
+    @BeforeEach
+    final void setup(EdcExtension extension) {
+        extension.setConfiguration(Map.of(VAULT_URL, VAULT_TEST_URL, VAULT_TOKEN, VAULT_TEST_TOKEN));
+    }
+
+    @Test
+    @DisplayName("Resolve a secret that exists")
+    void testResolveSecret_exists(Vault vault) {
```
paullatzelsperger, line 47: nit: do not use the `test`-prefix

extensions/hashicorp-vault/src/test/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultIntegrationTest.java
```diff
@@ -0,0 +1,107 @@
...
+        var key = UUID.randomUUID().toString();
+        var valueExpected = UUID.randomUUID().toString();
+
+        vault.storeSecret(key, valueExpected);
+        var secretValue = vault.resolveSecret(key);
+        Assertions.assertEquals(valueExpected, secretValue);
```
paullatzelsperger, line 53: nit: static imports, and use AssertJ (`assertThat(xyz).isNull()`) instead for consistency

extensions/hashicorp-vault/src/main/java/org/eclipse/dataspaceconnector/core/security/hashicorpvault/HashicorpVaultExtension.java
```diff
@@ -0,0 +1,94 @@
...
+public class HashicorpVaultExtension implements ServiceExtension {
+
+    @EdcSetting(value = "The URL of the Hashicorp Vault", required = true)
+    public static final String VAULT_URL = "edc.vault.hashicorp.url";
+
+    @EdcSetting(value = "The token used to access the Hashicorp Vault", required = true)
+    public static final String VAULT_TOKEN = "edc.vault.hashicorp.token";
+
+    @Inject
+    public OkHttpClient okHttpClient;
```
paullatzelsperger, line 40: why `public`?

paullatzelsperger (changes requested):
couple of minor things, commented directly in code.

ndr-brt:
@paullatzelsperger I don't really agree on removing the `Hashicorp` prefix from the class names, if we'd apply this to the rest of the project it will make all the project more hard to read and to go through (imagine having 10 `Client` classes...). 
Personally I would keep it as it is for all the other components.

paullatzelsperger:
> @paullatzelsperger I don't really agree on removing the Hashicorp prefix from the class names, if we'd apply this to the rest of the project it will make all the project more hard to read and to go through (imagine having 10 Client classes...).
Personally I would keep it as it is for all the other components.

I was mostly referring to the `...Payload` dto classes, because there it made the class name very long and unwieldy, and since they're internal/package private classes I didn't think it too much of a problem, but we can definitely re-add that prefix to `Client` and `Config`

paullatzelsperger (approved):

ndr-brt (commented):
Ok, so @bcronin90 if you could re-add the `Hashicorp` prefix on the `Client` and `Config` class, for me we are good to go :rocket:

ndr-brt:
Let's get this merged! :champagne:

PR #1539, opened 2022-06-24
Fix: do not use `QuerySpec` when SqlContractDefinitionStore#findAll()

### LINKED ISSUE #1515: SQL:  Connector sends out 50 contract offers max.
# Bug Report

## Describe the Bug

When a connector requests a catalog from another connector, it only gets the first 50 contract offers, even if the connector would be able to send out a lot more offers.

### Expected Behavior
- all contract offers can be accessed by another connector

### Observed Behavior
- another connector can only access the first 50 contract offers

## Steps to Reproduce
1. Setup Connector with SQL persistence
2. Setup the data so that the connector would generate more than 50 offers
3. Request catalog with another connector
  -> catalog contains only 50 offers 

## Context Information

- Requires SQL Extensions on offer provider side

## Detailed Description
-

## Possible Implementation
-

paullatzelsperger:
## What this PR changes/adds

Does not use a `QuerySpec` when executing `SqlContractDefinitionStore#findAll`, instead execute a plain `SELECT * FROM...`.

## Why it does that

all = no limits!

## Linked Issue(s)

Closes #1515 

## Checklist

### DIFF
--- a/extensions/sql/contract-definition-store-sql/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
+++ b/extensions/sql/contract-definition-store-sql/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
@@ -36,10 +36,10 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.sql.DataSource;
 
+import static java.lang.String.format;
 import static org.eclipse.dataspaceconnector.sql.SqlQueryExecutor.executeQuery;
 
 public class SqlContractDefinitionStore implements ContractDefinitionStore {
@@ -69,7 +69,15 @@ public class SqlContractDefinitionStore implements ContractDefinitionStore {
 
     @Override
     public @NotNull Collection<ContractDefinition> findAll() {
-        return findAll(QuerySpec.none()).collect(Collectors.toList());
+        return transactionContext.execute(() -> {
+
+            try (var connection = getConnection()) {
+                var queryStmt = format("SELECT * FROM %s", statements.getContractDefinitionTable());
+                return executeQuery(connection, this::mapResultSet, queryStmt);
+            } catch (SQLException exception) {
+                throw new EdcPersistenceException(exception);
+            }
+        });
     }
 
     @Override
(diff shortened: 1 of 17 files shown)


extensions/sql/contract-definition-store-sql/src/main/java/org/eclipse/dataspaceconnector/sql/contractdefinition/store/SqlContractDefinitionStore.java
```diff
@@ -69,7 +69,15 @@ ContractDefinition mapResultSet(ResultSet resultSet) throws Exception {
 
     @Override
     public @NotNull Collection<ContractDefinition> findAll() {
-        return findAll(QuerySpec.none()).collect(Collectors.toList());
+        return transactionContext.execute(() -> {
```
ndr-brt, line 72: I would make this more explicit as it was done in the `SqlAssetIndex` at line 132, passing to the other `findAll` method a `QuerySpec` instance with an insanely high `limit` value.
In this way the thing will be explicit (as I would expect this to change in the future), there will be no duplication and no SQL syntax will overflow in the store implementation

paullatzelsperger: There isn't really any duplication, there is just two (slightly) different SQL statements. In reality those two approaches would have the same effect, and once we have true pagination in IDS, we can remove the parameterless `findAll()` method completely.

What do you mean by "SQL syntax overflow"?

ndr-brt: having a string with `SELECT something FROM somewhere` in the `Store`/`Index`, we never did that anywhere else, all the SQL stuff is in the `Statements` class

paullatzelsperger: ah got it. How about we remove `findAll()` altogether then, and provide a static method `QuerySpec.max()` that creates a QuerySpec with `limit = Integer.MAX_VALUE`?

paullatzelsperger: [this commit](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/1539/commits/a73fa531f3b37a7175912eb5d1f1898514a3dc7b) removes the parameterless `findAll()`. WDYT?

ronjaquensel (approved):

ndr-brt (approved):
Totally agree on `findAll()`'s inline. Good for me then

PR #1565, opened 2022-06-29
Samples: add automated test to sample 04.0-file-transfer

### LINKED ISSUE #1269: Automated Test for Sample 04.0 (File Transfer)
# Feature Request

Add automated integration tests that run in CI for samples to prevent the
samples breaking due to changes in the EDC.

The scope of this story is to add automated integration tests for sample 04.0 (file transfer).

CI, samples

- Prevent the samples breaking due to changes in the EDC.
- Ensure samples can be followed as described.

## Solution Proposal

Provide an integration test using Rest Assured which is aligned to the steps of
the sample readme.

## Type of Issue

testing / reliability

## Checklist

pkirch:
## What this PR changes/adds

Automated integration tests that run in CI should be added for samples to prevent the samples breaking due to changes in the EDC. The scope of this story is to add automated integration tests for sample 04.0.

* added test class `FileTransferSampleTest.java`
* updated `samples/04.0-file-transfer/README.md` with missing sample output

## Why it does that

Every step from [sample readme](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/samples/04.0-file-transfer/README.md) is reproduced in the automated tests to reduce the risk from sample guidance diverging from actual code base.

## Further notes

* sample file `samples/04.0-file-transfer/filetransfer.json` was missing `edctype` set to `dataspaceconnector:datarequest`
* changed `FsConfigurationExtension.java` to allow multiple `EdcRuntimeExtension` instances which load configuration from filesystem
  * Adapted `AzureDataFactoryTransferIntegrationTest.java` to replicate previous behavior of `FsConfigurationExtension.java`. Without this change the check `Test Code (Style, Tests) / Azure-Cloud-Integration-Test` would fail.
* no update to changelog as these changes apply to samples and other fixes seem minor

## Linked Issue(s)

Closes #1269

## Checklist


paullatzelsperger (commented):
I don't understand why the changes to `DataRequest` and `DataAddress` are necessary. The changes also modifies behaviour (e.g. `properties`).

pkirch:
@paullatzelsperger wrote:
> I don't understand why the changes to `DataRequest` and `DataAddress` are necessary. The changes also modifies behaviour (e.g. `properties`).

The changes in `DataRequest` and `DataAddress` follow the builder pattern as used in other classes in `core-spi` to enable immutable copies with changed values. There is a need in the test in `readAndUpdateTransferJsonFile` to read in `DataRequest` and `DataAddress` and write back a copy with changed values. Using this approach, we reuse the existing model and verify the given `filetransfer.json` file of the sample is still compatible with the current EDC. E.g. doing it like this surfaced already that `filetransfer.json` was missing `"edctype": "dataspaceconnector:datarequest"`. Makes sense?

With changed behaviour you mean the use of `address.properties.putAll(properties)` instead `properties.forEach(this::property)`? That's a good catch! Indeed this "optimization" misses the null check now. I'll change that.

Apart from that I don't see changed behavior. Let me know if you see more.

pkirch:
> We won't add code to a class simply to make instantiating it in tests easier. If copying a data class is desired, then please add a utility method to that test.

I see the point in scope creep here. I have moved the copy code into the test. 

However, I have kept a fix and additional unit tests for `DataRequest` as there has been a shallow copy before and made a `DataRequest` instance mutable. If this should not be part of this PR, please let me know.

> I'm not talking about nullchecks or optimizations. `address.properties.putAll(properties)` _adds_ all properties (potentially replacing some), whereas before the properties were entirely _replaced_:
> 
> ```java
> public Builder properties(Map<String, String> value) {
>     request.properties = value;
>     return this;
> }
> ```
> 
> I would say that's a definite change in behaviour that is not desired at this time.

Indeed, this change in behavior was not intended. Reverted.

paullatzelsperger:
One more time: we do **not** need/want the `toBuilder()`, `copy()` methods at this time. Either revert those changes or retract this PR.

If anything, `DataRequest.copy()` can be removed, as it is not used.

pkirch:
> One more time: we do **not** need/want the `toBuilder()`, `copy()` methods at this time.

Newly introduced `toBuilder()` and `copy()` methods have been already removed in the last push; no more to revert here.

>  Either revert those changes or retract this PR.

I have removed the fix from `DataRequest` and the regression tests from `DataRequestTest`. Those files are not part of this PR anymore.

paullatzelsperger (approved):

samples/04.0-file-transfer/integration-tests/src/test/java/org/eclipse/dataspaceconnector/extension/sample/test/FileTransferSampleTest.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /**
+     * Assert that a POST request to initiate transfer process is successful.
+     * This method corresponds to the command in the sample: {@code curl -X POST -H "Content-Type: application/json" -H "X-Api-Key: password" -d @samples/04.0-file-transfer/filetransfer.json "http://localhost:9192/api/v1/data/transferprocess"}
+     *
+     * @throws IOException Thrown if there was an error accessing the transfer request file defined in {@link FileTransferSampleTest#TRANSFER_FILE_PATH}.
+     */
+    void requestTransferFile() throws IOException {
+        File transferJsonFile = getFileFromRelativePath(TRANSFER_FILE_PATH);
```
ndr-brt, line 182: Please use `var`

ndr-brt (approved):

PR #1569, opened 2022-06-30
feat: provide additional policy scopes

### LINKED ISSUE #1331: Contract policy evaluated during catalog request
# Bug Report

## Describe the Bug
When the connector responds to a catalog request, it checks both the access and the contract policies for each `ContractDefinition`. Therefore, the requesting connector will not see contract offers if the contract policy is not fulfilled at the time of the request.

### Expected Behavior
During catalog request, the connector evaluates whether the requesting agent fulfils a `ContractDefinition's` access policy, but not the contract policy, as the contract policy should be advertised to requesting agents.

### Observed Behavior
The connector evaluates whether the access and contract policies are fulfilled.

## Steps to Reproduce
1. Create an implementation of  `AtomicConstraintFunction<Permission>` that always returns false in sample 04.0's `transfer-file` module and register it with the `PolicyEngine` in the `TransferFileExtension`.
2. Bind the key used to register the function as well as the action type `USE` to all scopes using the `RuleBindingRegistry`.
3. Create a second policy in the `TransferFileExtension`, adding an action with type `USE` and a constraint using the key from steps 1 and 2 as the value for the left operand. Operator and right operand can be chosen arbitrarily.
4. Add the new policy as the `ContractDefinition's` contract policy, do not add it as the access policy.
5. Start both connectors as described in the sample's `README`.
6. Make a catalog request from consumer to provider and see an empty array of contract offers in the response: `curl -X GET -H 'X-Api-Key: password' "http://localhost:9192/api/v1/data/catalog?providerUrl=http://localhost:8282/api/v1/ids/data"`
7. See in the provider logs that there was a problem evaluating the constraint defined in the contract policy.

## Possible Implementation
The `ContractDefinitionServiceImpl` always evaluates access and contract policies when filtering `ContractDefinitions` for a `ParticipantAgent`. Here, only the access policies should be evaluated. The evaluation of the contract policies (e.g. during contract negotiation) should happen separately.

Also, the `ContractDefinitionServiceImpl` always uses the policy scope `contract.negotiation`, even though the service is also used during catalog requests.

### LINKED ISSUE #1446: Policy enforcement: definition of policy scopes
# Feature Request

Policy scopes need to be defined in the existing core modules and extensions, as policy enforcement needs to happen in different runtime contexts. 

core, DPF, extensions providing provisioning or transfer functionalities

In the existing core modules and extensions, policy scopes are missing. Currently, there is only the `contract.negotiation` scope defined. Accordingly, no policies are evaluated outside of the contract negotiation.

## Type of Issue

improvement

## Checklist

- [X] assigned appropriate label?

ronjaquensel:
## What this PR changes/adds

It introduces two new policy scopes. The first one is `contract.cataloging`, which is used during the cataloging phase. The previously existing scope `contract.negotiation` is now used during the negotiation itself and to re-evaluate the policies before a new transfer process is created. The other new scope is `provision.manifest.verify`, which is used to evaluate a generated `ResourceManifest` to ensure that it fulfils the given policy.

For the purpose of evaluating `ResourceManifests`, also two new function interfaces are introduced: `ResourceDefinitionRuleFunction` and `ResourceDefinitionAtomicConstraintFunction`. These are analogous to the existing function interfaces, but are bound to a specific sub-type of `ResourceDefinition` and can return a modified version of that definition.

## Why it does that

Currently, policy evaluation only takes place during the negotiation phase.

## Further notes

The interfaces for the `PolicyEngine` and the evaluation functions are moved to the `policy-spi` and the function interfaces from the `policy-evaluator` are moved to `policy-engine`, to avoid circular dependencies between `policy-evaluator`, `core-spi` and `transfer-spi`.

## Linked Issue(s)

Closes #1446 
Closes #1331

## Checklist


juliapampus (commented):
You wrote
> The first one is contract.offer-generation, which is used during the cataloging phase. 

Why not call it `contract.cataloging` then? I am not sure whether there could be also policy enforcement during the `contract.offering` phase, that may need to be separated in the future. Or call it `contract.offer.generation` if we may have subtypes of `contract.offer.**`? I know this is hard to decide...

On top of that, I would detail `provisioning`. Cause it's resource provisioning, we could name it `resource.provisioning`? If we need to separate "resource" actions. Or `provisioning.resource`.. don't know if this is too much at the moment but we should not block expandability.

ronjaquensel:
@juliapampus you're right, they should be renamed.

> Why not call it `contract.cataloging` then? I am not sure whether there could be also policy enforcement during the `contract.offering` phase, that may need to be separated in the future. Or call it `contract.offer.generation` if we may have subtypes of `contract.offer.**`? I know this is hard to decide...

Probably `contract.cataloging` would be the better choice, as the evaluation is technically done in the `ContractDefinitionService`, not the `ContractOfferService`. And `cataloging` would also imply that only the access policies, but not the contract policies are checked.

> On top of that, I would detail `provisioning`. Cause it's resource provisioning, we could name it `resource.provisioning`? If we need to separate "resource" actions. Or `provisioning.resource`.. don't know if this is too much at the moment but we should not block expandability.

As the evaluation is done after the `ResourceManifest` has been created but before the actual provisioning happens, how about `provision.manifest.verify` or something similar?

PR #1572, opened 2022-07-01
fix(sql): fixed JSON deserialization issue in transfer-process-store-sql extension

schoenenberg:
## What this PR changes/adds

We discovered an issue during the query of the transfer-process-store-sql, which prevented the extension to successfully query transfer-processes. This PR fixes that issue.

## Why it does that

The `@JsonPOJOBuilder` annotation requires by default builder methods in the style `withId`. This behaviour can be overridden with `@JsonPOJOBuilder(withPrefix = "")`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ in the detailed section referring to by linking pull requests or issues.
 * Fix race condition in `ContractNegotiationIntegrationTest` (#1505)
 * Fix for change in Cosmos DB behavior on missing sort fields (#1514)
 * Effectively removed default LIMIT in SQL Contract Def Store (#1515)
+* Fix for JSON Deserialization of `ResourceDefinition` and `ProvisionedResource` (#1572)
 
 ## [milestone-4] - 2022-06-07
 
--- a/extensions/sql/transfer-process-store-sql/src/test/java/org/eclipse/dataspaceconnector/sql/transferprocess/store/postgres/PostgresTransferProcessStoreTest.java
+++ b/extensions/sql/transfer-process-store-sql/src/test/java/org/eclipse/dataspaceconnector/sql/transferprocess/store/postgres/PostgresTransferProcessStoreTest.java
@@ -632,7 +632,6 @@ class PostgresTransferProcessStoreTest {
     @JsonDeserialize(builder = TestResourceDef.Builder.class)
     static class TestResourceDef extends ResourceDefinition {
 
-        @JsonPOJOBuilder(withPrefix = "")
         public static class Builder extends ResourceDefinition.Builder<TestResourceDef, Builder> {
             private Builder() {
                 super(new TestResourceDef());
@@ -649,7 +648,6 @@ class PostgresTransferProcessStoreTest {
     @JsonTypeName("dataspaceconnector:testprovisionedresource")
     static class TestProvisionedResource extends ProvisionedResource {
 
-        @JsonPOJOBuilder(withPrefix = "")
         public static class Builder extends ProvisionedResource.Builder<TestProvisionedResource, Builder> {
             private Builder() {
                 super(new TestProvisionedResource());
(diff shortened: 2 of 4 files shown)


schoenenberg:
Please add the `no-changelog` label

schoenenberg:
@paullatzelsperger updated tests and added Changelog entry.

Next time please do not alter my comments on GitHub and just answer in the PR.. I got really strange notifications because of that.

paullatzelsperger:
> Next time please do not alter my comments on GitHub and just answer in the PR.. I got really strange notifications because of that.

Sorry, I did not do that on purpose!

paullatzelsperger (changes requested):
what was the exact problem you encountered - could you describe the problem in detail? That should be covered in a linked (bug) issue, but since there was none (which is a violation of our contribution rules BTW), we'll clarify it here.

Also, there still is no test to verify the correct behaviour, that still needs to be added.

schoenenberg:
> what was the exact problem you encountered - could you describe the problem in detail? That should be covered in a linked (bug) issue, but since there was none (which is a violation of our contribution rules BTW), we'll clarify it here.

First, I am sorry to have violated the contribution rules. It was not my intention to violate them!

Second, the problem in detail: We have developed our own *Provisioning Extension* for the EDC and are in the process of setting up persistency and other stuff to be able to use it in pilot environments. For that we decided to persist EDC state with the existing Postgres extension. There we discovered the bug, that the EDC is not able to query the current state from the *transfer-process-store*. 

Our `ResourceDefinition` looked basically like this (just with some additional properties): https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/dd821c64f6a4b025e39cc3a651f3fb0540e3e23a/extensions/sql/transfer-process-store-sql/src/test/java/org/eclipse/dataspaceconnector/sql/transferprocess/store/postgres/PostgresTransferProcessStoreTest.java#L632-L644

I found out that the `@JsonPOJOBuilder` annotation requires a `withPrefix = ""` attribute to be able to use builder methods like `id`, `name`, ... instead of `withId`, `withName`, etc. At that point there were two solutions possible:
1. Add `@JsonPOJOBuilder(withPrefix = "")` to my implementation of `ResourceDefinition.Builder` to override the annotation from the extended `ResourceDefinition.Builder`.
2. Fix the abstract `ResourceDefinition.Builder` by replacing `@JsonPOJOBuilder` with `@JsonPOJOBuilder(withPrefix = "")`.

I decided to do 2. because of the following reasons:
- `ResourceDefinition.Builder` defines already two Builder-methods for `id` and `transferProcessId` in the `withPrefix = ""`-style. See here: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/dd821c64f6a4b025e39cc3a651f3fb0540e3e23a/spi/transfer-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/transfer/ResourceDefinition.java#L53-L61
- Already existing implementations are also missing the Annotation, e.g. https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/8004b0faa3135d7273785338a5746d67e1311726/extensions/aws/s3/s3-provision/src/main/java/org/eclipse/dataspaceconnector/aws/s3/provision/S3BucketResourceDefinition.java#L41-L68
- It took me some time to find the origin of the bug. To prevent others from having the same issue, I prefer to fix it at the root. 

While looking through the existing tests, I found that `ProvisionedResource.Builder` had the same issue. So I also added the fix for that.

> Also, there still is no test to verify the correct behaviour, that still needs to be added.

In my opinion the tests for that exist already, but the instances of `ResourceDefinition.Builder` and `ProvisionedResource.Builder` implement the fix I proposed in 1: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/dd821c64f6a4b025e39cc3a651f3fb0540e3e23a/extensions/sql/transfer-process-store-sql/src/test/java/org/eclipse/dataspaceconnector/sql/transferprocess/store/postgres/PostgresTransferProcessStoreTest.java#L630-L660 So removing the annotation on both results in a positive test. Reverting my fix would let those tests fail.

I hope you could follow me in this lengthy explanation. If you have further questions, please let me know.

paullatzelsperger:
I don't dispute the validity of the fix, in fact, the annotation value should be on the abstract class, and implementors can then override. 

What I meant was: there should be two test cases:
- one, where there is a subclass of both `ProvisionedResource` and `ResourceDefinition`, but _without_ the annotation present (so that it gets inherited). This test serializes, and deserializes using the `TypeManager` using appropriate assertions.
- another one, e.g. in the `SqlTransferProcessStoreTest`, which replicates the exact bug you encountered (I'm assuming it has to do with storing a `Transferprocess` that either has a `ResourceDefinition`, or a `PRovisionedResource` or both)

schoenenberg:
Working on it this week..

PR #1582, opened 2022-07-04
docs: add decision record for serialization contexts

juliapampus:
## What this PR changes/adds

Adds a decision record for issue #1581.

## Why it does that

Document thoughts and concepts.

## Further notes

PR containing the implementation will follow.

## Linked Issue(s)

Related #1581 

## Checklist

- [ ] added appropriate tests?
- [ ] performed checkstyle check locally?
- [ ] added/updated copyright headers?
- [ ] documented public classes/methods?
- [ ] added/updated relevant documentation?
- [ ] added relevant details to the changelog? (_skip with label `no-changelog`_)
- [ ] formatted title correctly? (_take a look at the [CONTRIBUTING](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/CONTRIBUTING.md#submit-a-pull-request) and [styleguide](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/styleguide.md) for details_)

PR #1586, opened 2022-07-04
feat: enhance TypeManager to support multiple serialization contexts

### LINKED ISSUE #1581: Support configuration of multiple serialization contexts
# Feature Request

Write-up provided by @jimmarino. 

The EDC runtime requires the ability to support configuration of multiple serialization contexts.

`TypeManager`

For example, an API may require different serialization characteristics than another API. Related to this, users may want to serialize additional metadata such as JSON-LD for certain types such as extensible Asset properties or IDS messages.

## Solution Proposal

The `TypeManager` will be extended to support multiple serialization contexts using the following approach:

1. A default context will define base configuration and be inherited by all other contexts. The existing TypeManager.getObjectMapper() will return the default context.

2. Serialization contexts will be created by invoking TypeManager.getObjectMapper(String). If a context is not already present, one will be created; otherwise the existing one will be returned. The returned `ObjectMapper` can have configuration applied to it during runtime initialization. 

3. The `TypeManager` will track created contexts and `ObjectMapper` instances. This will allow type registrations with the default context to be propagated to other `ObjectMapers` after they have been created.

4. The use of IDS and API contexts should be introduced.

### The TypeManager

The `TypeManager` will have the following methods added:

```

    /**
     * Returns the object mapper for the given serialization context, creating one based on the default mapper if required.
     */
    @NotNull
    public ObjectMapper getMapper(String key) {
        // ...
    }

    /**
     * Registers types with all contexts.
     */
    public void registerTypes(Class<?>... type) {
        // ...
    }

    /**
     * Registers types with all contexts.
     */
    public void registerTypes(NamedType... type) {
        // ...
    }

    /**
     * Registers types with a context.
     */
    public void registerTypes(String key, Class<?>... type) {
        // ...
    }

    /**
     * Registers types with a context.
     */
    public void registerTypes(String key, NamedType... type) {
        // ...
    }

    /**
     * Registers a serializer for the given type with a context.
     */
    public <T> void registerSerializer(String key, Class<T> type, JsonSerializer<T> serializer) {
        // ...
    }

    /**
     * Registers a serializer for the given type with the default context.
     */
    public <T> void registerSerializer(Class<T> type, JsonSerializer<T> serializer) {
        // ...
    }


```

### Registering a serializer

Serializers can be registered as follows:

```
    typeManager.registerSerializer("foo", Bar.class, new JsonSerializer<>() {
        @Override
        public void serialize(Bar value, JsonGenerator generator, SerializerProvider serializers) throws IOException {
            // handle serialization
        }
    });

    // ...

    var fooMapper = manager.getMapper("foo");
    var serialized = fooMapper.writeValueAsString(new Bar());
      

```

In the above example, a custom serializer is registered for the `Bar` type in the `foo` context.

### Writing a Decorating Serializer

A common use case for serialization contexts will be to decorate a serialized type with metadata. The following is an example of a serializer that can be registered with the `TypeManager` to add `@context` information for specific types. Note this example is user code and would __not__ be provided by the EDC:

```
    public class DecoratingSerializer<T> extends JsonSerializer<T> {
        private Class<T> type;

        public DecoratingSerializer(Class<T> type) {

            this.type = type;
        }

        public void serialize(Object value, JsonGenerator generator, SerializerProvider provider) throws IOException {
            generator.writeStartObject();
            var javaType = provider.constructType(type);
            var beanDescription = provider.getConfig().introspect(javaType);
            var staticTyping = provider.isEnabled(MapperFeature.USE_STATIC_TYPING);
            var serializer = BeanSerializerFactory.instance.findBeanOrAddOnSerializer(provider, javaType, beanDescription, staticTyping);
            serializer.unwrappingSerializer(null).serialize(value, generator, provider);
            generator.writeObjectField("@context", "some data");
            generator.writeEndObject();
        }
    }

```

The serializer can be registered for types as follows:

```
    manager.registerSerializer("foo", Bar.class, new DecoratingSerializer<>(Bar.class));
    manager.registerSerializer("foo", Baz.class, new DecoratingSerializer<>(Baz.class));
```

## Type of Issue
enhancement

## Checklist

juliapampus:
## What this PR changes/adds

Adds methods to register custom ObjectMappers and serialization contexts by type.

## Why it does that

Provide the possibility to support de/serialization of multiple contexts.

## Further notes

See decision record in #1582. Example will be provided with adding an IDS context (see #1541). 

## Linked Issue(s)

Closes #1581 

## Checklist


spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/TypeManager.java
```diff
@@ -16,53 +16,117 @@
...
 public class TypeManager {
     private final ObjectMapper objectMapper;
 
+    /**
+     * Concurrent support is not needed since this map is only populated a boot, which is single-threaded
+     */
+    private final Map<String, ObjectMapper> objectMappers = new HashMap<>();
+
+    /**
+     * Constructor without params.
```
paullatzelsperger, line 44: my inner monk w ould call this a "default constructor"

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/TypeManager.java
```diff
@@ -16,53 +16,117 @@
...
+    @NotNull
+    public ObjectMapper getMapper(String key) {
+        return objectMappers.computeIfAbsent(key, k -> objectMapper.copy());
+    }
+
+    /**
+     * Registers types with all contexts.
+     */
     public void registerTypes(Class<?>... type) {
         objectMapper.registerSubtypes(type);
```
paullatzelsperger, line 72: What do you think about adding the `objectMapper` to the `objectMappers`, for example using "`default`" as key? that would make line 72 and 80 obsolete.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/TypeManager.java
```diff
@@ -16,53 +16,117 @@
...
 
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * Manages system types and is used to deserialize polymorphic types.
  */
 public class TypeManager {
     private final ObjectMapper objectMapper;
```
paullatzelsperger, line 36: I'd rename `objectMapper` -> `defaultMapper`

paullatzelsperger (approved):

ronjaquensel (approved):

spi/core-spi/src/test/java/org/eclipse/dataspaceconnector/spi/types/TypeManagerTest.java
```diff
@@ -0,0 +1,124 @@
...
+import java.io.IOException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class TypeManagerTest {
+
+    @Test
+    void verifySerialization() throws JsonProcessingException {
+        var manager = new TypeManager();
+        manager.registerSerializer("foo", Bar.class, new JsonSerializer<>() {
```
ndr-brt, line 34: I'm late to the party but this test should be more strict as it succeed also if the serializer is not registered.

PR #1602, opened 2022-07-07
feat(build): automate release process

### LINKED ISSUE #1433: Build: improve build process and automation
# Feature Request

Currently, releasing a new version of EDC to MavenCentral involves a few steps:
- create a tag in git
- update the default version in build.gradle.kts
- update the changelog, add a new section
- create a GH release
- create a OSSRH snapshot (might be obsolete henceforth)
- create a release for MavenCentral by triggering a build in our JIPP instance

Whhile it is certainly workable, it would be nice to have some of these steps automated.

Build

Reduce error surface, make releases less manual

## Solution Proposal

- have a `release` branch in git
- have a GH action called "release version", that does the following:
  - user is prompted to enter a version string **OR**
  - automatically increase the version number
  - create a merge commit `main` -> `release` 
  - a tag is created on `main` with that version number
  - the merge-commit message also contains the version number
  - the JIPP instance is triggered to build and release that version
  - a GH release is created 

Updating the `CHANGELOG.md` and bumping the default version in `gradle.properties` would still have to be done manually _before_ triggering the `"release version"` action.

  
## Type of Issue
Improvement

## Checklist

paullatzelsperger:
## What this PR changes/adds

This PR introduces a new GitHub Actions workflow that automates the release process to OSSRH Staging and/or MavenCentral.

## Why it does that

To make the release process less manual and reduce the error surface. 

## Further notes

- The action is manual, only committers should be able to trigger it
- a valid SemVer must be entered
- The Jenkins publish action takes a long time and cannot be cancelled directly from GitHub.
- I haven't been able to test the auto-generated release notes, as I did not have any pull-requests in my fork

## Linked Issue(s)

Closes #1433 

## Checklist

### DIFF
--- a/.github/release.yaml
+++ b/.github/release.yaml
@@ -1,0 +1,16 @@
+changelog:
+  exclude:
+    labels:
+      - no-changelog
+    authors:
+      - dependabot[bot]
+  categories:
+    - title: Bugs
+      labels:
+        - bug
+    - title: New Features & Improvements
+      labels:
+        - "*"
+    - title: Documentation
+      labels:
+        - documentation
--- a/.github/workflows/release-edc.yml
+++ b/.github/workflows/release-edc.yml
@@ -1,0 +1,74 @@
+name: Create EDC Release
+on:
+  workflow_dispatch:
+    inputs:
+      edc_version:
+        description: 'Version string that is used for publishing. Appending -SNAPSHOT will create a snapshot release.'
+        required: true
+        type: string
+
+
+env:
+  EDC_VERSION: ${{ github.event.inputs.edc_version || inputs.edc_version }}
+
+jobs:
+  Prepare-Release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: releases
+
+      # create tag using Github's own API
+      - name: Create tag on current branch (main)
+        uses: actions/github-script@v5
+        with:
+          script: |
+            github.rest.git.createRef({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: 'refs/tags/v${{ env.EDC_VERSION }}',
+              sha: context.sha
+            })
+
+      # create merge commit
+      - name: Merge main -> releases
+        uses: everlytic/branch-merge@1.1.2
+        with:
+          github_token: ${{ github.token }}
+          source_ref: ${{ github.ref }}
+          target_branch: 'releases'
+          commit_message_template: 'Merge commit for release of version v${{ env.EDC_VERSION }}'
+
+      # Trigger EF Jenkins
+      - name: Trigger Release on EF Jenkins
+        uses: toptal/jenkins-job-trigger-action@master
+        with:
+          jenkins_url: "https://ci.eclipse.org/dataspaceconnector/"
+          jenkins_user: ${{ secrets.EF_JENKINS_USER }}
+          jenkins_token: ${{ secrets.EF_JENKINS_TOKEN }}
+          job_name: "EDC-Autobuild-Release"
+          job_params: |
+            {
+              "VERSION": "${{ env.EDC_VERSION }}"
+            }
+          job_timeout: "3600" # Default 30 sec. (optional)
+
+
+  Github-Release:
+    needs:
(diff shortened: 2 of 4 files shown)


.github/workflows/release-edc.yml
```diff
@@ -0,0 +1,76 @@
...
+jobs:
+  Prepare-Release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: releases
+
+      # create tag on the current branch using GitHub's own API
+      - name: Create tag on current branch (main)
```
ndr-brt, line 23: This confuses me, since at line 20 the selected branch seems to be "releases", not "main"

.github/workflows/release-edc.yml
```diff
@@ -0,0 +1,76 @@
...
+          jenkins_token: ${{ secrets.EF_JENKINS_TOKEN }}
+          job_name: "EDC-Autobuild-Release"
+          job_params: |
+            {
+              "VERSION": "${{ env.EDC_VERSION }}"
+            }
+          job_timeout: "3600" # Default 30 sec. (optional)
+
+
+  Github-Release:
```
ndr-brt, line 60: Would be possible to avoid this on SNAPSHOT releases?

.github/workflows/release-edc.yml
paullatzelsperger: correct. fixed.

.github/workflows/release-edc.yml
paullatzelsperger: We talking about "named" snapshots, e.g. `0.0.1-cxproduct-SNAPSHOT`. Technically they are snapshots, as the don't get published to MavenCentral, but in reality they will never get overwritten.

Rolling snapshots are not affected by this, they still get auto-created by periodically polling the repo, not GH Release is created for them and they overwrite one another.

How about making the Github Release job optional, and provide another boolean input in the prompt letting the user decide?

.github/workflows/release-edc.yml
```diff
@@ -0,0 +1,76 @@
+name: Create EDC Release
+on:
+  workflow_dispatch:
+    inputs:
+      edc_version:
+        description: 'Version string that is used for publishing. Appending -SNAPSHOT will create a snapshot release.'
```
juliapampus, line 6: Maybe add an example to remind people of the correct formatting?

```suggestion
        description: 'Version string that is used for publishing (e.g. "1.0.0", NOT "v1.0.0"). Appending -SNAPSHOT will create a snapshot release.'
```

.github/release.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - no-changelog
+    authors:
+      - dependabot[bot]
+  categories:
+    - title: Bugs
+      labels:
+        - bug
+    - title: New Features & Improvements
+      labels:
+        - "*"
```
ronjaquensel, line 15: Does this mean all labels including or excluding `bug` and `documentation`?

paullatzelsperger: I would assume so, but to be honest I don't know exactly - documentation is sparse and I wasn't able to test this on my fork, so I guess we'll have to run a test once the PR is merged.

ronjaquensel (approved):

.github/workflows/release-edc.yml
paullatzelsperger: update: so I added a condition that only runs the GH Release if not a `SNAPSHOT` release. If we need the user-input option later, we can always add it then.

juliapampus:
@paullatzelsperger With this, do you want to remove the "check-for-changelog-modifications" action? As there is no need to manually update the changelog anymore.

paullatzelsperger:
> @paullatzelsperger With this, do you want to remove the "check-for-changelog-modifications" action? As there is no need to manually update the changelog anymore. Or should we do it later, in a separate PR?

First, I would do this in a separate PR, because we first need to verify, that the auto-generated release-notes are serviceable, which can only be done, once this PR is merged.

Second, the question of getting rid of the changelog and the check is something we should discuss among the tech committers, although I don't expect too much pushback on it :)

ndr-brt (approved):

PR #1648, opened 2022-07-11
fix(api): validate canonical format

### LINKED ISSUE #1605: SQL Injection in DataManagement API
# Bug Report

## Describe the Bug

It's possible to manipulate a call to the DataManagementAPI to execute any SQL query. In the test this was done using the `/agreements/` endpoint. But as all SQL extension share the same code to build SQL queries, other endpoints might be affected, too.

### Expected Behavior
- No SQL injection possible

### Observed Behavior
- SQL injection possible

## Steps to Reproduce
1. Setup Connector with DataManagement API and SQL Extensions
2. Create a table called _test_ in the database
3. Execute the following API call to delete the table

```bash
curl -X GET "$providerDataMgmtUrl/$dataMgmtPath/contractagreements?filter=policy.id'LIKE/**/?/**/LIMIT/**/?/**/OFFSET/**/?;DROP/**/TABLE/**/test/**/--%20=%20ABC--" --header "$apiKey: $apiKeyValue" --header "Content-Type: application/json"
```

## Context Information

It looks like queries that target objects, that are stored as JSON in the database, are affected.

## Detailed Description
-

## Possible Implementation
-

paullatzelsperger:
## What this PR changes/adds

Query validation was added to all API endpoints that allow for dynamic querying. This is done by the `QueryValidator` object.

Validation is done based on introspecting Java classes and an optional class hierarchy tree.

This enables the `QueryValidator` to validate a query (more specifically: a query's left-hand operand) against the object graph, for example (Policy):
```
// valid query:
policy.permissions.action.constraint.leftExpression.value=someval

// invalid query:
policy.permissions.action.constraint.noexist=someval

```
the second example is invalid, because it does not express a valid traversal through the object graph.

## Why it does that

Avoids SQL injection, and potentially other problems, adds to stability.

## Further notes and limitations

- validating polymorphic types (e.g. `Constraint`) is possible, but limited to known subtypes, i.e. subclasses that are shipped with EDC. Extending them is currently not possible.
- `Map` types cannot be targeted, because their schema is not known, so for instance `dataAddress.properties.something=other` would be rejected. Adding `Map` support could be done in a future PR if ever needed.
- `Asset`s have a special validation: only "known" properties are allowed at the moment.
- the validation is done at the service level to enable it also for other store implementations. 
- the JSON representation of `Criterion` has changed to avoid problems with reflection:
  - `left` -> `operandLeft`
  - `op` -> `operator`
  - `right` -> `operandRight`
- added the `Result#merge` operator to merge two results + tests

## Linked Issue(s)

Closes #1605 

## Checklist

### DIFF
--- a/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetQueryValidator.java
+++ b/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetQueryValidator.java
@@ -31,8 +31,8 @@ class AssetQueryValidator extends QueryValidator {
             Asset.PROPERTY_CONTENT_TYPE
     );
 
-    AssetQueryValidator(Class<Asset> assetClass) {
-        super(assetClass);
+    AssetQueryValidator() {
+        super(Asset.class);
     }
 
     /**
--- a/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
+++ b/extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
@@ -48,7 +48,7 @@ public class AssetServiceImpl implements AssetService {
         this.contractNegotiationStore = contractNegotiationStore;
         this.transactionContext = transactionContext;
         this.observable = observable;
-        queryValidator = new AssetQueryValidator(Asset.class);
+        queryValidator = new AssetQueryValidator();
     }
 
     @Override
@@ -61,7 +61,7 @@ public class AssetServiceImpl implements AssetService {
         var result = queryValidator.validate(query);
 
         if (result.failed()) {
-            throw new EdcQueryException(format("Error validating schema: %s", String.join(", ", result.getFailureMessages())));
+            throw new EdcQueryException(format("Error validating schema: %s", result.getFailureDetail()));
         }
         return transactionContext.execute(() -> index.queryAssets(query).collect(toList()));
     }
--- a/spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/QueryValidator.java
+++ b/spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/QueryValidator.java
@@ -59,10 +59,7 @@ public class QueryValidator {
                 .map(Criterion::getOperandLeft)
                 .map(Object::toString)
                 .map(this::isValid)
-                .reduce((res1, res2) -> {
-                    res1.getFailureMessages().addAll(res2.getFailureMessages());
-                    return res1;
-                })
+                .reduce(Result::merge)
                 .orElse(Result.success());
     }
 
@@ -94,7 +91,7 @@ public class QueryValidator {
                     type = (Class<?>) genericType.getActualTypeArguments()[0];
                 }
             } else {
-                return Result.failure(format("Field %s not found on type %s", token, canonicalType));
+                return Result.failure(format("Field %s not found on type %s", token, type));
             }
         }
         return Result.success();
--- a/extensions/api/data-management/asset/src/test/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetQueryValidatorTest.java
+++ b/extensions/api/data-management/asset/src/test/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetQueryValidatorTest.java
@@ -1,0 +1,56 @@
+/*
+ *  Copyright (c) 2022 Microsoft Corporation
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.api.datamanagement.asset.service;
+
+import org.eclipse.dataspaceconnector.spi.query.QuerySpec;
+import org.eclipse.dataspaceconnector.spi.types.domain.asset.Asset;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class AssetQueryValidatorTest {
+
+    private final AssetQueryValidator validator = new AssetQueryValidator();
+
+    @ParameterizedTest
+    @ValueSource(strings = {
+            Asset.PROPERTY_ID,
+            Asset.PROPERTY_NAME,
+            Asset.PROPERTY_DESCRIPTION,
+            Asset.PROPERTY_VERSION,
+            Asset.PROPERTY_CONTENT_TYPE
+    })
+    void validate_validProperty(String key) {
+        var query = QuerySpec.Builder.newInstance().filter(key + "=someval").build();
+        assertThat(validator.validate(query).succeeded())
+                .isTrue();
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {
+            "asset_prop_id in (foo, bar)", // invalid key
+            "customProp=whatever", // no custom properties supported
+            "asset:prop:id.=something", // trailing dot
+            ".asset:prop:id=something" // leading dot
+    })
+    void validate_invalidProperty(String filter) {
+        var query = QuerySpec.Builder.newInstance()
+                .filter(filter)
+                .build();
+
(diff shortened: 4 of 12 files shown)


spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/QueryValidator.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Validates a {@link QuerySpec} whether it conforms to a particular schema (e.g. a Java class) or not
+     */
+    public Result<Void> validate(QuerySpec query) {
+        return query.getFilterExpression().stream()
+                .map(Criterion::getOperandLeft)
+                .map(Object::toString)
+                .map(this::isValid)
+                .reduce((res1, res2) -> {
+                    res1.getFailureMessages().addAll(res2.getFailureMessages());
```
ndr-brt, line 63: This will throw an NPE if the filter expression list contains at least one successful `Result`.
I would avoid this implementing it in a functional way: adding the `and` operator on the `Result` type, something like:
```java
    public Result<Void> and(Result<Void> other) {
        if (succeeded() && other.succeeded()) {
            return Result.success();
        } else {
            var messages = new ArrayList<String>();
            messages.addAll(Optional.ofNullable(this.getFailure()).map(Failure::getMessages).orElse(Collections.emptyList()));
            messages.addAll(Optional.ofNullable(other.getFailure()).map(Failure::getMessages).orElse(Collections.emptyList()));
            return Result.failure(messages);
        }
    }
```

so this lambda could be written just as:
```java
.reduce(Result::and)
```

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/query/QueryValidator.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import static java.lang.String.format;
+
+/**
+ * Validates that a particular query is valid, i.e. contains a left-hand operand, that conforms to the canonical
+ * format.
+ */
+public class QueryValidator {
```
ndr-brt, line 34: I would add some unit tests here as well, to verify all the conditions (invalid path, query map, field not found...)

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/sql/translation/EdcQueryException.java
```diff
@@ -0,0 +1,23 @@
...
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.sql.translation;
+
+import org.eclipse.dataspaceconnector.spi.EdcException;
+
+public class EdcQueryException extends EdcException {
```
ndr-brt, line 19: I don't know if I would add another exception, anyway this would need to be handled in the `EdcApiExceptionMapper` to return the expected 400.
Could be that the `query` method of every service returns a `ServiceResult<Collection<T>>`, in this way the error could be marked as `ServiceFailure.Reason.BAD_REQUEST` and be handled correctly to return a 400 by the controller

extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetQueryValidator.java
```diff
@@ -0,0 +1,48 @@
...
+class AssetQueryValidator extends QueryValidator {
+    private static final List<String> KNOWN_PROPERTIES = List.of(
+            Asset.PROPERTY_ID,
+            Asset.PROPERTY_NAME,
+            Asset.PROPERTY_DESCRIPTION,
+            Asset.PROPERTY_VERSION,
+            Asset.PROPERTY_CONTENT_TYPE
+    );
+
+    AssetQueryValidator(Class<Asset> assetClass) {
```
ndr-brt, line 34: This constructor could have no arguments and have the responsibility to pass the `Asset.class` object to the super constructor

extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
```diff
@@ -54,6 +58,11 @@ public Asset findById(String assetId) {
 
     @Override
     public Collection<Asset> query(QuerySpec query) {
+        var result = queryValidator.validate(query);
+
+        if (result.failed()) {
+            throw new EdcQueryException(format("Error validating schema: %s", String.join(", ", result.getFailureMessages())));
```
ndr-brt, line 64: Now there's the `getFailureDetail` method on the `AbstractResult` class that already to the failure messages join

extensions/api/data-management/asset/src/test/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImplTest.java
```diff
@@ -75,6 +79,36 @@ void query_shouldRelyOnAssetIndex() {
...
 
+    @ParameterizedTest
+    @ValueSource(strings = {
+            Asset.PROPERTY_ID,
+            Asset.PROPERTY_NAME,
+            Asset.PROPERTY_DESCRIPTION,
+            Asset.PROPERTY_VERSION,
+            Asset.PROPERTY_CONTENT_TYPE
+    })
+    void query_validFilter(String filter) {
```
ndr-brt, line 90: These seem like some tests to be put to the `AssetQueryValidator` class tests directly, And have a test here that mocks the validator and verifies its interaction. 
Would be the same for the other services, I have some doubt because the `subtypeMap`, but that's more a configuration than a behavior, so I'm not sure we need to test all of them, it will be enough a test one on the `QueryValidator`

extensions/sql/policy-store-sql/src/main/java/org/eclipse/dataspaceconnector/sql/policy/store/schema/postgres/PostgresDialectStatements.java
```diff
@@ -38,16 +38,16 @@ public String getFormatAsJsonOperator() {
 
     @Override
     public SqlQueryStatement createQuery(QuerySpec querySpec) {
-        if (querySpec.containsAnyLeftOperand("prohibitions")) {
+        if (querySpec.containsAnyLeftOperand("policy.prohibitions")) {
```
ndr-brt, line 41: since these statements are referring to the policy, it's necessary to have it made explicit in the left operand?
if I think about a query to search for policies probably I wouldn't think about putting the `policy` at the beginning of the operand. Same for the others, I'm not sure about the need for the root object name.

paullatzelsperger: Unfortunately prepending the `policy.` is necessary, because otherwise it wouldn't conform to the `PolicyDefinition.java` class anymore. Then, we'd have to add some special hack here, to make the reflective access work again, which is a bad idea IMO.

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/sql/translation/EdcQueryException.java
paullatzelsperger: returning a `ServiceResult` is probably better.

extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
paullatzelsperger: ah, didn't know that.

extensions/api/data-management/asset/src/test/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImplTest.java
paullatzelsperger: Actually, doing this by the book, I'd say we have to 
- test each service (which is already there)
- test the `QueryValidator` (to be done)
- test every variant, such as the `AssetQueryValidator` (to be done).

My argument for these (potentially duplicated) tests would be to maintain good coverage and regression stability: personally I want a test to break, if I change even the smallest thing :)

spi/core-spi/src/main/java/org/eclipse/dataspaceconnector/sql/translation/EdcQueryException.java
paullatzelsperger: created https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/1654

ndr-brt (approved):
Let's get this merged, ok with removing the NP check on `getFailureMessages` and the `@NotNull` on the `getContent`

PR #1657, opened 2022-07-12
Feature: refactor API services (query methods)

### LINKED ISSUE #1654: Refactor API services: use `ServiceResult` everywhere
All services used in the DM API controllers should be refactored, so that all service methods return a `ServiceResult` instead of returning objects and throwing exceptions. More specifically:
- refactor all `handleFailedResult` methods of all the controllers to have a case for `BAD_REQUEST`, throw an `IllegalArgumentException`
- possibly introduce some base class for those api controllers, and move `handleFailedResult` there
- return a dummy value (`null`?) in controller methods, in case the `ServiceResult` failed

paullatzelsperger:
## What this PR changes/adds

this PR changes the signature of all API services (the ones used by the controllers) to use `ServiceResult`s when querying.

## Why it does that

`ServiceResults` (or `Result`s in general) are a pattern in EDC and should be used wherever possible.

## Further notes

- moved the `handleFailureResult` methods into a helper class `ServiceResultHandler`.
- removed the `EdcQueryException` from the `EdcExceptionMapper` again.

## Linked Issue(s)

Closes #1654 

## Checklist

### DIFF
--- a/extensions/api/data-management/contractagreement/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractagreement/ContractAgreementApiController.java
+++ b/extensions/api/data-management/contractagreement/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractagreement/ContractAgreementApiController.java
@@ -25,6 +25,8 @@ import jakarta.ws.rs.core.MediaType;
 import org.eclipse.dataspaceconnector.api.datamanagement.contractagreement.model.ContractAgreementDto;
 import org.eclipse.dataspaceconnector.api.datamanagement.contractagreement.service.ContractAgreementService;
 import org.eclipse.dataspaceconnector.api.query.QuerySpecDto;
+import org.eclipse.dataspaceconnector.api.result.ServiceFailure;
+import org.eclipse.dataspaceconnector.api.result.ServiceResult;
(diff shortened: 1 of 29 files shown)


extensions/api/data-management/contractagreement/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractagreement/ContractAgreementApiController.java
```diff
@@ -64,7 +66,14 @@ public List<ContractAgreementDto> getAllAgreements(@Valid @BeanParam QuerySpecDt
 
         monitor.debug(format("get all contract agreements from %s", spec));
 
-        return service.query(spec).stream()
+        var queryResult = service.query(spec);
+
+        //will throw an exception
```
ndr-brt, line 71: Instead of having to add this comment and having a method that throws exception, could the method create an exception from the parameters and then would be up to the client (in this case the controller) to throw it?

paullatzelsperger: that would be a bit cleaner, because the method wouldn't "secretly" throw the exception, but we would get more duplicated code.

ndr-brt (approved):

Issue #1600, opened 2022-07-06
Clean-up `ids-api-multipart-dispatcher-v1`


ronjaquensel:
# Feature Request

The `ids-api-multipart-dispatcher-v1` module needs to be cleaned up.

## Which Areas Would Be Affected?
ids

## Why Is the Feature Desired?
Currently, there is a lot of Javadoc missing, some methods are duplicated and inheritance structures should be revised.

## Solution Proposal

- [x] #1667
- [x] #1695
- [x] #1697

## Type of Issue
cleanup

## Checklist

- [x] assigned appropriate label?
- [x] **Do NOT select a milestone or an assignee!**

PR #1695, opened 2022-07-14
feat: add check for ids message response type

juliapampus:
## What this PR changes/adds

Adds a type check for IDS mulitpart message responses. 

## Why it does that

Before, any IDS message was allowed as a response, although the further proccesing requires certain message types and contents. E.g. a state change would have been triggered in any case of a valid IDS response message, also for rejection messages.

## Further notes

Idea was not to check explicitly for rejection messages but to prevent processing unexpected responses.

## Linked Issue(s)

Relates #1600

## Checklist

### DIFF
--- a/data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/IdsMultipartSender.java
+++ b/data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/IdsMultipartSender.java
@@ -30,6 +30,7 @@ import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.ResponseBody;
 import org.eclipse.dataspaceconnector.ids.api.multipart.dispatcher.sender.response.IdsMultipartParts;
+import org.eclipse.dataspaceconnector.ids.api.multipart.dispatcher.sender.response.MultipartResponse;
 import org.eclipse.dataspaceconnector.ids.core.message.FutureCallback;
 import org.eclipse.dataspaceconnector.ids.core.message.IdsMessageSender;
 import org.eclipse.dataspaceconnector.ids.spi.IdsIdParser;
@@ -49,6 +50,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.net.http.HttpHeaders;
+import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 
@@ -202,8 +204,12 @@ abstract class IdsMultipartSender<M extends RemoteMessage, R> implements IdsMess
                         if (body == null) {
                             future.completeExceptionally(new EdcException("Received an empty body response from connector"));
                         } else {
-                            IdsMultipartParts parts = extractResponseParts(body);
-                            return getResponseContent(parts);
+                            var parts = extractResponseParts(body);
+                            var response = getResponseContent(parts);
+
+                            checkResponseType(response);
+
+                            return response.getPayload();
                         }
                     } catch (Exception e) {
                         future.completeExceptionally(e);
@@ -270,7 +276,14 @@ abstract class IdsMultipartSender<M extends RemoteMessage, R> implements IdsMess
      * @return an instance of the sub class's return type.
      * @throws Exception if parsing the response fails.
      */
-    protected abstract R getResponseContent(IdsMultipartParts parts) throws Exception;
+    protected abstract MultipartResponse<R> getResponseContent(IdsMultipartParts parts) throws Exception;
+
+    /**
+     * Return expected response type.
+     *
+     * @return the response type class.
+     */
+    protected abstract List<Class<? extends Message>> getAllowedResponseTypes();
 
     /**
      * Parses the multipart response. Extracts header and payload as input stream and wraps them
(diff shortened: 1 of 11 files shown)


data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/IdsMultipartSender.java
```diff
@@ -312,4 +323,20 @@ private IdsMultipartParts extractResponseParts(ResponseBody body) throws Excepti
                 .build();
     }
 
+    private R checkResponseType(R response) {
+        var type = getAllowedResponseTypes();
+
+        Objects.requireNonNull(type);
```
ndr-brt, line 329: as `getAllowedResponseTypes` result is always hardcoded, I would assume that's always non-null, maybe the `@NotNull` annotation could be used on the method and this guard avoided.

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/IdsMultipartSender.java
```diff
@@ -312,4 +323,20 @@ private IdsMultipartParts extractResponseParts(ResponseBody body) throws Excepti
...
+
+        if (response instanceof MultipartResponse) {
+            var header = ((MultipartResponse<?>) response).getHeader();
```
ndr-brt, line 332: I was wondering why the `R` type does not extends `MultipartResponse`, so I have a question, why `Catalog` is not enclosed in a `MultipartResponse` object? This would make the signature of the `IdsMultipartSender` more readable and this cast could be avoided, or am I missing something?

ndr-brt (commented):
Great to see this refactor evolving step by step :+1:

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/IdsMultipartSender.java
juliapampus: Currently , the method isn't returning a `MultipartResponse` because the `CatalogServiceImpl` and `BatchedRequestFetcher` expect a `Catalog` as return type for further processing, not just an `Object`.

ndr-brt: @juliapampus couldn't be the `IdsRemoteMessageDispatcher` that takes care of extracting the `Catalog` or every other object wrapped into `MultipartResponse` and returning it?

juliapampus: hmmm.. this could help, but how should the services then be able to implement error handling? Currently, the future is evaluated in almost every service and depending on throwables, responses etc. process continue differently. Even with using some kind of result object, the services expecting a specific object would have to be changed.

juliapampus: okay, so as we have this `AbstractResult` class. What about creating a new `MessageResult` and adapting the services that use this message dispatcher results? @ndr-brt

juliapampus: The dispatcher could then also, as proposed, take care of extracting the wrapped object and returning it in `success()`.

ndr-brt: @juliapampus why don't handle directly everything with the `CompletableFuture`? It's a "wrapping" type as `Result`.
The services want to receive a `CompletableFuture` with a certain content (e.g. `Catalog`), while the `MultipartResponse` it's a knowledge that should stay in the "multipart" world (if another protocol would exists, probably there would be another class wrapping the response).
So in the `IdsRemoteMessageDispatcher` the final line could be:
```java
return handler.send(message, context).thenApply(it -> (T) it.getPayload())
```
the `thenApply` works as a `map` function, it modifies the content of the wrapping type, in this case asynchronously

juliapampus: @ndr-brt Agree, but we should also not move multipart logic to the generic `IdsRemoteMessageDispatcher`. I therefore added the unpacking of the `MultipartResponse` to the `IdsMultipartSender` and changed the signature of the sender classes to return a simple object, not wrapped in a `MultipartResponse`.

juliapampus: I am not sure what the benefits may be, but as we currently want to make things easier, I would not introduce new interfaces. So I would keep this PR as-is and move to the next issue (#1697) as soon as this here is merged.

ronjaquensel (approved):

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/IdsMultipartSender.java
ndr-brt: @juliapampus fine for me

ndr-brt (approved):

PR #1702, opened 2022-07-18
iam: Change CredentialsVerifier contract.

### LINKED ISSUE #1570: Adapt CredentialsVerifier interface to IdentityHub use case
# Feature Request

Now the CredentialsVerifier declare the following method:
```java
Result<Map<String, Object>> verifyCredentials(String hubBaseUrl, PublicKeyWrapper publicKey);
```

We need to change this contract to work with IdentityHub.

When applying access policies, we need to first get the [VerifiableCredentials](https://www.w3.org/TR/vc-data-model) from IdentityHub, and get all the claims of the participant. For each claim, we need to know the issuer, because a policy can declare who should be the claim issuer.

Also, IdentityHub can return [VerifiableCredentials](https://www.w3.org/TR/vc-data-model) with various issuers, each issuer public key can be resolved to verify the signature, this mean we don't need the `PublicKeyWrapper publicKey` input.

A proposition would be to use this contract:

```java
Result<Collection<Claim>> getClaims(String participantDid);
```

The [claim](https://www.w3.org/TR/vc-data-model/#claims) would contain the claim information and the issuer. 
```java
class Claim {
String issuerDid;
String property;
String value;
}
```

identity-did-spi

This is desired to get the claims before applying policies.

## Solution Proposal

Use a contract like that instead:
```java
Result<Collection<Claim>> getClaims(String participantDid);

```
And this claim definition:
```java
class Claim {
String issuerDid;
String property;
String value;
}
```

## Type of Issue
Improvement

## Checklist

ouphi:
## What this PR changes/adds

Changed CredentialsVerifier to take a DidDocument as an input.

## Why it does that

DidDocument is a good input to verify credentials, as the way to get the credentials (identityHub service) is provided in the DidDocument.
For example, an IdentityHubCredentialVerifier would extract the identityHub URL from the did and query IdentityHub to get the verifiable credentials.

## Linked Issue(s)

Closes #1570 

## Checklist


jimmarino (approved):

paullatzelsperger (approved):

jimmarino:
@ouphi Can you fix the changelog and we can merge?

PR #1705, opened 2022-07-18
api: define common error message payload

### LINKED ISSUE #1490: api: define common error message payload
# Feature Request

Currently there are different way of returning an error payload based on the exception type.
- `EdcApiExceptionMapper` returns a list of objects that contain only a `message` field.
- `ValidationExceptionMapper`, that's the default one for `ValidationException` (thrown by the jersey validation library) returns a list of `ValidationErrorData` (that has these fields):
```java
protected String message;
protected String messageTemplate;
protected String path;
protected String invalidValue;
```

api

to have a consistent and shared error message that can be handled by clients

## Solution Proposal
Define a error payload class with some fields like: 
- `message`: will contain a description of the issue
- `type`: will describe the type of the issue, e.g. "NotNull", "ObjectAlreadyExists", "NotFormattedCorrectly"
- `path`: points to the name of the field that caused the error
- `invalidValue`: report the value that caused the error
- ... more can be added

## Type of Issue
improvement

## Checklist

ndr-brt:
## What this PR changes/adds

Extracts an `ApiErrorDetail` class and use it to describe api call errors to the clients

## Why it does that

To have an unique and shared way to describe details for failing requests

## Further notes

- I applied it to all the data-management apis, I wonder if it could be used to describe other api failures as well (as data-plane ones).
- remove unused (and "illegal") `http` dependency from `auth-spi` module.
- make `AssetService.query` return a `ServiceResult`, this slipped off from #1657 
- make all the exceptions that are going to be thrown in the filter/controllers extend `EdcApiException` and add a `getType` method on them to improve the result detail

## Linked Issue(s)

Closes #1490 

## Checklist


extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/EdcApiExceptionMapper.java
```diff
@@ -66,8 +67,11 @@ public Response toResponse(Throwable exception) {
 
         if (exception instanceof EdcException && verboseResponse) {
             var edcApiException = (EdcException) exception;
-            var responseBody = List.of(Map.of("message", edcApiException.getMessage()));
-            return Response.status(status).entity(responseBody).build();
+            var apiError = ApiErrorDetail.Builder.newInstance()
+                    .message(edcApiException.getMessage())
+                    .type(edcApiException.getClass().getSimpleName().replace("Exception", ""))
```
paullatzelsperger, line 72: this feels a bit brittle. How about we introduce an `EdcApiException` that contains a `type` field? Shouldn't be too much impact, as it would only touch the controllers.

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/EdcApiExceptionMapper.java
```diff
@@ -66,8 +67,11 @@ public Response toResponse(Throwable exception) {
...
+                    .build();
+            return Response.status(status).entity(List.of(apiError)).build();
```
paullatzelsperger, line 74: any particular reason this needs to be a `List`?

paullatzelsperger (commented):
I like the idea of having a standardized message payload, but it feels a bit off having `*Api` classes that define, e.g. a `List<AssetDto>` as return value, when in reality it could also be an `ApiErrorDetail` (rewritten by the exception mapper). 

Could we extend the `ApiErrorDetail` to a `ResponseDetail`, that contains a `result` and an `ErrorDetail`, similar to what we already have with `Result`s?

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/EdcApiExceptionMapper.java
ndr-brt: @paullatzelsperger this because the `ConstraintViolationException` contains a list of violations (as a validation could fail for different reasons), so the responses are consistent.

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/EdcApiExceptionMapper.java
ndr-brt: @paullatzelsperger fine with adding a `type` to the `EdcApiExtension`, but in the `EdcApiExceptionMapper` we are handling also an `EdcException`.
Probably the `EdcQueryException` would make more sense as an `Api` one, I'll work on that.

jimmarino (approved):

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/EdcApiExceptionMapper.java
paullatzelsperger: ah I see. makes sense.

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/EdcApiExceptionMapper.java
paullatzelsperger: we also handle some standard JDK exceptions, such as `IllegalArgumentException`, `UnsupportedOperationException`. 
What if the `exceptionMap` contains not just the `Response.Status`, such as `BAD_REQUEST`, but a tuple consisting of the status _and_ a type string?

paullatzelsperger (approved):

PR #1723, opened 2022-07-21
fix(catalog): introduce a mark for out-of-date offers

### LINKED ISSUE #1703: Federated catalog shows empty between clear stage and end of crawl
# Bug Report

## Describe the Bug
During the time between the federated catalog is clear before repopulated and the time is actually repopulated, the catalog is empty. 
Depending on the number of participants to be hit for retrieving the offers, this interval of time during which the catalog is empty can be quite important, which drastically degrades the user experience.

### Expected Behavior
The should be no period of unavailability of the catalog.

### Observed Behavior
If catalog is requested during the period of time between the clean and the end of a crawl, then it shows empty. 

## Steps to Reproduce
1. Creates a dataspace with few participants
2. Hit frequently the catalog endpoint
3. Observes that the catalog will be empty during some time before it is repopulated

## Possible Implementation
Possibly add a timestamp each time an item is stored in the cache. Then instead of clearing arbitrary the whole FederatedCacheStore at the beginning of each crawl, we could only clear items whose timestamp is older than a given duration.

paullatzelsperger:
## What this PR changes/adds

This PR solves the problem, where a catalog gets emptied at the beginning of a crawl run and items only trickling in as crawlers finish. This can lead to inconsistent UX for example in web frontends.

In order to alleviate that, I introduce the concept of a "mark" (or "dirty flag") which indicates that an item should be deleted.

At the beginning of every crawl run, the loader removes all marked items from the cache, and then proceeds to mark the remaining items. 
The mark is reset when the item is updated through the crawler. This way, only items that _didn't get updated_ retain the mark, and are thus deleted on the next crawl run.

## Why it does that

This counters inconsistencies with the cache.

## Linked Issue(s)

Closes #1703 

## Checklist

### DIFF
--- a/extensions/catalog/federated-catalog-cache/src/main/java/org/eclipse/dataspaceconnector/catalog/store/InMemoryFederatedCacheStore.java
+++ b/extensions/catalog/federated-catalog-cache/src/main/java/org/eclipse/dataspaceconnector/catalog/store/InMemoryFederatedCacheStore.java
@@ -32,7 +32,7 @@ import java.util.stream.Collectors;
  */
 public class InMemoryFederatedCacheStore implements FederatedCacheStore {
 
-    private final Map<String, ContractOffer> cache = new ConcurrentHashMap<>();
+    private final Map<String, MarkableEntry<ContractOffer>> cache = new ConcurrentHashMap<>();
     private final CriterionConverter<Predicate<ContractOffer>> converter;
     private final LockManager lockManager;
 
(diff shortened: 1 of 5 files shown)


paullatzelsperger:
@algattik @bscholtes1A could you take a quick look at this draft? Once you approve the general idea, I'll clean up the PR and maybe add some more tests.

extensions/catalog/federated-catalog-cache/src/main/java/org/eclipse/dataspaceconnector/catalog/store/InMemoryFederatedCacheStore.java
```diff
@@ -32,7 +32,7 @@
  */
 public class InMemoryFederatedCacheStore implements FederatedCacheStore {
 
-    private final Map<String, ContractOffer> cache = new ConcurrentHashMap<>();
+    private final Map<String, MarkableEntry<ContractOffer>> cache = new ConcurrentHashMap<>();
```
bscholtes1A, line 35: Several remarks:

1/ Is there really a need for a generic in the MarkableEntry? 

Also I would prefer to keep it immutable:

~~~
private static class MarkableEntry {
  private final ContractOffer contractOffer;
  private final isPotentiallyExpired;

  public MarkableEntry(boolean isPotentiallyExpired, ContractOffer contractOffer) {
    this.contractOffer = contractOffer;
    this.isPotentiallyExpired= isPotentiallyExpired;
  }
}
~~~

2/ Then in the `save` function becomes:

~~~
public void save(ContractOffer contractOffer) {
        lockManager.writeLock(() -> cache.put(contractOffer.getAsset().getId(), new MarkableEntry(false, contractOffer)));
    }
~~~

3/ And finally instead of having two methods for marking, and then deleting the deprecated, I would merge them in a single one:

~~~
@Override
public void deleteExpired() {
    lockManager.writeLock({
       cache.values().removeIf(me::isExpired);
       cache.replaceAll((k,v) -> v = new MarkableEntry(true, v.getContractOffer()));
       return null;
    }
}
~~~

paullatzelsperger: i'm fine with the first two points. I guess the in-mem impl is intended for development/testing purposes only, so I won't go into whether modifying or replacing a map entry is better, faster, more efficient, etc.

The reason why I kept the `markAll` and `removeMarked` methods separate is that the behaviour for different store implementations than be controlled by the owning object (here: the `LoaderImpl`); it makes it apparent through the interface that these are two distinct steps.

If we collate them into one method call, it would push business logic down into the store implementation, which I find dangerous.

bscholtes1A (changes requested):

bscholtes1A (approved):

PR #1729, opened 2022-07-22
feat: make rest api return error details on edc exceptions and validation error by default

### LINKED ISSUE #1492: api: return validation error message by default
# Feature Request

Currently, in case of failure after an api call, the detailed error message is returned only if a specific configuration setting is enabled.
This should be done by default, and a more "extended debug" mode could be enabled by another specific setting.

api

to have all the clients informed about the reason of the failure, and enable a debug mode that could show more details about failure (like stack trace), not to be enabled in production

## Solution Proposal
-

## Type of Issue
improvement

## Checklist

ndr-brt:
## What this PR changes/adds

Return the "api error detail" response payload by default

## Why it does that

The clients should be aware of the failure reason.

## Further notes

- having the possibility to include the stacktrace as it was hinted in the issue didn't seem a good idea, clients should only be aware of errors resolvable on their side, and the stacktrace won't help them at all, neither in a development environment
- Removed the "verbose" config settings as it has lost its value.

## Linked Issue(s)

Closes #1492 

## Checklist

### DIFF
--- a/extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/EdcApiExceptionMapper.java
+++ b/extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/EdcApiExceptionMapper.java
@@ -39,10 +39,8 @@ import static jakarta.ws.rs.core.Response.Status.UNAUTHORIZED;
 
 public class EdcApiExceptionMapper implements ExceptionMapper<Throwable> {
     private final Map<Class<? extends Throwable>, Response.Status> exceptionMap;
-    private final boolean verboseResponse;
 
-    public EdcApiExceptionMapper(boolean verboseResponse) {
-        this.verboseResponse = verboseResponse;
+    public EdcApiExceptionMapper() {
         exceptionMap = Map.of(
                 IllegalArgumentException.class, BAD_REQUEST,
                 NullPointerException.class, BAD_REQUEST,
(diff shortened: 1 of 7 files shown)


ronjaquensel (approved):

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/EdcApiExceptionMapper.java
```diff
@@ -63,13 +61,17 @@ public Response toResponse(Throwable exception) {
...
+
+            var errorDetail = ApiErrorDetail.Builder.newInstance()
                     .message(edcApiException.getMessage())
                     .type(edcApiException.getType())
                     .build();
-            return Response.status(status).entity(List.of(apiError)).build();
+
+            return Response.status(status)
+                    .entity(List.of(errorDetail))
+                    .build();
```
bscholtes1A, line 74: Not necessarily related to this PR. But we do not provide the error details (in the `entity` part of the Response) is the exception is not of type `EdcApiException`

paullatzelsperger (approved):

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/EdcApiExceptionMapper.java
ndr-brt: @bscholtes1A I think this is correct, as the other exceptions should cause a 5xx error (a server fault, so the client won't need any detail). 
I added an issue on this topic: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/1730

PR #1731, opened 2022-07-22
fix(DataPlane): Add system test for Client Data Pull with query parameters in the request

### LINKED ISSUE #1180: Data Plane consumer proxy filters request properties
# Bug Report

## Describe the Bug
This is consecutive to #1079, which adds some boolean in the http data address. These boolean are used as security to prevent some request properties (method, path, query params...) to be mapped in the `HttpDataSource`. 

This approach works fine with the provider proxy, as the booleans are directly obtained from the http `DataAddress` representing the actual data source. However, for the consumer proxy, the http data address is built from the `EndpointDataReference` returned by the provider Control Plane. In current implementation, none of the boolean is mapped into this http data address, resulting in all request attributes (method, path, query params...) to be filtered in the call performed by the consumer `HttpDataSource`. 

### Expected Behavior
All request attributes should be forwarded until the provide DPF, which is then in charge of applying the proper filters based on the actual data source address.

### Observed Behavior
Request attributes are systematically filtered by the consumer DPF.

## Possible Implementation
The Consumer proxy should be built in such a way that it does not filter any request properties. Filtering should _ONLY_ be done on provider side based on the properties of the actual data source address.

bscholtes1A:
## What this PR changes/adds

This PR adds a system test covering the use-case of a client data pull (see https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/585 for more details) with query parameters passed in input of the Data Plane public API.

## Why it does that

Ensure query parameters are properly conveyed all the way to the provider data source (if its allows it).

## Linked Issue(s)

Closes #1180 

## Checklist

### DIFF
--- a/system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/dataspaceconnector/test/e2e/AbstractEndToEndTransfer.java
+++ b/system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/dataspaceconnector/test/e2e/AbstractEndToEndTransfer.java
@@ -19,6 +19,8 @@ import org.eclipse.dataspaceconnector.spi.types.domain.HttpDataAddress;
 import org.junit.jupiter.api.Test;
 
 import java.time.Duration;
+import java.util.Map;
+import java.util.UUID;
 
 import static io.restassured.RestAssured.given;
 import static org.assertj.core.api.Assertions.assertThat;
@@ -37,7 +39,7 @@ public abstract class AbstractEndToEndTransfer {
     void httpPullDataTransfer() {
         PROVIDER.registerDataPlane();
         CONSUMER.registerDataPlane();
-        String definitionId = "1";
+        var definitionId = "1";
         createAssetAndContractDefinitionOnProvider("asset-id", definitionId, "HttpData");
 
         var catalog = CONSUMER.getCatalog(PROVIDER.idsEndpoint());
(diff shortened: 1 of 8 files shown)


system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/dataspaceconnector/test/e2e/AbstractEndToEndTransfer.java
```diff
@@ -50,29 +52,63 @@ void httpPullDataTransfer() {
...
-            given()
-                    .baseUri(CONSUMER.backendService().toString())
-                    .when()
-                    .get("/api/consumer/data")
-                    .then()
-                    .statusCode(200)
-                    .body("message", equalTo("some information"));
-        });
+        // retrieve the data reference
+        await().atMost(timeout).untilAsserted(() -> given()
```
ndr-brt, line 64: As the `EndpointDataReference` object is used below, this could be extacted in a method and moved in the `Participant` class, like it has been done for the `getContractAgreementId` method, this will make the tests more readable and it will encapsulate the url of the call.

system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/dataspaceconnector/test/e2e/AbstractEndToEndTransfer.java
```diff
@@ -50,29 +52,63 @@ void httpPullDataTransfer() {
...
+                .when()
+                .get("/api/consumer/dataReference/" + dataRequestId)
+                .then()
+                .statusCode(200)
+                .extract()
+                .body()
+                .as(EndpointDataReference.class);
+
+        // pull the data
+        await().atMost(timeout).untilAsserted(() -> given()
```
ndr-brt, line 85: Maybe also this part could be included in the same method I proposed above: the method on participant could be like `getData` or `pullData` and it could hide all the edr details

bscholtes1A: Updated, could you please have another look?

ndr-brt (approved):

PR #1744, opened 2022-07-26
feat: separate unexpected from expected exceptions

### LINKED ISSUE #1730: api: clearly separate "expected" exceptions from "unexpected" ones
# Feature Request

Currently, the `EdcApiExceptionMapper`, it's the gateway to handle exceptions raised in the `jersey` context.
The mapper handles `EdcApiException`s and also other `java.lang.*` exceptions as:
- `IllegalArgumentException`
- `NullPointerException`
- `UnsupportedOperationException`

making them (at least the first 2) return a 4xx error code, that's, by definition, a "client error", so something that has been caused by the client, but, a NPE it's not a client fault, it's a server fault.

We need to clearly separate them and make all the `java.lang` "unexpected" exception return a 5xx error.

api

To improve error handling and reporting

## Solution Proposal
Make all the `java.lang` exceptions return a 5xx error, maybe printing a `severe` log, to permit developers to signal the behavior as a bug.
The `IllegalArgumentException` should not be thrown in the api context, a more detailed `EdcApiException` should be used, that will make the server return an error detail that could be better interpreted by the client.

## Type of Issue
improvement

## Checklist

ndr-brt:
## What this PR changes/adds

Distinct between expected (subclasses of `EdcApiException`) from unexpected (any other) exceptions.
The latter, as unexpected, will return a 5xx error, by default a 500 Internal Server Error, as it's a generic "server failure" error.

## Why it does that

To improve error handling.

## Further notes

- introduced a `InvalidRequestException` that replaces all the `IllegalArgumentException` in the `data-management` api modules
- separated the exception mapper in 3 parts: `EdcApiExceptionMapper`, `ValidationExceptionMapper` and `UnexpectedExceptionMapper`

## Linked Issue(s)

Closes #1730

## Checklist


extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/model/NegotiationInitiateRequestDto.java
```diff
@@ -46,6 +49,14 @@ public ContractOfferDescription getOffer() {
         return offer;
     }
 
+    @AssertTrue(message = "connectorId, connectorAddress, protocol, offerId and assetId cannot be blank")
+    @JsonIgnore
+    public boolean isValid() {
```
paullatzelsperger, line 54: Why do we need this? all properties are already validated using `@NotNull` annotations, right? Even the nested ones?

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/model/NegotiationInitiateRequestDto.java
```diff
@@ -14,7 +14,10 @@
 
 package org.eclipse.dataspaceconnector.api.datamanagement.contractnegotiation.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import jakarta.validation.constraints.AssertTrue;
 import jakarta.validation.constraints.NotNull;
+import org.apache.commons.lang3.StringUtils;
```
paullatzelsperger, line 20: Where does this come from? We should be extremely careful about pulling in apache commons, and I honestly think its not necessary to pull it in, just to check a few strings for nullity.

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/UnexpectedExceptionMapper.java
```diff
@@ -0,0 +1,47 @@
...
+import jakarta.ws.rs.ext.ExceptionMapper;
+
+import java.util.Map;
+
+import static jakarta.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static jakarta.ws.rs.core.Response.Status.NOT_IMPLEMENTED;
+
+/**
+ * Exception mapper that catches all the exceptions that are not handled by the other mappers, mainly:
+ * - java.lang.*Exception
```
paullatzelsperger, line 28: actually subclasses of `java.lang.Throwable` according to the class' type definition

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/JerseyRestService.java
```diff
@@ -83,8 +84,10 @@ private void registerContext(String contextAlias, List<Object> controllers) {
         resourceConfig.registerClasses(controllers.stream().map(Object::getClass).collect(toSet()));
         resourceConfig.registerInstances(new Binder());
         resourceConfig.registerInstances(new TypeManagerContextResolver(typeManager));
+
         resourceConfig.registerInstances(new EdcApiExceptionMapper());
```
paullatzelsperger, line 88: registering the exception mappers directly in the JerseyRestService will have two effects:
1. it will prevent users from implementing their own error handling. I imagine this may be problematic for some.
2. it will cause _all_ APIs to use these mappers, even quasi-internal ones, such as the Dataplane API, that are not meant for public consumption.

How about we register these three mappers in the `DataManagementApiConfigurationExtension` to make the JerseyRestService oblivious toward how errors need to be handled? That way the DM-API would still have proper error handling, but we don't force it on every other API.

extensions/http-provisioner/src/main/java/org/eclipse/dataspaceconnector/transfer/provision/http/webhook/HttpProvisionerWebhookApiController.java
```diff
@@ -58,27 +57,27 @@ public void callProvisionWebhook(@PathParam("processId") String transferProcessI
...
+
+        transferProcessManager.enqueueCommand(command);
 
     }
 
     @Override
     @POST
     @Path("/{processId}/deprovision")
     public void callDeprovisionWebhook(@PathParam("processId") String transferProcessId, @Valid DeprovisionedResource resource) {
+        if (resource == null || resource.getProvisionedResourceId() == null) {
```
paullatzelsperger, line 74: nit: the `resource == null` part could be replaced by putting the `@NotNull` annotation on the method parameter (both here and in the interface).

extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/AssetApiController.java
```diff
@@ -67,7 +68,7 @@ public void createAsset(@Valid AssetEntryDto assetEntryDto) {
         var dataAddressResult = transformerRegistry.transform(assetEntryDto.getDataAddress(), DataAddress.class);
 
         if (assetResult.failed() || dataAddressResult.failed()) {
-            throw new IllegalArgumentException("Request is not well formatted");
+            throw new InvalidRequestException("Request is not well formatted");
```
paullatzelsperger, line 71: We're using this message string in several locations, maybe we could add a static method like `throw InvalidRequestException.requestFormat()` to have the message text in a central place?

paullatzelsperger (commented):
comments inline

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/model/NegotiationInitiateRequestDto.java
paullatzelsperger: [update]: Apache Commons Lang gets pulled in by the Swagger Jakarta integration and Rest Assured. However, I still think we should severely restrict the use of it in our code base.

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/model/NegotiationInitiateRequestDto.java
ndr-brt: I moved the `isValid` method from here:
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/501883cc32f7d390b1c3bbad562c551210dbfa88/extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/ContractNegotiationApiController.java#L175

plus, there's something wrong with the jersey-bean-validation, the validation is not working every time, before this PR we had all the NPE becoming 400, but changing this behavior made me notice that actually there are a lot of NPE thrown because of non-working validation, I need to investigate more

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/model/NegotiationInitiateRequestDto.java
ndr-brt: I moved the `isValid` method from here:
https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/501883cc32f7d390b1c3bbad562c551210dbfa88/extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/ContractNegotiationApiController.java#L175

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/mapper/UnexpectedExceptionMapper.java
ndr-brt: the description was meant to say, all the exceptions belonging to `java.lang` package, because not all the subclasses of `Throwable` (that are, in fact, all the possible exceptions) are handled here, as some are handled by other mappers

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/JerseyRestService.java
ndr-brt: 1. actually not, because every newly registered mapper will like be "chosen before" the other (with the same "distance" from the actual exception type), so everyone can register its own exception mapper
2. I think that's not a problem, as long as the "exception throwing" is correct that's a sane way to unify error behaviors (then every other module could have its own exception superclass and its own exception handler).

As this was already the used approach, I will propose to discuss this in another issue.

extensions/http-provisioner/src/main/java/org/eclipse/dataspaceconnector/transfer/provision/http/webhook/HttpProvisionerWebhookApiController.java
ndr-brt: As I said before, this validation thing seems to not work very well, try to remove that null check and run the `HttpProvisionerWebhookApiControllerIntegrationTest`.

extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/AssetApiController.java
ndr-brt: it's used 3 times and in two of them there's an additional detail. I agree that we should avoid this kind of duplication but I don't see a real advantage here.

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/JerseyRestService.java
paullatzelsperger: ok, fine for me.

extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/AssetApiController.java
paullatzelsperger: could it be a `.requestFormat(@Nullable String additionalDetail)`, or an override?

extensions/http-provisioner/src/main/java/org/eclipse/dataspaceconnector/transfer/provision/http/webhook/HttpProvisionerWebhookApiController.java
paullatzelsperger: if you pick the "right" `@NotNull` annotation, and put it on the interface and the class it should work. at least it worked for me...?

extensions/api/data-management/contractnegotiation/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/contractnegotiation/model/NegotiationInitiateRequestDto.java
paullatzelsperger: ok, then we should go through the code and cut out all those Apache Commons occurrences. can be done later though.

extensions/http-provisioner/src/main/java/org/eclipse/dataspaceconnector/transfer/provision/http/webhook/HttpProvisionerWebhookApiController.java
ndr-brt: @paullatzelsperger oh, now I remember, the problem here is that the `DeprovisionedResource` object is used, and it's located in the `transfer-spi` module, so it cannot have the annotation registered. I would left this as it is and provide a dedicated DTO in another PR

paullatzelsperger: ahh, you're right. 
we can also leave the `resource == null` check in place its not really a problem :) AFAIK the `DeprovisionedResource` was originally intended to be serializable, so I'd be ok with just leaving it.

extensions/api/data-management/asset/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/AssetApiController.java
ndr-brt: @paullatzelsperger actually this reminded me that I had an idea, provide a list of messages to the `EdcApiException` so there could be a better mapping from `Result` to `EdcApiException` to a `List<ApiErrorDetail>`, this will make the exception accepting a list of messages from the failure, without having to specify any further message (if I see a 400 error it's pretty obvious that the "Request is not well formatted").

paullatzelsperger (approved):

PR #1753, opened 2022-07-28
feat: add https call enforcer, it can be enabled with a config setting

### LINKED ISSUE #1734: Enforce outgoing HTTPS calls
# Feature Request

As specified in discussion #1678 we should avoid "not secure" `http` calls.
Having the EDC enforcing HTTPS calls by default would improve the security level.
I think would be superfluous to explain why HTTPS should be used over HTTP, but [here's an explaination article by cloudflare just for reference](https://www.cloudflare.com/learning/ssl/why-use-https/).

Non-secure http calls should be permitted only for testing/local development purposes enabling a dedicate config setting.

http client

To improve the connector security.

## Solution Proposal
This could be implemented having an okhttp `Interceptor` implemented on the registered `OkHttpClient` that would cancel the request and log an error if the call is not `https`.
With a dedicated settings this interceptor should not be registered and a warning log printed.

## Type of Issue
improvement

## Checklist

ndr-brt:
## What this PR changes/adds

Adds a "HTTPS enforcer" interceptor that can be enabled by setting the `edc.http.enforce-https` to `true`.
If enabled, it will throw an exception every time an `http` call is made by the http client.

## Why it does that

To provide an additional security setting that will avoid insecure http calls.

## Further notes

- Extracted a new `HttpClientServiceExtension` as the `CoreServicesExtension` was getting too big.
- Avoided deprecated warning on the "publish test result" CI job
- Bumped `okHttp` version to [4.10.0](https://square.github.io/okhttp/changelogs/changelog_4x/#version-4100)

## Linked Issue(s)

Closes #1734

## Checklist


jimmarino:
I'm not a big fan of having multiple ServiceExtension implementations per Gradle module. Usually, it is a sign that things aren't fully modularized. Can it be pulled out or combined with the other one?

ndr-brt:
@jimmarino could be useful to have like an OkHttpClient factory that takes care of initializing it giving a configuration and then the `@Provider` method could go back to the `CoreServicesExtension`? This will make things crispier without having an additional extension

core/base/src/main/java/org/eclipse/dataspaceconnector/core/HttpClientExtension.java
```diff
@@ -0,0 +1,81 @@
...
+                .connectTimeout(30, TimeUnit.SECONDS)
+                .readTimeout(30, TimeUnit.SECONDS);
+
+        ofNullable(okHttpEventListener).ifPresent(builder::eventListener);
+
+        boolean enforceHttps = context.getSetting(EDC_HTTP_ENFORCE_HTTPS, false);
+        if (enforceHttps) {
+            builder.addInterceptor(new EnforceHttps());
+        } else {
+            context.getMonitor().info("HTTPS enforcement it's not enabled, it should be in a production environment");
```
paullatzelsperger, line 64:
```suggestion
            context.getMonitor().info("HTTPS enforcement it not enabled, please do not use this in a production environment");
```

paullatzelsperger (approved):
minor textual improvement, other than that good to go.

core/base/src/main/java/org/eclipse/dataspaceconnector/core/HttpClientExtension.java
ndr-brt: @paullatzelsperger the double negation make it pretty confusing, what about "please enable it in a production environment"?

paullatzelsperger: strictly speaking not a double negative (that would be e.g. "don't not use it..."), but your suggestion is fine too :)

ndr-brt:
@jimmarino introduced the factory and removed the additional extension, could you take a look? thanks

jimmarino (approved):

PR #1756, opened 2022-07-28
feat(api): add custom validation framework

### LINKED ISSUE #1736: Custom validator for specific `AssetEntryDto` fields
# Feature Request

Currently, on the `POST /assets` call, the validation act on the `data-management/asset` relies on the `jakarta-validation` annotations set on the `AssetEntryDto` class.
Since the `Asset` entity is composed by a free-form dictionary, it would be useful to have a way to extend a custom dataspace-specific validation (e.g. for a specific dataspace a certain information should be mandatory or have a particular format).

data-management api

To give the possibility to specify a dataspace-specific validation for the `Asset` properties

## Solution Proposal
The simplest implementation could be to introduce a new `AssetEntryDtoValidator` service that would have a "no-op" default implementation and that it could be overrode by an implementation provided by extension.
This validator would be called on the controller before the `transformation` call, and return a 400 error if the validation returned a `failed` result.
This concept could be extended to the other api endpoints as well, but I would address this to other issues.

## Type of Issue
improvement

## Checklist

paullatzelsperger:
## What this PR changes/adds

This PR adds a convenient way for users to register their own validation functions for API methods.

## Why it does that

Extensible validation may be desired by users, especially when dealing with extensible types.

## Further notes

- Contrary to what is stated in the DR, all registration methods now accept a `ValidationFunction`, which is a marker for `Function<Object[], Result<Void>>`. This was done to improve readability.

## Example Usage
```java
public class CustomValidationExtension implements ServiceExtension {

    @Inject
    private ValidationFunctionRegistry validationFunctionRegistry;

    @Override

        // lets validate that every Asset must have an "owner" property
        validationFunctionRegistry.registerForType(AssetEntryDto.class, objects -> {
            var assetDto = Stream.of(objects).filter(ob -> AssetEntryDto.class.isAssignableFrom(ob.getClass())).findFirst();

            if (assetDto.isEmpty()) return Result.failure("Method parameters did not contain an AssetEntryDto!");

            var dto = (AssetEntryDto) assetDto.get();

            if (!dto.getAsset().getProperties().containsKey("owner")) {
                return Result.failure("owner property not present");
            }

            return dto.getAsset().getProperties().get("owner") == null ? Result.failure("owner was null") : Result.success();
        });

        // now lets add some other validation logic, specifically for the "createAsset" method
        try {
            var method = AssetApiController.class.getDeclaredMethod("createAsset", AssetEntryDto.class);
            validationFunctionRegistry.registerForMethod(method, methodArguments -> {
                // some elaborate validation logic
                return Result.failure("createAsset: elaborate validation failed");
            });
        } catch (NoSuchMethodException e) {
            throw new RuntimeException(e);
        }
    }
}

```

## Linked Issue(s)

Closes #1736 

## Checklist


extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/JerseyExtension.java
```diff
@@ -42,11 +48,19 @@ public void initialize(ServiceExtensionContext context) {
 
         jerseyRestService = new JerseyRestService(jettyService, typeManager, configuration, monitor);
 
+        provider = new ResourceInterceptorProvider();
+        jerseyRestService.registerInstance(() -> new ResourceInterceptorBinder(provider));
```
ndr-brt, line 52: Is there a particular reason why the binder is passed through a `Supplier` and not directly?

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/validation/ValidationFunctionRegistryImpl.java
```diff
@@ -0,0 +1,42 @@
...
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.extension.jersey.validation;
+
+import java.lang.reflect.Method;
+
+public class ValidationFunctionRegistryImpl implements ValidationFunctionRegistry {
```
ndr-brt, line 19: Seems like an avoidable indirection layer, could `ResourceInterceptorProvider` directly implement `ValidationFunctionRegistry`?

ndr-brt (commented):
Looks really promising!
A test showing the a custom validation implemented would help the comprehension, I know that's still a draft, but it would be a nice to have in the final version of the PR

paullatzelsperger:
> Looks really promising!
> 
> A test showing the a custom validation implemented would help the comprehension, I know that's still a draft, but it would be a nice to have in the final version of the PR

That's a cool idea - I'll add an integration test for that

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/JerseyExtension.java
paullatzelsperger: yes, because there must be a new instance registered for every resource, otherwise Jersey will vomit all over itself

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/validation/ValidationFunctionRegistryImpl.java
paullatzelsperger: technically we could do that, but the idea was to make the `ResourceInterceptorProvider` a generic hook point that could be used for other things than _just_ validation, so it is a generic layer, whereas the `ValidationFunctionRegistry` "binds" it specifically to the validation use case. 

Also, I liked the idea of having a clean user-facing interface.

paullatzelsperger: hmmm, so I gave it a try, and I actually like it. check the respective [commit](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/1756/commits/4b835e8bfffa994db6a2e1510b2a6b6bcfe5748b)

florianrusch-zf (changes requested):
LGTM so far.

Just one additional point: We should maybe also describe at least in the documentation, that the failure message is "important" to be a good one, as it will be returned to the client in the response body.

extensions/http/jersey/src/test/java/org/eclipse/dataspaceconnector/extension/jersey/validation/ResourceInterceptorProviderTest.java
```diff
@@ -0,0 +1,116 @@
...
+        public boolean isClassBased() {
+            return true;
+        }
+
+        @Override
+        protected Object getHandlerInstance() {
+            return this;
+        }
+    }
+}
```
florianrusch-zf, line 116: Missing empty line at EOF
```suggestion
}

```

extensions/http/jersey/src/test/java/org/eclipse/dataspaceconnector/extension/jersey/validation/ResourceInterceptorTest.java
```diff
@@ -0,0 +1,109 @@
...
+        interceptor.invoke(proxy, getAnswerMethod(), new Object[0]);
+
+        var functionOrder = inOrder(fct1, fct2);
+        functionOrder.verify(fct1).apply(any());
+        functionOrder.verify(fct2).apply(any());
+
+    }
+
+
+}
```
florianrusch-zf, line 109: Missing empty line at EOF
```suggestion
}

```

florianrusch-zf (changes requested):

paullatzelsperger:
> ... the failure message is "important" to be a good one, as it will be returned to the client in the response body.

All failure messages should be treated as "good ones". No point in stating that explicitly.

jimmarino (approved):

extensions/http/jersey/src/main/java/org/eclipse/dataspaceconnector/extension/jersey/validation/ValidationFunctionRegistryImpl.java
ndr-brt: @paullatzelsperger maybe the `ValidationFunctionRegistry` could be called `InterceptorFunctionRegistry`?
That would make it more generic looking.

paullatzelsperger: I thought about it, and considering that everything else is called "interceptor"-something I'm fine with it.

ndr-brt (approved):
Great, LGTM :rocket:

PR #1778, opened 2022-08-04
chore: make JwtUtils public

paullatzelsperger:
## What this PR changes/adds

Makes the `JwtUtils` (formerly known as `VerifiableCredentialsFActory`) public again and moved to the `identity-did-crypto` module

## Why it does that

Downstream projects, e.g. `RegistrationService`, use it (and won't compile otherwise).

## Further notes

- the package name `org.eclipse.dataspaceconnector.iam.did.crypto.JwtUtils` is not right, but I wanted to avoid a widespread impact on the code base. We'll refactor this during an upcoming directory- and package overhaul.
- a related PR in the RegistrationService repo will follow.

## Checklist


jimmarino (approved):

algattik (changes requested):
Actually this was intentional. As the JavaDoc states, It is a "convenience/helper class to generate and verify Signed JSON Web Tokens (JWTs) for communicating between connector instances.". The `RegistrationService` now uses Nimbus directly to generate JWT VCs with different code (for example, no `audience` claim).

Is there still a need to be public then?

paullatzelsperger:
> Actually this was intentional. As the JavaDoc states, It is a "convenience/helper class to generate and verify Signed JSON Web Tokens (JWTs) for communicating between connector instances.". The `RegistrationService` now uses Nimbus directly to generate JWT VCs with different code (for example, no `audience` claim).
> 
> Is there still a need to be public then?

The current `main` branch of the RegistrationService still has references to the `VerifiableCredentialFactory`:
- [DidJwtAuthenticationFilter](https://github.com/eclipse-dataspaceconnector/RegistrationService/blob/main/extensions/registration-service/src/main/java/org/eclipse/dataspaceconnector/registration/auth/DidJwtAuthenticationFilter.java)
- [ClientUtils](https://github.com/eclipse-dataspaceconnector/RegistrationService/blob/main/client-cli/src/main/java/org/eclipse/dataspaceconnector/registration/cli/ClientUtils.java)

so the `RegistrationService` `main` does not compile with `EDC` `main`

algattik (approved):
Understood, thank you.
Note that RegistrationService `main` should not be adapted to compile against post-M5 changes, since the target is decided to be M5.

But it's good to prepare post-M5 EDC to be ready for RegistrationService changes when the time comes.

Issue #1781, opened 2022-08-04
transfer process stopped working after update


kedorlaomer:
# Bug Report

## Describe the Bug
The commit [988215e7094778495938ff79e6bbfc038b9e8981](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/commit/988215e7094778495938ff79e6bbfc038b9e8981) broke our usage of the API endpoints `/api/v1/data/transferprocess` and `/api/v1/data/catalog`.

### Expected Behavior
Afterwards, the endpoint `/api/v1/data/catalog` returns 200/OK and a benevolent JSON in its body.

### Observed Behavior
The endpoint returns 500/Internal Server Error and a HTML body stating Internal Server Error

## Steps to Reproduce
1. Start negotiation in order to get agreement and register asset (works)
2. Transfer process via `/api/v1/data/transferprocess` (which returns 200/OK)
3. Look for asset in catalog via `/api/v1/data/catalog`.

## Context Information

- Used version: 988215e7094778495938ff79e6bbfc038b9e8981 (also happens for as-of-now current 3ff940b720f44826df28e893fb31344eb6faacef)
- OS: Windows 10

## Detailed Description
The log gives stack trace  
> org.eclipse.dataspaceconnector.spi.EdcException: Received unexpected response type.
>	at org.eclipse.dataspaceconnector.ids.api.multipart.dispatcher.sender.IdsMultipartSender.checkResponseType(IdsMultipartSender.java:332)
>	at org.eclipse.dataspaceconnector.ids.api.multipart.dispatcher.sender.IdsMultipartSender.lambda$send$0(IdsMultipartSender.java:210)
>	at org.eclipse.dataspaceconnector.ids.core.message.FutureCallback.onResponse(FutureCallback.java:45)

This error message is thrown already after `/api/v1/data/transferprocess`; the 500 of `/api/v1/data/catalog` is likely only a consequence.

juliapampus:
I cannot reproduce the exception. I used sample 4.0 to test the negotiation and transfer and thought that it may be because the IDS object mapper is used for payload de/serialization in `MultipartEndpointDataReferenceRequestSender` but not in `EndpointDataReferenceHandler`. However, this seems not to be the issue.

[...]

PR #1790, opened 2022-08-05
chore(IdsMultipartSender): use default object mapper for data reference serialization

juliapampus:
## What this PR changes/adds

Resets the used `ObjectMapper` for serializing the `MultipartEndpointDataReferenceRequestSender` payload to the default one, since `EndpointDataReference` is not an IDS object and `EndpointDataReferenceHandler` uses the same for reading this value.

Details the edc exception message in `IdsMultipartSender::checkResponseType`.

## Why it does that

The IDS object mapper should only be used for de/serializing ids objects, not edc objects.

## Further notes

--

## Linked Issue(s)

Relates #1781 

## Checklist

### DIFF
--- a/data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/IdsMultipartDispatcherServiceExtension.java
+++ b/data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/IdsMultipartDispatcherServiceExtension.java
@@ -78,6 +78,7 @@ public class IdsMultipartDispatcherServiceExtension implements ServiceExtension 
         var idsWebhookAddress = idsApiConfiguration.getIdsWebhookAddress();
 
         var objectMapper = context.getTypeManager().getMapper("ids");
+        var typeManager = context.getTypeManager();
 
         var dispatcher = new IdsMultipartRemoteMessageDispatcher();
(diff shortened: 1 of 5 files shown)


ndr-brt (commented):
Why couldn't be the `EndpointDataReferenceHandler` use the same mapper (the ids specific) as the sender? I know that's not an ids specific message, but using the same mapper thorugh all the senders/handlers would make a lot of sense

juliapampus:
> Why couldn't be the `EndpointDataReferenceHandler` use the same mapper (the ids specific) as the sender? I know that's not an ids specific message, but using the same mapper thorugh all the senders/handlers would make a lot of sense

My first thought was that only IDS objects should be serialized to JSON-LD because that is currently the "protocol specification". If you want to use (and deserialize) the payload of the IDS message elsewhere (other extension), it would not be clean to have to use an IDS `ObjectMapper` there. Same goes for, when the IDS protocol changes and methods expect the reference still as JSON-LD just as leftover, but not for a particular reason. But I also understand your point. Don't know.

ronjaquensel (approved):
I agree that using two different object mappers in the same class is not an ideal solution, but for the reasons you described @juliapampus I feel it would be cleaner to only (de)serialize IDS objects with the IDS object mapper. Also, the JSON-LD generated by the IDS object mapper would contain a reference to the `idsa` context, right? This context wouldn't contain the EDR entity, which also doesn't seem right then.

juliapampus:
> Also, the JSON-LD generated by the IDS object mapper would contain a reference to the `idsa` context, right? This context wouldn't contain the EDR entity, which also doesn't seem right then.

Right, this info was missing :D so for me it makes no sense to de/serialize an edc object to ids. @ndr-brt Fine for you?

ndr-brt (approved):
Let's go :+1:

PR #1792, opened 2022-08-05
fix: Fix usage of default @Provider methods in IdentityDidCoreExtension

bscholtes1A:
## What this PR changes/adds

Make some `@Provider` methods non-default in `IdentityDidCoreExtension` as this is causing non-deterministic dependency injection in `IdentityHub` project.

## Why it does that

Default `@Provider` methods depending on injected services might be invoked while these services are not yet initialized.

## Checklist


docs/developer/dependency_resolution.md
```diff
@@ -111,9 +111,13 @@ public class DefaultsExtension implements ServiceExtension {
 Provider methods configured with `isDefault=true` are only invoked, if the respective service (here: `FooStore`) is not
 provided by any other extension.
 
+> **WARNING**: Please do not abuse of `isDefault=true`! Especially we strongly discourage injecting services in default
```
paullatzelsperger, line 114: How about:
> "The `isDefault` attribute should be used sparingly, it typically is not needed for  technology extensions. Default provider methods are invoked during the _inject_-phase, i.e. while resolving `@Inject`-ed fields, so the code inside the provider method should not reference other fields, that are also `@Inject`-ed because they might not yet be initialized.
Their intended use is to provide default implementations for services, not to realize extensibility."

extensions/iam/decentralized-identity/identity-did-core/src/main/java/org/eclipse/dataspaceconnector/iam/did/IdentityDidCoreExtension.java
```diff
@@ -109,29 +108,19 @@ public DidStore defaultDidDocumentStore() {
...
-    }
-
     private void registerParsers(PrivateKeyResolver resolver) {
 
         // add EC-/PEM-Parser
-        resolver.addParser(ECKey.class, (encoded) -> {
+        resolver.addParser(ECKey.class, encoded -> {
             try {
-                return (ECKey) ECKey.parseFromPEMEncodedObjects(encoded);
+                return (ECKey) JWK.parseFromPEMEncodedObjects(encoded);
```
marcgs, line 116: How is this related to this change?

extensions/iam/decentralized-identity/identity-did-core/src/main/java/org/eclipse/dataspaceconnector/iam/did/DidResolverRegistryExtension.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.eclipse.dataspaceconnector.iam.did.resolution.DidResolverRegistryImpl;
+import org.eclipse.dataspaceconnector.iam.did.spi.resolution.DidResolverRegistry;
+import org.eclipse.dataspaceconnector.spi.system.Provider;
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtension;
+
+
+public class DidResolverRegistryExtension implements ServiceExtension {
+
+    @Provider
```
algattik, line 25: now I am confused, shouldn't we use `isDefault=true` here? Otherwise it seems to me there is no point configuring this service through DI, since extenders cannot provide another implementation? -> in that case we can just use `new` in the extension that requires the service

bscholtes1A: Updated, thanks!

paullatzelsperger (approved):

PR #1797, opened 2022-08-08
feat: split Dtos in InputDto and OutputDto for Asset and ContractDefinition

### LINKED ISSUE #1796: Separate `ContractDefinitionDto` and `AssetDto` in input and output parts
# Feature Request

Currently, `ContractDefinitionDto` and `AssetDto` contain the `createdAt` attribute, this is misleading as the openapi documentation reports that the field is part of the input payload on the `POST /` calls, but that's not true, as it will be valued by the EDC.

The proposal is to split the `Dto`s in two parts: `InputDto` and `OutputDto`, the first won't contain the `createdAt` field, the latter will.

data management api

Improve api documentation, provide a separation from input and output dto that, in fact, are two different objects.

## Solution Proposal
The current `BaseDto` object could be duplicated and renamed in `BaseInputDto` and `BaseOutputDto`, the first won't contain the `createdAt` field.

## Type of Issue
improvement

## Checklist

ndr-brt:
## What this PR changes/adds

Split 
- `AssetDto` in `AssetInputDto` and `AssetOutputDto`
- `ContractDefinitionDto` in `ContractDefinitionInputDto` and `ContractDefinitionOutputDto`

## Why it does that

To improve APIs, as currently the spec are stating that a `createdAt` field can be passed from the client, but that's not true as that field is valued by the EDC itself

## Further notes

- remove duplication between `TransformerContextImpl` and `DtoTransformerContext` that were the pretty much the same class.
- improve validation on the `id`s, as they shouldn't be mandatory from the client but they should be created by the EDC as random UUID by default (as done for `PolicyDefinition`, `ContractNegotiation`, `TransferProcess` and `DataRequest`)
- make the asset and contract definition transformers "NPE proof". This would lead to a subsequent refactor in another PR as all the transformer should handle the "null input" correctly

## Linked Issue(s)

Closes #1796 

## Checklist


ndr-brt:
> I have not yet looked at this in greater detail, but my inner monk somehow would immediately feel more comfortable if they were named `RequestDto` and `ResponseDto`

@paullatzelsperger makes total sense to me. Done

paullatzelsperger (approved):

PR #1817, opened 2022-08-11
fix: validate offer policy equality with definition policy on negotiation

### LINKED ISSUE #1791: Connector accepts **any** Contract Offer
# Bug Report

## Describe the Bug
A connector accepts any contract offers. The content of the contract offer is not checked.

=> It is possible to change a contract in any way. It is not possible to create a contract for assets, the connector would not have access to. 

### Expected Behavior

- Connector accepts only contracts it has offered before.

### Observed Behavior

- Connector accepts **any** contracts.

## Steps to Reproduce
1. Connector_A: Create Asset

```json
{
  "asset": {
    "properties": {
      "asset:prop:id": "1"
    }
  },
  "dataAddress": {
    "properties": {
      "type": "HttpData",
      "baseUrl": "<url>"
    }
  }
}
```

3. Connector_A: Create Policy

```json
{
  "uid": "1",
  "policy": {
    "permissions": [
      {
        "edctype": "dataspaceconnector:permission",
        "action": {
          "type": "USE"
        },
        "constraints": [
          {
            "edctype": "AtomicConstraint",
            "leftExpression": {
              "edctype": "dataspaceconnector:literalexpression",
              "value": "Pay_Me"
            },
            "rightExpression": {
              "edctype": "dataspaceconnector:literalexpression",
              "value": "1000 Euros"
            },
            "operator": "EQ"
          }
        ]
      }
    ]
  }
}
```

4. Connector_A: Create Contract Definition

```json
{
  "id": "1",
  "criteria": [
    {
      "operandLeft": "asset:prop:id",
      "operator": "=",
      "operandRight": "1"
    }
  ],
  "accessPolicyId": "<something public>",
  "contractPolicyId": "1"
}
```

4. Connector_B: Initiate contract negotiation via POST at `/contractnegotiations` API, but does not want to pay (leaves the constraints out).

```json
{
  "connectorId": "foo",
  "connectorAddress": "${Connector_A}/api/v1/ids/data",
  "offer": {
    "offerId": "1:foo",
    "assetId": "1",
    "policy": {
      "uid": "2",
      "permissions": [
        {
          "edctype": "dataspaceconnector:permission",
          "action": {
            "type": "USE"
          },
          "target": "1",
          "constraints": []
        }
      ]
    }
  }
}
```

5. Check contract agreement in database of Connector_A

| agr_id | asset_id | policy |
|:---|:---|:---|
|1:6e9e400c-a411-4cef-8135-047ffd532061 | 1 | {"permissions":[{"edctype":"dataspaceconnector:permission","uid":null,"target":"1","action":{"type":"USE","includedIn":null,"constraint":null},"assignee":null,"assigner":null,"constraints":[],"duties":[]}],"prohibitions":[],"obligations":[],"extensibleProperties":{},"inheritsFrom":null,"assigner":null,"assignee":null,"target":null,"@type":{"@policytype":"set"}} |

## Context Information

- Used version: Milestone 5

## Detailed Description

Catalog from Connector_A:

```json
{
  "id": "default",
  "contractOffers": [
    {
      "id": "1:95cab8a1-b3c5-4d59-b8b7-f8fcc6e4697f",
      "policy": {
        "permissions": [
          {
            "edctype": "dataspaceconnector:permission",
            "uid": null,
            "target": "1",
            "action": {
              "type": "USE",
              "includedIn": null,
              "constraint": null
            },
            "assignee": null,
            "assigner": null,
            "constraints": [
              {
                "edctype": "AtomicConstraint",
                "leftExpression": {
                  "edctype": "dataspaceconnector:literalexpression",
                  "value": "Pay_Me"
                },
                "rightExpression": {
                  "edctype": "dataspaceconnector:literalexpression",
                  "value": "1000 Euros"
                },
                "operator": "EQ"
              }
            ],
            "duties": []
          }
        ],
        "prohibitions": [],
        "obligations": [],
        "extensibleProperties": {},
        "inheritsFrom": null,
        "assigner": null,
        "assignee": null,
        "target": null,
        "@type": {
          "@policytype": "set"
        }
      },
      "asset": {
        "properties": {
          "asset:prop:description": "Product EDC Demo Asset",
          "ids:byteSize": null,
          "asset:prop:id": "1",
          "ids:fileName": null
        }
      },
      "assetId": null,
      "provider": "urn:connector:provider",
      "consumer": "urn:connector:consumer",
      "offerStart": null,
      "offerEnd": null,
      "contractStart": null,
      "contractEnd": null
    }
  ]
}
```

## Possible Implementation

-

ndr-brt:
## What this PR changes/adds

Add an equality check between contract offer policy and contract negotiation policy in `ContractValidationService`

## Why it does that

To avoid injection of malicious policies.

## Further notes

- this check is needed since currently the negotiation is automatically, and no counter offers are supported
- the equality check is done using the `ObjectMapper`'s `ObjectNode` as it is way less disruptive than overriding the `equals` method on every policy model class 
- the `target` has been excluded from the check for reasons explained in the issue: #1791 
- in `ContractValidationServiceImpl` the way the target asset is retrieved has been simplified, since the `id` must always be present, it can be used as the only criteria to get the `Asset`
- I'm not a supporter of "unhappy path e2e tests" but I needed this to light the way to the solution, I would keep it as it is a critical case, but it could be deleted in the future.
- **question**: is the `assetId` in the `ContractOffer` still useful? seems like isn't used by anoyone. I left it as it is, but in my opinion it could be removed.

## Linked Issue(s)

Closes #1791 

## Checklist

### DIFF
--- a/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractServiceExtension.java
+++ b/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/ContractServiceExtension.java
@@ -25,6 +25,7 @@ import org.eclipse.dataspaceconnector.contract.observe.ContractNegotiationObserv
 import org.eclipse.dataspaceconnector.contract.offer.ContractDefinitionServiceImpl;
 import org.eclipse.dataspaceconnector.contract.offer.ContractOfferServiceImpl;
 import org.eclipse.dataspaceconnector.contract.policy.PolicyArchiveImpl;
+import org.eclipse.dataspaceconnector.contract.policy.PolicyEquality;
 import org.eclipse.dataspaceconnector.contract.validation.ContractValidationServiceImpl;
 import org.eclipse.dataspaceconnector.spi.EdcSetting;
 import org.eclipse.dataspaceconnector.spi.agent.ParticipantAgentService;
@@ -156,7 +157,8 @@ public class ContractServiceExtension implements ServiceExtension {
         var contractOfferService = new ContractOfferServiceImpl(agentService, definitionService, assetIndex, policyStore);
         context.registerService(ContractOfferService.class, contractOfferService);
 
-        var validationService = new ContractValidationServiceImpl(agentService, definitionService, assetIndex, policyStore, clock, policyEngine);
+        var policyEquality = new PolicyEquality(context.getTypeManager());
+        var validationService = new ContractValidationServiceImpl(agentService, definitionService, assetIndex, policyStore, clock, policyEngine, policyEquality);
         context.registerService(ContractValidationService.class, validationService);
 
         var waitStrategy = context.hasService(NegotiationWaitStrategy.class) ? context.getService(NegotiationWaitStrategy.class) : new ExponentialWaitStrategy(DEFAULT_ITERATION_WAIT);
--- a/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/policy/PolicyEquality.java
+++ b/core/contract/src/main/java/org/eclipse/dataspaceconnector/contract/policy/PolicyEquality.java
@@ -1,0 +1,41 @@
+/*
+ *  Copyright (c) 2022 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.contract.policy;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import org.eclipse.dataspaceconnector.policy.model.Policy;
+import org.eclipse.dataspaceconnector.spi.types.TypeManager;
+
+import java.util.function.BiPredicate;
+
+public class PolicyEquality implements BiPredicate<Policy, Policy> {
+
+    private final ObjectMapper mapper;
+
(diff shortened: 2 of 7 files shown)


jimmarino (commented):
Let's discuss when I'm back. This touches on issues in the IDS protocol that need to be fixed and the concept of a policy infoset that can be used to derive signatures and equality. For now can't we just throw away the policy on the provider and always regenerate it?

ndr-brt:
> Let's discuss when I'm back. This touches on issues in the IDS protocol that need to be fixed and the concept of a policy infoset that can be used to derive signatures and equality. For now can't we just throw away the policy on the provider and always regenerate it?

@jimmarino do you mean to completely ignore the policy in the offer and to use the policy in the definition? Doing like this there will be inconsistency between the negotiation data on the consumer (it'd have the "modified" policy) and ond the provider (it'd have the original one). Wouldn't be better to explicitly decline the negotiation?

DominikPinsel (approved):

jimmarino:
@ndr-brt  The provider should never accept the consumer contract offer, even if it is "electronically equal". It should only accept one it has generated. If the consumer has modified the offer in any way, it is a "bug" or malicious intent on the consumer's part. When we implement signing and non-reputation, we can deal with this negative case more robustly.

ndr-brt:
> @ndr-brt  The provider should never accept the consumer contract offer, even if it is "electronically equal". It should only accept one it has generated. If the consumer has modified the offer in any way, it is a "bug" or malicious intent on the consumer's part. When we implement signing and non-reputation, we can deal with this negative case more robustly.

@jimmarino if we implement that logic (the provider will accept the policy it created, regardless of what the consumer sent) it will mean that the consumer request will rely only on the definition id, and every change to the policy will be accepted but without any error, and that's confusing by a consumer perspective. Is this acceptable? If so, I would open a PR with this behavior to have either this or the other merged asap because we need some kind of restriction for malicious intents for the forthcoming C-X release

paullatzelsperger (approved):
This seems to be a problem in need of a timely fix, so I'm fine with _how_ the change is implemented.
We'll probably want to revisit the topic during/after the upcoming improvements to IDS.

ronjaquensel (approved):

Issue #1809, opened 2022-08-10
Project structure: core modules


ndr-brt:
# Feature Request

All the `core` modules should be put in the `core` folder divided by component

## Which Areas Would Be Affected?
core

## Why Is the Feature Desired?
project structure review

## Type of Issue
_improvement

## Checklist

- [x] assigned appropriate label?
- [x] **Do NOT select a milestone or an assignee!**

PR #1832, opened 2022-08-16
feat: restructure `spi` by component

### LINKED ISSUE #1808: Project structure: spi modules
# Feature Request

All the `spi` modules should be put in the `spi` folder divided by component

spi

project structure review

## Type of Issue
_improvement

## Checklist

ndr-brt:
## What this PR changes/adds

move all the `spi` folders into the `spi` folder, grouped by the component they are needed for.

```
spi
├── common
│   ├── auth-spi
│   ├── catalog-spi
│   ├── core-spi
│   ├── identity-did-spi
│   ├── oauth2-spi
│   ├── transaction-datasource-spi
│   ├── transaction-spi
│   ├── transport-spi
│   └── web-spi
├── control-plane
│   ├── contract-spi
│   ├── control-plane-spi
│   ├── data-plane-transfer-spi
│   ├── policy-spi
│   └── transfer-spi
├── data-plane
│   └── data-plane-spi
├── data-plane-selector
│   └── selector-spi
└── federated-catalog
    └── federated-catalog-spi

```

## Why it does that

## Further notes

- this PR only moves modules, it does not change packages or class names, so it shouldn't conflict with other PRs, as the modifications regard only gradle files
- got rid of the `spi` BOM module as it will lose meaning as no extension should refer to the "generic" spi but to the component-relative one or to a specific one
- `extensions/dataloading` seems to be an `spi` more than an extension, will create an issue to inline it (as it was already discussed)
- `common/token-generation-lib` and `token-validation-lib` seem to contain `spi` code that's actually extended by some extensions, I propose to introduct a `token-spi` module, related issue will be created
- all the `policy` thing is pretty confusing, I will create a discussion about how to improve its structure, now we have:
  - `policy-spi` that only contains two interfaces, one used only by the transfer process, and the other is the repository for `PolicyDefinition`
  - `core-spi`/`policy` package, that contains interfaces for the policy-engine
  - `policy-evaluator` that contains all the policy related model classes (no logic)
  - `policy-engine` that contains the `PolicyEvaluator` class and accessors
  - `base-core`/`policy` package, that contains the policy-engine implementation
- My gut feeling is that `core-spi` contains too different kind of classes, need to be reviewed, probably a better view will come when #1809 will be done

## Linked Issue(s)

Closes #1808 

## Checklist


ronjaquensel (approved):

paullatzelsperger (approved):
It won't make much sense going through this file-by-file, so I'll just address your notes:
- `extensions/dataloading` could be deleted altogether (in a separate PR), as far as I know we're not really using it. 
- I'm not too keen on those "lib" modules because they tell little about the content: does it contain SPIs? does it contain impls? does it contain models? maybe we can correct that when we're extracting the `token-spi`.
- IIRC the `policy-evaluator` "lib" was intended to have zero dependencies onto other EDC modules, that's why it is so isolated. The idea was that it could theoretically be used as a standalone lib in other projects. If that is not relevant anymore, we could merge it with the rest of the policy module(s).
- agree on `core-spi`, especially all the model classes 
- FYI: the federated catalog stuff will get moved out to its own repo in the mid future

PR #1841, opened 2022-08-17
feat: remove h2 from all tests

### LINKED ISSUE #1754: Housekeeping: make Postgres tests also run the H2-based tests
# Feature Request

For many SQL-implementations we have two test classes: one that uses H2, and one that uses an actual Postgres instance. This was done to enable the SQL test to run locally and very fast without an actual Postgres instance.

Due to to fact how JUnit Test tags work, the Postgres test cannot extend the H2 test, because then it would always run, even if the `-DincludeTags="PostgresqlDbIntegrationTest"` is not specified, thus requiring all developers to have a local Postgres instance.

We have observed on multiple occasions that bugs occur on Postgres, and not on H2, especially when using Postgres-specific features, such as the `JSON` datatype.

build, CI

Test coverage, test consistency

## Solution Proposal
There are basically two ways I can think of to solve this:
1. Move the actual test code into a "test method" class, and invoke all methods from the test classes:
   ```java
   @Test
   void someTest(){
       testMethods.runSomeTest();
   }
   ``` 
   The upside is that test code is centrally located. The downside is that we still have to duplicate all test methods.
1. ditch H2, run everything on Postgres. The upside of this is that we always have realistic tests, don't need to run them twice. The downside is, that for locally running the test, users have to run Postgres and specify the `-DincludeTags="PostgresqlDbIntegrationTest"` flag

As a semi-related issue we could move all the Postgres initialization code into a JUnit `PostgresExtension` and add an `@ExtendWith` on the test class.

## Type of Issue
cleanup

## Checklist

paullatzelsperger:
## What this PR changes/adds

removes all usages of H2 from all test code.

## Why it does that

Executing some tests against H2, and others against Postgres has the inherent risk of causing inconsistencies, therefor it was decided (#1839) to remove H2 entirely.

## Further notes

- test scenarios are defined in the abstract base class for the tests
- H2 is still used in the `common-sql` and `transaction` modules, as they are truly independent of database technology and simply deal with JDBC.

## Linked Issue(s)

Closes #1754 

## Checklist


ndr-brt (approved):
Just a note: at this point these abstract `*StoreTest` could be used to test in memory and cosmos as well? They seems pretty agnostic and at the end the behavior we want out of the different implementation is the same.

paullatzelsperger:
> Just a note: at this point these abstract `*StoreTest` could be used to test in memory and cosmos as well? They seems pretty agnostic and at the end the behavior we want out of the different implementation is the same.

that's correct, then we could dub it a "test specification" and probably the correct location for it would probably be in the same module as the *store itself, right?

ndr-brt:
> > Just a note: at this point these abstract `*StoreTest` could be used to test in memory and cosmos as well? They seems pretty agnostic and at the end the behavior we want out of the different implementation is the same.
> 
> that's a fine idea, then we could dub it a "test specification" and probably the correct location for it would probably be in the same module as the *store itself, right? I'll look into that.

I would say so, so everyone that wants to provide an implementation will just need to extend that class and make the implementation work accordingly

ronjaquensel (approved):

paullatzelsperger:
> > > Just a note: at this point these abstract `*StoreTest` could be used to test in memory and cosmos as well? They seems pretty agnostic and at the end the behavior we want out of the different implementation is the same.
> > 
> > 
> > that's a fine idea, then we could dub it a "test specification" and probably the correct location for it would probably be in the same module as the *store itself, right? I'll look into that.
> 
> I would say so, so everyone that wants to provide an implementation will just need to extend that class and make the implementation work accordingly

Created https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/1845, because this will bring with quite a few changes:
- somehow consolidate `AssetIndex`, `AssetLoader` and `DataAddressResolver`
- move the abstract test classes into the `testFixtures` directory of the SPI module
- add a dependency onto that
- potentially fix quite a few tests
- ...

Issue #1596, opened 2022-07-05
Project structure review


ndr-brt:
# Feature Request

Review the overall project structure

## Which Areas Would Be Affected?
all

## Why Is the Feature Desired?
> to have a more clear division from a component perspective. 

## Solution Proposal
The current EDC project structure reached a state where some things cannot be obtained (like permitting "extenders" to implement a custom api layer using the core services) and in general a not-so-clear state.

My proposal would be to have a more clear division from a component perspective. 
> A component is a particular EDC combination of dependencies that respond to a particular need, e.g.: `control-plane`, `data-plane`, `data-plane-selector`.
This would help developers to extricate themselves in the codebase more easily and separate responsibilities in a better way.

### Spi
The `spi` module should be refactored as:
```
spi/catalog-spi
spi/control-plane
spi/control-plane/asset-spi
spi/control-plane/contract-spi
spi/control-plane/policy-spi
spi/control-plane/transfer-spi
spi/core-spi
spi/data-plane-spi
spi/data-plane-selector-spi
spi/transport-spi
spi/web-spi
```
Every component should have its own `spi` module, then other "common" spi could exist as well (like `core`, `catalog`, `transport` and `web`)

Some details:
- the `asset-spi` would be extracted from the `core-spi`
- the `*Service` interfaces should be moved to the `control-plane/*-spi`

### Core
Responsibility of the core modules should be to provide a working base version of every component, so it should provide a default implementation for every interface present in the relative `spi`module, so these three module will be introduced:
- `control-plane-core`: would contain all the default implementations of the stores and the services and an extension that register them as default
- `data-plane-core`: is the current `data-plane-framework` extension, will register all the default implementations
- `data-plane-selector-core`: is the current `selector-core` module

All of them should extend the `base-core` module. The `defaults` module would probably disappear (replaced by the `<component>-core` ones)

### Extensions

This separation could be raised also in the `extensions` folder, as some extensions are needed only by one of the components, for example:
- `control-plane`: data management api, provisioners (s3, blobstorage, http...), "stores" sql/cosmos implementations (for assets, contracts, transfer process, policies...), data-plane-selector client, ...
- `data-plane`: implementations of the source/sink interfaces like `data-plane-http`, `data-plane-s3`, and so on.., data plane specific api (`data-plane-api` module)
- `data-plane-selector`: implementations of the `DataPlaneInstanceStore`
[...]

PR #1842, opened 2022-08-17
feat: restructure `core` by component

### LINKED ISSUE #1809: Project structure: core modules

ndr-brt:
## What this PR changes/adds

move all the `core` folders into the `core` folder, grouped by the component they are needed by:
```
core
├── common
│   ├── base
│   ├── boot
│   ├── policy-engine
│   └── policy-evaluator
├── control-plane
│   ├── contract
│   ├── control-plane-core
│   └── transfer
├── data-plane
│   ├── data-plane-core
│   └── data-plane-framework
├── data-plane-selector
│   └── data-plane-selector-core
└── federated-catalog
    ├── federated-catalog-cache
    └── federated-catalog-core
```

## Why it does that

## Further notes

- All the chickens are coming home to roost :chicken: :rooster: :chicken: 
- As already reported in #1832: we need to untangle the policy knots, let's talk about this in the discussion: #1836
- I'm not really sure why `boot` and `base` are separated, they could be the same module, and they are used pretty much always together (unifying them plus refactoring the policy ones could avoid that ugly `common` folder)
- inlined `selector-store` into `selector-core`, as it was only a default provider for the `DefaultDataPlaneInstanceStore`
- inlined `catalog` and substituted by `federated-catalog-core`
- introduced `control-plane-core` and inlined `defaults`, as it was pretty straightforward. Seems like the `core` module of a specific component is the right place where to declare default components to-be registered by extensions
- I have a doubt about `selector-api` and `selector-client`, they are extensions used to build a `data-plane-selector`, but at the same time could a data-plane-selector exist without them? if not, I would put them as core modules for `data-plane-selector`, but they are completely fine as exception as they are.

## Linked Issue(s)

Closes #1809 

## Checklist


paullatzelsperger (commented):
mostly addressing your "further notes"

- > All the chickens are coming home to roost 🐔 🐓 🐔

    best comment so far.
- `boot` vs `base`: they're separate because they serve distinct purposes: `boot` takes care of runtime boostrap and extension loading, whereas `base` provides some basic (but optional) features. Some runtimes will only require the `boot` module for bootstrapping, but will not need the `base` module. Thus having them separate still makes sense to me.
- `selector-store`: as per the (updated) [documentation](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/docs/developer/default_provider_methods.md), a default provider for a service should be in its own extension, even if it does not use any other injected fields. 
- `control-plane-core`: I would rename the `ControlPlaneCoreExtension` to something like `DefaultStoresExtension` or something to clearly state what it's doing.
- `selector-api` and `selector-client`: architecturally, a `data-plane-selector` should not need them. For example when running the selector embedded. They are only needed when the selector runs in its own process (e.g. as pod in a K8s). If there is currenlty a dependency it should be cleaned up (in another PR, I can help out in that regard)
- `federated-catalog-core` is empty, except for a build file. on purpose? 
- minor nit: in some files the Copyright seems to have been captured by BMW :) e.g.  `ControlPlaneCoreExtension` - the original Copyright is by Microsoft. could have happened in others as well.

ndr-brt:
> * `boot` vs `base`: they're separate because they serve distinct purposes: `boot` takes care of runtime boostrap and extension loading, whereas `base` provides some basic (but optional) features. Some runtimes will only require the `boot` module for bootstrapping, but will not need the `base` module. Thus having them separate still makes sense to me.

ok

> * `selector-store`: as per the (updated) [documentation](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/docs/developer/default_provider_methods.md), a default provider for a service should be in its own extension, even if it does not use any other injected fields.
> * `control-plane-core`: I would rename the `ControlPlaneCoreExtension` to something like `DefaultStoresExtension` or something to clearly state what it's doing.
> * `federated-catalog-core` is empty, except for a build file. on purpose?

These three points are strictly related, in my opinion every component should have a "core" extension that defines the base services and the default ones, for the control plane is `ControlPlaneCoreExtension`, and for the data-plane-selector is `DataPlaneSelectorExtension`.
`data-plane` and `federated-catalog` have no default services, that's why their `-core` module is just a BOM build file.

Having a `-core` module that gathers the base needed dependencies and that provides the default services will permit extenders to just add that dependency on their build file and then extensions on top of it (this idea could be part of #1596).

Seems a clean pattern to me

> * `selector-api` and `selector-client`: architecturally, a `data-plane-selector` should not need them. For example when running the selector embedded. They are only needed when the selector runs in its own process (e.g. as pod in a K8s). If there is currenlty a dependency it should be cleaned up (in another PR, I can help out in that regard)

Ok

> * minor nit: in some files the Copyright seems to have been captured by BMW :) e.g.  `ControlPlaneCoreExtension` - the original Copyright is by Microsoft. could have happened in others as well.

Fixed

paullatzelsperger:
I like the `-core` module pattern, that is fine. 
However:
 
> > * `selector-store`: as per the (updated) [documentation](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/docs/developer/default_provider_methods.md), a default provider for a service should be in its own extension, even if it does not use any other injected fields.

The problem there is that it is the _same_ extension (`DataPlaneSelectorExtension`) that does the `@Inject`-ing and `@Provider`-ing of the same service (i.e. `DataPlaneInstanceStore`), which - according to the documentation - should be avoided. For now it would be sufficient to just have another extension in the `data-plane-selector-core` module, that only contains the provider method. As a future enhancement, that should to be cleaned up further, e.g. by renaming the store to `InMemory-`, and providing it from another module etc.

> > * `federated-catalog-core` is empty, except for a build file. on purpose?
> 
> These three points are strictly related, in my opinion every component should have a "core" extension that defines the base services and the default ones, for the control plane is `ControlPlaneCoreExtension`, and for the data-plane-selector is `DataPlaneSelectorExtension`. `data-plane` and `federated-catalog` have no default services, that's why their `-core` module is just a BOM build file.

OK that's fine, but now we have
```
federated-catalog-cache <-- contains code
federated-catalog-core <-- contains only the bom
```
could we just merge/move everything from `-cache` over to `-core`?

paullatzelsperger (approved):

PR #1867, opened 2022-08-23
feat: introduce jwt-spi and jwt-core modules

### LINKED ISSUE #1834: Project structure: introduce `token-spi` module
# Feature Request

As reported in #1832 , `common/token-generation-lib` and `common/token-validation-lib` contain interfaces that are actually implemented by extensions, so probably they deserve to be put in an `spi` module.

This new `token-spi` module will contain:
- `JwtDecorator`
- `JwtDecoratorRegistry`
- `TokenValidationRule`

As these 3 interfaces are being implemented by different modules

common, spi

cleanup

## Solution Proposal
-

## Type of Issue
cleanup

## Checklist

ndr-brt:
## What this PR changes/adds

refactor `token-validation-lib` and `token-generation-lib` to the `spi`/`core` pattern

## Why it does that

## Further notes
- I called the modules `jwt-spi` and `jwt-core` as they are specific to JWT (using nimbus-jose jwt related classes)
- As pointed out in the issue #1834 there is some duplication that could be avoided, but it was too much for this PR, will be something to be fixed in the future
- With the new project structure seems that the modules in the `common` folder could be moved in `core/common`, as in the end they are core modules as well.

## Linked Issue(s)

Closes #1834 

## Checklist


core/common/jwt-core/build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2022 Amadeus
```
jimmarino, line 2: I think the copyrights need to be additive, not replaced

jimmarino (changes requested):

core/common/jwt-core/build.gradle.kts
ndr-brt: The fact is that I removed the two preexistent `-lib` modules (`token-validation-lib` and `token-generation-lib`) and created `jwt-core`. For some reasons git detected this as a file rename, but, in fact, this is a completely different file.

ndr-brt:
@jimmarino , yesterday @bscholtes1A pointed out to the fact that the `jwt-spi` module I introduced needs the `nimbus-jose` library, I don't know if this is accettable, the alternative would be to introduce some interfaces/classes to represent the JWT concept and replace/wrap the nimbus ones

jimmarino:
> @jimmarino , yesterday @bscholtes1A pointed out to the fact that the `jwt-spi` module I introduced needs the `nimbus-jose` library, I don't know if this is accettable, the alternative would be to introduce some interfaces/classes to represent the JWT concept and replace/wrap the nimbus ones

@ndr-brt In this case, it may be onerous to wrap and unwrap Nimbus classes. Are there a lot of classes that need to be wrapped?

ndr-brt:
> @ndr-brt In this case, it may be onerous to wrap and unwrap Nimbus classes. Are there a lot of classes that need to be wrapped?

In the spi we are using three nimbus-jose classes:
- `JWSHeader`
- `JWTClaimsSet`
- `SignedJWT`

the first two of them are actually used paired on the `JwtDecorator.decorate` method so they can be wrapped by a single class.
I think this is doable, maybe we can address this to another issue for M7, wdyt @jimmarino ?

jimmarino:
> > @ndr-brt In this case, it may be onerous to wrap and unwrap Nimbus classes. Are there a lot of classes that need to be wrapped?
> 
> In the spi we are using three nimbus-jose classes:
> 
> * `JWSHeader`
> * `JWTClaimsSet`
> * `SignedJWT`
> 
> the first two of them are actually used paired on the `JwtDecorator.decorate` method so they can be wrapped by a single class. I think this is doable, maybe we can address this to another issue for M7, wdyt @jimmarino ?

Sounds good to me.

ndr-brt:
#1879 created.
Are we good to go here? @jimmarino @bscholtes1A

bscholtes1A (approved):

PR #1916, opened 2022-09-05
Fix: Add controlplane dependency for transfer onto contract

bcronin90:
## What this PR changes/adds

Add a dependency for controlplane transfer onto contract

## Why it does that

We are having a problem downstream that PolicyArchive is not injected, even though it's required by an upstream dependency:
`Field "policyArchive" of type [interface org.eclipse.dataspaceconnector.spi.policy.store.PolicyArchive] required by org.eclipse.dataspaceconnector.transfer.core.CoreTransferExtension
`
## Further notes

_List other areas of code that have changed but are not necessarily linked to the main feature. This could be method signature changes, package declarations, bugs that were encountered and were fixed inline, etc._

## Linked Issue(s)

## Checklist

<sub>Brendan Cronin brendan.cronin@mercedes-benz.com, Mercedes Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/LEGAL_IMPRINT.md)</sub>

### DIFF
--- a/core/control-plane/transfer/build.gradle.kts
+++ b/core/control-plane/transfer/build.gradle.kts
@@ -22,6 +22,7 @@ plugins {
 dependencies {
     api(project(":spi:control-plane:policy-spi"))
     api(project(":spi:control-plane:transfer-spi"))
+    api(project(":core:control-plane:contract"))
     implementation(project(":common:state-machine-lib"))
     implementation(project(":common:util"))
     implementation("io.opentelemetry:opentelemetry-extension-annotations:${openTelemetryVersion}")


bcronin90:
@ndr-brt Please tell me if this makes any sense at all. I'm having a bit of an issue correctly identifying all the edc-internal dependency structures.

jimmarino (changes requested):
The correct way to do this is at the launcher level.

Transfer core should not depend on the core contract infrastructure. `PolicyArchive` was introduced as a way to decouple the implementations. For collocated control plane/DPF deployments, the two core modules will be included at the launcher level. 

For separate process deployments, we need to implement a standalone version of `PolicyArchive`, which hasn't been done yet. That implementation would be loaded by the launcher module as well.

PR #1919, opened 2022-09-06
refactor: remove inheritance for multipart senders

### LINKED ISSUE #1697: Refactor IdsMultipartSender
My idea would be quite different...
currently the behaviors that are described by abstract methods in the `IdsMessageSender` are:
- `retrieveRemoteConnectorAddress`
- `buildMessageHeader`
- `buildMessagePayload`
- `getResponseContent`

The first one is straightforward, and I think a `RemoteMessage` should know about the destination, so it will be enough to add a `getRemoteConnectorAddress` method on the `RemoteMessage` interface.

For the three others, I would start the refactor extracting an interface called like `MultipartSenderDelegate` (or something better :D ) with these three methods, and I would make all the `Senders` implement it.
At this point a delegate could be injected into the `IdsMultipartSender` class, that would stop being an abstract class.

_Originally posted by @ndr-brt in https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/1607#issuecomment-1179035961_

ronjaquensel:
## What this PR changes/adds

Refactors the `IdsMultipartSender` and its sub-classes. The former sub-classes now implement the `MultipartSenderDelegate` interface and are injected into the `IdsMultipartSender`, which is now not abstract anymore.

## Further notes

Removes the interfaces `IdsMessageSender` and `IdsRemoteMessageDispatcher`, as they only had the implementation for IDS Multipart. May be reintroduced in some form, when an additional IDS protocol is added.

## Linked Issue(s)

Closes #1697 

## Checklist


data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/MultipartSenderDelegate.java
```diff
@@ -0,0 +1,72 @@
...
+    Message buildMessageHeader(M request, DynamicAttributeToken token) throws Exception;
+
+    /**
+     * Builds the IDS multipart payload for the request.
+     *
+     * @param request the request.
+     * @return the message payload.
+     * @throws Exception if building the message payload fails.
+     */
+    default String buildMessagePayload(M request) throws Exception {
```
ndr-brt, line 51: I would avoid the `default` implementation here and let the subclasses overriding it explicitly

ndr-brt, line 51: I think the signature for now works better, but I noticed that every subclass in fact serialize the body to json, that should be something in charge of the `IdsMultipartSender`, because these delegates should be protocol-agnostic, and the serialization method is definitely protocol-dependent.
However I'm aware that currently is not possible (because the EDR not being an IDS object) and probably not needed as far as we don't support other protocols. But this is something we could take in consideration for the future.

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/IdsMultipartDispatcherServiceExtension.java
```diff
@@ -80,14 +83,16 @@ public void initialize(ServiceExtensionContext context) {
...
+        dispatcher.register(new IdsMultipartSender<>(monitor, httpClient, identityService, objectMapper, new MultipartArtifactRequestSender(senderContext, vault)));
+        dispatcher.register(new IdsMultipartSender<>(monitor, httpClient, identityService, objectMapper, new MultipartDescriptionRequestSender(senderContext)));
+        dispatcher.register(new IdsMultipartSender<>(monitor, httpClient, identityService, objectMapper, new MultipartContractOfferSender(senderContext)));
+        dispatcher.register(new IdsMultipartSender<>(monitor, httpClient, identityService, objectMapper, new MultipartContractAgreementSender(senderContext)));
+        dispatcher.register(new IdsMultipartSender<>(monitor, httpClient, identityService, objectMapper, new MultipartContractRejectionSender(senderContext)));
+        dispatcher.register(new IdsMultipartSender<>(monitor, httpClient, identityService, objectMapper, new MultipartCatalogDescriptionRequestSender(senderContext)));
+        dispatcher.register(new IdsMultipartSender<>(monitor, httpClient, identityService, objectMapper, new MultipartEndpointDataReferenceRequestSender(senderContext, typeManager)));
```
ndr-brt, line 95: In the future would be great to clearly separate the "sender" from the "delegates" so that the dispatcher would be able to choose the sender by the protocol and the delegate by message type.
Not needed now, just to see if we're on the same track.

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/IdsMultipartSender.java
```diff
@@ -101,7 +91,7 @@ private static URI createConnectorIdUri(String connectorId) {
      */
     @Override
     public CompletableFuture<R> send(M request, MessageContext context) {
```
ndr-brt, line 93: the `MessageContext` is not used, may worth removing it to keep the interface simpler

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/type/MultipartArtifactRequestSender.java
```diff
@@ -140,12 +124,12 @@ protected String buildMessagePayload(DataRequest request) throws Exception {
...
-    protected MultipartResponse<String> getResponseContent(IdsMultipartParts parts) throws Exception {
-        return parseMultipartStringResponse(parts, getObjectMapper());
+    public MultipartResponse<String> getResponseContent(IdsMultipartParts parts) throws Exception {
+        return parseMultipartStringResponse(parts, context.getObjectMapper());
     }
 
     @Override
-    protected List<Class<? extends Message>> getAllowedResponseTypes() {
+    public List<Class<? extends Message>> getAllowedResponseTypes() {
         return List.of(ResponseMessageImpl.class, RequestInProcessMessageImpl.class); // TODO remove ResponseMessage.class
```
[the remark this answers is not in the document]
ronjaquensel: good catch, could be removed now.

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/MultipartSenderDelegate.java
ronjaquensel: done

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/IdsMultipartSender.java
ronjaquensel: done

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/IdsMultipartDispatcherServiceExtension.java
ronjaquensel: Do you mean that we'd have one dispatcher for IDS messages that keeps a list of senders for the protocol and a list of delegates for the message type? I'm not sure the same delegates can be used for different IDS protocols, as e.g. for IDS REST the IDS message header information would be mapped to HTTP headers. But maybe we could rewrite the `IdsMultipartRemoteMessageDispatcher` to choose the delegate depending on the message type and hand that over to the `IdsMultipartSender` as a parameter, what do you think?

ndr-brt: @ronjaquensel Maybe I went too far with the speculation, it's something that could raise when another protocol will pop up (if it ever would).
In the meantime I think your proposal to choose the delegate depending on the message type and sending it to the `IdsMultipartSender` makes sense, doing that we we'll end up with a single sender instance and the dispatcher would just choose the delegate.

ronjaquensel: refactored it

spi/common/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/message/RemoteMessageDispatcher.java
```diff
@@ -36,6 +36,6 @@
      * @param context      the message context
      * @return a future that can be used to retrieve the response when the operation has completed
      */
-    <T> CompletableFuture<T> send(Class<T> responseType, RemoteMessage message, MessageContext context);
+    <T, M extends RemoteMessage> CompletableFuture<T> send(Class<T> responseType, M message, MessageContext context);
```
paullatzelsperger, line 39: `responseType` and `context` are unused. could we remove them?

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/MultipartSenderDelegate.java
paullatzelsperger: well, some of the implementors don't serialize anything (i.e. return `null`), others simply return a non-formatted string (`MultipartContractRejectionSender` returns a reason phrase). 
However I agree that the concrete serialization should be a) harmonized and b) externalized.

Doesn't the `TypeManager` already have the ability to serialize to different serialization contexts?

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/SenderDelegateContext.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Contains services and context information required by sender delegate classes.
+ */
+public class SenderDelegateContext {
+    private final URI connectorId;
+    private final ObjectMapper objectMapper;
+    private final IdsTransformerRegistry transformerRegistry;
+    private final String idsWebhookAddress;
+
+    public SenderDelegateContext(@NotNull URI connectorId,
```
paullatzelsperger, line 32: minor nit: the `@NotNull` is purely for IntelliJ's code inspection feature. Unless there is a `Objects.requireNotNull` assertion in code, nothing will in fact be checked at runtime.

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/test/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/IdsMultipartSenderTest.java
```diff
@@ -47,9 +45,9 @@ void should_fail_if_token_retrieval_fails() {
 
         var objectMapper = IdsTypeManagerUtil.getIdsObjectMapper(new TypeManager());
 
-        var sender = new TestIdsMultipartSender("any", mock(OkHttpClient.class), objectMapper, mock(Monitor.class), identityService, mock(IdsTransformerRegistry.class));
+        var sender = new IdsMultipartSender(mock(Monitor.class), mock(OkHttpClient.class), identityService, objectMapper);
 
-        var result = sender.send(new TestRemoteMessage(), () -> "any");
+        var result = sender.send(new TestRemoteMessage(), new TestSenderDelegate());
```
paullatzelsperger, line 50: Could you use a `mock` instead of the `TestSenderDelegate`? Seems it does nothing, and a mock would make the intent clearer IMO.

spi/common/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/message/RemoteMessageDispatcher.java
ronjaquensel: If we remove them from the dispatcher, they should probably also be removed from the dispatcher registry. `context` we can definitely remove. When removing `responseType`, we would lose the type of the completable future returned by the `send` method, meaning we'd always get back a `CompletableFuture<Object>`. Since we only have one implementation of `RemoteMessageDispatcher` at the moment, this shouldn't be a problem, but we may need to reconsider this when another protocol is added, to ensure that the expected response type is returned.

@paullatzelsperger what do you think?

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/MultipartSenderDelegate.java
ronjaquensel: @paullatzelsperger The `TypeManager` can serialize to different contexts. The problem when moving the serialization to the `IdsMultipartSender` is that it would then need to decide which context to use depending on the delegate. This also doesn't seem right to me.

spi/common/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/message/RemoteMessageDispatcher.java
ndr-brt: I think the return type can be obtained at runtime, so in fact the `responseType` currently does nothing, if we remove it, the dispatchment should work fine, in the `CatalogServiceImpl`, that's the only place where we actually bind a return type that's not `Object`, the type it's set from the method return type that's `CompletableFuture<Catalog>`, so at the end that will be the type returned. An eventual cast exception could be thrown in the `IdsMultipartRemoteMessageDispatcher` when the delegate is obtained, but that could happen also in the current main branch so I think it's ok to remove it.
Maybe it could be done in another PR to not put too much effort in this PR that brings already a lot of value.

ndr-brt (approved):
To me we are good to go, some cleanup could be done but I would address that in further PRs

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/test/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/IdsMultipartSenderTest.java
ronjaquensel: done

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/SenderDelegateContext.java
ronjaquensel: good call, added non-null checks

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/dispatcher/sender/IdsMultipartSender.java
```diff
@@ -61,12 +61,13 @@ public class IdsMultipartSender {
...
 
-    public IdsMultipartSender(@NotNull Monitor monitor, @NotNull OkHttpClient httpClient, @NotNull IdentityService identityService,
+    public IdsMultipartSender(@NotNull Monitor monitor, @NotNull OkHttpClient httpClient,
+                              @NotNull IdentityService identityService,
                               @NotNull ObjectMapper objectMapper) {
-        this.monitor = monitor;
-        this.httpClient = httpClient;
-        this.identityService = identityService;
-        this.objectMapper = objectMapper;
+        this.monitor = Objects.requireNonNull(monitor);
```
ndr-brt, line 67: I lost this modification, I would remove this checks and the `@NotNull` annotation as they provide no value, the not-null constraint is enforced by the injection model at the extension level

ronjaquensel: removed them

ronjaquensel:
```
I think the return type can be obtained at runtime, so in fact the responseType currently does nothing, if we remove it, the dispatchment should work fine, in the CatalogServiceImpl, that's the only place where we actually bind a return type that's not Object, the type it's set from the method return type that's CompletableFuture<Catalog>, so at the end that will be the type returned. An eventual cast exception could be thrown in the IdsMultipartRemoteMessageDispatcher when the delegate is obtained, but that could happen also in the current main branch so I think it's ok to remove it.
Maybe it could be done in another PR to not put too much effort in this PR that brings already a lot of value.
```

@ndr-brt from my side I'd be fine with doing the refactoring of the dispatcher and registry interfaces in another PR. I would then open an issue for this.

@paullatzelsperger also fine for you?

Only question I have to merge is regarding the CodeQL warnings (2 because of the unused parameters and 1 because the remote address is just taken from user input). They can be disabled, but as far as I understand that would disable them for the whole project, not just this PR (not sure we want that), otherwise they would probably pop up again in the pipeline on the main branch. Any opinions on this?

paullatzelsperger:
> @paullatzelsperger also fine for you?

yes

> Only question I have to merge is regarding the CodeQL warnings (2 because of the unused parameters and 1 because the remote address is just taken from user input). They can be disabled, but as far as I understand that would disable them for the whole project, not just this PR (not sure we want that), otherwise they would probably pop up again in the pipeline on the main branch. Any opinions on this?

I would go ahead and try it. Not that big of a deal either way, IMO.

paullatzelsperger (approved):

PR #1950, opened 2022-09-12
feat(test): harmonize ContractDefinitionStore tests to use a common te…

### LINKED ISSUE #1939: Subtask: Consolidate tests for `ContractDefinitionStore`

wolf4ood:
## What this PR changes/adds

Add a common base test class for `ContractDefinitionStore` which all implementors should extends

## Why it does that

To provide a "least common denominator" for tests, i.e. a minimum subset of test scenarios for each `ContractDefinitionStore`.

## Further notes

* The impl `ContractDefinitionStore#deleteById` of `CosmosContractDefinitionStore` has been changed by returning null instead of throwing an exception if the `ContractDefinition` is not found
* The following base tests were ignored in InMemory and Comos impl since they do not support yet querying nested fields.
   * find_queryMultiple
   * find_queryBySelectorExpression_left
   * find_queryBySelectorExpression_right
   * find_queryBySelectorExpression_entireCriterion
* The following base tests were ignored in the Posgres impl since it does not support yet sorting.
  * findAll_verifySorting_invalidProperty
  * findAll_verifySorting

## Linked Issue(s)

Closes #1939 

## Checklist


extensions/control-plane/store/cosmos/contract-definition-store-cosmos/src/main/java/org/eclipse/dataspaceconnector/contract/definition/store/CosmosContractDefinitionStoreExtension.java
```diff
@@ -47,7 +47,7 @@ public void initialize(ServiceExtensionContext context) {
         var client = clientProvider.createClient(vault, configuration);
         var cosmosDbApi = new CosmosDbApiImpl(configuration, client);
 
-        var store = new CosmosContractDefinitionStore(cosmosDbApi, context.getTypeManager(), (RetryPolicy<Object>) context.getService(RetryPolicy.class), configuration.getPartitionKey());
+        var store = new CosmosContractDefinitionStore(cosmosDbApi, context.getTypeManager(), (RetryPolicy<Object>) context.getService(RetryPolicy.class), configuration.getPartitionKey(), context.getMonitor());
```
paullatzelsperger, line 50: minor nit: since you're touching this line of code, you could convert the `context.get...` calls to `@Inject`-ed members.

spi/control-plane/contract-spi/src/testFixtures/java/org/eclipse/dataspaceconnector/contract/offer/store/ContractDefinitionStoreTestBase.java
```diff
@@ -0,0 +1,464 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.eclipse.dataspaceconnector.contract.offer.store.TestFunctions.createContractDefinition;
+import static org.eclipse.dataspaceconnector.contract.offer.store.TestFunctions.createContractDefinitions;
+import static org.eclipse.dataspaceconnector.spi.asset.AssetSelectorExpression.SELECT_ALL;
+
+public abstract class ContractDefinitionStoreTestBase {
+
+    public ContractDefinitionStoreTestBase() {
+        System.setProperty("contractdefinitionstore.supports.nestedfields", String.valueOf(supportNestedFields()));
```
paullatzelsperger, line 51: Here, the situation is, that the Cosmos impl uses a cache, which in turn uses utility that uses Reflection to construct the `Predicate` for matching items to the query. That utility cannot (yet) match collection types, only scalars. Thus, it is not about nested fields here, but rather about collection types. Could you adapt the property name accordingly?

I'll create another issue to follow up on this.

paullatzelsperger (commented):
A few minor things. Also, please make sure you're using the correct code style plus formatting template, see [here](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/styleguide.md): I saw a few `private` methods above `protected` methods, etc.

paullatzelsperger (approved):

PR #1978, opened 2022-09-16
feat(asset): use vault for sensitive data (#1977)

### LINKED ISSUE #1977: Protect sensitive data when stored persistently and prevent exposing secrets to consumers
### Discussed in https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/discussions/1675

<sup>Originally posted by **DanielaWuensch** July 10, 2022</sup>
# Feature Request
Allow HttpDataAddress elements to be tagged with ```vault:``` and do not store corresponding values in the DB, but instead in the vault. The HttpDataAddress will hold the hash of the value as a value in the DB and when exposed to the consumer.

On the dataplane reverse logic is applied. Once HttpDataAddress contains a ```vault:``` tagged element, the value will be used as key to lookup the real value in the vault. The vault content replaces the address value and the tag is removed.

data mgt api, http data plane

Security Requirement

## Solution Proposal
sensitive data should not be exposed to other connectors if not required

## Type of Issue
non-functional requirement

## Checklist
sensitive data not exposed in data address to consumer connectors
</div>

lholthof:
## What this PR changes/adds

By tagging items in dataAddress with ```vault:``` the value is persisted in the vault and vault's key (hash of value) is used as value for persistence of the address.

On data plane the reverse logic is applied. Once ```vault:``` tagged item is received, the value is used as vault key and the real value is replacing the content in the address. The tag is removed and this might overwrite redundant items, vault contents win.

## Why it does that

To not expose the sensitive data of http data addresses to consumer connectors.

## Linked Issue(s)

Closes #1977

## Checklist

- [x ] formatted title correctly? (_take a look at the [CONTRIBUTING](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/CONTRIBUTING.md#submit-a-pull-request) and [styleguide](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/styleguide.md) for details_)

### DIFF
--- a/spi/common/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/DataAddress.java
+++ b/spi/common/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/DataAddress.java
@@ -10,6 +10,7 @@
  *  Contributors:
  *       Microsoft Corporation - initial API and implementation
  *       Siemens AG - enable read property and return a default value is missing
+ *       SAP SE - use vault for sensitive data
  *
  */
 
@@ -55,6 +56,10 @@ public class DataAddress {
 
     public String getProperty(String key) {
         return properties.get(key);
+    }
+
+    public String setProperty(String key, String value) {
+        return properties.put(key, value);
     }
 
     public String getProperty(String key, String defaultValue) {
--- a/extensions/control-plane/api/data-management/asset-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/AssetApiExtension.java
+++ b/extensions/control-plane/api/data-management/asset-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/AssetApiExtension.java
@@ -11,6 +11,7 @@
  *       ZF Friedrichshafen AG - Initial API and Implementation
  *       Microsoft Corporation - name refactoring
  *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - improvements
+ *       SAP SE - use vault for sensitive data
  *
  */
 
@@ -29,6 +30,7 @@ import org.eclipse.dataspaceconnector.spi.asset.AssetIndex;
 import org.eclipse.dataspaceconnector.spi.contract.negotiation.store.ContractNegotiationStore;
 import org.eclipse.dataspaceconnector.spi.event.EventRouter;
 import org.eclipse.dataspaceconnector.spi.observe.asset.AssetObservableImpl;
+import org.eclipse.dataspaceconnector.spi.security.Vault;
 import org.eclipse.dataspaceconnector.spi.system.Inject;
 import org.eclipse.dataspaceconnector.spi.system.Provides;
 import org.eclipse.dataspaceconnector.spi.system.ServiceExtension;
@@ -64,6 +66,9 @@ public class AssetApiExtension implements ServiceExtension {
     @Inject
     Clock clock;
 
+    @Inject
+    Vault vault;
+
     @Override
     public String name() {
         return "Data Management API: Asset";
@@ -76,7 +81,7 @@ public class AssetApiExtension implements ServiceExtension {
         var assetObservable = new AssetObservableImpl();
         assetObservable.registerListener(new AssetEventListener(clock, eventRouter));
 
-        var assetService = new AssetServiceImpl(assetIndex, contractNegotiationStore, transactionContext, assetObservable);
+        var assetService = new AssetServiceImpl(assetIndex, contractNegotiationStore, transactionContext, assetObservable, vault);
         context.registerService(AssetService.class, assetService);
 
         transformerRegistry.register(new AssetRequestDtoToAssetTransformer());
--- a/extensions/control-plane/api/data-management/asset-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
+++ b/extensions/control-plane/api/data-management/asset-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/asset/service/AssetServiceImpl.java
@@ -9,22 +9,28 @@
  *
  *  Contributors:
  *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *       SAP SE - use vault for sensitive data
  *
  */
 
 package org.eclipse.dataspaceconnector.api.datamanagement.asset.service;
 
+import jakarta.xml.bind.DatatypeConverter;
 import org.eclipse.dataspaceconnector.api.result.ServiceResult;
+import org.eclipse.dataspaceconnector.spi.EdcException;
 import org.eclipse.dataspaceconnector.spi.asset.AssetIndex;
 import org.eclipse.dataspaceconnector.spi.contract.negotiation.store.ContractNegotiationStore;
 import org.eclipse.dataspaceconnector.spi.observe.asset.AssetObservable;
 import org.eclipse.dataspaceconnector.spi.query.Criterion;
 import org.eclipse.dataspaceconnector.spi.query.QuerySpec;
 import org.eclipse.dataspaceconnector.spi.query.QueryValidator;
+import org.eclipse.dataspaceconnector.spi.security.Vault;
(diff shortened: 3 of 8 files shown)


paullatzelsperger:
I do not agree with modifying the existing asset service. The better way of going about this would have been to write an extension and handle this at the persistence layer by creating an asset store that can handle `vauld:` prefixes. The asset service does not need to know about this. 

Also, I'm missing the decision record for this feature, which should have included a technical proposal.

paullatzelsperger (changes requested):
see my comment in the PR. In my opinion this PR needs to be closed and re-written.

jimmarino:
Please refer back to the original discussion where I mentioned that sensitive data should not be stored on the DataAddress. For an example of how to do this in the codebase, have a look at the provisioners.

spi/common/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/DataAddress.java
```diff
@@ -57,6 +58,10 @@ public String getProperty(String key) {
         return properties.get(key);
     }
 
+    public String setProperty(String key, String value) {
```
jimmarino, line 61: This should not be mutable

jimmarino (changes requested):
As I mentioned in the discussion and comments, sensitive data should not be stored in the data address. We already have a practice for dealing with this - see the provisions which pass sensitive data types explicitly.

This PR needs to be adapted to correspond to that approach. Also, there is no need to Hash data when stored in a Vault as we can assume the latter to be cryptographically secure.

bscholtes1A:
Hello,
I do not understand the purpose of this PR. 
The `HttpDataAddress` already has an attribute called `secretName`, which can be used to carry the name of a secret stored in the vault. 

In `HttpRequestParamsSupplier` this `secretName` is used to fetch the secret value from the vault: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/ec694a4b41f82dd8750071e92f75b59bc840e8dd/extensions/data-plane/data-plane-http/src/main/java/org/eclipse/dataspaceconnector/dataplane/http/pipeline/HttpRequestParamsSupplier.java#L114

Using this approach, no sensitive data is never exposed to the consumer.

lholthof:
Hi,

thanks for your feedback. I tried to structure your points a bit in the following enumeration and tried to explain the items.

1)	**Modify existing asset service vs create new extension to intercept on persistence layer.**
Intercepting on persistence layer would solve that sensitive data is not stored in the db, but still the external perspective on an asset would contain such data. As the external perspective is actually changed, I see the asset service is the perfect place to implement this logic and it should not depend on the persistence. In addition, the feature can be used on top of the existing functionality and is optional, hence it is not breaking anything.
2)	**Contribution practise, create DR first.**
We raised a discussion to align before implementing. This alignment seemed to be stuck to me. We scheduled an alignment meeting with Jim, which got cancelled. Therefore, I just put my suggestion together in this PR and I am happy to resume a solution oriented discussion, now with my implemented suggested solution. I appreciate if your collaborate on it and if a DR is required, I am happy to feed my thoughts into it.
3)	**Non-private @Inject fields**
@paullatzelsperger, which field do you mean? The only @Inject-ed field is located in `AssetApiExtension` and injects the `vault`. It’s done with default visibility as all other injection in this class as well. Do you suggest to set it to private anyways?
4)	**Sensitive data should not be stored on Asset/DataAddress**
The nature of every field in a `HttpDataAddress` can be interpreted as sensitive. Every detail represents the internal connection between provider's data-plane and provider's backend-service. Therefore I see a strong need to be able to exclude those details from exposing to the consumer. With this PR everybody can judge to tag the contents with `vault:` and prevent exposure in that way. It can be applied to all the other fields besides `authCode` as well and is not limited to just this one.
5)	**DataAddress should not be mutable**
DataAddress was mutable already before I added `setProperty`, see `setKeyName `and `setType`. Why do you see a need to treat it differently for `setProperty`? 
6)	**Use typed fields to pass sensitive data like provisioners**
As of now the `DataAddress ` just holds a `Map<String, String>` containing the required information to access the backend-service. So I did not see a chance to bring in dedicated types without major/breaking changes.
The provisioners is new to me, can you please point me to any documentation about it? Looking at the code I assume it involves additional communication during the transferprocess creation, which derives e.g. credentials from the backend-service. To me this does not seem to fit the current idea in Catena-X. But maybe I can explore this option in more detail, once I understood.
7)	**Hash**
@jimmarino, I think this is a misunderstanding. The hash is just created as a key for the vault entry. The hash replaced a generated GUID at that stage. The benefit from it is that, considering that one backend might host a lot of different assets, not each asset creates a new vault entry with the same content. The most coordinates in the data address might be the same across a lot of different assets. Same contents will generate the same hash and this prevents redundant storage in the vault and makes local caching of the vault efficient.
8)	**Use HttpDataAddress secretName**
The existing secret mechanism is limited to the `authCode ` only which is not sufficient in my point of view. A URL, proxy-parameters and other headers can be seen as sensitive data as well. 
In addition the existing `secretName` just holds a name of an existing vault entry. @bscholtes1A, do I get this right or is there any documentation out which might help me to understand this better? This means the backend-service registering an asset has to deal with an additional API, once to store the authCode value into the vault and once to create the assets linking the vault key as a `secretName`. With this PR the backend-service can just perform one call to the data mgt api and the asset creation will take care of separating the sensitive data into the vault.

I see benefit in this PR, as it improves security of the current way assets are used within Catena-X. After clarifying the above points I am happy to incorporate your suggestions. I am looking forward to your feedback.

Regards, Lukas

paullatzelsperger:
@lholthof please note:
> Modify existing asset service vs create new extension to intercept on persistence layer.
Intercepting on persistence layer would solve that sensitive data is not stored in the db, but still the external perspective on an asset would contain such data. As the external perspective is actually changed, I see the asset service is the perfect place to implement this logic and it should not depend on the persistence. In addition, the feature can be used on top of the existing functionality and is optional, hence it is not breaking anything.

`DataAddresses` are inherently private, they will never get exposed to the consumer, only `Assets` will. They (DA's) will also never get exposed through a client API. That is why Assets may only carry metadata. Things like database names, table names, blob names, s3 bucket names and the like go into the `DataAddress`, **NOT** the `Asset`. Credentials should be stored out-of-band in a vault. That is a design principle of EDC and we'll not discuss that further. The `AssetService` is therefore not the place to do this, if any sort of special handling is needed (I'm not saying it's a good idea!) this must be done transparently. 

> We raised a discussion to align before implementing. This alignment seemed to be stuck to me. We scheduled an alignment meeting with Jim, which got cancelled. Therefore, I just put my suggestion together in this PR and I am happy to resume a solution oriented discussion, now with my implemented suggested solution. I appreciate if your collaborate on it and if a DR is required, I am happy to feed my thoughts into it.

No need to justify, just to follow procedure. D-R first, implementation later. Every feature of some substance must be preceded by a decision record. There are fundamental problems with your suggestion though, so I would start on a clean sheet with the input you got from various people.

> The nature of every field in a HttpDataAddress can be interpreted as sensitive. Every detail represents the internal connection between provider's data-plane and provider's backend-service. Therefore I see a strong need to be able to exclude those details from exposing to the consumer. With this PR everybody can judge to tag the contents with vault: and prevent exposure in that way. It can be applied to all the other fields besides authCode as well and is not limited to just this one.

When proxying a (private) provider API, the `HttpDataAddress` gets encrypted and wrapped in another `HttpDataAddress` that gets encoded in an `EndpointDataReference`. The first (outer) HDA is there to identify the provider's proxy URL, the inner (encrypted) HDA is unreadable by the consumer, round-trips to the provider and ultimately gets unwrapped/decrypted by the provider. It is used to target the private API. This can be found in the [documentation](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/extensions/control-plane/data-plane-transfer/data-plane-transfer-sync/README.md)

I will say that currently we do not have an _actual_ `DataEncrypter` at the moment, only a `NoopDataEncrypter`, but there is this [issue](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/1504). Maybe that's something you could contribute?

From your comments I get the feeling that a deeper understanding of the code base and the underlying concepts is needed. The PR you brought forward violates key concepts and paradigms of our code base, and will therefore not get merged.

PR #2004, opened 2022-09-23
feat(store): implements sorting in Postgres stores

### LINKED ISSUE #1958: Feature: implement sorting in Postgres stores
# Feature Request

The DataManagement API supports sorting through the `QuerySpec`, but currently only the in-mem and Cosmos stores actually use the `sortField` and `sortOrder` field.

All Postgres implementations

To achieve feature parity and actually support the full DataManagement API at the persistence level.

## Solution Proposal
The `...Statements.java` classes typically have a `createQuery` method, which constructs the `SELECT` statement. To add statement parameters, it uses the `SqlQueryStatement` class.

That `SqlQueryStatement` class could be extended to also add `ORDER BY ... [ASC|DESC]` clause to the SQL select statement, and add the sort field and the sort order to the statement parameters. 

Note: The order-by clause should only be added to the SQL statement, if it was specified in the `QuerySpec`.

## Type of Issue
_i.e., new feature, improvement, cleanup, etc._

## Checklist

wolf4ood:
## What this PR changes/adds

Adds sorting feature to Postgres stores. It translates `sortField` and `sortOrder` of `QuerySpec` into `ORDER BY` clause

## Why it does that

To achieve feature parity and actually support the full DataManagement API at the persistence level.

## Linked Issue(s)

Closes #1958 

## Checklist


extensions/common/sql/common-sql/src/main/java/org/eclipse/dataspaceconnector/sql/translation/SqlQueryStatement.java
```diff
@@ -34,17 +36,22 @@ public class SqlQueryStatement {
...
     private static final String WHERE_TOKEN = "WHERE";
     private static final String AND_TOKEN = "AND";
+
+    private static final String ORDER_BY_TOKEN = "ORDER BY %s %s";
+
     private final String selectStatement;
     private final List<String> whereClauses = new ArrayList<>();
     private final List<Object> parameters = new ArrayList<>();
 
+    private Optional<String> sortField = Optional.empty();
```
paullatzelsperger, line 46: Please do not use `Optional` as field types

extensions/common/sql/common-sql/src/main/java/org/eclipse/dataspaceconnector/sql/translation/TranslationMapping.java
```diff
@@ -54,6 +54,25 @@ public String getStatement(String canonicalPropertyName) {
         return entry.toString();
     }
 
+    /**
+     * Returns the mapping for the input fieldName if it's a String
+     *
+     * @param fieldName The input
+     * @return The column name mapped to the input fieldName
+     */
+    public String getSimpleFieldMapping(String fieldName) {
```
paullatzelsperger, line 63: Why do we need this? Doesn't `getStatement` to the exact same thing?

paullatzelsperger (changes requested):

extensions/common/sql/common-sql/src/main/java/org/eclipse/dataspaceconnector/sql/translation/TranslationMapping.java
wolf4ood: That was my first implementation since we have it. Then i got a security check complains on `JsonFieldMapping` which could open to SQL injection on order by.  So i figured it wanted a cast to simple type like `String`

paullatzelsperger (approved):

PR #2016, opened 2022-09-26
refactor: move duplicated code to `ids-core`

### LINKED ISSUE #1727: ids: move duplicated code to `ids-core`
# Feature Request

There is some duplicated code between the `ids-api-multipart-dispatcher-v1` and `ids-api-multipart-endpoint-v1` modules, e.g.: 

* code for acquiring a `DynamicAttributeToken` for the messages
* `resolveConnectorId()` in both extension classes

This code should be moved to `ids-core` to be shared by the other modules.

ids

To avoid duplications.

## Solution Proposal

Move code to `ids-core`.

## Type of Issue

cleanup

## Checklist

ronjaquensel:
## What this PR changes/adds

Moves code that was duplicated between `ids-api-multipart-dispatcher-v1` and `ids-api-multipart-endpoint-v1` to `ids-core`. Adds a utility class for resolving the connector's IDS ID setting and introduces a service for handling `DynamicAttributeTokens`.

## Why it does that

To avoid duplications.

## Linked Issue(s)

Closes #1727  

## Checklist

### DIFF
--- a/data-protocols/ids/ids-core/src/main/java/org/eclipse/dataspaceconnector/ids/core/service/DynamicAttributeTokenServiceImpl.java
+++ b/data-protocols/ids/ids-core/src/main/java/org/eclipse/dataspaceconnector/ids/core/service/DynamicAttributeTokenServiceImpl.java
@@ -25,7 +25,7 @@ import org.eclipse.dataspaceconnector.spi.iam.TokenRepresentation;
 import org.eclipse.dataspaceconnector.spi.result.Result;
 
 import java.net.URI;
-import java.util.HashMap;
+import java.util.Map;
 
 public class DynamicAttributeTokenServiceImpl implements DynamicAttributeTokenService {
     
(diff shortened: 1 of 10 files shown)


juliapampus (approved):

data-protocols/ids/ids-core/src/main/java/org/eclipse/dataspaceconnector/ids/core/util/ConnectorIdUtil.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public static String resolveConnectorId(ServiceExtensionContext context) {
+        var value = context.getSetting(EDC_IDS_ID, DEFAULT_EDC_IDS_ID);
+    
+        // Hint: use stringified uri to keep uri path and query
+        var result = IdsId.from(value);
+        if (result.succeeded()) {
+            var idsId = result.getContent();
+            if (idsId.getType() == IdsType.CONNECTOR) {
+                return idsId.getValue();
```
ndr-brt, line 50: I have some doubts about the usage of this connectorId... could this be returned as an `IdsId` and passed around?
Because it seems like in the "sender delegates" we are valuing the `_issuerConnector_` and the `_senderAgent_` of the message with the URI parsing of the value, while in the various responses in the `ResponseUtil` we are putting the whole `urn` parsed at URI. Shouldn't them be always the same (as the latter)?

data-protocols/ids/ids-core/src/main/java/org/eclipse/dataspaceconnector/ids/core/service/DynamicAttributeTokenServiceImpl.java
```diff
@@ -0,0 +1,71 @@
...
+        var tokenResult = identityService.obtainClientCredentials(tokenParameters);
+        
+        if (tokenResult.failed()) {
+            return Result.failure("Failed to obtain identity token.");
+        }
+        
+        return Result.success(new DynamicAttributeTokenBuilder()
+                ._tokenFormat_(TokenFormat.JWT)
+                ._tokenValue_(tokenResult.getContent().getToken())
+                .build());
```
ndr-brt, line 55: This block could be replaced using the `map` function of `Result`:
```java
        return identityService.obtainClientCredentials(tokenParameters)
                .map(credentials -> new DynamicAttributeTokenBuilder()
                    ._tokenFormat_(TokenFormat.JWT)
                    ._tokenValue_(credentials.getToken())
                    .build()
                );
```

so eventual errors would be passed to the called directly (that is actually logging them and failing the send process)

data-protocols/ids/ids-core/src/main/java/org/eclipse/dataspaceconnector/ids/core/service/DynamicAttributeTokenServiceImpl.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+    
+    @Override
+    public Result<ClaimToken> verifyDynamicAttributeToken(DynamicAttributeToken token, URI issuerConnector, String audience) {
+        // Prepare DAT validation: IDS token validation requires issuerConnector
+        var additional = new HashMap<String, Object>();
+        additional.put("issuerConnector", issuerConnector);
```
ndr-brt, line 62: This could be an inliner
`var additional = Map.<String, Object>of("issuerConnector", issuerConnector);`
(as a nice side effect, the map will be immutable)

data-protocols/ids/ids-core/src/main/java/org/eclipse/dataspaceconnector/ids/core/util/ConnectorIdUtil.java
ronjaquensel: Good catch, had not noticed that before. But you're absolutely right, they should always be the same and in the URN notation. I'll open an issue and address this in a separate PR, as it's probably gonna touch a lot of classes.

PR #2015, opened 2022-09-26
feat: IDS catalog request filtering

maciejkizlich-zf:
## What this PR changes/adds

Extend existing CatalogRequest object with list of criteria used to filter Assets within Catalog response.

## Why it does that

In a large scale environment, with multiple EDCs holding the information about hundreds of thousands of Assets, instead of looping thought all of the ContractOffers, there is a need to narrow the results down to even single offer. For example, an Asset might hold the 'special' type of data (like registry or database) and should be searchable by its type.

## Linked Issue(s)

Closes #1966

## Checklist

- [x] added appropriate tests?
- [x] performed checkstyle check locally?
- [x] added/updated copyright headers?
- [x] documented public classes/methods?
- [x] added/updated relevant documentation?
- [ ] assigned appropriate label? (exclude from changelog with label `no-changelog`)
- [ ] formatted title correctly? (_take a look at the [CONTRIBUTING](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/CONTRIBUTING.md#submit-a-pull-request) and [styleguide](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/styleguide.md) for details_)

Issue #2009, opened 2022-09-23
Potential Bug: Empty criteria list in ContractDefinition results in valid data offers


juliapampus:
# Bug Report

## Describe the Bug
Potential misbehavior: When a `ContractDefinition` is created with an `accessPolicyId`, `contractPolicyId` and an empty `Criterion` list, this `ContractDefinition` is applied to all `Assets` and results in a `Catalog` with `ContractOffers` that each link the `Policy` with the `contractPolicyId` and every single `Asset`.

Example:
- Create 1 `Policy`
- Create 1 `ContractDefinition` (1 `accessPolicyId`, 1 `contractPolicyId`, empty `Criterion` list) 
- 50 `Assets`

### Expected Behavior
`Catalog` with 0 `ContractOffers` as the `ContractDefinition` is invalid and does not point to a single `Asset`.

### Observed Behavior
`Catalog` with 50 `ContractOffers`, each `Asset` linking to the same `Policy` from the `ContractDefinition`.

As a result: 50 `ContractOffers` (empty `Criterion` list) and 50 registered `Assets` result in 50x50 `ContractOffers` in the `Catalog`.

## Possible Implementation
_Note: This is a suspicion at this point, and needs to be investigated._

Issue #1173, opened 2022-04-19
Story: implement cross-cutting caching strategy


paullatzelsperger:
# Feature Request

Various data backends we have (`"*stores"`) should have a caching layer to reduce communication with the actual database and to speed up requests to the connector.

## Which Areas Would Be Affected?
All data backends

## Why Is the Feature Desired?
The main drivers for this feature request are: 
1. access time: keeping certain objects in - and accessing them from - memory will speed up request turnaround times greatly. 
2. implementation simplicity: things like filtering and sorting can be implemented in a generic and _much easier_ way when operating on a memory-based collection as opposed to: a database table.

## Solution Proposal
I propose devising a caching mechanism with interchangeable eviction strategies, that allow for adapting this to particular needs, and even delegate out to dedicated cache services such as Redis, which may come in handy in clustered environments.
Several implementations could come out-of-the-box:
- write-through: every write hits the cache _and_ the data backend, every read only hits the cache. only suitable for smaller collections
- batch-write ("write back"): similar to write-through, only persists cache periodically. 
- ringbuffer: only a certain amount of elements is held in memory and "old" elements get evicted. Eviction could happen due to age or access frequency. Suitable for very large collections, e.g. the `AssetIndex`.
- shared cache: delegation to e.g. a Redis service. Best for clustered deployments.

## Type of Issue
Improvement Story

**THIS IS NOT YET READY FOR DEVELOPMENT, NEEDS FURTHER DETAILS**

## Checklist

- [ ] assigned appropriate label?
- [x] **Do NOT select a milestone or an assignee!**

PR #2018, opened 2022-09-26
fix: use the `range` to limit offers, not definitions

### LINKED ISSUE #2008: Potential bug: paging is not applied correctly during a catalog request
# Bug Report

## Describe the Bug
There could be a bug in the `ContractOfferServiceImpl`, where the `Range` parameter (i.e. the paging) is applied incorrectly

### Expected Behavior
Paging should limit the number of assets (or rather: `ContractOffers`) that are returned.

### Observed Behavior
The current implementation looks like this:
```java
    @Override
    @NotNull
    public Stream<ContractOffer> queryContractOffers(ContractOfferQuery query, Range range) {

        return definitionService.definitionsFor(agent, range) //<-- this might be incorrect
                .flatMap(definition -> {
                    var assets = assetIndex.queryAssets(definition.getSelectorExpression());
                    return Optional.of(definition.getContractPolicyId())
                            .map(policyStore::findById)
                            .map(policy -> assets.map(asset -> createContractOffer(definition, policy.getPolicy(), asset)))
                            .orElseGet(Stream::empty);
                });
    }
```

which indicates that the `Range` limits the stream coming from `definitionsFor()` rather than the `assets` variable.

_Note: this is a suspicion at this point, and needs to be investigated_

## Possible Implementation
Apply the `Range` parameter to the resulting `Stream<ContractOffer>`

paullatzelsperger:
## What this PR changes/adds

In the current implementation of the `ContractOfferSerivceImpl`, the `Range` parameter is used to limit `ContractDefinitions` rather than the resulting `ContractOffers`. This is a bug and could lead to more contractOffers than specified by `Range`.

This PR dynamically composes the `Stream<Asset>` based on the `from` and `to` parameters of the `Range`, skipping as many `Assets` as necessary, and applying the `skip` and `limit` parameters to the respective database query.

In doing that, it was necessary to add a method `AssetIndex.count(QuerySpec)` to obtain the number of Assets referenced by a particular `AssetSelectorExpression`.

## Why it does that

To guarantee the correct/expected behaviour of the paging.

## Further notes

- This solution works regardless of the underlying implementation of the Stream returned by the AssetIndex
- There are other pending changes to be made (#2015 and #2009) which will also influence this area of code. 

Closes #2008 

## Checklist

### DIFF
--- a/core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
+++ b/core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
@@ -12,7 +12,7 @@
  *       Microsoft Corporation - Refactoring
  *       Fraunhofer Institute for Software and Systems Engineering - extended method implementation
  *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - improvements
- *       ZF Friedrichshafen AG - enable asset filtering
+ *
  */
 
 package org.eclipse.dataspaceconnector.contract.offer;
@@ -24,6 +24,7 @@ import org.eclipse.dataspaceconnector.spi.contract.ContractId;
 import org.eclipse.dataspaceconnector.spi.contract.offer.ContractDefinitionService;
 import org.eclipse.dataspaceconnector.spi.contract.offer.ContractOfferQuery;
 import org.eclipse.dataspaceconnector.spi.contract.offer.ContractOfferService;
+import org.eclipse.dataspaceconnector.spi.message.Range;
 import org.eclipse.dataspaceconnector.spi.policy.store.PolicyDefinitionStore;
 import org.eclipse.dataspaceconnector.spi.query.QuerySpec;
 import org.eclipse.dataspaceconnector.spi.types.domain.asset.Asset;
@@ -32,11 +33,9 @@ import org.eclipse.dataspaceconnector.spi.types.domain.contract.offer.ContractOf
 import org.jetbrains.annotations.NotNull;
 
 import java.net.URI;
-import java.util.Optional;
-import java.util.stream.Collectors;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Stream;
-
-import static java.util.stream.Stream.concat;
 
 /**
  * Implementation of the {@link ContractOfferService}.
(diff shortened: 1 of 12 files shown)


core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
```diff
@@ -55,15 +57,21 @@ public ContractOfferServiceImpl(ParticipantAgentService agentService, ContractDe
     @NotNull
     public Stream<ContractOffer> queryContractOffers(ContractOfferQuery query, Range range) {
         var agent = agentService.createFor(query.getClaimToken());
-
-        return definitionService.definitionsFor(agent, range)
+        var offers = definitionService.definitionsFor(agent)
```
ndr-brt, line 60: I know that there aren't many ways to do this, but could this lead to long-blocked threads? 
I mean, for every page we have to fetch all the definitions - that could be not a big issue, also in the case if there are a lot of them, since they are really little.
But, on the last page request here the service will fetch all the definitions, query/evaluate all the policies, fetch all the assets, and then `skip` and `limit` are applied, as they are located after a `flatMap`, so the time needed to fetch the last page is actually the same time needed to fetch all the catalog in a single call.

e.g. (speculative) having 100.000 (pretty standard value) definitions means that there will be 200.000 queries to the policy store (access and contract) and 100.000 to the asset index. let's assume that the stores are really fast and the policy evaluation take no-time, if we count 1ms for every definition the single page request could take a time around 1 minute.

I don't remember if the issue with the "full catalog" fetch was with this operation of with serdes/transmission, but I think this implementation could lead poor performances and potentially to OOM exceptions.

core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
```diff
@@ -55,15 +57,21 @@ public ContractOfferServiceImpl(ParticipantAgentService agentService, ContractDe
...
-                            .map(policy -> assets.map(asset -> createContractOffer(definition, policy.getPolicy(), asset)))
+                            .map(policy -> assetIndex.queryAssets(spec).map(asset -> createContractOffer(definition, policy.getPolicy(), asset)))
                             .orElseGet(Stream::empty);
-                });
+                })
+                // todo: change this: take limit/offset from query
+                .skip(range.getFrom())
+                .limit(range.getTo() - range.getFrom())
+                .collect(Collectors.toList());
+        return offers.stream();
```
ndr-brt, line 73: I would avoid to collect and stream again

ndr-brt (commented):
I'm worried about the performances implications of this

core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
paullatzelsperger: yeah, i had that for debugging. should be removed.

core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
paullatzelsperger: >... skip and limit are applied, as they are located after a flatMap
well, from a logical standpoint, that is the only place where `skip` and `limit` _can_ be applied, because we need the flat list of assets/offers first. 
Maybe we could collect first all the `Assets`, slice out the page and _then_ convert them into offers? That would still rely on the store not materializing the stream prematurely, but would save calls to the `PolicyDefinitionStore`.

Also, the way the `range` was applied before, was flat-out wrong, because it limited the number of `contractDefinitions`, so _some_ action needed to be taken.

The problem with the full list was that jetty ran into timeouts (and potentially body-size limits) when using anything other than in-mem.

> I mean, for every page we have to fetch all the definitions ...
We could add a cache for the definitions and policies (see #1173), if that needs to be optimized.

paullatzelsperger: The implementation was updated in commit https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/2018/commits/ea755913749a7e98e5c82891cd951fc46e8e7a25, as was the PR descriptions

ndr-brt (approved):
I think this is probably the most efficient way to do this at the moment.
The next thing where we could investigate is actually supporting streaming in the `SqlExecutor` (because now we are reading everything in a list in memory on which the stream is opened).
Will open an issue about.

wolf4ood (approved):

juliapampus (approved):

PR #2019, opened 2022-09-27
fix(ContractOfferServiceImpl): check for existing criteria list in AssetSelectorExpression

### LINKED ISSUE #2009: Potential Bug: Empty criteria list in ContractDefinition results in valid data offers

juliapampus:
## What this PR changes/adds

Moves/Adds business logic to `ContractOfferServiceImpl`:
(1) Query `Assets` with `QuerySpec.none()` if `AssetSelectorExpression == AssetSelectorExpression.SELECT_ALL`
(2) Filter `ContractDefinitions` with empty or null criteria list if `AssetSelectorExpression != AssetSelectorExpression.SELECT_ALL`

Removes check (1) from `InMemoryAssetIndex` and `CosmosAssetIndex`.

## Why it does that

For database queries, no filter criteria means "select all" (equal to "*"). In terms of data sovereignty and our domain model, `ContractDefinitions` that provide no selector expression are not pointing to an `Asset` and should therefore not result in a valid `ContractOffer` as part of the conntector's `Catalog`.

## Further notes

Unit tests added for `InMemoryAssetIndex`. As the logic is added to the `ContractOfferServiceImpl`, this behavior will exist for all `AssetIndex` implementations. 

Tested locally with sample 4.0: Registering a `ContractDefinition` in `FileTransferExtension` with an empty or null criteria list results in a catalog with 0 `contractOffers` during catalog querying. (This was not the case before the added filter.)

Local style check hinted to missing license headers and redundant line spaces --> fixed this.

## Linked Issue(s)

Closes #2009 

## Checklist

### DIFF
--- a/core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
+++ b/core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
@@ -13,6 +13,8 @@
  *       Fraunhofer Institute for Software and Systems Engineering - extended method implementation
  *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - improvements
  *       ZF Friedrichshafen AG - enable asset filtering
+ *       Fraunhofer Institute for Software and Systems Engineering - add criteria validation
+ *
  */
 
 package org.eclipse.dataspaceconnector.contract.offer;
@@ -20,6 +22,7 @@ package org.eclipse.dataspaceconnector.contract.offer;
 import org.eclipse.dataspaceconnector.policy.model.Policy;
 import org.eclipse.dataspaceconnector.spi.agent.ParticipantAgentService;
 import org.eclipse.dataspaceconnector.spi.asset.AssetIndex;
+import org.eclipse.dataspaceconnector.spi.asset.AssetSelectorExpression;
 import org.eclipse.dataspaceconnector.spi.contract.ContractId;
 import org.eclipse.dataspaceconnector.spi.contract.offer.ContractDefinitionService;
 import org.eclipse.dataspaceconnector.spi.contract.offer.ContractOfferQuery;
@@ -60,10 +63,22 @@ public class ContractOfferServiceImpl implements ContractOfferService {
         var agent = agentService.createFor(query.getClaimToken());
 
         return definitionService.definitionsFor(agent, query.getRange())
+                // ignore definitions with missing criteria (only if not SELECT_ALL)
+                .filter(definition -> (definition.getSelectorExpression().getCriteria() != null &&
+                        !definition.getSelectorExpression().getCriteria().isEmpty()) ||
+                        definition.getSelectorExpression() == AssetSelectorExpression.SELECT_ALL)
                 .flatMap(definition -> {
-                    var assetFilterQuery = QuerySpec.Builder.newInstance()
-                            .filter(concat(definition.getSelectorExpression().getCriteria().stream(), query.getAssetsCriteria().stream()).collect(Collectors.toList())).build();
-                    var assets = assetIndex.queryAssets(assetFilterQuery);
+                    Stream<Asset> assets;
+                    if (definition.getSelectorExpression() == AssetSelectorExpression.SELECT_ALL) {
+                        // select everything if the special constant is used
+                        assets = assetIndex.queryAssets(QuerySpec.none());
+                    } else {
+                        var assetFilterQuery = QuerySpec.Builder.newInstance()
+                                .filter(concat(definition.getSelectorExpression().getCriteria().stream(),
+                                        query.getAssetsCriteria().stream()).collect(Collectors.toList())).build();
+                        assets = assetIndex.queryAssets(assetFilterQuery);
+                    }
+
                     return Optional.of(definition.getContractPolicyId())
                             .map(policyStore::findById)
                             .map(policy -> assets.map(asset -> createContractOffer(definition, policy.getPolicy(), asset)))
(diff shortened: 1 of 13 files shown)


juliapampus:
As discussed, potentially blocked by #2015 and blocking #2018, as all changes address the same method.

core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
```diff
@@ -57,8 +60,19 @@ public Stream<ContractOffer> queryContractOffers(ContractOfferQuery query, Range
         var agent = agentService.createFor(query.getClaimToken());
 
         return definitionService.definitionsFor(agent, range)
+                // ignore definitions with missing criteria (only if not SELECT_ALL)
+                .filter(definition -> (definition.getSelectorExpression().getCriteria() != null &&
```
ndr-brt, line 64: I think this filter does nothing, as when the `criteria` field is empty, then the `selectorExpression` equals to `SELECT_ALL`. Because `SELECT_ALL` means no criterion are provided.
(this pass:)
```java
    @Test
    void equality() {
        assertThat(AssetSelectorExpression.Builder.newInstance().build()).isEqualTo(AssetSelectorExpression.SELECT_ALL);
    }
```

Furthermore, the `selectorExpression` of the definition cannot be null.

core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
```diff
@@ -57,8 +60,19 @@ public Stream<ContractOffer> queryContractOffers(ContractOfferQuery query, Range
...
+                        !definition.getSelectorExpression().getCriteria().isEmpty()) ||
+                        definition.getSelectorExpression() == AssetSelectorExpression.SELECT_ALL)
                 .flatMap(definition -> {
-                    var assets = assetIndex.queryAssets(definition.getSelectorExpression());
+                    Stream<Asset> assets;
+                    if (definition.getSelectorExpression() == AssetSelectorExpression.SELECT_ALL) {
```
ndr-brt, line 69: Pay attention on using `==` to equal on object instances, however I see no use for this `if` block

paullatzelsperger (approved):
LGTM. I'll just rebase #2018 and apply your added checks analogously.

ndr-brt (commented):
I'm absolutely in favor of pull out the `AssetSelectorExpression` evaluation out of the repository (I would also inline the `Stream<Asset> queryAssets(AssetSelectorExpression expression);` method, as the `AssetSelectorExpression` is something related to the `ContractDefinition` and that it should be converted to a `QuerySpec` before being used to fetch assets.

But the `SELECT_ALL` is only a constant that is actually not used (it's impossible to generate it a runtime), so personally I would remove it completely.

core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
juliapampus: The `selectorExpression` cannot be null but the `criteriaList` can. There is no additional check in the builder class that validates whether the expresion is `SELECT_ALL` and otherwise the criteria list cannot be empty or null.

core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
juliapampus: This if block adds the logic that `SELECT_ALL` overwrites any filters as it applies a `ContractDefinition` to any `Asset`. As above, the methods adds logic that the `AssetSelectorExpression` does not provide itself.

juliapampus:
> I'm absolutely in favor of pull out the `AssetSelectorExpression` evaluation out of the repository (I would also inline the `Stream<Asset> queryAssets(AssetSelectorExpression expression);` method, as the `AssetSelectorExpression` is something related to the `ContractDefinition` and that it should be converted to a `QuerySpec` before being used to fetch assets.
> 
> But the `SELECT_ALL` is only a constant that is actually not used (it's impossible to generate it a runtime), so personally I would remove it completely.

As I originally did not implement this `AssetSelectorExpression`, I did not want to remove provided functionality. However, I agree that this class could be implemnted differently. Nevertheless, as we interpret missing criteria as "select none", we should provide a possibility to explicitly "select all" without the need to provide a list of assetIds.

core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
juliapampus: > Because `SELECT_ALL` means no criterion are provided.

This should not be the case. I missed this test. `SELECT_ALL` means "select all". No criteria means "select none".

juliapampus:
Did not mange to get the QL workflow working again.... new PR: #2044

PR #2031, opened 2022-09-29
fix: use IdsId as type for connector ID

### LINKED ISSUE #2022: ids: inconsistent handling of connector ID
# Bug Report

## Describe the Bug
For the connector ID, some classes (e.g. `MultipartSenderDelegate` implementations) use only the `value` part of the connector's IDS ID. Other classes (e.g. `ResponseUtil`) use the whole URN. Thus, the connector's ID is not the same in all outgoing messages.

>I have some doubts about the usage of this connectorId... could this be returned as an `IdsId` and passed around?
>Because it seems like in the "sender delegates" we are valuing the `_issuerConnector_` and the `_senderAgent_` of the message with
>the URI parsing of the value, while in the various responses in the `ResponseUtil` we are putting the whole `urn` parsed at URI.
>Shouldn't them be always the same (as the latter)?

_Originally posted by @ndr-brt in https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/2016#discussion_r981175267_

### Expected Behavior
The connector always sets the ID for itself in outgoing messages (initial messages and responses).

### Observed Behavior
The connector sets different IDs for itself in initial messages and responses (*id-value* and *urn:connector:id-value*).

## Possible Implementation
Handle the connector's ID as an `IdsId` instance, use `toUri()` when setting it in IDS messages.

ronjaquensel:
## What this PR changes/adds

It changes the type of the `connectorId` field from `String`/`URI` to `IdsId` in all IDS classes. Whenever the ID is mapped to Infomodel classes, `IdsId.toUri()` is used.

## Why it does that

To have a consistent handling of the connector's ID in IDS classes, as currently the connector uses different ID formats for its own ID (*id-value* and *urn:connector:id-value*).

## Further notes

Uses `ConnectorIdUtil.resolveConnectorId()` to set the connector's ID in `ConnectorServiceSettings`, as I missed this when removing the code duplications.

## Linked Issue(s)

Closes #2022 

## Checklist


data-protocols/ids/ids-api-multipart-endpoint-v1/src/main/java/org/eclipse/dataspaceconnector/ids/api/multipart/handler/ArtifactRequestHandler.java
```diff
@@ -161,7 +161,7 @@ public boolean canHandle(@NotNull MultipartRequest multipartRequest) {
                 .id(message.getId().toString())
                 .protocol(MessageProtocol.IDS_MULTIPART)
                 .dataDestination(dataDestination)
-                .connectorId(connectorId)
+                .connectorId(connectorId.toUri().toString())
```
juliapampus, line 164: Maybe we can add a `toString` method to the `IdsId` to capsulated this call?

ronjaquensel: done.

juliapampus (approved):

PR #2041, opened 2022-10-03
feat: Stream over ResultSet on sql queries

### LINKED ISSUE #2027: Support real streaming on sql queries
# Feature Request

Currently, all the sql queries pass through `SqlExecutorQuery` that fetches all the `ResultSet` in a `List` on which it opens a `Stream`.
This is not the optimal solution, as we could stream over the `ResultSet` directly.

sql layer

performances

## Solution Proposal
-

## Type of Issue
improvement

## Checklist

ndr-brt:
## What this PR changes/adds

Modify `SqlQueryExecutor.executeQuery` to stream over the `ResultSet` without collect the results in memory before.
This is done using a `Splititerator`.
Note that the stream must be always closed explicitly to permit sql resource closure (`ResultSet`, `Statement`, ...).
The closure of the connection can be chosen with the `closeConnection` flag.

## Why it does that

To improve performances

## Further notes

- introduced `DoorKeeper` that's a component that takes care of closing the `AutoCloseable` components registered on it
- changed services api to return a `ServiceResult<Stream<T>>` instead of `ServiceResult<List<T>>`, then the streams have been closed explicitly in the controllers
- moved `PostgresLocalInstance` to `common-sql` test fixtures, as it seems a better place to find it.
- simplified postgresql integration tests, moving the datasource creation into `PostgresLocalInstance` utility class.
- removed two tests from `SqlQueryExecutorTest`, to be honest I don't get those test class, it seems overmocking.

## Linked Issue(s)

Closes #2027

## Checklist


extensions/common/sql/lease-sql/src/main/java/org/eclipse/dataspaceconnector/sql/lease/SqlLeaseContext.java
```diff
@@ -106,9 +107,9 @@ public void acquireLease(String entityId) {
      */
     public @Nullable SqlLease getLease(String entityId) {
         var stmt = statements.getFindLeaseByEntityTemplate();
-        var leases = executeQuery(connection, this::mapLease, stmt, entityId);
-
-        return leases.stream().findFirst().orElse(null);
+        try (var leases = SqlQueryExecutor.executeQuery(connection, false, this::mapLease, stmt, entityId)) {
```
paullatzelsperger, line 110: i don't think streaming the `Lease`s is necessary here, because the query will return at most one `Lease` (i.e. the one for the given entity, or null)

extensions/control-plane/store/sql/asset-index-sql/src/main/java/org/eclipse/dataspaceconnector/sql/assetindex/SqlAssetIndex.java
```diff
@@ -229,14 +231,9 @@ private Object fromPropertyValue(String value, String type) throws ClassNotFound
...
-        var assetCount = transactionContext.execute(() -> executeQuery(connection, this::mapRowCount, assetStatements.getCountAssetByIdClause(), assetId).iterator().next());
-
-        if (assetCount <= 0) {
-            return false;
-        } else if (assetCount > 1) {
-            throw new IllegalStateException("Expected result set size of 0 or 1 but got " + assetCount);
-        } else {
-            return true;
+        var sql = assetStatements.getCountAssetByIdClause();
+        try (var stream = SqlQueryExecutor.executeQuery(connection, false, this::mapRowCount, sql, assetId)) {
```
paullatzelsperger, line 235: I don't think this has to be a `Stream`. It executes a simple `SELECT COUNT (*)` query, which returns an int. Seems we're killing rats with a cannon.

extensions/control-plane/store/sql/asset-index-sql/src/main/java/org/eclipse/dataspaceconnector/sql/assetindex/SqlAssetIndex.java
```diff
@@ -152,44 +161,48 @@ public void accept(AssetEntry item) {
...
-            var dataAddressList = transactionContext.execute(() -> executeQuery(connection, this::mapDataAddress, assetStatements.getFindDataAddressByIdTemplate(), assetId));
-            return single(dataAddressList);
-        } catch (Exception e) {
-            if (e instanceof EdcPersistenceException) {
-                throw (EdcPersistenceException) e;
-            } else {
-                throw new EdcPersistenceException(e.getMessage(), e);
+        return transactionContext.execute(() -> {
+            var sql = assetStatements.getFindDataAddressByIdTemplate();
+            try (var stream = SqlQueryExecutor.executeQuery(getConnection(), true, this::mapDataAddress, sql, assetId)) {
```
paullatzelsperger, line 192: I don't think this has to be a `Stream`, as it returns at most one result.

extensions/control-plane/store/sql/asset-index-sql/src/main/java/org/eclipse/dataspaceconnector/sql/assetindex/SqlAssetIndex.java
```diff
@@ -152,44 +161,48 @@ public void accept(AssetEntry item) {
...
-            var asset = findById(assetId);
-            if (asset == null) {
-                return null;
-            }
+        return transactionContext.execute(() -> {
+            try (var connection = getConnection()) {
+                var asset = findById(assetId);
+                if (asset == null) {
+                    return null;
+                }
```
paullatzelsperger, line 176: please do not remove the transaction block, as there may be a transaction already in progress, to which the delete statement is attached.

paullatzelsperger (commented):
Having a streaming database API is great, but there are situations where using a `Stream` (as opposed to the conventional API) makes things more complicated than they need to be, for example when performing query-by-id, or delete-by-id. To me it makes sense only in statements, where multiple results are expected.

extensions/control-plane/store/sql/asset-index-sql/src/main/java/org/eclipse/dataspaceconnector/sql/assetindex/SqlAssetIndex.java
ndr-brt: @paullatzelsperger actual I didn't removed, but I "enlarged" to contain both the `findBy` and the `delete` queries, as they should stay in the same transaction context, and that was not the case.

ndr-brt:
@paullatzelsperger what we could do could is to add a "executeQuerySimple` that would execute do that "first or else null" logic, but since we already took care of the stream complexity, I can't see any advantage in keeping and maintaining the old `executeQuery` method - that does not stream - just for the "get one" queries.

ndr-brt:
@paullatzelsperger `executeQuerySingle` method added

extensions/common/sql/common-sql/src/main/java/org/eclipse/dataspaceconnector/sql/SqlQueryExecutor.java
```diff
@@ -45,37 +51,82 @@ public static int executeQuery(Connection connection, String sql, Object... argu
...
+            } catch (Exception ex) {
+                sqlEx.addSuppressed(ex);
+            }
+
+            throw new EdcPersistenceException(sqlEx);
         }
     }
 
+    @NotNull
+    private static <T> Spliterators.AbstractSpliterator<T> createSplititerator(ResultSetMapper<T> resultSetMapper, ResultSet resultSet) {
```
paullatzelsperger, line 112:
```suggestion
    private static <T> Spliterators.AbstractSpliterator<T> createSpliterator(ResultSetMapper<T> resultSetMapper, ResultSet resultSet) {
```

extensions/common/sql/common-sql/src/main/java/org/eclipse/dataspaceconnector/sql/DoorKeeper.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.eclipse.dataspaceconnector.spi.EdcException;
+
+import java.util.Stack;
+
+/**
+ * The DoorKeeper takes care to close all the registered {@link AutoCloseable} components from the last to the first
+ * Wrapping eventually checked exceptions with an unchecked {@link EdcException}
+ */
+public class DoorKeeper implements AutoCloseable {
```
paullatzelsperger, line 25: minor nit: the `DoorKeeper` could be package-private, since it's only used by the `SqlQueryExecutor`. Or we could make it generally available, in which case I'd move it to one of the core packages.

extensions/common/sql/lease-sql/src/main/java/org/eclipse/dataspaceconnector/sql/lease/SqlLeaseContext.java
```diff
@@ -106,9 +107,7 @@ public void acquireLease(String entityId) {
      */
     public @Nullable SqlLease getLease(String entityId) {
         var stmt = statements.getFindLeaseByEntityTemplate();
-        var leases = executeQuery(connection, this::mapLease, stmt, entityId);
-
-        return leases.stream().findFirst().orElse(null);
+        return SqlQueryExecutor.executeQuerySingle(connection, false, this::mapLease, stmt, entityId);
```
paullatzelsperger, line 110: minor nit: static import

paullatzelsperger (approved):
LGTM

PR #2044, opened 2022-10-04
fix(ContractOfferServiceImpl): check for existing criteria list in AssetSelectorExpression

### LINKED ISSUE #2009: Potential Bug: Empty criteria list in ContractDefinition results in valid data offers

juliapampus:
## What this PR changes/adds

Moves/Adds business logic to `ContractOfferServiceImpl`:
(1) Query `Assets` with `QuerySpec.none()` if `AssetSelectorExpression == AssetSelectorExpression.SELECT_ALL`
(2) Filter `ContractDefinitions` with empty or null criteria list if `AssetSelectorExpression != AssetSelectorExpression.SELECT_ALL`

Removes check (1) from `InMemoryAssetIndex` and `CosmosAssetIndex`.

## Why it does that

For database queries, no filter criteria means "select all" (equal to "*"). In terms of data sovereignty and our domain model, `ContractDefinitions` that provide no selector expression are not pointing to an `Asset` and should therefore not result in a valid `ContractOffer` as part of the conntector's `Catalog`.

## Further notes

Unit tests added for `InMemoryAssetIndex`. As the logic is added to the `ContractOfferServiceImpl`, this behavior will exist for all `AssetIndex` implementations. 

Tested locally with sample 4.0: Registering a `ContractDefinition` in `FileTransferExtension` with an empty or null criteria list results in a catalog with 0 `contractOffers` during catalog querying. (This was not the case before the added filter.)

Local style check hinted to missing license headers and redundant line spaces --> fixed this.

## Linked Issue(s)

Closes #2009 

## Checklist

### DIFF
--- a/core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
+++ b/core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
@@ -13,6 +13,8 @@
  *       Fraunhofer Institute for Software and Systems Engineering - extended method implementation
  *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - improvements
  *       ZF Friedrichshafen AG - enable asset filtering
+ *       Fraunhofer Institute for Software and Systems Engineering - add criteria validation
+ *
  */
 
 package org.eclipse.dataspaceconnector.contract.offer;
@@ -20,6 +22,7 @@ package org.eclipse.dataspaceconnector.contract.offer;
 import org.eclipse.dataspaceconnector.policy.model.Policy;
 import org.eclipse.dataspaceconnector.spi.agent.ParticipantAgentService;
 import org.eclipse.dataspaceconnector.spi.asset.AssetIndex;
+import org.eclipse.dataspaceconnector.spi.asset.AssetSelectorExpression;
 import org.eclipse.dataspaceconnector.spi.contract.ContractId;
 import org.eclipse.dataspaceconnector.spi.contract.offer.ContractDefinitionService;
 import org.eclipse.dataspaceconnector.spi.contract.offer.ContractOfferQuery;
@@ -60,10 +63,22 @@ public class ContractOfferServiceImpl implements ContractOfferService {
         var agent = agentService.createFor(query.getClaimToken());
 
         return definitionService.definitionsFor(agent, query.getRange())
+                // ignore definitions with missing criteria (only if not SELECT_ALL)
+                .filter(definition -> (definition.getSelectorExpression().getCriteria() != null &&
+                        !definition.getSelectorExpression().getCriteria().isEmpty()) ||
+                        definition.getSelectorExpression() == AssetSelectorExpression.SELECT_ALL)
                 .flatMap(definition -> {
-                    var assetFilterQuery = QuerySpec.Builder.newInstance()
-                            .filter(concat(definition.getSelectorExpression().getCriteria().stream(), query.getAssetsCriteria().stream()).collect(Collectors.toList())).build();
-                    var assets = assetIndex.queryAssets(assetFilterQuery);
+                    Stream<Asset> assets;
+                    if (definition.getSelectorExpression() == AssetSelectorExpression.SELECT_ALL) {
+                        // select everything if the special constant is used
+                        assets = assetIndex.queryAssets(QuerySpec.none());
+                    } else {
+                        var assetFilterQuery = QuerySpec.Builder.newInstance()
+                                .filter(concat(definition.getSelectorExpression().getCriteria().stream(),
+                                        query.getAssetsCriteria().stream()).collect(Collectors.toList())).build();
+                        assets = assetIndex.queryAssets(assetFilterQuery);
+                    }
+
                     return Optional.of(definition.getContractPolicyId())
                             .map(policyStore::findById)
                             .map(policy -> assets.map(asset -> createContractOffer(definition, policy.getPolicy(), asset)))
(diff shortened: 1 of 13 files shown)


core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
```diff
@@ -60,10 +63,22 @@ public Stream<ContractOffer> queryContractOffers(ContractOfferQuery query) {
         var agent = agentService.createFor(query.getClaimToken());
 
         return definitionService.definitionsFor(agent, query.getRange())
+                // ignore definitions with missing criteria (only if not SELECT_ALL)
+                .filter(definition -> (definition.getSelectorExpression().getCriteria() != null &&
```
ndr-brt, line 67: I think this filter does nothing, as when the `criteria` field is empty, then the `selectorExpression` equals to `SELECT_ALL`. Because `SELECT_ALL` means no criterion are provided.
(this passes:)
```java
    @Test
    void equality() {
        assertThat(AssetSelectorExpression.Builder.newInstance().build()).isEqualTo(AssetSelectorExpression.SELECT_ALL);
    }
```

Furthermore, the `selectorExpression` of the definition cannot be null, neither the `criteria` list inside it

core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
```diff
@@ -60,10 +63,22 @@ public Stream<ContractOffer> queryContractOffers(ContractOfferQuery query) {
...
+                        !definition.getSelectorExpression().getCriteria().isEmpty()) ||
+                        definition.getSelectorExpression() == AssetSelectorExpression.SELECT_ALL)
                 .flatMap(definition -> {
-                    var assetFilterQuery = QuerySpec.Builder.newInstance()
-                            .filter(concat(definition.getSelectorExpression().getCriteria().stream(), query.getAssetsCriteria().stream()).collect(Collectors.toList())).build();
-                    var assets = assetIndex.queryAssets(assetFilterQuery);
+                    Stream<Asset> assets;
+                    if (definition.getSelectorExpression() == AssetSelectorExpression.SELECT_ALL) {
```
ndr-brt, line 72: Pay attention on using `==` to equal on object instances (likely it will never be true).

ndr-brt (commented):
I reported my comments from #2019 as otherwise they would have been lost.

I went through this PR again and I don't understand why we would need to avoid the filter if the contract definition has a "select all" selector (empty criteria). If the consumer wants a catalog filtered by a particular asset id, it won't need to find in there other offers related to other assets, right?

I would also remove the `SELECT_ALL` constant as it is not used by anyone, it could be replaced by a `isSelectAll()` method on `AssetSelectorExpression` that checks if the `criteria` list is empty or not.

core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
juliapampus: resolved

core/control-plane/contract/src/main/java/org/eclipse/dataspaceconnector/contract/offer/ContractOfferServiceImpl.java
juliapampus: Still, either I don't get it or I don't agree. 

> I would also remove the SELECT_ALL constant as it is not used by anyone, it could be replaced by a isSelectAll() method on AssetSelectorExpression that checks if the criteria list is empty or not.

We should not follow common DB querying here, as accidentally missing criteria could result in an information leak (what is critically in terms of data sovereignty). So we should not interpret missing criteria as "select all".. what would simplify the implementation from a logically thinking, I agree, but not represent our business logic.

> Because SELECT_ALL means no criterion are provided.

This is also not true because both properties can be set in parallel, don't they? There is no such logic in the `AssetSelectorExpression`, nor is it needed. In my understanding, `SELECT_ALL` has just a higher priority then the criteria list as it must explicitly being set.

juliapampus: > Furthermore, the selectorExpression of the definition cannot be null, neither the criteria list inside it

Is this a suggestion or a fact? Because I checked and you're right, a expression cannot be null, still there is currently no logic that checks whether the criteria list is empty or whether one of `SELECT_ALL` or the criteria list is null or empty. Or did I miss anything?

PR #2046, opened 2022-10-05
chore(ContractOffer): remove field assetId

### LINKED ISSUE #2030: `ContractOffer.assetId` field is not used
# Feature Request

the `assetId` field of the `ContractOffer` class is not used by anyone, nor read nor wrote. 
Actually that information is kept into the `asset` field.
The proposal is to remove the `assetId` field to avoid misunderstanding (I tried more than once to get the id from that field... failing miserably!)

spi

cleanup

## Solution Proposal
-

## Type of Issue
-

## Checklist

wolf4ood:
## What this PR changes/adds

Removes `assetId` from `ContractOffer`

## Why it does that

Cleanup, since `assetId` is not used  by anyone

## Linked Issue(s)

Closes #2030 

## Checklist


wolf4ood:
Hi @ndr-brt 
in this way the field will be removed also from the API which could cause a breaking change, i guess it could be an issue right?

ndr-brt (approved):
Yes, it is an api breaking change, but since that field value was always null (nobody was valuing it), that's not a big deal.

paullatzelsperger:
Even if it was used in downstream projects, e.g. MVD, we'll just have to adapt there.

PR #2047, opened 2022-10-05
feat(DataSourceRegistry): provides a default impl for DataSourceRegistry

### LINKED ISSUE #2032: Provide a default `DataSourceRegistry`
# Feature Request

Currently it's mandatory to add a `transaction-*` dependency when a `sql` database implementation is used, since the `DataSourceRegistry` is registered only by those extensions.
Since we have a `NoopTransactionContext` that's being registered by default when no other implementations are provided, we should have the same for the `DataSourceRegistry`, without having to force the user to introduce a transaction manager (e.g. in test environment)

default services

improve developer experience

## Solution Proposal
-

## Type of Issue
improvement

## Checklist

wolf4ood:
## What this PR changes/adds

 Provide a default `DataSourceRegistry` 

## Why it does that

improve developer experience

## Linked Issue(s)

Closes #2032 

## Checklist

### DIFF
--- a/core/common/base/build.gradle.kts
+++ b/core/common/base/build.gradle.kts
@@ -28,6 +28,8 @@ dependencies {
     api(project(":spi:common:core-spi"))
     api(project(":spi:common:policy-engine-spi"))
     api(project(":spi:common:transaction-spi"))
+    api(project(":spi:common:transaction-datasource-spi"))
+
     implementation(project(":core:common:policy-engine"))
     implementation(project(":core:common:util"))
 
--- a/core/common/base/src/main/java/org/eclipse/dataspaceconnector/core/CoreDefaultServicesExtension.java
+++ b/core/common/base/src/main/java/org/eclipse/dataspaceconnector/core/CoreDefaultServicesExtension.java
@@ -34,6 +34,8 @@ import org.eclipse.dataspaceconnector.spi.system.vault.NoopPrivateKeyResolver;
 import org.eclipse.dataspaceconnector.spi.system.vault.NoopVault;
 import org.eclipse.dataspaceconnector.spi.transaction.NoopTransactionContext;
 import org.eclipse.dataspaceconnector.spi.transaction.TransactionContext;
+import org.eclipse.dataspaceconnector.spi.transaction.datasource.DataSourceRegistry;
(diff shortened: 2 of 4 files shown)


extensions/common/sql/common-sql/src/main/java/org/eclipse/dataspaceconnector/sql/SqlDefaultServicesExtension.java
```diff
@@ -0,0 +1,44 @@
...
+import org.eclipse.dataspaceconnector.spi.transaction.datasource.DataSourceRegistry;
+import org.eclipse.dataspaceconnector.sql.datasource.DefaultDataSourceRegistry;
+
+/**
+ * Default SQL services. It provides a default {@link DataSourceRegistry} without a transaction manager
+ */
+@Extension(value = SqlDefaultServicesExtension.NAME)
+public class SqlDefaultServicesExtension implements ServiceExtension {
+
+    public static final String NAME = "Default SQL services";
```
ndr-brt, line 30: Just a nit, on the other `DefaultServicesExtension`s we used the `<extension name> Default Services` form. I'm not sure what's the best one, but in the meantime we could stick with that convention.

ndr-brt (approved):
I didn't think about introducing a simple data registry, that fits out nicely

extensions/common/sql/common-sql/src/main/java/org/eclipse/dataspaceconnector/sql/SqlDefaultServicesExtension.java
```diff
@@ -0,0 +1,44 @@
...
+    public static final String NAME = "SQL Default Services";
+
+    @Override
+    public String name() {
+        return NAME;
+    }
+
+
+    @Provider(isDefault = true)
+    public DataSourceRegistry dataSourceRegistry(ServiceExtensionContext context) {
```
paullatzelsperger, line 39: Could we move this to `CoreDefaultServicesExtension`? That would introduce a dependency onto `transaction-datasource-spi` in `core:common:base`, but to me that's OK, since the `NoopTransactionContext` (and thus a dependency onto `transaction-spi`) is already registered there.

paullatzelsperger (approved):

PR #2053, opened 2022-10-06
fix(ContractOffer): add asset not null check when building a ContractOffer

### LINKED ISSUE #2052: `ContractOffer.asset` should be a required field (not null)
# Bug Report

## Describe the Bug
With `ContractOffer.Builder` is possible to build a `ContractOffer` with a null asset field. 
But this violate the `@NotNull` annotation on the getter

```java
@NotNull
public Asset getAsset() {
     return asset;
}
```

### Expected Behavior

It should throw an exception if the field is null

### Observed Behavior

This is valid:

```java
 ContractOffer.Builder.newInstance().id("some-id")
     .policy(Policy.Builder.newInstance().build())
     .build()
```

## Possible Implementation

Add a validation check on `ContractOffer.Builder#build` on asset being null

wolf4ood:
## What this PR changes/adds

Adds the check on asset field when building a ContractOffer

## Why it does that

For consistency with the `getAsset` API, since the Asset is mandatory in the `ContractOffer`

## Further notes

While fixing this the style of the `ContractOffer.Builder` has been changed following our usual builder pattern

## Linked Issue(s)

Closes #2052 

## Checklist

### DIFF
--- a/data-protocols/ids/ids-api-multipart-dispatcher-v1/src/test/java/org/eclipse/dataspaceconnector/ids/api/multipart/client/MultipartDispatcherIntegrationTest.java
+++ b/data-protocols/ids/ids-api-multipart-dispatcher-v1/src/test/java/org/eclipse/dataspaceconnector/ids/api/multipart/client/MultipartDispatcherIntegrationTest.java
@@ -145,7 +145,7 @@ class MultipartDispatcherIntegrationTest {
 
     @Test
     void testSendContractOfferMessage(RemoteMessageDispatcherRegistry dispatcher) {
-        var contractOffer = ContractOffer.Builder.newInstance().id("id").policy(Policy.Builder.newInstance().build()).build();
+        var contractOffer = contractOffer("id");
         when(transformerRegistry.transform(any(), any()))
                 .thenReturn(Result.success(getIdsContractOffer()));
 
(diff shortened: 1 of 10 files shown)


spi/control-plane/contract-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/contract/offer/ContractOffer.java
```diff
@@ -202,6 +202,10 @@ public Builder policy(Policy policy) {
         public ContractOffer build() {
             Objects.requireNonNull(id);
 
+            if (asset == null) {
```
paullatzelsperger, line 205: an alternative would be `Objects.requireNonNull(asset, "Asset must not be null")`, which would throw a NPE, but it would make the code a bit more readable, and we have it in many other places in the code base as well.

data-protocols/ids/ids-api-multipart-dispatcher-v1/src/test/java/org/eclipse/dataspaceconnector/ids/api/multipart/client/MultipartDispatcherIntegrationTest.java
```diff
@@ -145,7 +145,11 @@ void testSendArtifactRequestMessage(RemoteMessageDispatcherRegistry dispatcher)
 
     @Test
     void testSendContractOfferMessage(RemoteMessageDispatcherRegistry dispatcher) {
-        var contractOffer = ContractOffer.Builder.newInstance().id("id").policy(Policy.Builder.newInstance().build()).build();
+        var contractOffer = ContractOffer.Builder.newInstance()
```
paullatzelsperger, line 148: minor nit: building the offer could be moved to a separate method, to avoid code duplication. We often create two methods, one returning the object itself, when just a generic default is needed, and another one, returning the `Builder`, so that we can easily customize the object.

paullatzelsperger (commented):
minor remarks directly in code.

spi/control-plane/contract-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/contract/offer/ContractOffer.java
wolf4ood: Since i'm here should i change also the policy check?

paullatzelsperger: yes pls :)

paullatzelsperger (approved):

PR #2060, opened 2022-10-07
feat(api): add a `POST` endpoint to request the catalog

### LINKED ISSUE #2051: Investigate whether multi-filter-expressions are supported by DMgmtAPI
# Bug Report

## Describe the Bug
It needs to be determined (and potentially fixed), whether is possible to issue filter expressions, that consist of multiple statements, via the DataManagement API. 

For example, when querying the catalog, we would want to transmit a query param similar to `someProp=someValue,someOtherProp=someOtherValue`. This would also be relevant for all the other DMgmtApi endpoints that accept a query param.

### Expected Behavior
A `QuerySpecDto` with multiple filter expressions gets converted into a `QuerySpec` with the correct filter properties.

### Observed Behavior
none yet, needs to be investigated

- OS: [e.g. iOS, Windows]
- ...

## Detailed Description
Currently the `QuerySpecDto` gets assembled out of various query params using the `@BeanParam` annotation. This was done to make it a `GET` endpoint. 
The `QuerySpecDtoToQuerySpecTransformer` passes the `filter` String directly into the builder method of the `QuerySpec`, which in turn assumes that the filter expression consists of only 3 elements, which seems to be incorrect. 

## Possible Implementation
If we allow those endpoints to be `POST` endpoints, we could easily extend the `QuerySpecDto` to contain a `List<Criterion>` instead of a filter string. That would avoid having to parse a complex filter string that contains multiple criteria, because the criteria would ingress as JSON and thus can be deserialized.

**That would be a breaking API change!**

paullatzelsperger:
## What this PR changes/adds

This PR adds another endpoint called `POST /catalog/request` to the DMgmtApi, which accepts a `CatalogRequestDto` in the body.

## Why it does that

switching over to `POST` allows to pass a more complex object to the endpoint, in particular an actual request object instead of just a filter string.

## Further notes

- the existing `GET /catalog` endpoint was deprecated
- the body is mandatory, as it also contains the `providerUrl`. 
- moved the `CriterionDto` and its transformers to `api-core`
- using a dedicated request object (or modifying/recycling the `QuerySpecDto`) and `POST`ing instead of `GET`ting in all the other endpoints as well should be done in (a) separate PR(s)
- unrelated: added/fixed some copyright headers

## Linked Issue(s)

Closes #2051 

## Checklist

### DIFF
--- a/extensions/control-plane/api/data-management/catalog-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/catalog/CatalogApiController.java
+++ b/extensions/control-plane/api/data-management/catalog-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/catalog/CatalogApiController.java
@@ -71,16 +71,12 @@ public class CatalogApiController implements CatalogApi {
     @POST
     @Path("/request")
     public void requestCatalog(@RequestBody(required = true) CatalogRequestDto requestDto, @Suspended AsyncResponse response) {
+        var result = transformerRegistry.transform(requestDto, QuerySpec.class);
 
-        var query = QuerySpec.Builder.newInstance()
-                .filter(requestDto.getFilter())
-                .limit(requestDto.getLimit())
-                .offset(requestDto.getOffset())
-                .sortOrder(requestDto.getSortOrder())
-                .sortField((requestDto.getSortField()))
-                .build();
-
-        performQuery(requestDto.getProviderUrl(), query, response);
+        if (result.failed()) {
+            throw new InvalidRequestException(result.getFailureMessages());
+        }
+        performQuery(requestDto.getProviderUrl(), result.getContent(), response);
     }
 
     private void performQuery(String providerUrl, QuerySpec spec, AsyncResponse response) {
--- a/extensions/control-plane/api/data-management/catalog-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/catalog/CatalogApiController.java
+++ b/extensions/control-plane/api/data-management/catalog-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/catalog/CatalogApiController.java
@@ -14,15 +14,18 @@
 
 package org.eclipse.dataspaceconnector.api.datamanagement.catalog;
 
+import io.swagger.v3.oas.annotations.parameters.RequestBody;
 import jakarta.validation.Valid;
 import jakarta.ws.rs.BeanParam;
 import jakarta.ws.rs.GET;
+import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.Produces;
 import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.container.AsyncResponse;
 import jakarta.ws.rs.container.Suspended;
 import jakarta.ws.rs.core.MediaType;
+import org.eclipse.dataspaceconnector.api.datamanagement.catalog.model.CatalogRequestDto;
 import org.eclipse.dataspaceconnector.api.datamanagement.catalog.service.CatalogService;
 import org.eclipse.dataspaceconnector.api.query.QuerySpecDto;
 import org.eclipse.dataspaceconnector.api.transformer.DtoTransformerRegistry;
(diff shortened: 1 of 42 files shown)


extensions/control-plane/api/data-management/catalog-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/catalog/CatalogApiController.java
```diff
@@ -61,6 +64,26 @@ public void getCatalog(@jakarta.validation.constraints.NotNull(message = PROVIDE
...
 
+        performQuery(providerUrl, spec, response);
+    }
+
+    @Override
+    @POST
+    @Path("/request")
+    public void requestCatalog(@RequestBody(required = true) CatalogRequestDto requestDto, @Suspended AsyncResponse response) {
+
+        var query = QuerySpec.Builder.newInstance()
```
bscholtes1A, line 75: Is it expected to not use a transformer for mapping the CatalogRequestDto to a QuerySpec?

bscholtes1A (approved):

ronjaquensel (approved):

extensions/control-plane/api/data-management/catalog-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/catalog/CatalogApiController.java
ndr-brt: Same opinion, I'd say it's better to follow the pattern and introduce a transformer.

extensions/control-plane/api/data-management/catalog-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/catalog/model/CatalogRequestDto.java
```diff
@@ -0,0 +1,119 @@
...
+import static java.util.Objects.requireNonNull;
+
+@JsonDeserialize(builder = CatalogRequestDto.Builder.class)
+public class CatalogRequestDto {
+    @PositiveOrZero(message = "offset must be greater or equal to zero")
+    private Integer offset = 0;
+    @Positive(message = "limit must be greater than 0")
+    private Integer limit = 50;
+    private SortOrder sortOrder = SortOrder.ASC;
+    private List<Criterion> filter = new ArrayList<>();
```
ndr-brt, line 38: Better use the `CriterionDto`, currently it's used in the `contractdefinition-api` module, but I would move it upstream to `api-core`, that will permit criteria validation.

extensions/control-plane/api/data-management/catalog-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/catalog/CatalogApiController.java
paullatzelsperger: yes, the reason I didn't yet do it, is because in my view the `CatalogRequestDto` is only here temporarily, as long as we have the `QuerySpecDto` and the `CatalogRequestDto` in parallel. They are very similar, the only difference is that the `QuerySpecDto` has a `String filterExpression`, whereas the `CatalogRequestDto` has a `List<Criterion> filter`. 

Once we have established the pattern to use a `POST` in all the endpoints, we can then modify the `QuerySpecDto` to have the `List<Criterion>`, by which point it would replace the `CatalogRequestDto`.

TL;DR: I was too lazy adding the transformer, because `CatalogRequestDto` is only a temporary class.

extensions/control-plane/api/data-management/catalog-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/catalog/model/CatalogRequestDto.java
paullatzelsperger: ah good catch. Forgot we had that :)

extensions/control-plane/api/data-management/catalog-api/src/main/java/org/eclipse/dataspaceconnector/api/datamanagement/catalog/CatalogApiController.java
paullatzelsperger: @ndr-brt @bscholtes1A since the `CatalogRequestDto` now contains a `CriterionDto`, for which we already have a transformer, I added a `CatalogRequestDtoToQuerySpecTransformer` as well.

ndr-brt (approved):

PR #2084, opened 2022-10-12
feat(api): add `POST` endpoint to all APIs

### LINKED ISSUE #2072: Datamanagement API: implement `POST` endpoints for query
# Feature Request

Currently many endpoints of the DataManagement API use `GET` endpoints to obtain collections of items, for example `GET /assets`. 
These endpoints accept the properties of a `QuerySpecDto` as query parameters using the `@BeanParam` deserialization feature.

We should switch over to using endpoints that follow the `POST /<resource>/request` pattern, and accept the `QuerySpecDto` as request body.

Data Management API

Currently the `QuerySpecDto#filter` field is only capable of handling _one_ filter expression, whereas the `QuerySpec` is able to handle multiple. 
Transmitting a JSON string as query param is cumbersome - it is easier to just transmit it via request body.

## Solution Proposal
- deprecate the current `GET /<resource>` endpoints
- add new endpoint `POST /<resource>/request` 
- add a `List<Criterion> filterExpression` field to `QuerySpecDto` and deprecate `QuerySpecDto#filter`
- transmit the `QuerySpecDto` as request body
- `CatalogApi` only: A `POST` endpoint was already introduced in #2060, so in order to stay consistent we only need to revert to having the `providerUrl` as query parameter and put only the  `QuerySpecDto` in the body 

## Type of Issue
refactoring

## Subtasks
- [ ] `AssetApi`
- [ ] `CatalogApi`
- [ ] `ContractAgreementApi`
- [ ] `ContractDefinitionApi`
- [ ] `ContractNegotiationApi`
- [ ] `PolicyDefinitionApi`
- [ ] `TransferProcessApi`

_[edit] as this is very similar work for all the APIs, I would do all of it in one PR_

## Checklist

paullatzelsperger:
## What this PR changes/adds

adds a `POST /<resource>/request` endpoint to all DataManagementAPIs to enable fully-fledged query capabilities.
for every API endpoint I did the following:
- added an additional method `queryAll...`, annotated with `@POST` and `@Path("/request")`
- marked the old `@GET` method deprecated. Both methods execute the same code.
- added tests
- minor adaptations in `QuerySpecDto` and `CriterionDto`

## Why it does that

To achieve feature parity with IDS in regards to querying.

## Further notes

.

## Linked Issue(s)

Closes #2072 

## Checklist


extensions/common/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/query/QuerySpecDto.java
```diff
@@ -51,6 +59,7 @@ public Integer getLimit() {
         return limit;
     }
 
+    @Deprecated
```
ndr-brt, line 62: At this point I would deprecate the `QuerySpec.Builder.filter(String)` builder method too, to switch to the `QuerySpec.Builder.filter(String)` as well, as it contains some logic that doesn't worth maintaining anymore

ndr-brt (approved):
neat!

extensions/common/api/api-core/src/main/java/org/eclipse/dataspaceconnector/api/query/QuerySpecDto.java
paullatzelsperger: done

juliapampus (approved):

PR #2099, opened 2022-10-14
fix(configuration): strip whitespaces in properties and env values

### LINKED ISSUE #2101: Strip whitespaces from property settings and environment variables
# Feature Request

The current implementation of the configuration regarding the properties or environment variables also allows trailing or ending whitespaces, that could end up in malformed behaviour or crashing the connector at all.

core-util

Some properties including whitespaces could lead to malformed behaviour of several extensions, others are just throwing exceptions during startup.

**Examples: (\*indicates a whitespace)**

Whitespaces in url properties
Will end up in a java.lang.IllegalArgumentException trying to map to an URI
edc.oauth.token.url=https://daps1.int.demo.catena-x.net/token****

Whitespaces in privateKey name property 
Will end up in a java.lang.NullPointerException: privateKey is marked non-null but is null
edc.oauth.private.key.alias=daps_key.pkcs8***

If the related property belongs to a specific extension the error will occur when the respective property will be addressed within the extension during runtime. 

If the property belongs to a base component inside the connector the error will prevent the connector from a proper boot up.

Please feel free to reproduce the behaviour on your own by inserting some whitespaces to some properties.

## Solution Proposal
To prevent that and keep the originated pattern, eg. key=value, i suggest to get rid of trailing or ending whitespaces at his point.

## Type of Issue
feature / improvement

domreuter:
## What this PR changes/adds

Make sure that properties or env variables not starting or ending with whitespaces.

## Why it does that

Ensure that properties with trailing or ending whitespaces are not crashing the components and extensions or end up in a malfunction. Therefore remove all trailing or ending whitespaces if there are any.

## Linked Issue(s)

Closes #2101 

## Checklist

- [notneeded] documented public classes/methods?
- [not needed] added/updated relevant documentation?

### DIFF
--- a/core/common/util/src/main/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctions.java
+++ b/core/common/util/src/main/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctions.java
@@ -35,7 +35,7 @@ public class ConfigurationFunctions {
         if (!StringUtils.isNullOrBlank(value)) {
             return value.strip();
         }
-        return defaultValue.strip();
+        return defaultValue;
     }
 
 }
--- a/core/common/util/src/main/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctions.java
+++ b/core/common/util/src/main/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctions.java
@@ -9,7 +9,7 @@
  *
  *  Contributors:
  *       Microsoft Corporation - initial API and implementation
- *
+ *       SAP SE - Improvements
  */
 
 package org.eclipse.dataspaceconnector.common.configuration;
@@ -28,14 +28,14 @@ public class ConfigurationFunctions {
     public static String propOrEnv(String key, String defaultValue) {
         String value = System.getProperty(key);
         if (!StringUtils.isNullOrBlank(value)) {
-            return value;
+            return value.strip();
         }
         String upperKey = key.toUpperCase().replace('.', '_');
         value = System.getenv(upperKey);
         if (!StringUtils.isNullOrBlank(value)) {
-            return value;
+            return value.strip();
         }
-        return defaultValue;
+        return defaultValue.strip();
     }
 
 }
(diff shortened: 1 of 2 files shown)


core/common/util/src/main/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctions.java
```diff
@@ -28,14 +28,14 @@
...
+            return value.strip();
         }
         String upperKey = key.toUpperCase().replace('.', '_');
         value = System.getenv(upperKey);
         if (!StringUtils.isNullOrBlank(value)) {
-            return value;
+            return value.strip();
         }
-        return defaultValue;
+        return defaultValue.strip();
```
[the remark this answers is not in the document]
jimmarino: This does not need to change the default value as it would be a significant programming error that should be reported via tests.

core/common/util/src/test/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctionsTest.java
```diff
@@ -64,23 +73,20 @@ public void returnSystemProperty(String key, String expected) {
...
     @SetEnvironmentVariable(key = ENV_VAR_2, value = "")
     @SetEnvironmentVariable(key = EDC_VAR_3, value = "    ")
+    @SetEnvironmentVariable(key = EDC_VAR_4, value = VALUE_1 + "    ")
     @MethodSource("envVarsSource")
     public void returnEnv(String key, String expected) {
         String resultValue = ConfigurationFunctions.propOrEnv(key, DEFAULT);
         assertThat(resultValue).isEqualTo(expected);
     }
 
-    @Test
```
jimmarino, line 73: My preference is for explicit tests rather than moving this to the parameterized tests because it verifies a specific case.

jimmarino (changes requested):
Thanks for your contribution. I made some comments for changes inline.

core/common/util/src/test/java/org/eclipse/dataspaceconnector/common/configuration/ConfigurationFunctionsTest.java
domreuter: Both variants cover the exact same thing, therefore it is more about which version you personally prefer. But i am fine with splitting it up in a separate test.

ndr-brt (commented):
As I reported in the issue: https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/2101#issuecomment-1283949547
this is not a good practice, I would suggest to add an explicit validation on the setting.

PR #2116, opened 2022-10-20
feat(transferprocess): introduced a comunication channel from data-plane to control-plane

### LINKED ISSUE #2082: Provide callback capabilities from DataPlane to ControlPlane
# Feature Request

In some situations, for example the one highlighted in #1287, the `data-plane` must inform the `control-plane` of certain ongoings, status updates, etc. Therefore it needs a way to call back "into" the control plane. For now, HTTP should be used for that.

Relates to #2081, Implicitly closes #1287

control plane, data plane

Enable two-way communication between DP and CP to enable status updates.

## Solution Proposal
The `control-plane` attaches a callback URL (=webhook) to the outgoing `DataFlowRequest`, and exposes an API to receive these callbacks.

The `data-plane` then invokes that callback URL.
## Type of Issue
improvement

## Checklist

wolf4ood:
## What this PR changes/adds

Adds a way to callback the control-plane from the data-plane once the transfer has finished (success or failed)

## Why it does that

Enable two-way communication between DP and CP to enable status updates. In
this particular implementation it will allow to mark the TransferProcess as completed instead
of being stuck in `InProgress` forever in the provider side

## Further notes

- `FailTransferCommand` and `CompleteTransferCommand` has been added as command with their handler in order to mark a transfer process failed or completed.
- `ControlPlaneApiUrl` interface has been added in `transfer-spi` for attaching the callback URL to the `DataFlowRequest`
- `control-plane-api`  module provides the `ControlPlaneApiUrl` and exposes the Rest API for sending the `FailTransferCommand` and `CompleteTransferCommand` to the TPM
- `control-plane-api-client-spi` module provides an interface for calling back the Control Plane
- `control-plane-api-client`  module provides an HTTP implementation for calling the Control Plane Rest API
- The client is then injected into the DPMImpl for calling back the CP when the transfer at DP level has finished (success or failed)

For now it's backward compatible since the injected `ControlPlaneApiUrl` and the client in the DPMimpl are not required by default. If they are not provided the behavior will be the same as today.

## Linked Issue(s)

Closes #2082 

## Checklist

### DIFF
--- a/core/control-plane/control-plane-api-client/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/client/transferprocess/TransferProcessHttpClient.java
+++ b/core/control-plane/control-plane-api-client/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/client/transferprocess/TransferProcessHttpClient.java
@@ -1,0 +1,111 @@
+/*
+ *  Copyright (c) 2020, 2022 Microsoft Corporation
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.controlplane.api.client.transferprocess;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import dev.failsafe.Failsafe;
+import dev.failsafe.RetryPolicy;
+import okhttp3.MediaType;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import org.eclipse.dataspaceconnector.controlplane.api.client.transferprocess.model.TransferProcessFailRequest;
+import org.eclipse.dataspaceconnector.spi.controlplane.api.client.transferprocess.TransferProcessApiClient;
+import org.eclipse.dataspaceconnector.spi.monitor.Monitor;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.DataFlowRequest;
+import org.jetbrains.annotations.NotNull;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * Implementation of {@link TransferProcessApiClient} which talks to the Control Plane Transfer Process via HTTP APIs
+ */
+public class TransferProcessHttpClient implements TransferProcessApiClient {
+
+    public static final MediaType TYPE_JSON = MediaType.parse("application/json");
+
+    private final RetryPolicy<Object> retryPolicy;
+    private final OkHttpClient okHttpClient;
+
+    private final ObjectMapper mapper;
+    private final Monitor monitor;
+
+    public TransferProcessHttpClient(OkHttpClient okHttpClient, RetryPolicy<Object> retryPolicy, ObjectMapper mapper, Monitor monitor) {
+        this.okHttpClient = okHttpClient;
+        this.retryPolicy = retryPolicy;
+        this.mapper = mapper;
+        this.monitor = monitor;
+    }
+
+
+    @Override
+    public void completed(DataFlowRequest dataFlowRequest) {
+        sendRequest(dataFlowRequest, "complete", null);
+    }
+
+    @Override
+    public void failed(DataFlowRequest dataFlowRequest, String reason) {
+        sendRequest(dataFlowRequest, "fail", TransferProcessFailRequest.Builder.newInstance().errorMessage(reason).build());
+    }
+
+
+    private void sendRequest(DataFlowRequest dataFlowRequest, String action, Object body) {
+
+        if (dataFlowRequest.getCallbackAddress() != null) {
+            try {
+                var rq = createRequest(buildUrl(dataFlowRequest, action), body);
+                try (var response = executeRequest(rq)) {
+                    if (!response.isSuccessful()) {
+                        monitor.severe(String.format("Failed to send callback request: received %s from the TransferProcess API", response.code()));
+                    }
+                }
+
+            } catch (Exception e) {
+                monitor.severe("Failed to send callback request", e);
+            }
+        } else {
+            monitor.warning(String.format("Missing callback address in DataFlowRequest %s", dataFlowRequest.getId()));
+        }
+    }
+
+    @NotNull
+    private String buildUrl(DataFlowRequest dataFlowRequest, String action) throws URISyntaxException {
+        var url = new URI(dataFlowRequest.getCallbackAddress().toString() + "/").resolve(String.format("./transferprocess/%s/%s", dataFlowRequest.getProcessId(), action)).normalize();
+        return url.toString();
+    }
+
+    @NotNull
+    private Response executeRequest(Request rq) {
+        return Failsafe.with(retryPolicy).get(() -> okHttpClient.newCall(rq).execute());
+    }
+
+
+    private Request createRequest(String url, Object body) throws JsonProcessingException {
+        RequestBody requestBody = null;
+        if (body != null) {
+            requestBody =
+                    RequestBody.create(mapper.writeValueAsString(body), TYPE_JSON);
+        } else {
+            requestBody = RequestBody.create("", null);
+        }
+        return new Request.Builder()
+                .url(url)
+                .post(requestBody)
+                .build();
+    }
+}
--- a/core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
+++ b/core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
@@ -1,0 +1,109 @@
+/*
+ *  Copyright (c) 2020, 2022 Microsoft Corporation
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.controlplane.api;
+
+import org.eclipse.dataspaceconnector.controlplane.api.transferprocess.TransferProcessControlApiController;
+import org.eclipse.dataspaceconnector.controlplane.api.transferprocess.model.TransferProcessFailStateDto;
+import org.eclipse.dataspaceconnector.runtime.metamodel.annotation.Extension;
+import org.eclipse.dataspaceconnector.runtime.metamodel.annotation.Inject;
+import org.eclipse.dataspaceconnector.runtime.metamodel.annotation.Provider;
+import org.eclipse.dataspaceconnector.spi.EdcException;
+import org.eclipse.dataspaceconnector.spi.WebServer;
+import org.eclipse.dataspaceconnector.spi.WebService;
+import org.eclipse.dataspaceconnector.spi.system.Hostname;
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtension;
+import org.eclipse.dataspaceconnector.spi.system.ServiceExtensionContext;
+import org.eclipse.dataspaceconnector.spi.transfer.TransferProcessManager;
+import org.eclipse.dataspaceconnector.spi.transfer.callback.ControlPlaneApiUrl;
+import org.jetbrains.annotations.NotNull;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+import static java.lang.String.format;
+
+/**
+ * {@link ControlPlaneApiExtension } exposes HTTP endpoints for internal interaction with the Control Plane
+ */
+@Extension(value = ControlPlaneApiExtension.NAME)
+public class ControlPlaneApiExtension implements ServiceExtension {
+
+    public static final String NAME = "Control Plane API";
+
+    public static final String CONTROL_PLANE_API_CONFIG = "web.http.controlplane";
+    public static final String CONTROL_PLANE_CONTEXT_ALIAS = "controlplane";
+
+    public static final int DEFAULT_CONTROL_PLANE_API_PORT = 8384;
+    public static final String DEFAULT_CONTROL_PLANE_API_CONTEXT_PATH = "/api/v1/controlplane";
+
+    private int port = DEFAULT_CONTROL_PLANE_API_PORT;
+    private String path = DEFAULT_CONTROL_PLANE_API_CONTEXT_PATH;
+
+    @Inject
+    private WebServer webServer;
+    @Inject
+    private WebService webService;
+    @Inject
+    private TransferProcessManager transferProcessManager;
+    @Inject
+    private Hostname hostname;
+
+    @Override
+    public String name() {
+        return NAME;
+    }
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        var monitor = context.getMonitor();
+        var alias = CONTROL_PLANE_CONTEXT_ALIAS;
+
+        var config = context.getConfig(CONTROL_PLANE_API_CONFIG);
+        if (config.getEntries().isEmpty()) {
+            monitor.warning(format("Settings for [%s] and/or [%s] were not provided. Using default" +
+                    " value(s) instead.", CONTROL_PLANE_API_CONFIG + ".path", CONTROL_PLANE_API_CONFIG + ".path"));
+            webServer.addPortMapping(alias, port, path);
+        } else {
+            path = config.getString("path", path);
+            port = config.getInteger("port", port);
+        }
+
+        context.getTypeManager().registerTypes(TransferProcessFailStateDto.class);
+
+        monitor.info(format("Control Plane API will be available at [path=%s], [port=%s].", path, port));
+
+        webService.registerResource(alias, new TransferProcessControlApiController(transferProcessManager));
+    }
+
+
+    @Provider
+    public ControlPlaneApiUrl controlPlaneApiUrl(ServiceExtensionContext context) {
+        var s = getApiUrl();
+        try {
+            var url = new URL(s);
+            return () -> url;
+        } catch (MalformedURLException e) {
+            context.getMonitor().severe("Error creating callback endpoint", e);
+            throw new EdcException(e);
+        }
+    }
+
+    @NotNull
+    private String getApiUrl() {
+        return String.format("http://%s:%s%s", hostname.get(), port, path);
+    }
+
+
+}
--- a/core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/TransferProcessControlApiController.java
+++ b/core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/TransferProcessControlApiController.java
@@ -1,0 +1,58 @@
+/*
+ *  Copyright (c) 2020, 2022 Microsoft Corporation
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.controlplane.api.transferprocess;
+
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotNull;
+import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.core.MediaType;
+import org.eclipse.dataspaceconnector.controlplane.api.transferprocess.model.TransferProcessFailStateDto;
+import org.eclipse.dataspaceconnector.spi.transfer.TransferProcessManager;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.command.CompleteTransferCommand;
+import org.eclipse.dataspaceconnector.spi.types.domain.transfer.command.FailTransferCommand;
+
+
+@Consumes({ MediaType.APPLICATION_JSON })
+@Produces({ MediaType.APPLICATION_JSON })
+@Path(TransferProcessControlApiController.PATH)
+public class TransferProcessControlApiController implements TransferProcessControlApi {
+
+    public static final String PATH = "/transferprocess";
+    private final TransferProcessManager transferProcessManager;
+
+
+    public TransferProcessControlApiController(TransferProcessManager transferProcessManager) {
+        this.transferProcessManager = transferProcessManager;
+    }
+
+
+    @POST
+    @Path("/{processId}/complete")
+    @Override
+    public void complete(@PathParam("processId") String processId) {
+        transferProcessManager.enqueueCommand(new CompleteTransferCommand(processId));
+    }
+
+    @POST
+    @Path("/{processId}/fail")
+    @Override
+    public void fail(@PathParam("processId") String processId, @NotNull @Valid TransferProcessFailStateDto request) {
+        transferProcessManager.enqueueCommand(new FailTransferCommand(processId, request.getErrorMessage()));
+    }
+}
--- a/core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/model/TransferProcessFailStateDto.java
+++ b/core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/model/TransferProcessFailStateDto.java
@@ -1,0 +1,58 @@
+/*
+ *  Copyright (c) 2020, 2022 Microsoft Corporation
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *
+ */
+
+package org.eclipse.dataspaceconnector.controlplane.api.transferprocess.model;
+
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import jakarta.validation.constraints.NotNull;
+
+@JsonDeserialize(builder = TransferProcessFailStateDto.Builder.class)
+public class TransferProcessFailStateDto {
+    @NotNull
+    private String errorMessage;
+
+    private TransferProcessFailStateDto() {
+
+    }
+
+    public String getErrorMessage() {
+        return errorMessage;
+    }
+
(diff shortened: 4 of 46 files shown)


core/data-plane/data-plane-framework/src/main/java/org/eclipse/dataspaceconnector/dataplane/framework/manager/DataPlaneManagerImpl.java
```diff
@@ -152,16 +152,37 @@ private void processDataFlowRequest(DataFlowRequest request) {
...
+            transferProcessClient.failed(request, result.getFailureDetail());
+        }
+    }
+
     public static class Builder {
-        private DataPlaneManagerImpl manager;
+        private final DataPlaneManagerImpl manager;
+
+        private Builder() {
+            manager = new DataPlaneManagerImpl();
```
paullatzelsperger, line 183: unrelated, but the CTor of the DataPlaneManagerImpl could be made `private`, to force the use of the Builder

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
```diff
@@ -0,0 +1,125 @@
...
+        monitor.info(format("Control Plane API will be available at [path=%s], [port=%s].", path, port));
+
+        webService.registerResource(alias, new TransferProcessControlApiController(transferProcessManager));
+        webService.registerResource(alias, new AuthenticationRequestFilter(authService));
+    }
+
+
+    @Provider
+    public ControlPlaneApiUrl controlPlaneApiUrl(ServiceExtensionContext context) {
+        String s = getApiUrl();
```
bscholtes1A, line 100: var s

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
```diff
@@ -0,0 +1,125 @@
...
+    @NotNull
+    private String getApiUrl() {
+        var s = hostname.get();
+
+        if (!s.startsWith("http")) { // a hostname should never have a protocol prefix, but just to be safe
+            s = "http://" + s;
+        }
+        if (!s.matches(".*:([0-9]){1,5}$")) { // a hostname also shouldn't have a port, but again, to be sure
+            s += ":" + port;
+        }
```
bscholtes1A, line 119: Shouldnt these checks be performed in the `Hostname` class?

core/control-plane/control-plane-api-client/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/client/transferprocess/TransferProcessHttpClient.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @NotNull
+    private String buildUrl(DataFlowRequest dataFlowRequest, String action) throws URISyntaxException {
+        var url = new URI(dataFlowRequest.getCallbackAddress() + "/").resolve(String.format("./transferprocess/%s/%s", dataFlowRequest.getProcessId(), action)).normalize();
+        return url.toString();
+    }
+
+    @NotNull
+    private Response executeRequest(Request rq) throws IOException {
```
ndr-brt, line 101: `IOException` is actually not throwable here (handled by `Failsafe`)

bscholtes1A (changes requested):

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
```diff
@@ -0,0 +1,125 @@
...
+ * {@link ControlPlaneApiExtension } exposes HTTP endpoints for internal interaction with the Control Plane
+ */
+@Provides(ControlPlaneApiUrl.class)
+@Extension(value = ControlPlaneApiExtension.NAME)
+public class ControlPlaneApiExtension implements ServiceExtension {
+
+    public static final String NAME = "Control Plane API";
+
+    public static final String CONTROL_PLANE_API_CONFIG = "web.http.controlplane";
+    public static final String CONTROL_PLANE_CONTEXT_ALIAS = "controlplane";
```
ndr-brt, line 50: Maybe "control" would be more generic and fit with the context already defined by the data-plane (DataPlaneExtension), so if control-plane and data-plane are deployed together the control api will be exposed on the same context.
I would change it also at line 49

core/control-plane/transfer-core/src/main/java/org/eclipse/dataspaceconnector/transfer/core/command/handlers/SingleTransferProcessCommandHandler.java
```diff
@@ -56,9 +56,10 @@ public void handle(SingleTransferProcessCommand command) {
      * If the {@link TransferProcess} was indeed modified, implementors should return {@code true}, otherwise {@code false}
      *
      * @param process The {@link TransferProcess}
+     * @param command The {@link SingleTransferProcessCommand}
      * @return true if the process was actually modified, false otherwise.
      */
-    protected abstract boolean modify(TransferProcess process);
+    protected abstract boolean modify(TransferProcess process, T command);
```
paullatzelsperger, line 62: This is not needed - only the `FailTransferCommandHandler` requires access to the command object, and that can be achieved by hooking into the `handle` method:

```java
// in FailTransferCommandHandler
@Override
public void handle(FailTransferCommand command) {
        // store for later use
        this.failCommand = command;
        // delegate back up the chain
        super.handle(command);
    }

@Override
protected boolean modify(TransferProcess process){
        // use this.failCommand
}
```

core/control-plane/control-plane-api-client/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/client/transferprocess/TransferProcessHttpClient.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Override
+    public void failed(DataFlowRequest dataFlowRequest, String reason) {
+        withUrlGuard(dataFlowRequest, () ->
+                sendRequest(dataFlowRequest, "fail", TransferProcessFailRequest.Builder.newInstance().errorMessage(reason).build()));
+    }
+
+
+    public void withUrlGuard(DataFlowRequest dataFlowRequest, Runnable callback) {
```
paullatzelsperger, line 72: why not perform the url check directly in the `sendRequest` method?

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
```diff
@@ -0,0 +1,125 @@
...
+    public String name() {
+        return NAME;
+    }
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        var monitor = context.getMonitor();
+        var alias = CONTROL_PLANE_CONTEXT_ALIAS;
+
+        var config = context.getConfig(CONTROL_PLANE_API_CONFIG);
```
ndr-brt, line 79: This piece of code (getting the context config, adding port mapping and printing log) is used also in other classes as `IdsApiConfigurationExtension`, `DataManagementApiConfigurationExtension`, `HttpWebhookExtension`... could it be a good moment to extract an utility method? Otherwise a dedicated issue could be just enough for now.

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/TransferProcessControlApiController.java
```diff
@@ -0,0 +1,63 @@
...
+        Objects.requireNonNull(processId);
+
+        transferProcessManager.enqueueCommand(new CompleteTransferCommand(processId));
+    }
+
+    @POST
+    @Path("/{processId}/fail")
+    @Override
+    public void fail(@PathParam("processId") String processId, @Valid TransferProcessFailStateDto request) {
+        Objects.requireNonNull(processId);
```
paullatzelsperger, line 59: Better to use `jakarta.validation.constraints` annotations, e.g. `@NotNull`

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
ndr-brt: I agree, in `CoreServicesExtension` we should ensure that this is only an hostname (so no protocol nor port added)

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/model/TransferProcessFailStateDto.java
```diff
@@ -0,0 +1,56 @@
...
+package org.eclipse.dataspaceconnector.controlplane.api.transferprocess.model;
+
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+@JsonDeserialize(builder = TransferProcessFailStateDto.Builder.class)
+public class TransferProcessFailStateDto {
+    private String errorMessage;
```
paullatzelsperger, line 24: Should be annotated with `jakarta.validation.constraints.NotNull`

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/TransferProcessControlApiController.java
```diff
@@ -0,0 +1,63 @@
...
+    public TransferProcessControlApiController(TransferProcessManager transferProcessManager) {
+        this.transferProcessManager = transferProcessManager;
+    }
+
+
+    @POST
+    @Path("/{processId}/complete")
+    @Override
+    public void complete(@PathParam("processId") String processId) {
+        Objects.requireNonNull(processId);
```
paullatzelsperger, line 50: Not 100% sure, but I think path params cannot be `null` anyway, so this would be redundant.

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
```diff
@@ -0,0 +1,125 @@
...
+            path = config.getString("path", path);
+            port = config.getInteger("port", port);
+        }
+
+        context.getTypeManager().registerTypes(TransferProcessFailStateDto.class);
+
+        monitor.info(format("Control Plane API will be available at [path=%s], [port=%s].", path, port));
+
+        webService.registerResource(alias, new TransferProcessControlApiController(transferProcessManager));
+        webService.registerResource(alias, new AuthenticationRequestFilter(authService));
```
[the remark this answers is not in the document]
ndr-brt: Agree, at least they could be exposed on the same context, using the same port and base path

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/model/TransferProcessFailStateDto.java
```diff
@@ -0,0 +1,56 @@
...
+
+package org.eclipse.dataspaceconnector.controlplane.api.transferprocess.model;
+
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+@JsonDeserialize(builder = TransferProcessFailStateDto.Builder.class)
+public class TransferProcessFailStateDto {
```
paullatzelsperger, line 23: Seems like this class and `TransferProcessFailRequest` are redundant. could we delete one?

spi/common/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/transfer/DataFlowRequest.java
```diff
@@ -37,6 +37,7 @@ public class DataFlowRequest implements Polymorphic, TraceCarrier {
     private DataAddress sourceDataAddress;
     private DataAddress destinationDataAddress;
 
+    private String callbackAddress;
```
ndr-brt, line 40: Could it be an URI/URL directly?

core/data-plane/data-plane-framework/src/main/java/org/eclipse/dataspaceconnector/dataplane/framework/DataPlaneFrameworkExtension.java
```diff
@@ -108,6 +112,7 @@ public void initialize(ServiceExtensionContext context) {
                 .pipelineService(pipelineService)
                 .transferServiceRegistry(transferServiceRegistry)
                 .store(registerStore(context))
+                .transferProcessClient(transferProcessApiClient)
```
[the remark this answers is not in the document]
ndr-brt: agree

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
ndr-brt: I got it, but these checks could be moved to one single point, and validate a config setting seems like a sane approach instead of introducing a whole lot of logic aroung

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private int port = DEFAULT_CONTROL_PLANE_API_PORT;
+    private String path = DEFAULT_CONTROL_PLANE_API_CONTEXT_PATH;
+
+    @Inject
+    private WebServer webServer;
+    @Inject
+    private WebService webService;
+    @Inject
+    private AuthenticationService authService;
```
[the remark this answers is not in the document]
wolf4ood: I agree auth keys are used for external comunication. There is no clear boundary that i found from External to IPC communication and also authentication

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/model/TransferProcessFailStateDto.java
wolf4ood: We could i only didn't know where to put this common payload class. 
The two modules `control-plane-api` and `control-plane-api-client` don't share a common module so i was not sure

core/control-plane/transfer-core/src/main/java/org/eclipse/dataspaceconnector/transfer/core/command/handlers/SingleTransferProcessCommandHandler.java
wolf4ood: Isn't the handler registered only once at startup time?

core/control-plane/control-plane-api-client/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/client/transferprocess/TransferProcessHttpClient.java
wolf4ood: yeah we could i didn't want to introduce the third layer of if try nesting but it's fine to place it there :)

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
wolf4ood: I would raise a refactoring issue, for this, otherwise i fear it would overcomplicare this PR

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
wolf4ood: So we can just assume here that the Hostname is clean, got it :)

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
bscholtes1A: I am afraid this would break some existing behaviors... As here it would lead to registering the injected `authService` under the `control` context (see l. 94), while this context is already used when the data plane validates the access tokens against the control plane. See here:

https://user-images.githubusercontent.com/88310985/157418442-da5cc5aa-6b54-42b3-b836-e4f0a3f46fcf.png

ndr-brt: @bscholtes1A yes, you're right, this could fall in the same context of [the comment I left about the `AuthenticationService`](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/2116#discussion_r1003121732), maybe the authentication should be related to the context so we could have an authentication dedicated to the intra-connector communication. To be honest for the moment I would remove the authentication on this extension and create a dedicated issue for it.

bscholtes1A: Completely agreed, let's remove the Auth Service for now and uniformize this Control Plane API later on.

core/control-plane/transfer-core/src/main/java/org/eclipse/dataspaceconnector/transfer/core/command/handlers/SingleTransferProcessCommandHandler.java
paullatzelsperger: yes, the handler is registered only once, but the `handle` method should be invoked whenever the handler receives a command

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
wolf4ood: Ok so do you propose to remove the auth filter and also from the client remove the auth key right?

core/control-plane/transfer-core/src/main/java/org/eclipse/dataspaceconnector/transfer/core/command/handlers/SingleTransferProcessCommandHandler.java
wolf4ood: Yes but saving state between handle calls could lead to strange bug due being the same instance right?

spi/common/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/transfer/DataFlowRequest.java
wolf4ood: Does Jackson support URL or do we have a custom ser/de for that?

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
wolf4ood: Raise here https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/issues/2128 :)

core/control-plane/transfer-core/src/main/java/org/eclipse/dataspaceconnector/transfer/core/command/handlers/SingleTransferProcessCommandHandler.java
paullatzelsperger: the state, i.e. the command, would only be valid between `handle` and `modify`, and should be set to `null` after that, that's correct.

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/model/TransferProcessFailStateDto.java
paullatzelsperger: Ah, that's right. 
Technically there should be a shared module, something like a `control-plane-api-model`, but for now that would only contain the `TransferProcessFailStateDto`. Not sure that is worth it. That could be revisited, should the control plane's internal API ever become "wider".
So I guess we could leave it for now.

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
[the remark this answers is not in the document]
jimmarino: @wolf4ood Can you create another issue to track this for later?

spi/common/core-spi/src/main/java/org/eclipse/dataspaceconnector/spi/types/domain/transfer/DataFlowRequest.java
ndr-brt: It should be supported as well as URI, just tried it out in an unit test

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
ndr-brt: Yes, since this API is not supposed to be exposed to the outside, as it is only for intra-connector communication, I don't see it as a big issue to leave it without auth for now, we should bring endpoint authentication to an higher level

core/control-plane/transfer-core/src/main/java/org/eclipse/dataspaceconnector/transfer/core/command/handlers/SingleTransferProcessCommandHandler.java
ndr-brt: @paullatzelsperger I don't think it's a good idea to make the `CommandHandler` stateful, I like the @wolf4ood solution more, we expect that the command brings some information needed to be stored in the entity, so having the command passed to the `modify` method makes total sense.

paullatzelsperger: @ndr-brt I don't quite understand what the problem is in carrying the command over from `handle()` to `modify()` granted that it gets nullified at the end of `modify()`. It seems much less intrusive than adding a method parameter that is only used in one particular situation.

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/ControlPlaneApiExtension.java
bscholtes1A: @wolf4ood  could you please rename the context to `control` now that the AuthService has been removed?

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/TransferProcessControlApiController.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @POST
+    @Path("/{processId}/complete")
+    @Override
+    public void complete(@PathParam("processId") String processId) {
+        transferProcessManager.enqueueCommand(new CompleteTransferCommand(processId));
+    }
+
+    @POST
+    @Path("/{processId}/fail")
```
bscholtes1A, line 53: MINOR: Should be `failed` instead of `fail`, no?

core/data-plane/data-plane-framework/src/main/java/org/eclipse/dataspaceconnector/dataplane/framework/DataPlaneFrameworkExtension.java
```diff
@@ -129,6 +135,16 @@ public void shutdown() {
         }
     }
 
+    @NotNull
+    private TransferProcessApiClient registerTransferProcessClient(ServiceExtensionContext context) {
+        if (transferProcessApiClient != null) {
+            return transferProcessApiClient;
+        }
+        var noopClient = new NoopTransferProcessClient();
+        context.registerService(TransferProcessApiClient.class, noopClient);
```
bscholtes1A, line 144: So we are missing the @Provides annotation at the top of the class then ? 

Couldn't we use a default Provider for the NoopTransferClient? 

@ndr-brt  what do you think?

core/data-plane/data-plane-framework/src/main/java/org/eclipse/dataspaceconnector/dataplane/framework/manager/DataPlaneManagerImpl.java
```diff
@@ -212,13 +238,13 @@ public Builder store(DataPlaneStore store) {
...
+        public Builder transferProcessClient(TransferProcessApiClient transferProcessClient) {
+            manager.transferProcessClient = transferProcessClient;
+            return this;
         }
 
-        private Builder() {
-            manager = new DataPlaneManagerImpl();
-            this.manager.telemetry = new Telemetry(); // default noop implementation
+        public DataPlaneManagerImpl build() {
+            return manager;
```
bscholtes1A, line 247: Assert here that `transferProcessClient` is not null.

core/data-plane/data-plane-framework/src/main/java/org/eclipse/dataspaceconnector/dataplane/framework/manager/DataPlaneManagerImpl.java
```diff
@@ -152,16 +156,38 @@ private void processDataFlowRequest(DataFlowRequest request) {
...
+
+                onTransferFinished(request, result, exception);
+
+
             });
         }
     }
 
+    private void onTransferFinished(DataFlowRequest request, StatusResult<Void> result, Throwable exception) {
+        if (transferProcessClient != null) {
```
bscholtes1A, line 173: Not needed if already asserted in the builder.

bscholtes1A (changes requested):

core/control-plane/control-plane-api/src/main/java/org/eclipse/dataspaceconnector/controlplane/api/transferprocess/TransferProcessControlApiController.java
ndr-brt: In this case the DataFlowRequest failed on data-plane, but this call is meant to fail the transfer process on the control plane (like /complete completes it), so to me "fail" fits well

core/data-plane/data-plane-framework/src/main/java/org/eclipse/dataspaceconnector/dataplane/framework/DataPlaneFrameworkExtension.java
ndr-brt: @bscholtes1A right. To be coherent with the convention, I'd create a `DataPlaneDefaultServices` where that noop `TransferProcessClient` is provided by default

ndr-brt (approved):

bscholtes1A (approved):

PR #2121, opened 2022-10-21
docs: add decision record about EDC version catalogs

paullatzelsperger:
## What this PR changes/adds

Adds a decision record about the usage of version catalogs in EDC.

## Why it does that

Documentation and decision tracking.

## Further notes

- the implementation will be spread out over the GradlePlugins repo, EDC, and all the other projects (IH, RS, MVD,...)

## Checklist

### DIFF
--- a/docs/developer/decision-records/2022-10-21-gradle-versioncatalogs/README.md
+++ b/docs/developer/decision-records/2022-10-21-gradle-versioncatalogs/README.md
@@ -50,6 +50,30 @@ Version catalogs are lightweight, much more so than platforms (which actually in
 should be understood as the EDC project team's recommendation. Clients can then use the platform feature to apply them
 and restrict the dependency graph or simply choose to override them at their own risk.
 
+### A word on naming
+
+Version catalogs automatically convert separated aliases into hierarchical structures, so `azure-resourcemanager-auth`
+would be converted into `azure.resourcemanager.auth`. As a general rule of thumb those aliases should include the
+project name and the module that is being imported, for example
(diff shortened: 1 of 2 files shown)


jimmarino (approved):
Looks good. Can you include a section that sets the naming convention for version catalogs, i.e., when creating them? This will avoid naming divergence.

ronjaquensel (approved):

ndr-brt (approved):
Looks good, I only have some doubts, likely given by my missing knowledge about the version catalogs.
- if the versions are handled centrally in the `EDC Build Plugin` how we know if a version update doesn't break any component? Would we need a CI job that builds all the projects with the new version before merging the "dependency bump" PR?
- would adding a new external dependency require a PR to the build plugin first right? This actually could be good because it makes the process steeper.

paullatzelsperger:
> if the versions are handled centrally in the `EDC Build Plugin` how we know if a version update doesn't break any component? Would we need a CI job that builds all the projects with the new version before merging the "dependency bump" PR?

Version catalogs don't enforce the use of versions (that what `platform`s are for), so it is possible to bump and override a version locally (i.e. in the respective project) for testing. These kinds of build triggers are not possible with Github out of the box, but we could investigate, whether such an action exists. Jenkins can do that though.

> * would adding a new external dependency require a PR to the build plugin first right? This actually could be good because it makes the process steeper.

Technically no, because every project can still have "it's own" dependencies. I would not necessarily enforce the adoption of every 3rd party dep into the version catalog, only those that are used in multiple places/modules/projects.

As a second step we could investigate the usage of `platforms`, which actually influence the dependency graph and could be used to _enforce_ dependencies.

PR #2135, opened 2022-10-25
feat: implement oauth2 provision module

### LINKED ISSUE #2137: transfer data from/to http endpoints with oauth2
# Feature Request

Currently the `data-plane-http` can access enpoint that requires authentication through details that can be stored in the vault.
However in an OAuth2 context where the tokens have a certain validity this is not doable, as a token will expire sooner or later, so there's the need to request the token prior the transfer.

provision

Enabling OAuth2 autenticated data source/sink endpoints

## Solution Proposal
Introduce an `oauth2-provision` extension that will take care of obtaining a token from an oauth2 server, storing it in the vault where the data-plane will use it to get/put the data.

diagram

## Type of Issue
improvement

ndr-brt:
## What this PR changes/adds

Implement an `OAuth2` module, that will permit data-plane to authenticate with an oauth2 access token to get the data from an HTTP endpoint.

## Why it does that

Permit authentication to an HTTP data source endpoint for data-transfer, both source and sink sides.

## Further notes

- Made `FsVault` reload the vault file if a secret is not found. Needed to make the E2E test work.
- Made `HttpRequestParametersSupplier` try to deserialize the secret as a map to extract the token before, if this doesn't work, it uses the string as authentication header (current behavior)
- Pulled up `dataAddressBuilder` field to `ProvisionedDataAddressResource`

## Linked Issue(s)

Closes #2137

## Checklist


extensions/control-plane/provision/oauth2-provision/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## How it works
+![](docs/diagram.png)
+
+The token will be requested in the provision phase and stored in the vault, then it will be used by the `data-plane-http`
+extension and put in the request.
+
+## How to use it
+
+The extension works all the `HttpData` addresses that contain the properties (defined in
```
bscholtes1A, line 13: typo? "works for all"

bscholtes1A (approved):

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParamsSupplier.java
```diff
@@ -30,9 +32,11 @@
 public abstract class HttpRequestParamsSupplier implements Function<DataFlowRequest, HttpRequestParams> {
 
     private final Vault vault;
+    private final TypeManager typeManager;
 
-    protected HttpRequestParamsSupplier(Vault vault) {
+    protected HttpRequestParamsSupplier(Vault vault, TypeManager typeManager) {
         this.vault = vault;
+        this.typeManager = typeManager;
```
paullatzelsperger, line 39: we're not using any functionality specifically from the `TypeManager`, so using _just_ an `ObjectMapper` would be better IMO.

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/types/domain/HttpDataAddress.java
```diff
@@ -222,7 +224,7 @@ public Builder nonChunkedTransfer(boolean nonChunkedTransfer) {
         }
 
         public Builder copyFrom(DataAddress other) {
-            other.getProperties().forEach(this::property);
+            Optional.ofNullable(other).map(DataAddress::getProperties).orElse(emptyMap()).forEach(this::property);
```
paullatzelsperger, line 227: nit: static import for `ofNullable`

spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/transfer/spi/types/SecretToken.java
```diff
@@ -27,5 +25,4 @@ public interface SecretToken extends Polymorphic {
 
     long getExpiration();
```
paullatzelsperger, line 27: The `SimpleTokenTest` could thus also be deleted entirely.

system-tests/e2e-transfer-test/backend-service/src/main/java/org/eclipse/edc/test/e2e/Oauth2TokenController.java
```diff
@@ -0,0 +1,47 @@
...
+    @POST
+    @Path("/token")
+    public Map<String, String> getToken(@FormParam("client_id") String clientId, @FormParam("client_secret") String clientSecret) {
+        monitor.info("Oauth2 token requested");
+        if (!"clientId".equals(clientId) || !"clientSecret".equals(clientSecret)) {
+            var message = format("Cannot validate token request, parameters are not valid: client_id %s - client_secret %s", clientId, clientSecret);
+            monitor.severe(message);
+            throw new InvalidRequestException(message);
+        }
+        return Map.of("access_token", "atoken"); // TODO: there's a way to easily provide an actual jwt token?
```
paullatzelsperger, line 45: you could try Nimbus' `PlainJWT` class. Haven't verified myself, but could be a starting point.

extensions/control-plane/provision/provision-oauth2/README.md
```diff
@@ -0,0 +1,18 @@
+# OAuth2 Provision
+
+Permits to attach an OAuth2 token to the transfer, needed when the source or the sink require it.
+
+## How it works
+![](docs/diagram.png)
+
+The token will be requested in the provision phase and stored in the vault, then it will be used by the `data-plane-http`
```
paullatzelsperger, line 8: If I'm honest, I would improve the README in the following aspects:
- In what situation would one use this extension? Meaning: is it used to authenticate the provider against a _private_ API in his own environment, or is it used to authenticate a consumer against a publicly exposed API of the provider (i.e. the proxy usecase)?
- To me it is not clear, whether the `consumer` or the `provider` (or both?) have to request the OAuth token. In any case, it would be beneficial to clarify this, in the diagram and/or the README.
- Which Oauth2 flows does this apply to? And if it applies to all, or if the flow is irrelevant to us ("out-of-band"), that should be noted too IMO.
- What happens if the token expires between the provisioning and the transfer phase? That could happen if more resources are to be provisioned, and one of them takes longer than the expiry time of the token (e.g. auth code flow)?

extensions/control-plane/provision/provision-oauth2/src/main/java/org/eclipse/edc/connector/provision/oauth2/Oauth2DataAddressValidator.java
```diff
@@ -0,0 +1,42 @@
...
+ * Validates {@link DataAddress}, returns true if the Address has the fields needed for the OAuth2 provisioning
+ */
+public class Oauth2DataAddressValidator implements Predicate<DataAddress> {
+
+    private final Predicate<DataAddress> isHttpDataType = dataAddress -> "HttpData".equals(dataAddress.getType());
+    private final Predicate<DataAddress> hasClientId = dataAddress -> dataAddress.hasProperty(CLIENT_ID);
+    private final Predicate<DataAddress> hasClientSecret = dataAddress -> dataAddress.hasProperty(CLIENT_SECRET);
+    private final Predicate<DataAddress> hasTokenUrl = dataAddress -> dataAddress.hasProperty(TOKEN_URL);
+
+    private final Predicate<DataAddress> isValid = isHttpDataType.and(hasClientId).and(hasClientSecret).and(hasTokenUrl);
```
paullatzelsperger, line 35: nifty!

extensions/control-plane/provision/provision-oauth2/src/main/java/org/eclipse/edc/connector/provision/oauth2/Oauth2SecretToken.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private final String token;
+
+    public Oauth2SecretToken(@JsonProperty("token") String token) {
+        this.token = token;
+    }
+
+    @Override
+    public long getExpiration() {
+        return 0;
```
paullatzelsperger, line 33: OAuth Tokens can indeed have an expiry time encoded in the `expires_in` field, at least in the auth code flow. Check [here](https://www.oauth.com/oauth2-servers/access-tokens/access-token-lifetime/) and the [RFC](https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2) for reference. Clients can choose to ignore it, and "wait" for a HTTP 401, but that seems a bit wasteful to me. 
Also, wouldn't we need to store additional information, such as a refresh token?

system-tests/e2e-transfer-test/backend-service/src/main/java/org/eclipse/edc/test/e2e/Oauth2TokenController.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public Oauth2TokenController(Monitor monitor) {
+        this.monitor = monitor;
+    }
+
+    @POST
+    @Path("/token")
+    public Map<String, String> getToken(@FormParam("client_id") String clientId, @FormParam("client_secret") String clientSecret) {
+        monitor.info("Oauth2 token requested");
+        if (!"clientId".equals(clientId) || !"clientSecret".equals(clientSecret)) {
```
paullatzelsperger, line 40: nit: `"clientId"` and `"clientSecret"` should be constants

extensions/control-plane/provision/provision-oauth2/docs/diagram.dot
```diff
@@ -0,0 +1,31 @@
+digraph {node [ shape=square ];
```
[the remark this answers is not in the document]
ndr-brt: That's because I'm not familiar with plantuml :smiling_face_with_tear:  I ended up keeping only the png file, that will be just enough for the documentation.

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParamsSupplier.java
ndr-brt: Although we heavily rely on jackson (it actually can't be replaced), having an dedicated wrapper class is useful as it provides a neater interface and makes our services talk about our language, not the library's one.

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/types/domain/HttpDataAddress.java
ndr-brt: I'm not a big fan of static imports when they hide details, in this case `ofNullable` does not explain well what's about to do. I would keep it as it is if that's not a problem

extensions/control-plane/provision/provision-oauth2/src/main/java/org/eclipse/edc/connector/provision/oauth2/Oauth2SecretToken.java
ndr-brt: Unfortunately `expires_in` [is noted as recommended but not required](https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/) so I wouldn't rely on that information, same for `refresh_token`, in that case more specific to the server implementation extension could be implemented.

paullatzelsperger: sure, but that would mean it's an optional value, with a default of `0`, and maybe a `Map<String, String> additionalFields`. The way it is now, we _can never have_ an expiry, or a refresh token.

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/types/domain/HttpDataAddress.java
paullatzelsperger: To my, it is pretty self-explanatory what's going on. Also, with modern IDEs it you'll get a hint anyways. 
But ok, I'm not that hung up on it :)

extensions/control-plane/provision/provision-oauth2/src/main/java/org/eclipse/edc/connector/provision/oauth2/Oauth2SecretToken.java
ndr-brt: As long we aren't using them I can't see any use for extracting them, otherwise we could give the impression of actually supporting these features. I stated that in the README.

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParamsSupplier.java
paullatzelsperger: The `TypeManager` is more than just a wrapper for an `ObjectMapper`, it handles type registration, delegation to multiple serialization contexts, etc.

I would thus argue, that its interface contains methods that have nothing to do with just serialization and deserialization, the Oauth2 module does not need multiple serialization contexts, or care about type registrations.

As you said, we cannot easily replace Jackson anyway, so why not just embrace it and use the `ObjectMapper` directly. It would be a more focused class.

extensions/control-plane/provision/provision-oauth2/docs/diagram.dot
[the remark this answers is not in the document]
paullatzelsperger: It is not difficult to get into, and there's a serviceable plugin for IntelliJ with live updates and everything.

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParamsSupplier.java
ndr-brt: Since this approach is widely used I think we should take a clear decision and say, for example "never pass TypeManager to services, use ObjectMapper instead" and then change this everywhere.

paullatzelsperger: > Since this approach is widely used I think we should take a clear decision and say, for example "never pass TypeManager to services, use ObjectMapper instead" and then change this everywhere.

Absolutely agree, and IIRC there was already talk about this in the past, I think in the context of the SQL stores. Don't recall whether it was a comment on a PR, or a discussion, but that somehow rings a bell. 
How about we come up with some sort of "coding guidelines", which would clearly state that, amongst many others?

extensions/control-plane/provision/provision-oauth2/src/main/java/org/eclipse/edc/connector/provision/oauth2/Oauth2SecretToken.java
paullatzelsperger: ok, since you explicitly mentioned the client credentials flow, it should be OK. It would  mean that a new token has to be requested every time, which seems legit.

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParamsSupplier.java
ndr-brt: I don't know, I always liked the `TypeManager` approach :grin: but it could be.
I'm ok for addressing this to another issue.

paullatzelsperger: Frankly, having coding guidelines is something that's been lurking in the back of my mind for quite a while now, and I have already started to write some, but abandoned it due to a lack of time. Maybe I can dig out that branch once again.

paullatzelsperger (approved):

PR #2139, opened 2022-10-26
feat(api): adds @ApiTest tag for controller integration tests

### LINKED ISSUE #2136: Move the API controller integration test to a separate test suite
# Feature Request

All the `*IntegrationTest` tests should be annotated with a new annotation `@ApiTest`, which is an `@IntegrationTest` and prevents all API i-tests from being executed during a "normal" `./gradlew test` run.

Build/Test, all Integration tests

Saving time: during a profiling session of the build we detected that those integration tests take the lion's share of the test duration, since they spin up a connector runtime. While there is value to them, it is sufficient to execute them on-demand (using `-DincludeTags`) or on CI.

## Solution Proposal
- Add a new annotation `@ApiTest`, that declares a `@Tag("ApiTest")` and put it on every controller integration test.
- Add a separate CI step to execute these integration tests.

## Type of Issue
improvement

## Further Notes
There are other tests, that spin up a connector runtime, that _aren't_ API i-tests, but the majority (with respect to time) is consumed by the API tests, so lets focus on them first, and extend the scope later if needed.

wolf4ood:
## What this PR changes/adds

Adds `@ApiTest` on all controller integration test for executing them on demand via `-DincludeTags="ApiTest"` 
and creates a separated CI job for executing them

## Why it does that

Reduce `./gradlew test` execution time

## Linked Issue(s)

Closes #2136 

## Checklist

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -361,6 +361,20 @@ jobs:
         with:
           command: ./gradlew test jacocoTestReport -DincludeTags="ComponentTest"
 
+  API-Tests:
+    env:
+      JACOCO: true
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-build
+
+      - name: Component Tests
+        uses: ./.github/actions/run-tests
+        with:
+          command: ./gradlew test jacocoTestReport -DincludeTags="ApiTest"
+
+
   OpenTelemetry-Integration-Tests:
     runs-on: ubuntu-latest
     steps:
@@ -475,6 +489,7 @@ jobs:
     needs:
       - Unit-Tests
       - Component-Tests
+      - API-Tests
     runs-on: ubuntu-latest
     if: always()
     steps:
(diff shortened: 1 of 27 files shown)


jimmarino (approved):

core/common/util/src/testFixtures/java/org/eclipse/dataspaceconnector/common/util/junit/annotations/ApiTest.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Common annotation for integration testing.  It applies integration-test Junit Tag.
+ */
+@Target({ ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@IntegrationTest
+@Tag("ApiTest")
+public @interface ApiTest {
```
ndr-brt, line 31: I was wondering... could these test annotation stay in the `junit` module instead of the abused "common-util"? Seems like a better place for them

wolf4ood: It make sense, 
probably it make sense also to move that module in core instead of extensions?

I see a lot of core modules that depends on `testImplementation(project(":extensions:common:junit"))` for extension testing

ndr-brt: That could be good, but in case I'd open an issue for that to gather some opinions

Issue #2143, opened 2022-10-27
Add `canGenerate` method on `Consumer`/`Provider` `ResourceDefinitionGenerator`


ndr-brt:
# Feature Request

Currently it's up to the implementation of the `ResourceDefinitionGenerator` decide if the `ResourceDefinition` needs to be generated.
This could be handled in a different way, like it has been done on the `Provider` interface: adding a method that states if a certain request needs to have a resource generated or not. This will make the API and the implementation breezier.

## Which Areas Would Be Affected?
transfer

## Why Is the Feature Desired?
refactoring

## Solution Proposal
Add a method with this interface to the generator
```java
boolean canGenerate(DataRequest dataRequest, Policy policy);
```

Then on the `ResourceManifestGenerator` it will be used to filter out the generators:
```java
var definitions = consumerGenerators.stream()
                .filter(generator -> generator.canGenerate(dataRequest, policy))
                .map(generator -> generator.generate(dataRequest, policy))
                .collect(Collectors.toList());
```

## Type of Issue
refactoring

jimmarino:
+1 we should be careful to note that an explicit design is to allow multiple resource generates to be engaged as opposed to just one (as is the case with other canXXX() methods).

PR #2150, opened 2022-10-28
refactor: adds canGenerate method on consumer/provider ResoruceDefinitionGenerator

diegogomez-zf:
## What this PR changes/adds

Refactors the method generate() from every generator implemented using the interfaces Consumer- and ProviderResourceDefinitionGenerator. A method "canGenerate()" contains conditions to generate a `ResourceDefinition` and "generate()" will perform the true generation of the `ResourceDefinition`

## Why it does that

Currently it's up to the implementation of the ResourceDefinitionGenerator decide if the ResourceDefinition needs to be generated. This could be handled in a different way, like it has been done on the Provider interface: adding a method that states if a certain request needs to have a resource generated or not. This will make the API and the implementation breezier.

## Linked Issue(s)

#2143 

## Checklist


paullatzelsperger:
@diegogomez-zf to have the PR close the issue automatically, please use ["closing keywords"](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/transfer/spi/provision/ConsumerResourceDefinitionGenerator.java
```diff
@@ -28,9 +28,18 @@
...
      *
      * @param dataRequest the data request associated with transfer process
-     * @param policy the contract agreement usage policy for the asset being transferred
+     * @param policy      the contract agreement usage policy for the asset being transferred
      */
     @Nullable
     ResourceDefinition generate(DataRequest dataRequest, Policy policy);
 
+    /**
+     * Checks if a data request can be generated based on a policy.
```
paullatzelsperger, line 37: ...checks if a `ResourceDefinition` can be generated

spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/transfer/spi/provision/ProviderResourceDefinitionGenerator.java
```diff
@@ -28,11 +28,21 @@
...
+     * @param dataRequest  the data request associated with transfer process
      * @param assetAddress the asset data address
-     * @param policy the contract agreement usage policy for the asset being transferred
+     * @param policy       the contract agreement usage policy for the asset being transferred
      */
     @Nullable
     ResourceDefinition generate(DataRequest dataRequest, DataAddress assetAddress, Policy policy);
 
+    /**
+     * Checks if a data request can be generated based on a policy.
```
paullatzelsperger, line 39: ...checks if a `ResourceDefinition` can be generated...

extensions/control-plane/provision/provision-aws-s3/src/test/java/org/eclipse/edc/connector/provision/aws/s3/S3ConsumerResourceDefinitionGeneratorTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @BeforeEach
+    void setUp() {
+        generator = new S3ConsumerResourceDefinitionGenerator();
+    }
+
+    @Test
+    void generate() {
+
+        DataAddress destination = DataAddress.Builder.newInstance().type(S3BucketSchema.TYPE)
```
ndr-brt, line 43: please use `var`, same for the other `destination` instances

extensions/control-plane/provision/provision-aws-s3/src/test/java/org/eclipse/edc/connector/provision/aws/s3/S3ConsumerResourceDefinitionGeneratorTest.java
```diff
@@ -0,0 +1,90 @@
...
+        var dr = DataRequest.Builder.newInstance().dataDestination(destination).assetId(asset.getId()).processId("process-id").build();
+        var policy = Policy.Builder.newInstance().build();
+
+        var definition = generator.generate(dr, policy);
+
+        assertThat(definition).isInstanceOf(S3BucketResourceDefinition.class);
+        var objectDef = (S3BucketResourceDefinition) definition;
+        assertThat(objectDef.getBucketName()).isEqualTo("test-name");
+        assertThat(objectDef.getRegionId()).isEqualTo(Region.US_EAST_1.id());
+        assertThat(objectDef.getId()).matches(regexPattern);
```
ndr-brt, line 57: to avoid the regex, this can be asserted as:
`assertThat(objectDef.getId()).satisfies(UUID::fromString);`

(same for the other tests)

ndr-brt (changes requested):

paullatzelsperger (commented):
I am a bit confused about this, especially when reading the description of the linked issue which says:
> ... adding a method that states if a certain request needs to have a resource generated or not.

The way I understand it, the semantics of that differ significantly from the current impementation, which _always_ generates the `ResourceDefinition` _if it can_, whereas the issue description calls for determining whether the resource _needs to be generated_. I absolutely realize that I'm being pedantic, but I would advocate adapting the issue description so that it fits to the PR, or changing the PR such that it actually impements what the issue specifies.

paullatzelsperger (changes requested):
as per comments. clicked the "comment" button by accident

extensions/control-plane/provision/provision-blob/src/main/java/org/eclipse/edc/connector/provision/azure/blob/ObjectStorageConsumerResourceDefinitionGenerator.java
```diff
@@ -40,4 +38,9 @@ public class ObjectStorageConsumerResourceDefinitionGenerator implements Consume
         }
         return ObjectStorageResourceDefinition.Builder.newInstance().id(id).accountName(account).containerName(container).build();
     }
+
+    @Override
+    public boolean canGenerate(DataRequest dataRequest, Policy policy) {
+        return AzureBlobStoreSchema.TYPE.equals(dataRequest.getDestinationType());
```
paullatzelsperger, line 44: this is not the same condition as it was before. The fact that `getDestinationType()` is always `null` when `getDataDestination()` is also `null` is an implementation detail and you cannot assume that is always the case.

extensions/control-plane/provision/provision-aws-s3/src/main/java/org/eclipse/edc/connector/provision/aws/s3/S3ConsumerResourceDefinitionGenerator.java
```diff
@@ -30,18 +30,16 @@ public class S3ConsumerResourceDefinitionGenerator implements ConsumerResourceDe
 
     @Override
     public ResourceDefinition generate(DataRequest dataRequest, Policy policy) {
```
jimmarino, line 32: This code should not be removed

paullatzelsperger (changes requested):

jimmarino (changes requested):
This PR inadvertently modifies the existing generator behavior and removes code and comments that need to be preserved. Please carefully revisit your changes and fix those issues. Thx.

extensions/control-plane/provision/provision-gcs/src/test/java/org/eclipse/edc/connector/provision/gcp/GcsConsumerResourceDefinitionGeneratorTest.java
```diff
@@ -0,0 +1,83 @@
...
+    private final Pattern regexPattern = Pattern.compile("([a-f0-9]{8}(-[a-f0-9]{4}){4}[a-f0-9]{8})");
+    private GcsConsumerResourceDefinitionGenerator generator;
+
+    @BeforeEach
+    void setUp() {
+        generator = new GcsConsumerResourceDefinitionGenerator();
+    }
+
+    @Test
+    void generate() {
```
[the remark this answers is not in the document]
ndr-brt: actually they can't be null, so I won't bother testing these cases

extensions/control-plane/provision/provision-aws-s3/src/test/java/org/eclipse/edc/connector/provision/aws/s3/S3ConsumerResourceDefinitionGeneratorTest.java
```diff
@@ -0,0 +1,90 @@
...
+    private final Pattern regexPattern = Pattern.compile("([a-f0-9]{8}(-[a-f0-9]{4}){4}[a-f0-9]{8})");
+    private S3ConsumerResourceDefinitionGenerator generator;
+
+    @BeforeEach
+    void setUp() {
+        generator = new S3ConsumerResourceDefinitionGenerator();
+    }
+
+    @Test
+    void generate() {
```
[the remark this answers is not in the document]
ndr-brt: actually they can't be null, so I won't bother testing these cases

extensions/control-plane/provision/provision-aws-s3/src/main/java/org/eclipse/edc/connector/provision/aws/s3/S3ConsumerResourceDefinitionGenerator.java
ndr-brt: What's the purpose of this? `dataDestination` cannot never be null, and I don't see a good reason to have a default region value (in this case `US_EAST_1`), I think it's something we inherited from the past and it needed to be fixed accordingly.

extensions/control-plane/provision/provision-blob/src/main/java/org/eclipse/edc/connector/provision/azure/blob/ObjectStorageConsumerResourceDefinitionGenerator.java
ndr-brt: `dataDestination` and `destinationType` cannot be null (as per validation in `DataRequest.Builder.build()` method)

ndr-brt (approved):

extensions/control-plane/provision/provision-aws-s3/src/main/java/org/eclipse/edc/connector/provision/aws/s3/S3ConsumerResourceDefinitionGenerator.java
jimmarino: I was referring to the note about the need to integrate the policy engine and the change to existing behavior that may impact users. I would prefer we keep a default region (does it really need to be removed?). However, we decide to change the existing behavior, it should be noted in the PR and also in the release notes.

extensions/control-plane/provision/provision-blob/src/main/java/org/eclipse/edc/connector/provision/azure/blob/ObjectStorageConsumerResourceDefinitionGenerator.java
paullatzelsperger: > dataDestination and destinationType cannot be null (as per validation in DataRequest.Builder.build() method)

what if that ever changes? suddenly you'd need to touch code (i.e. the res-def-gen) that has nothing to do with it. as @jimmarino said, defensive programming is not a bad trait.

extensions/control-plane/provision/provision-gcs/src/test/java/org/eclipse/edc/connector/provision/gcp/GcsConsumerResourceDefinitionGeneratorTest.java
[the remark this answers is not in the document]
paullatzelsperger: At the very least I would expect an `Object.requireNotNull` to assert that, and a test for it. testing only the happy path is dangerous.

extensions/control-plane/provision/provision-aws-s3/src/main/java/org/eclipse/edc/connector/provision/aws/s3/S3ConsumerResourceDefinitionGenerator.java
ndr-brt: The problem is that the current behavior was to set the default region if the `dataDestination` was null, and it always is, so what's the point? Could be good to modify this logic in a more reasonable: "if bucket is not specified on the data destination, then set a default one? wdyt @jimmarino

extensions/control-plane/provision/provision-blob/src/main/java/org/eclipse/edc/connector/provision/azure/blob/ObjectStorageConsumerResourceDefinitionGenerator.java
ndr-brt: To be honest we never have been defensive, and that's good in my opinion, if a null object appears here it's an upstream problem, so having a NPE exception thrown is correct (is the same thing that is done by the `requireNotNull` BTW).
I wouldn't expect this to change, that's why we define interfaces, having a component that expects a null input is only added complexity, even more so here that we're talking about a `DataRequest` and a `Policy`, can't imagine a scenario where we accept them as null.

extensions/control-plane/provision/provision-gcs/src/test/java/org/eclipse/edc/connector/provision/gcp/GcsConsumerResourceDefinitionGeneratorTest.java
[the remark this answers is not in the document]
ndr-brt: We don't do that in any other component I think, as I wrote in the last comment, it would be useless defensive stuff in a component that would never ever be called with null arguments.

extensions/control-plane/provision/provision-aws-s3/src/main/java/org/eclipse/edc/connector/provision/aws/s3/S3ConsumerResourceDefinitionGenerator.java
jimmarino: @ndr-brt yes, that makes more sense. IMO it should be done in a follow-on PR. Also, please keep the comments about policy.

extensions/control-plane/provision/provision-blob/src/main/java/org/eclipse/edc/connector/provision/azure/blob/ObjectStorageConsumerResourceDefinitionGenerator.java
jimmarino: Agreed, that's why I suggested `Objects.requireNotNull` and not to handle the null case - it doesn't pollute the code and provides a better indication of where the issue is.

ndr-brt: @jimmarino the point here is that having a null `DataRequest` or a null `Policy` would make the calling this call useless, we should validate them at a manager level and, if some of those are null we should fail the transfer process, but taking care of nulls in any component is definitely a bad practice

jimmarino: @ndr-brt That definitely needs to be solved, but in a separate PR.

extensions/control-plane/provision/provision-aws-s3/src/test/java/org/eclipse/edc/connector/provision/aws/s3/S3ConsumerResourceDefinitionGeneratorTest.java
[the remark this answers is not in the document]
diegogomez-zf: The Generators are already covered with the `Object.requireNotNull` so that no method can receive null parameters

extensions/control-plane/provision/provision-gcs/src/test/java/org/eclipse/edc/connector/provision/gcp/GcsConsumerResourceDefinitionGeneratorTest.java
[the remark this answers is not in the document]
diegogomez-zf: The Generators are already covered with the `Object.requireNotNull` so that no method can receive null parameters

extensions/control-plane/provision/provision-aws-s3/src/main/java/org/eclipse/edc/connector/provision/aws/s3/S3ConsumerResourceDefinitionGenerator.java
diegogomez-zf: @jimmarino the previous condition related to the S3ConsumerResourceDefinitionGenerator is back again but since I think the discussion is related to the case when someone does not specify a region, I opted to refine the condition with `getDataDestination().getProperty(S3BucketSchema.REGION)==null` instead of `if(getDataDestination()=null)` and I made a unit test that evaluates the case when a person does not add bucket name as property (returning NullPointerExcp). 

I think it's obvious and already spoken but I will write it here. The current validators inside `DataRequest` do not allow to create an object with `dataDestination` null. I can return to the original condition but that would mean, if I keep the  `getDataDestination()=null` it would not be possible to test that case

extensions/control-plane/provision/provision-blob/src/main/java/org/eclipse/edc/connector/provision/azure/blob/ObjectStorageConsumerResourceDefinitionGenerator.java
diegogomez-zf: The Generators already contain the Object.requireNotNull condition so that no method can receive null parameters

extensions/control-plane/provision/provision-aws-s3/src/test/java/org/eclipse/edc/connector/provision/aws/s3/S3ConsumerResourceDefinitionGeneratorTest.java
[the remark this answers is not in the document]
paullatzelsperger: @diegogomez-zf we're writing tests to ensure that and to guard against regression.

paullatzelsperger (approved):

PR #2153, opened 2022-11-01
docs: correct port of the file transfer process for the data consumer (9192)

Julian-Marco:
## What this PR changes/adds

Correct port of the file transfer process for the data consumer (9192)

## Why it does that

The change was necessary because the documentation contained the wrong port (9191)

### DIFF
--- a/samples/05-file-transfer-cloud/README.md
+++ b/samples/05-file-transfer-cloud/README.md
@@ -139,7 +139,7 @@ the response will also contain an agreement id, that is required in the next ste
 To initiate the data transfer, execute the statement below. Please take care of setting the contract agreement id obtained at previous step.
 
 ```bash
-curl --location --request POST 'http://localhost:9191/api/v1/data/transferprocess' \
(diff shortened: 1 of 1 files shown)


ndr-brt (approved):

ndr-brt:
Hey @Julian-Marco thanks for the fix, you need to change the PR name according to the convention and add at least one label, if you don't have enough permissions to do that we can do that

Julian-Marco:
Hey @ndr-brt! I changed the title (hopefully in a correct kind of way), but I don't have the permission to give the PR a label, so feel free to do it for me.

ndr-brt:
@Julian-Marco please take a look at the [pr etiquette](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/pr_etiquette.md) to get the pr title rules.
Label added

paullatzelsperger:
I realize that this is a really minor fix, but in the future please stick to the established process as per PR etiquette: a bug issue should be created first, and the implementation PR should follow.

paullatzelsperger (approved):

PR #2169, opened 2022-11-07
doc: amend coding guidelines and rename file

### LINKED ISSUE #2160: Provide clear coding guidelines
# Feature Request

We should provide a document in which we clearly outline the most important Java coding guidelines that contributors must adhere to.
By that I mean things that cannot be enforced through a checkstyle check or an IntelliJ style. As a starting point for a discussion I would propose

- General Code Style
  - use `var` instead of explicit types (helps clarity)
  - avoid `final` in method args and local variables
  - use `final` in field declarations
  - avoid `static` fields except in constants or when absolutely necessary. (you should be able to provide a reason).
  - use `interfaces` as containers for constants
  - use "minimally required types" (or "smallest possible API"), e.g. use `ObjectMapper` instead of `TypeManager`, `String` instead of an object containing the String, etc.
  - use either `public` members, which are documented and tested, or `private` members.
  - avoid package-private members if only needed for testing
  - avoid `protected` members unless they're intended to be overridden.
  - use package-private classes if they're not needed outside the package, e.g. implementation classes
  - avoid using `enum`s for anything other than named integer enumerations.
  - avoid using static classes as much as possible. Exceptions to this are helper functions and test utils, etc.
  - avoid any kind of variable prefixes, no underscores, no `m_` for members, etc.
  - avoid unnecessary `this.` (exception: in builders)
  - use static imports, as long as it is still clear what's being done. For example, use `of(item1, item2).map(it -> it.someOperation)...` instead of `Stream.of(item1, item2)`, or `format("...",arg1)` instead of `String.format("...", arg1)`. Helps with clarity and readability.
  - avoid `Optional` as method return type or method argument. 
  - avoid cryptic variable names, especially in long methods. Instead, try to write them out, at least to a reasonable extent.
  
- Testing
  - avoid testing multiple things in a single test methods, unless they're really simple, like SerDes.
  - use `[METHOD]_when[CONDITION]_should[EXPECTATION]` naming scheme for test methods, e.g. `verifyInput_whenNull_shouldThrowNpe()`
  - use test functions classes to provide common helper methods, e.g. to instantiate an object
  

- Builders:
   - use the Builder pattern when:
     1. there are any number of optional constructor args
     2. there are more than 3 constructor args
     3. the object should be serializable
     4. inheriting from an object that fulfills any of the above. In this case use derived builders as well.
   - Builders are to be named just `Builder` and should be static nested classes.
   - create a `public static Builder newInstance(){ ... }` method, which is used to instantiate the builder.
   - use single-field builders: a `Builder` instantiates the object it builds in its constructor, and sets the properties in its builder methods. The `build()` method then only performs verification [optional] and returns the instance.
   - use `private` constructors for the objects that the Builder builds
   - use Jackson annotations on Builders only

Documentation

Having a clearly defined coding guideline will help us in keeping a homogenous and consistent code base. 

## Solution Proposal
Create a documentation item, that is referenced from the contribution guidelines.

## Type of Issue
documentation

paullatzelsperger:
## What this PR changes/adds

Amends the existing `architecture-principles` with some more detailed coding guidelines, renames the file to `coding-guidelines.md`.

## Why it does that

To provide the basis for a homogenous and consistent coding style across the code base. 

## Further notes
.

## Linked Issue(s)

Closes #2160 

## Checklist

- [ ] formatted title correctly? (_take a look at the [CONTRIBUTING](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/CONTRIBUTING.md#submit-a-pull-request) and [Etiquette for pull requests](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/pr_etiquette.md) for details_)


docs/developer/architecture/coding-principles.md
```diff
@@ -27,26 +27,64 @@
...
-### IV. Testing
+### IV. General Coding Style
+
+1. use `var` instead of explicit types (helps with clarity)
+2. avoid `final` in method args and local variables
+3. use `final` in field declarations
+4. avoid `static` fields except in constants or when absolutely necessary. (you should be able to provide a reason).
+5. use `interfaces` as containers for constants
+6. use "minimally required types" (or "smallest possible API"), e.g. use `ObjectMapper` instead of `TypeManager`
+   , or use a `String` instead of a more complex object containing the String, etc.
```
jimmarino, line 38: These are probably two separate issues. The string issue is "Don't create wrapper types for Strings and other types that may be passed and referenced explicitly"

docs/developer/architecture/coding-principles.md
```diff
@@ -27,26 +27,64 @@
...
+7. use either `public` members, which are documented and tested, or `private` members.
+8. avoid package-private members, especially if only needed for testing
+9. avoid `protected` members unless they're intended to be overridden.
+10. use package-private classes if they're not needed outside the package, e.g. implementation classes
+11. avoid using `enum`s for anything other than named integer enumerations.
+12. avoid using static classes as much as possible. Exceptions to this are helper functions and test utils, etc.
```
jimmarino, line 44: Static inner classes are also good

docs/developer/architecture/coding-principles.md
```diff
@@ -27,26 +27,64 @@
...
+13. avoid any kind of variable prefixes, no underscores, no `m_` for members, etc.
```
jimmarino, line 45: Should be "Use only camel case and no prefixes for naming"

docs/developer/architecture/coding-principles.md
```diff
@@ -27,26 +27,64 @@
...
+14. avoid unnecessary `this.` (exception: in builders)
+15. use static imports, as long as it is still clear what's being done. For example,
+    use `of(item1, item2).map(it -> it.someOperation)...` instead of `Stream.of(item1, item2)`, or `format("...",arg1)`
+    instead of `String.format("...", arg1)`. Helps with clarity and readability.
+16. avoid `Optional` as method return type or method argument. Use `null` in signatures.
```
jimmarino, line 50: "except when designing a fluent API"

docs/developer/architecture/coding-principles.md
```diff
@@ -27,26 +27,64 @@
...
+17. avoid cryptic variable names, especially in long methods. Instead, try to write them out, at least to a reasonable
```
jimmarino, line 51: Do we need to say this, it's kind of like "don't write bad code"? This document should highlight our conventions, not how to write good code. Conventions come up when there is more than one coding idiom and we require one approach.

docs/developer/architecture/coding-principles.md
```diff
@@ -27,26 +27,64 @@
...
+4. Use classes with static test functions to provide common helper methods, e.g. to instantiate an object.
+5. Use `[METHOD]_when[CONDITION]_should[EXPECTATION]` as naming template for test methods,
+   e.g. `verifyInput_whenNull_throwsNpe()` as opposed to `testInputNull()`
+
+### VI. Data Objects
+
+1. Use the `Builder` pattern when:
+    - there are any number of optional constructor args
+    - there are more than 3 constructor args
+    - the object should be serializable
```
jimmarino, line 69: I would remove the serializable statement because it is not the reason builders are used. The builder pattern is for immutability and handling the issue of parameters.

docs/developer/architecture/coding-principles.md
```diff
@@ -27,26 +27,64 @@
...
+    - inheriting from an object that fulfills any of the above. In this case use derived builders as well.
+2. Builders should be named just `Builder` and be static nested classes.
+3. Create a `public static Builder newInstance(){...}` method to instantiate the builder
+4. Builders have non-public constructors
+5. use single-field builders: a `Builder` instantiates the object it builds in its constructor, and sets the properties
+   in its builder methods. The `build()` method then only performs verification (optional) and returns the instance.
+6. use `private` constructors for the objects that the builder builds.
+7. use Jackson annotations on builders only, as opposed to: the object.
```
jimmarino, line 77: Jackson annotations are still required on the object, e.g. type info

jimmarino (changes requested):

docs/developer/architecture/coding-principles.md
paullatzelsperger: I meant annotations to configure ser/des. Theoretically users could put `@JsonProperty` annotations etc. in several places, but we want them to be on the builders. I'll clarify.

docs/developer/architecture/coding-principles.md
paullatzelsperger: developers sometimes have a tendency to abbreviate var names, e.g. `gat` instead of `generatedAccessToken`, I know I've been guilty of this, and with refactoring capabilities of modern IDEs it can happen all the more easily. That's why I put it in.

docs/developer/architecture/coding-principles.md
```diff
@@ -26,27 +26,67 @@
...
-
-1. Do not create constructors that take many parameters; instead use the builder pattern. An example
-   is `TransferProcess`.
-2. Note that the motivation behind use of builders is not for immutability (although that may be good in certain
+2. Prefer unit tests over all other test types: unit > integration/component > e2e
+3. When appropriate, prefer composing services via the constructor so that dependencies can be mocked as opposed to
+   instantiating dependencies directly.
+4. Use classes with static test functions to provide common helper methods, e.g. to instantiate an object.
+5. Use `[METHOD]_when[CONDITION]_should[EXPECTATION]` as naming template for test methods,
+   e.g. `verifyInput_whenNull_throwsNpe()` as opposed to `testInputNull()`
```
ronjaquensel, line 63: Shouldn't this be `verifyInput_whenNull_shouldThrowNpe()`to match the pattern?

ronjaquensel (approved):

docs/developer/architecture/coding-principles.md
paullatzelsperger: ha, you're right, thanks!

docs/developer/architecture/coding-principles.md
```diff
@@ -26,27 +26,67 @@
 ## III. Simplicity
 
 1. Avoid layers of indirection when they are not needed (e.g. "pass-through methods").
-
-### IV. Testing
+2. Avoid needlessly wrapping objects, especially primitive datatypes.
+
+### IV. General Coding Style
```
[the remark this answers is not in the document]
paullatzelsperger: it has been that way before, don't know if on purpose or by accident. @jimmarino?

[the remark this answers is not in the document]
ndr-brt: seems like by accident I'd say, I would align them to a common level

docs/developer/architecture/coding-principles.md
```diff
@@ -26,27 +26,67 @@
...
+6. Use "minimally required types" (or "smallest possible API"), e.g. use `ObjectMapper` instead of `TypeManager`
+   , or use a `String` instead of a more complex object containing the String, etc.
+7. Use either `public` members, which are documented and tested, or `private` members.
+8. Avoid package-private members, especially if only needed for testing
+9. Avoid `protected` members unless they're intended to be overridden.
+10. Use package-private classes if they're not needed outside the package, e.g. implementation classes
+11. Avoid using `enum`s for anything other than named integer enumerations.
+12. Avoid using static classes as much as possible. Exceptions to this are helper functions and test utils, etc. as well as static inner classes.
+13. Use only camel case and no prefixes for naming.
+14. Avoid unnecessary `this.` (exception: in builders)
```
ndr-brt, line 47: I would avoid them in builders as well, I mean, as long as there isn't a name overlapping (in that case the `this` will be necessary) I would avoid it.

docs/developer/architecture/coding-principles.md
```diff
@@ -26,27 +26,67 @@
...
+15. Use static imports, as long as it is still clear what's being done. For example,
```
ndr-brt, line 48: I commented this a lot of time, I'm absolutely in favor of static imports when the method name is self-explainatory (es. `Assertions.asserThat` -> `assertThat`), but not for other cases like `Stream.of` or `Optional.of`, they wouldn't be explainatory, moreover, if two static methods with the same name (as the `of` methods I just listed) are used in the same class they can't be both imported statically. So I would leave this decision driven by readability and code comprehension.

docs/developer/architecture/coding-principles.md
paullatzelsperger: that was what I meant. I'll make it explicit though

docs/developer/architecture/coding-principles.md
paullatzelsperger: >  if two static methods with the same name

that should be implicit though - it just won't compile anyway :)

> So I would leave this decision driven by readability and code comprehension

that is what I meant by "as long as it still clear...", but I can use your phrasing to be more precise.

ndr-brt (approved):

jimmarino (approved):

PR #2173, opened 2022-11-08
fix: remove usages of Apache Commons Lang

### LINKED ISSUE #1747: Chore: find and remove all occurrences and usages of Apache Commons Lang
Go through the code and remove all occurrences and usages of Apache commons lang. It is a transitive dependency that was introduced by the Swagger Jakarta Integration. Relying on transitive dependencies is bad.

_Originally posted by @paullatzelsperger in https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/pull/1744#discussion_r930800492_

iwasakims:
## What this PR changes/adds

removing usage of org.apache.commons:commons-lang3:3.12.0 pulled as transitive dependency of org.mock-server:mockserver-netty:5.14.0.

## Why it does that

Relying on transitive dependency should be avoided.

## Linked Issue(s)

Closes #1747

## Checklist

- [n/a] added appropriate tests?
- [n/a] added/updated copyright headers?
- [n/a] documented public classes/methods?
- [n/a] added/updated relevant documentation?

### DIFF
--- a/system-tests/tests/src/test/java/org/eclipse/edc/test/system/remote/FileTransferAsClientSimulation.java
+++ b/system-tests/tests/src/test/java/org/eclipse/edc/test/system/remote/FileTransferAsClientSimulation.java
@@ -15,10 +15,7 @@
 package org.eclipse.edc.test.system.remote;
 
 import io.gatling.javaapi.core.Simulation;
-import org.apache.commons.lang3.StringUtils;
 import org.eclipse.edc.test.system.FileTransferSimulationConfiguration;
-
-import java.util.Objects;
 
 import static io.gatling.javaapi.core.CoreDsl.atOnceUsers;
 import static io.gatling.javaapi.core.CoreDsl.global;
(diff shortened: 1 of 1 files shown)


system-tests/tests/src/test/java/org/eclipse/edc/test/system/remote/FileTransferAsClientSimulation.java
```diff
@@ -51,6 +48,11 @@ public FileTransferAsClientSimulation() {
     }
 
     private static String getFromEnv(String env) {
-        return Objects.requireNonNull(StringUtils.trimToNull(System.getenv(env)), env);
```
ndr-brt, line 54: My question here is: did we need this `trimToNull` at all? I'd say that `System.getenv` with a `isBlank` check would already do the job. The variables are set in the `minikube-test.sh` and I can't see a case where they have leading-trailing whitespaces to trim or be an empty string.
I would change the implementation of this method to 
`Objects.requireNonNull(System.getenv(env), env)`

ndr-brt (changes requested):

system-tests/tests/src/test/java/org/eclipse/edc/test/system/remote/FileTransferAsClientSimulation.java
iwasakims: Though bailing out fast on empty/invalid value may help to find the cause on future bug, I think it should be enough to check the null value here as you suggest. I will update the code.

ndr-brt (approved):

Issue #2194, opened 2022-11-11
API: move all the `management` api to the `management` context


ndr-brt:
# Feature Request

Currently the API that are used by the planes to communicate together are not defined in the same context, as they should be.
The API that are involved are:

- `data-management-api`
- `data-plane-selector`
- `federated-catalog`
- `http-provisioner`
- `observability-api`
- ...

## Which Areas Would Be Affected?
api

## Why Is the Feature Desired?
improve api structure

## Solution Proposal
-

## Type of Issue
improvement

PR #2195, opened 2022-11-11
feat: rename `data-management-api` to `management-api`

### LINKED ISSUE #2177: API: rename `data-management-api` to `management-api`
# Feature Request

The `data` term seems to be overused in the EDC domain and it leads to misunderstandings.
the `data-management-api` module could be renamed to `management-api`, same for the web context `data` that could become `management`.

api

better naming

## Solution Proposal
-

## Type of Issue
refactoring

ndr-brt:
## What this PR changes/adds

Renames all the occurrencies of "data management api" to "management api".
Provides a way to handle config changes through deprecation.

## Why it does that

To avoid confusion, as `data` is an overused term in our domain.

## Further notes

- in the `ManagementApiConfigurationExtension` I provided a way to transit from the current `web.http.data` to `web.http.management`, this is not done because we want to support breaking changes, but it was done to help developer to switch to the new convention. I marked the deprecation as "from milestone8", so it can be removed after milestone 8 release.
- I decided to split the previous issue in 2 parts (the second one is #2194) to provide a more step-by-step change.

## Linked Issue(s)

Closes #2177

## Checklist


docs/developer/decision-records/2022-10-31-aggregate-service-layer/README.md
```diff
@@ -14,7 +14,7 @@ For `AggregateService` we intend a component that provides a clear and simple in
 `Asset`s, `ContractNegotiation`s and so on) encapsulating some logic and features, like business validation, 
 transactional context, exception handling, ...
 
-Currently they are located into the `data-management-api`. Moving them to a separate layer will permit other kinds of
+Currently they are located into the `management-api`. Moving them to a separate layer will permit other kinds of
```
juliapampus, line 17: IMO, decision records should be excluded from refactorings as they represent a snapshot of a specific date in time.

ndr-brt: ops, that was an overshight

extensions/control-plane/api/management-api/catalog-api/src/main/java/org/eclipse/edc/connector/api/datamanagement/catalog/CatalogApiExtension.java
```diff
@@ -15,7 +15,7 @@
 package org.eclipse.edc.connector.api.datamanagement.catalog;
```
ronjaquensel, line 15: The package should be renamed as well. As far as I see it's `management` in the api configuration module, but none of the others.

ndr-brt: good catch!

extensions/control-plane/api/management-api/management-api-configuration/README.md
```diff
@@ -10,20 +10,20 @@ Further, this module registers an `AuthenticationRequestFilter` and an `EdcApiEx
...
 ```properties
-web.http.data.port=9191
-web.http.data.path=/api/v1/data
+web.http.management.port=9191
+web.http.management.path=/api/v1/management
 web.http.port=8181
 web.http.path=/api
 ```
 
 Please note that since a specific context is registered (here: `data`), then the default one **must** be registered
```
ronjaquensel, line 19:
```suggestion
Please note that since a specific context is registered (here: `management`), then the default one **must** be registered
```

extensions/control-plane/api/management-api/management-api-configuration/README.md
```diff
@@ -10,20 +10,20 @@ Further, this module registers an `AuthenticationRequestFilter` and an `EdcApiEx
...
 explicitly!
 
-_Caution: failing to provide a `web.http.data.port` and `web.http.data.path` configuration will cause
+_Caution: failing to provide a `web.http.management.port` and `web.http.management.path` configuration will cause
 the `AuthenticationRequestFilter` and `EdcApiExceptionMapper` to be registered in the _default_ context causing it to
 fire for EVERY request on that context including IDS communication._
```
ronjaquensel, line 24: The last part here about IDS could actually be removed, as the IDS API configuration always creates a separate context for the IDS API.

extensions/control-plane/api/management-api/management-api-configuration/src/main/java/org/eclipse/edc/connector/api/management/configuration/ManagementApiConfigurationExtension.java
```diff
@@ -0,0 +1,109 @@
...
+@Extension(value = ManagementApiConfigurationExtension.NAME)
+public class ManagementApiConfigurationExtension implements ServiceExtension {
+
+    public static final String NAME = "Management API configuration";
+
+    private static final String DEPRECATED_MANAGEMENT_SETTINGS_GROUP_SUFFIX = "data";
+    private static final String DATA_MANAGEMENT_CONTEXT_ALIAS = "management";
+    private static final int DEFAULT_DATA_MANAGEMENT_API_PORT = 8181;
+    private static final String DEPRECATED_DEFAULT_DATA_MANAGEMENT_API_CONTEXT_PATH = "/api";
+    private static final String DEFAULT_DATA_MANAGEMENT_API_CONTEXT_PATH = "/api/v1/management";
```
ronjaquensel, line 45: the `DATA` prefix could also be removed from the constants

ronjaquensel (approved):

extensions/control-plane/api/management-api/management-api-configuration/src/main/java/org/eclipse/edc/connector/api/management/configuration/ManagementApiConfigurationExtension.java
ndr-brt: it's actually used to handle the deprecation: `web.http.data` will be still usable for some time printing a warning, so the users will have the time to switch to the new settings group.

juliapampus (approved):

paullatzelsperger (approved):

PR #2222, opened 2022-11-21
fix: Map ContractAgreement end date through ContractOffer

### LINKED ISSUE #722: Make ContractAgreement duration configurable
At the time of writing there is still some placeholder code for the duration of contract agreements. I assume the default duration of contracts should be configurable in the `ContractDefinition`.

bscholtes1A:
## What this PR changes/adds

Add possibility to configure contract agreement end date based on a new field provided in `ContractDefinition`. If not specified, then contract duration will be automatically set to one year, as it was before this PR.

## Why it does that

Contract duration must be configurable.

## Linked Issue(s)

Closes #722 

## Checklist

- [ ] formatted title correctly? (_take a look at the [CONTRIBUTING](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/CONTRIBUTING.md#submit-a-pull-request) and [Etiquette for pull requests](https://github.com/eclipse-dataspaceconnector/DataSpaceConnector/blob/main/pr_etiquette.md) for details_)


extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/model/ContractDefinitionRequestDto.java
```diff
@@ -39,6 +40,9 @@ public class ContractDefinitionRequestDto {
     @NotNull(message = "criteria cannot be null")
     private List<CriterionDto> criteria = new ArrayList<>();
 
+    @Positive(message = "contractValidityDuration must be positive")
+    private long contractValidityDuration;
```
ndr-brt, line 44: I was wondering if this could be represented in a more structured way, because there could be different levels of duration, someone could need a contract valid for some minutes (so like 600 seconds), but it could be also 1 or more years (1 year is 31536000 seconds), it will make the api not so user friendly, so my proposal could be to introduce a structured type, at least on the `Dto` level, like:
```java
public class DurationDto {
  public int amount;
  public String unit;
}
```

and unit could accept values as "minutes", "hours", "days", "months", "years".
Then we could decide if in the `ContractDefinition` class we want to translate it into seconds or into a java `Duration` object, maybe for serialization the first one would be simpler.

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/types/offer/ContractOffer.java
```diff
@@ -192,12 +192,11 @@ public Builder policy(Policy policy) {
             return this;
         }
 
-
         public ContractOffer build() {
             Objects.requireNonNull(contractOffer.id);
             Objects.requireNonNull(contractOffer.asset, "Asset must not be null");
             Objects.requireNonNull(contractOffer.policy, "Policy must not be null");
-
+            Objects.requireNonNull(contractOffer.contractEnd, "ContractEnd must not be null");
```
ndr-brt, line 199: I have a question... is it correct to have the `contractEnd` on the `ContractOffer`? 
I think this is something that we inherited from the ids infomodel, but on the offer there should be reported only the duration, not the end date, because the offer could be negotiated some time later. 
I'd say the the end date should be calculated when the `ContractAgreement` is created. 
But maybe this is a discussion that needs to be brought to an higher level, I'd ask also to @juliapampus and @jimmarino in relation with the new protocol.

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/types/offer/ContractDefinition.java
```diff
@@ -36,9 +38,16 @@
...
+
+    /**
+     * If not specified, each {@link ContractOffer} defines a 1-year contract.
+     */
+    private static final long DEFAULT_CONTRACT_VALIDITY_DURATION = TimeUnit.DAYS.toSeconds(365);
+
     private String accessPolicyId;
     private String contractPolicyId;
     private AssetSelectorExpression selectorExpression;
+    private long contractValidityDuration = DEFAULT_CONTRACT_VALIDITY_DURATION;
```
ndr-brt, line 50: `contract` is implicit, and `validity` and `duration` are pretty much equivalent in this context, I would call this field just `validity`, or `validitySeconds` to make the unit explicit.

extensions/control-plane/api/management-api/contract-negotiation-api/src/main/java/org/eclipse/edc/connector/api/management/contractnegotiation/transform/NegotiationInitiateRequestDtoToDataRequestTransformer.java
```diff
@@ -46,6 +47,7 @@ public Class<ContractOfferRequest> getOutputType() {
                 .consumer(URI.create("urn:connector:consumer"))
                 .provider(URI.create("urn:connector:provider"))
                 .policy(object.getOffer().getPolicy())
+                .contractEnd(ZonedDateTime.now()) // dummy end date, actual contract end will be set by the provider
```
ndr-brt, line 50: I think this field should be reported on the `ContractOfferDescription` and reported here, then needs to be validated by the provider in the `ContractValidationServiceImpl.validateInitialOffer`, like it has been done with the policy.

bscholtes1A: First part done. 
For the check in `ContractValidationService`, it would be preferable to clarify first if ContractOffer should convey a duration instead of the actual end date. As it is it would force to compare a end date with a duration in the contract validation...

extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/model/ContractDefinitionRequestDto.java
bscholtes1A: Done

extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/model/ContractDefinitionResponseDto.java
```diff
@@ -30,6 +30,7 @@ public class ContractDefinitionResponseDto extends BaseResponseDto {
     private String accessPolicyId;
     private String contractPolicyId;
     private List<CriterionDto> criteria = new ArrayList<>();
+    private long validity;
```
ndr-brt, line 33: It should be of type `DurationDto`, conversion is not really immediate but it shouldn't be that hard to implement

bscholtes1A: Done as per last commit

extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/model/ContractDefinitionRequestDto.java
```diff
@@ -38,6 +48,8 @@ public class ContractDefinitionRequestDto {
     @Valid
     @NotNull(message = "criteria cannot be null")
     private List<CriterionDto> criteria = new ArrayList<>();
+    @NotNull(message = "validity cannot be null")
+    private DurationDto validity = DEFAULT_VALIDITY;
```
paullatzelsperger, line 52: I wonder why need the `DurationDto` at all. What value does it add that a `long` would not also give?

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/types/offer/ContractOffer.java
paullatzelsperger, line 199: A contract with a `contractEnd`, but no `contractStart` seems ... odd. Both should be enforced IMO.

core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/validation/ContractValidationServiceImpl.java
```diff
@@ -99,7 +104,10 @@ public Result<ContractOffer> validateInitialOffer(ClaimToken token, ContractOffe
         var validatedOffer = ContractOffer.Builder.newInstance()
                 .id(offer.getId())
                 .asset(targetAsset)
+                .consumer(offer.getConsumer())
+                .provider(offer.getProvider())
                 .policy(contractPolicyDef.getPolicy())
+                .contractEnd(ZonedDateTime.ofInstant(clock.instant().plusSeconds(contractValidityDuration), clock.getZone()))
```
paullatzelsperger, line 110: `contractStart` should be set here too.

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/types/offer/ContractOffer.java
```diff
@@ -99,7 +99,7 @@ public ZonedDateTime getContractStart() {
         return contractStart;
     }
 
-    @Nullable
+    @NotNull
     public ZonedDateTime getContractEnd() {
         return contractEnd;
```
paullatzelsperger, line 104: AFAIK @ndr_brt has already commented it, but at this point in time, I would strongly suggest to remove the `ZonedDateTime` in favor of UTC

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/types/offer/ContractOffer.java
jimmarino: We have been looking at fixing this in the IDS spec. My view is that both the end and duration should be specified as policy constraints, and only a timestamp should exist on the agreement.

bscholtes1A: As discussed, we keep the current fix minimal in terms of changes so that we can migrate easily to the policy-approach when new IDS protocol is ready ;)

core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -362,7 +356,7 @@ private boolean processConsumerApproving(ContractNegotiation negotiation) {
         var agreement = ContractAgreement.Builder.newInstance()
                 .id(ContractId.createContractId(definitionId))
                 .contractStartDate(clock.instant().getEpochSecond())
-                .contractEndDate(clock.instant().plus(365, ChronoUnit.DAYS).getEpochSecond()) // TODO Make configurable (issue #722)
+                .contractEndDate(lastOffer.getContractEnd().toEpochSecond())
```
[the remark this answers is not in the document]
bscholtes1A: For now ZonedDateTime in `ContractOffer` is mainly used for the mapping to XMLGreogrianCalendar (IDS stuff). 
We will very likely get rid of that with the new IDS protocol.

core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/validation/ContractValidationServiceImpl.java
bscholtes1A: done

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/types/offer/ContractOffer.java
bscholtes1A: As mentioned above, `ZonedDateTime` is only here for IDS-mapping purposes. Will get rid of that when new protocol is ready.

ndr-brt (approved):

PR #2243, opened 2022-11-25
feat: Add basic HttpDataAddress validation

### LINKED ISSUE #2235: Add validation on `HttpDataAddress`
# Feature Request

Currently, in the `HttpDataAddress` there's no validation in place, that means if, for example, we create an asset with a data address with a not-valid url or with invalid characters, that won't give any error until the first data transfer on it.
Would it be nice to have such a validation, both on asset and transfer process creation.

api

fail fast

## Solution Proposal
the validation could be put in place into the transformers (`DataAddressDtoToDataAddressTransformer`, `TransferRequestDtoToDataRequestTransformer`) to give a clear 400 error to the client, but also on the `DataAddress`/`HttpDataAddress` builder to avoid such errors coming from other kind of Api

ndr-brt:
## What this PR changes/adds

Introduce a new `DataAddressValidator` component, that takes case of validate any `DataAddress` that is going to be ingested by the system, currently in two different points: on `Asset` creation and on `TransferProcess` initialization.

Currently it validates the `baseUrl` field of `HttpData` data address, that must be a valid 

## Why it does that

To fail fast in case of not well formatted data addresses.

## Further notes

- this is a simple version, but it could be made extensible if needed to give the opportunity to add validation for other types of `DataAddress` (e.g. s3, gcs, ...)

## Linked Issue(s)

Closes #2235

## Checklist


bscholtes1A (approved):

paullatzelsperger (approved):
Looks good to me. At some point, likely in another PR, we could generalize these validators into something like a `ValidatorRegistry`, into which we can register validators, and that allows object validation by type. That would avoid having multiple validator objects in a service. We could even use `javax.validation` classes under the hood.

PR #2276, opened 2022-12-01
feat(httpProxy): add module for http dynamic receivers

### LINKED ISSUE #2265: Add support for dynamic HTTP Receiver callback endpoints
# Feature Request

We have the use-case to have multiple client applications on consumer side which will fetch data through the EDC via the HttpProxy method. These are separate application with different use-case based on different data etc. Therefor, they mostly don't know anything about each other.

As far as I know (please correct me if I'm wrong!) we just have the possibility to configure only one specific URL on the EDC (config property `edc.receiver.http.endpoint`) where the information about the DataPlane access (`EndpointDataReference`) will be sent to. This is fine if we have just one client application which is using the EDC.

With having multiple client applications which are independent, we would need the possibility to somehow define the URL where the `EndpointDataReference` (EDR) object will be sent to.

For example:
- Application A wants to get the access to the data, send the `EDR` object to: `https://app-a:8080/callback`
- Application B wants to get the access to the data, send the `EDR` object to: `https://app-b:8080/callback`
- ...

DPF

Already described above

## Solution Proposal
One idea could be, that every time a `TransferProcess` will be created by the client, we configure in the `TransferProcessDto` object the URL where the EDC should send the `EDR` to.

But I'm not so deep into this topic, maybe there are better ways to make this dynamically configurable by the client application.

wolf4ood:
## What this PR changes/adds

Add a new module `http-dynamic-receiver` for sending the EDR to a custom url/callback defined in transfer creation phase by the client.

## Why it does that

Flexibility for client applications.

## Further notes

For supporting dynamic callback of EDR, a new field `properties` has been added to `TransferProcess` in order to store the callback for the EDR in the custom property `receiver.http.endpoint`.  

Once the consumer needs to forward the EDR to the external API, it will look for the correlated transfer process and it will fetch the `receiver.http.endpoint` property if present.

For security reason the Auth headers are still configurable via EDC properties and not stored in the EDC at runtime.

The only limitation for now is that all the client application needs to share the same header authorization when 
working with the EDR APIs.

## Linked Issue(s)

Closes #2265

## Checklist


bscholtes1A:
Shouldn't this new extension be rather into `extensions:control-plane:transfer` ?

wolf4ood:
@bscholtes1A yes it's a WIP, i will move it into the new folder

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/src/main/java/org/eclipse/edc/connector/receiver/http/dynamic/HttpDynamicEndpointDataReferenceReceiver.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @Override
+    public CompletableFuture<Result<Void>> send(@NotNull EndpointDataReference edr) {
+
+        var processId = transferProcessStore.processIdForDataRequestId(edr.getId());
+
+        if (processId == null) {
+            throw new EdcException(format("Failed to found processId for DataRequestId %s", edr.getId()));
+        }
```
paullatzelsperger, line 69:
```suggestion
         throw new EdcException(format("Could not find transfer process id for DataRequestId %s", edr.getId()));
```

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/src/main/java/org/eclipse/edc/connector/receiver/http/dynamic/HttpDynamicEndpointDataReferenceReceiver.java
```diff
@@ -0,0 +1,163 @@
...
+        var transferProcess = transferProcessStore.find(processId);
+
+        if (transferProcess == null) {
+            throw new EdcException(format("Failed to found transfer process for id %s", processId));
+        }
```
paullatzelsperger, line 74:
```suggestion
            throw new EdcException(format("Could not find transfer process for id %s", processId));
```

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/src/main/java/org/eclipse/edc/connector/receiver/http/dynamic/HttpDynamicEndpointDataReferenceReceiver.java
```diff
@@ -0,0 +1,163 @@
...
+        var requestBody = RequestBody.create(typeManager.writeValueAsString(edr), JSON);
+        var requestBuilder = new Request.Builder().url(endpoint).post(requestBody);
+        if (!isNullOrBlank(authKey) && !isNullOrBlank(authToken)) {
+            requestBuilder.header(authKey, authToken);
+        }
+        try (var response = with(retryPolicy).get(() -> httpClient.newCall(requestBuilder.build()).execute())) {
+            if (response.isSuccessful()) {
+                var body = response.body();
+                if (body == null) {
+                    throw new EdcException(format("Received empty response body when receiving endpoint data reference at uri: %s", endpoint));
```
paullatzelsperger, line 101: Shouldn't this return a failed future? Same as line 105

florianrusch-zf (changes requested):
Had just a quick look through the source code

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/README.md
```diff
@@ -0,0 +1,36 @@
...
+  }
+}
+```
+
+## Configuration
+
+| Parameter name                        | Description                                              | Mandatory | Default value |
+|---------------------------------------|----------------------------------------------------------|-----------|---------------|
+| `edc.receiver.http.dynamic.auth-key`  | The header name that will be sent with the EDR request.  | false     | null          |
+| `edc.receiver.http.dynamic.auth-code` | The header value that will be sent with the EDR request. | false     | null          |
```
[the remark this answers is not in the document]
jimmarino: Thi stable is no longer needed and we should replace it with the generation of settings documentation.

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/src/main/java/org/eclipse/edc/connector/receiver/http/dynamic/HttpDynamicEndpointDataReferenceReceiver.java
```diff
@@ -0,0 +1,163 @@
...
+                }
```
florianrusch-zf, line 102: Why is this a problem?

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/README.md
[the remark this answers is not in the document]
florianrusch-zf: Hey @jimmarino that sounds even better!! Never heard about this so far. Could you provide a link where I can find some more informations about this, e.g. how to generate the documentation?

extensions/control-plane/store/sql/transfer-process-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/transferprocess/store/SqlTransferProcessStore.java
```diff
@@ -253,7 +253,8 @@ private void insert(Connection conn, TransferProcess process) {
                 toJson(process.getProvisionedResourceSet()),
                 toJson(process.getContentDataAddress()),
                 process.getType().toString(),
-                toJson(process.getDeprovisionedResources()));
+                toJson(process.getDeprovisionedResources()),
+                toJson(process.getProperties()));
```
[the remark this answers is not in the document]
wolf4ood: in `TransferProcessStoreTestBase#create` 
I've added some properties in the test that should cover this case

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/src/main/java/org/eclipse/edc/connector/receiver/http/dynamic/HttpDynamicEndpointDataReferenceReceiver.java
wolf4ood: Indeed it should

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/src/main/java/org/eclipse/edc/connector/receiver/http/dynamic/HttpDynamicEndpointDataReferenceReceiver.java
wolf4ood: It has the same behavior if the fixed http receiver

https://github.com/eclipse-edc/Connector/blob/a2995869307c2f1a1609017209be71e3688e71c7/extensions/control-plane/transfer/transfer-pull-http-receiver/src/main/java/org/eclipse/edc/connector/receiver/http/HttpEndpointDataReferenceReceiver.java#L66-L68

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/README.md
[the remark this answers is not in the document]
florianrusch-zf: @wolf4ood ahhh I just saw that `authKey` & `authCode` comes from the config...these properties should of course also be configurable in the transfer process!

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/src/test/java/org/eclipse/edc/connector/receiver/http/dynamic/TestFunctions.java
```diff
@@ -0,0 +1,66 @@
...
+
+public class TestFunctions {
+
+    public static TransferProcess createTransferProcess(String id, Map<String, String> properties) {
+        return TransferProcess.Builder.newInstance()
+                .id(id)
+                .state(TransferProcessStates.IN_PROGRESS.code())
+                .type(TransferProcess.Type.CONSUMER)
+                .dataRequest(DataRequest.Builder.newInstance()
+                        .destinationType("file")
```
ndr-brt, line 38: It looks confusing that we're testing an `HttpProxy` feature and the destination type here is `file`, that confirms that having the receiver endpoint in the destination address could have more sense (so this method would accept a `DataAddress` object)

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/README.md
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "transferType": {
+    "contentType": "application/octet-stream",
+    "isFinite": true
+  },
+  "managedResources": false,
+  "connectorAddress": "http://localhost:8282/api/v1/ids/data",
+  "connectorId": "consumer",
+  "properties": {
+    "receiver.http.endpoint" : "http://localhost:9999"
```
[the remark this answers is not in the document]
wolf4ood: That was my first implementation but since the `dataDestination` it's sent and stored in the provider side also i was not sure if it was a leak on private information (consumer side)

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/README.md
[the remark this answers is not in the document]
wolf4ood: The main issue as below I'm not sure we don't want to store those security information in the database.

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/src/main/java/org/eclipse/edc/connector/receiver/http/dynamic/HttpDynamicEndpointDataReferenceReceiver.java
wolf4ood: For the response i agree it did make sense to fail in this case.

we should change also `HttpEndpointDataReferenceReceiver`

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/README.md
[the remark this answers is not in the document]
wolf4ood: They were dynamic at first but then those those properties are store in the database and I'm not sure we want to store secrets in the database

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/src/test/java/org/eclipse/edc/connector/receiver/http/dynamic/TestFunctions.java
wolf4ood: i agree that make more sense but as above it could be a leak of information since `dataDestination` is sent to the provider and unless we provide some encryption mechanism the provider will have the consumer token info in plain sight

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/src/main/java/org/eclipse/edc/connector/receiver/http/dynamic/HttpDynamicEndpointDataReferenceReceiver.java
ndr-brt: I think this is because an our misuse of the okhttp client. As the `Response.body` documentation reports:
```
Returns a non-null value if this response was passed to Callback.onResponse or returned from Call.execute. Response bodies must be closed and may be consumed only once.
This always returns null on responses returned from cacheResponse, networkResponse, and priorResponse.
```
In our case that object will always be non-null, so, ok to check it just to avoid the warning, but if we don't have to fetch the body I would remove that control completely.

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/README.md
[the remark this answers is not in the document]
ndr-brt: Oh, you're right, what a pity, this is resolved :ok_hand:

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/src/test/java/org/eclipse/edc/connector/receiver/http/dynamic/TestFunctions.java
ndr-brt: Yes, you're right, could the `DataAddress` be passed to this method? To make it more meaningful

extensions/control-plane/transfer/transfer-pull-http-dynamic-receiver/README.md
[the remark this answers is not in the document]
ndr-brt: I guess we should do as in other parts of the connector where in the configuration there's the vault key and then it is stored in there. Maybe to be addressed in another issue?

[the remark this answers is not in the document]
ndr-brt: Duplicate conversation, will report the comment I posted in the other place:
> I guess we should do as in other parts of the connector where in the configuration there's the vault key and then it is stored in there. Maybe to be addressed in another issue?

ndr-brt (approved):

tuncaytunc-zf:
Hi @wolf4ood we had a discussion about this issue and decided, that it would be good to use the former configuration "edc.receiver.http.endpoint" as fallback when the "receiver.http.endpoint" property not set in TransferProcess. See the story in Catena-X board (https://jira.catena-x.net/browse/A1IDSC-371).
As far as I can see you are throwing an error if "receiver.http.endpoint" is not set, could you please use the "edc.receiver.http.endpoint" instead and throw an error only if both are not set. It would be great if you could make this adjustment too.

wolf4ood:
@tuncaytunc-zf 
I've incorporated in the PR the fallback URLt configuration time

tuncaytunc-zf (approved):

Issue #1566, opened 2022-06-29
Use `dev.failsafe` okhttp module on all the http calls to provide "short retries"


ndr-brt:
# Feature Request

With #1561 we are upgrading the `failsafe` library to the latest version, this could be a good opportunity to verify if every http call made by the EDC is covered by a `RetryPolicy`.
With the latest version, `failsafe` provides an [okttp module](https://failsafe.dev/okhttp/) that could be used for the purpose

## Which Areas Would Be Affected?
core

## Why Is the Feature Desired?
improve EDC resiliency

## Solution Proposal
-

## Type of Issue
improvement

## Checklist

- [x] assigned appropriate label?
- [x] **Do NOT select a milestone or an assignee!**

ndr-brt:
Maybe the best way to tackle this would be to introduce a custom defined `EdcHttpClient` service that would provide failsafe calls out of the box, because then every `OkHttpClient` will need to go together with a `RetryPolicy`. The interface could be something like:

```java
public interface EdcHttpClient {
    // for sync requests
    Response execute(Request request);

    // for async requests
    void enqueue(Callback callback);
}
```

PR #2285, opened 2022-12-05
docs: provide decision record for EdcHttpClient

ndr-brt:
## What this PR changes/adds

Add decision record for `EdcHttpClient` that's a wrapper around `OkHttpClient` and `Failsafe` that will provide retry mechanism out-of-the-box. Ref. #1566 

## Why it does that

document

## Further notes

-

## Linked Issue(s)

Ref #1566

## Checklist

### DIFF
--- a/docs/developer/decision-records/2022-12-05-edc-http-client/README.md
+++ b/docs/developer/decision-records/2022-12-05-edc-http-client/README.md
@@ -1,0 +1,35 @@
+# EDC Http Client
+
+## Decision
+
+An `EdcHttpClient` will be introduced to wrap `OkHttpClient` and `Failsafe` functionalities.
+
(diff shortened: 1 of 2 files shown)


bscholtes1A:
Would be good if this wrapper service also performs some logic common to every call, such as checking whether response body is null, log if this is the case...
Would also greatly simplify the testing.

ndr-brt:
> Would be good if this wrapper service also performs some logic common to every call, such as checking whether response body is null, log if this is the case... Would also greatly simplify the testing.

I think this is something we could introduce later on, I would start with the simplest approach but then it could grow based on needs

jimmarino (approved):

bscholtes1A (approved):

PR #2298, opened 2022-12-06
fix: Unable to set the CommonsConnectionPoolConfig Settings #2295

### LINKED ISSUE #2295: Unable to set the CommonsConnectionPoolConfig Settings
# Bug Report

## Describe the Bug
The `edc.datasource.*.pool.*` properties are ignored by the `CommonsConnectionPoolServiceExtension`. This is caused by the `Settings` in `CommonsConnectionPoolConfigKeys` are formulated in camel case, but the `ConfigImpl` will always transform the provided settings to lower case. For instance:
https://github.com/eclipse-edc/Connector/blob/c2bc86f0e4eafbf925e145b69ef5de9d3cd0a359/extensions/common/sql/sql-pool/sql-pool-apache-commons/src/main/java/org/eclipse/edc/sql/pool/commons/CommonsConnectionPoolConfigKeys.java#L22

### Expected Behavior
CommonsConnectionPoolServiceExtension using `edc.datasource.*.pool.*` properties.

### Observed Behavior
`edc.datasource.*.pool.*` properties ignored.

## Steps to Reproduce
- For instance setting `edc.datasource.policy.pool.maxTotalConnections=1`

## Possible Implementation
Changing the `Settings` in `CommonsConnectionPoolConfigKeys` to lower case.

efiege:
## What this PR changes/adds

Changes property names of `edc.datasource.*.pool.*` to lower case.

## Why it does that

`edc.datasource.*.pool.*` are not loaded, cause the values are written in camel case instead of lower case.

## Linked Issue(s)

Closes #2295 

## Checklist

### DIFF
--- a/extensions/common/sql/sql-pool/sql-pool-apache-commons/src/main/java/org/eclipse/edc/sql/pool/commons/CommonsConnectionPoolConfigKeys.java
+++ b/extensions/common/sql/sql-pool/sql-pool-apache-commons/src/main/java/org/eclipse/edc/sql/pool/commons/CommonsConnectionPoolConfigKeys.java
@@ -41,7 +41,7 @@ interface CommonsConnectionPoolConfigKeys {
     String POOL_TEST_CONNECTION_WHILE_IDLE = "pool.testconnectionwhileidle";
 
     @Setting(required = false)
-    String POOL_TEST_QUERY = "pool.testQuery";
+    String POOL_TEST_QUERY = "pool.testquery";
 
     @Setting(required = true)
     String URL = "url";
--- a/extensions/common/sql/sql-pool/sql-pool-apache-commons/src/main/java/org/eclipse/edc/sql/pool/commons/CommonsConnectionPoolConfigKeys.java
+++ b/extensions/common/sql/sql-pool/sql-pool-apache-commons/src/main/java/org/eclipse/edc/sql/pool/commons/CommonsConnectionPoolConfigKeys.java
@@ -9,6 +9,7 @@
  *
  *  Contributors:
  *       Daimler TSS GmbH - Initial API and Implementation
+ *       sovity GmbH - Renaming Settings to Lower Case
  *
  */
 
@@ -19,25 +20,25 @@ import org.eclipse.edc.runtime.metamodel.annotation.Setting;
 interface CommonsConnectionPoolConfigKeys {
 
     @Setting(required = false)
-    String POOL_MAX_IDLE_CONNECTIONS = "pool.maxIdleConnections";
+    String POOL_MAX_IDLE_CONNECTIONS = "pool.maxidleconnections";
 
     @Setting(required = false)
-    String POOL_MAX_TOTAL_CONNECTIONS = "pool.maxTotalConnections";
+    String POOL_MAX_TOTAL_CONNECTIONS = "pool.maxtotalconnections";
 
     @Setting(required = false)
-    String POOL_MIN_IDLE_CONNECTIONS = "pool.minIdleConnections";
+    String POOL_MIN_IDLE_CONNECTIONS = "pool.minidleconnections";
 
     @Setting(required = false)
-    String POOL_TEST_CONNECTION_ON_BORROW = "pool.testConnectionOnBorrow";
+    String POOL_TEST_CONNECTION_ON_BORROW = "pool.testconnectiononborrow";
 
     @Setting(required = false)
-    String POOL_TEST_CONNECTION_ON_CREATE = "pool.testConnectionOnCreate";
+    String POOL_TEST_CONNECTION_ON_CREATE = "pool.testconnectiononcreate";
 
     @Setting(required = false)
-    String POOL_TEST_CONNECTION_ON_RETURN = "pool.testConnectionOnReturn";
+    String POOL_TEST_CONNECTION_ON_RETURN = "pool.testconnectiononreturn";
 
     @Setting(required = false)
-    String POOL_TEST_CONNECTION_WHILE_IDLE = "pool.testConnectionWhileIdle";
+    String POOL_TEST_CONNECTION_WHILE_IDLE = "pool.testconnectionwhileidle";
 
     @Setting(required = false)
     String POOL_TEST_QUERY = "pool.testQuery";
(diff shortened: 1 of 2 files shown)


paullatzelsperger (changes requested):
there is not a single test case highlighting both the corrected functionality and that the fix was necessary in the first place.

ndr-brt (commented):
I pretty much agree that there should be a convention for settings key, e.g. upper case characters should be avoided (my taste, nothing we decided on an organization level).

Anyway I see another issue here:
I'm not sure that the "ENV_FORMAT to property.format" conversion should be done in the `ConfigImpl` constructor, that should be done only when we load configurations from the environment variables, I would move that behavior into a specific method in the `ConfigFactory`, that should fix the issue without changing the setting keys.

paullatzelsperger (changes requested):
Why do you need to instantiate the `CommonConnectionPoolServiceExtension` in the `CommomsConnectionPoolConfigTest`? If the objective was to test that the _extension_ performs as expected, then the respective test code should be moved to the extension's test or into a new test class. 
Secondly, that test code should stick to the following pattern:
- create mocks for collaborators, eg,. for `DataSourceRegistry`
- register the mocks with the context
- instantiate the extension using the object factory and the context (you could look at the `ObservabilityApiExtensionTest` how that works)
- call its `initialize` method 
- verify collaborators and mocks, see that all config properties and their defaults were used correctly.

On a related note I agree with @ndr_brt that the connection pool extension is in dire need of refactoring, because it violates almost every coding guideline we have (final classes were they aren't needed, wrong use of builders,...). If you are interested in doing that, a decision record outlining the planned changes would be necessary.

efiege:
You are right, it would make more sense to put the test code in the extensions test.

Here is why I did not stick to the pattern you mentioned:

Too show that the properties are not loading correctly following needs to be verified:
1. The extensions constants are used to retrieve the properties
2. The properties values are not returned as expected when the extensions constants are used

The error results from the `CommonConnectionPoolServiceExtension` implementation not working properly with the `ConfigFactory` implementation. To prove this I needed both implementations in the tests. But when I use the injected `ServiceExtensionContext` I have no control over what `Config` is returned on the contexts `getConfig` method. So it would not be possible to show (1.) that the extensions constants are used to retrieve the config properties using the `Config`.

To prove that the properties are not loading as expected, the provided tests verify the following:
- When the `CommonConnectionPoolServiceExtension` implementation is used together with the `ConfigFactory` implementation the get property methods are called on the `Config` with the extensions constants
- When the constants are used to retrieve the properties on the same `Config` the corresponding property value is returned

SebastianOpriel:
It is a pity that this technically valid solution was not adopted and now got closed due to inactivity and another workaround by @ndr-brt. As stated in this [document](https://catena-x.net/fileadmin/user_upload/Standard-Bibliothek/Update_PDF_Feb._2023/3_Sovereign_Data_Exchange_v1.0/CX_-_0018_EDC_PC_Self-Sovereign_Data_Exchange_v_1.0.0.pdf) one shall contribute, instead of creating an own solution. Such examples as this PR will demotivate companies who are willing to contribute, with the result of the opposite: resolving such issues in ones own forked EDC.

jimmarino:
> It is a pity that this technically valid solution was not adopted and now got closed due to inactivity and another workaround by @ndr-brt. As stated in this [document](https://catena-x.net/fileadmin/user_upload/Standard-Bibliothek/Update_PDF_Feb._2023/3_Sovereign_Data_Exchange_v1.0/CX_-_0018_EDC_PC_Self-Sovereign_Data_Exchange_v_1.0.0.pdf) one shall contribute, instead of creating an own solution. Such examples as this PR will demotivate companies who are willing to contribute, with the result of the opposite: resolving such issues in ones own forked EDC.

Please read the contribution guidelines carefully. Two committers requested changes that were never made, so this PR was closed after two weeks. A new PR can be submitted if the technical issues raised by the committers are addressed.

SebastianOpriel:
> the document you referenced is not affiliated with the Eclipse EDC project

My bad, you are right. Nevertheless, there is a great overlap of contributors to Eclipse EDC project and those who are involved in Catena-X. I assumed, that opinions won't be too far away. But as said - fair enough.


> Two committers requested changes that were never made

In this case, there was an explanation afterwards, regarding why the code was structured in this way. I guess every contributor would be happy to have some more feedback afterwards, like "Your points are valid" or "We stick to our opinion" and "but to have this PR accepted, please make changes stated above". This would help and lower frustration.

ndr-brt:
> > Two committers requested changes that were never made
> 
> In this case, there was an explanation afterwards, regarding why the code was structured in this way. I guess every contributor would be happy to have some more feedback afterwards, like "Your points are valid" or "We stick to our opinion" and "but to have this PR accepted, please make changes stated above". This would help and lower frustration.

to be honest, my review was completely ignored so from my point of view this PR wasn't good for merging.

SebastianOpriel:
@ndr-brt Maybe I missed some point here, but directly after your initial message, there was a question on how to address the issue - without an answer from your side:
> @ndr-brt Should I add another enhancement issue for that? Currently fromMap method in the ConfigFactory is used in the whole project with both environment variables and java properties format

Please do not get me wrong. I am not willing to blame someone or start a discussion. Initially I simply wanted to point out, that we took effort, wanting to contribute to the project (and also took feedback of Paul and added tets), but with an result, which was not motivating. We are willing to contribute and same as you, want to push EDC to great success!

SebastianOpriel:
I am aware of the process and I agree with you in most points and know that focus is paid to more crucial parts of development. Lets start over with new PRs and keep collaborating :-)

ndr-brt:
> @ndr-brt Maybe I missed some point here, but directly after your initial message, there was a question on how to address the issue - without an answer from your side:
> 
> > @ndr-brt Should I add another enhancement issue for that? Currently fromMap method in the ConfigFactory is used in the whole project with both environment variables and java properties format
> 
> Please do not get me wrong. I am not willing to blame someone or start a discussion. Initially I simply wanted to point out, that we took effort, wanting to contribute to the project (and also took feedback of Paul and added tets), but with an result, which was not motivating. We are willing to contribute and same as you, want to push EDC to great success!

I likely lost the notification, next time please ping again

PR #2321, opened 2022-12-09
fix: remove `DataPlanePublicApiUrl`

### LINKED ISSUE #2320: bug: URL contained in `DataPublicApiUrl` might be different from actual Data Plane Public API URL
# Bug Report

## Describe the Bug
For now `DataPlanePublicApiUrl` is built as such:

```java
    @Provider
    public DataPlanePublicApiUrl dataPlanePublicApiUrl() {
        try {
            var url = new URL(String.format("http://%s:%s%s", hostname.get(), configuration.getPort(), configuration.getPath()));
            return () -> url;
        } catch (MalformedURLException e) {
            throw new EdcException(e);
        }
    }
```

The URL that this provides is not intended to be used in production-grade deployments, as:

- it uses `http`
- the public API will likely be exposed through a gateway or an Ingress, which uses it's own address (for example when using Ingress there is no port in the URL). 

### Observed Behavior
`DataPlanePublicApiUrl` should give the _real_ URL of the public API (the one exposed on the public network)

## Possible Implementation
As discussed with @paullatzelsperger, it is preferred to remove `DataPublicApiUrl` as it might be confusing.

bscholtes1A:
## What this PR changes/adds

Remove `DataPlanePublicApiUrl`.

## Why it does that

Current way `DataPlanePublicApiUrl` is built might not be relevant depeding on how connector is deployed (use of gateway, ingress)

## Linked Issue(s)

Closes #2320 

## Checklist


jimmarino (approved):

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/DataPlaneApiExtension.java
```diff
@@ -102,20 +99,25 @@ public void initialize(ServiceExtensionContext context) {
...
+        webService.registerResource(publicApiConfig.getContextAlias(), publicApiController);
     }
 
-    @Provider
-    public DataPlanePublicApiUrl dataPlanePublicApiUrl() {
+    @Provider(isDefault = true)
+    public DataPlanePublicApiUrl dataPlanePublicApiUrl(ServiceExtensionContext context) {
         try {
-            var url = new URL(String.format("http://%s:%s%s", hostname.get(), configuration.getPort(), configuration.getPath()));
+            var config = publicApiConfig(context);
```
paullatzelsperger, line 110: the `webServiceConfigurer` could be `null` by the time the default provider method is invoked.

paullatzelsperger (commented):
I do not quite understand why this could not be done through configuration?

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/DataPlaneApiExtension.java
bscholtes1A: > I do not quite understand why this could not be done through configuration?

`DataPlanePublicApiUrl` removed as discussed

ndr-brt (approved):
:ok_hand:

PR #2332, opened 2022-12-12
fix: set correct `consumer` and `provider` ids on `ContractOffer`

### LINKED ISSUE #753: IDS: `ContractOffer` is missing values for `provider` and `consumer`
### Problem statement
in the `ContractValidationServiceImpl` there is a check in place that requires that both the `provider` and `consumer` fields of a `ContractOffer` be not-null:
```java
private boolean isMandatoryAttributeMissing(ContractOffer offer) {
    // TODO add contractStart and contractEnd check as soon as the Infomodel serializer is used
    return offer.getProvider() == null || offer.getConsumer() == null;
}
```
However, those values are never set, which causes **all contract negotiations to be declined**.

### Proposed solution
The `provider` and `consumer` fields must be set during the creation of the `ContractOffer`, which is done in the `ContractOfferServiceImpl`. 

Suggestion:
- `provider`: the offering connector's ID as URN
- `consumer`: the ID of the `ParticipantAgent` for whom the offer is generated

ndr-brt:
## What this PR changes/adds

Set the correct `provider` (connector id) and `consumer` (issuer connector) fields on the `ContractOffer`

## Why it does that

likely these fields will be removed from the contract offer in the future, but in the meantime better value them correctly since they are provided
Fix bug #753 

## Further notes
- introduced a `DescriptionRequest` data class that contains some fields necessary for the description request
- refactored `DescriptionRequestHandler` because it was hard to read with those 2 switch-case blocks.

## Linked Issue(s)

Closes #753 

## Checklist

### DIFF
--- a/data-protocols/ids/ids-core/src/main/java/org/eclipse/edc/protocol/ids/service/CatalogServiceImpl.java
+++ b/data-protocols/ids/ids-core/src/main/java/org/eclipse/edc/protocol/ids/service/CatalogServiceImpl.java
@@ -20,8 +20,7 @@ import org.eclipse.edc.catalog.spi.Catalog;
 import org.eclipse.edc.connector.contract.spi.offer.ContractOfferQuery;
 import org.eclipse.edc.connector.contract.spi.offer.ContractOfferResolver;
 import org.eclipse.edc.protocol.ids.spi.service.CatalogService;
-import org.eclipse.edc.spi.iam.ClaimToken;
-import org.eclipse.edc.spi.query.QuerySpec;
+import org.eclipse.edc.protocol.ids.spi.types.container.DescriptionRequest;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.Objects;
@@ -39,19 +38,17 @@ public class CatalogServiceImpl implements CatalogService {
         this.contractOfferResolver = Objects.requireNonNull(contractOfferResolver);
     }
 
-    /**
-     * Provides the dataCatalog object, which may be used by the IDS self-description of the connector.
-     *
-     * @return data catalog
-     */
     @Override
-    @NotNull
-    public Catalog getDataCatalog(ClaimToken claimToken, QuerySpec querySpec) {
+    public @NotNull Catalog getDataCatalog(@NotNull DescriptionRequest descriptionRequest) {
+        var querySpec = descriptionRequest.getQuerySpec();
 
         var query = ContractOfferQuery.Builder.newInstance()
-                .claimToken(claimToken)
+                .claimToken(descriptionRequest.getClaimToken())
                 .assetsCriteria(querySpec.getFilterExpression())
-                .range(querySpec.getRange()).build();
+                .range(querySpec.getRange())
+                .provider(descriptionRequest.getProvider())
+                .consumer(descriptionRequest.getConsumer())
+                .build();
 
         try (var offers = contractOfferResolver.queryContractOffers(query)) {
             return Catalog.Builder.newInstance()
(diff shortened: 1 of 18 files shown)


core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/negotiation/command/handlers/DeclineNegotiationCommandHandlerTest.java
```diff
@@ -49,7 +49,6 @@ void handle_negotiationExists_declineNegotiation() {
                 .protocol("test-protocol")
                 .state(ContractNegotiationStates.REQUESTED.code())
                 .build();
-        var originalTime = negotiation.getUpdatedAt();
         var command = new DeclineNegotiationCommand(negotiationId);
```
paullatzelsperger, line 52: why was this deleted? I don't agree with removing test assertions unless there is a good reason.

data-protocols/ids/ids-core/src/main/java/org/eclipse/edc/protocol/ids/service/CatalogServiceImpl.java
```diff
@@ -39,19 +38,17 @@ public CatalogServiceImpl(
         this.contractOfferResolver = Objects.requireNonNull(contractOfferResolver);
     }
 
-    /**
```
paullatzelsperger, line 42: If you delete comments, please be sure to amend the one that is on the inteface method, otherwise we could lose information.

system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/AbstractEndToEndTransfer.java
```diff
@@ -48,8 +48,7 @@ void httpPullDataTransfer() {
         var assetId = UUID.randomUUID().toString();
         createResourcesOnProvider(assetId, noConstraintPolicy(), UUID.randomUUID().toString(), httpDataAddressProperties());
 
-        var catalog = CONSUMER.getCatalog(PROVIDER.idsEndpoint());
-        assertThat(catalog.getContractOffers()).hasSizeGreaterThan(0);
```
paullatzelsperger, line 52: Not a big fan of moving test assertions into util methods, because the make the test code less readable and tedious to maintain. 
If there are recurring assertions, at the very least I would create a dedicated static method in some test util class with a descriptive name.

ndr-brt: Generally I agree with you, but these are end to end tests, they should demonstrate the system from an hi-level perspective, showing everything in the main test method will make the test pretty impossible to read.
These assertions are not meaningful to the scope of the test, but they verifies that no regression are going to be introduced on some really important details.

data-protocols/ids/ids-core/src/main/java/org/eclipse/edc/protocol/ids/service/CatalogServiceImpl.java
ndr-brt: It's documented on the interface, added a missing sentence there.

core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/negotiation/command/handlers/DeclineNegotiationCommandHandlerTest.java
ndr-brt: Yes, I was just annoyed by this test and thought that behavior was already tested on the `ContractNegotiation`, but in fact it's not, anyway I made it not flaky anymore.

system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/AbstractEndToEndTransfer.java
paullatzelsperger: I'd argue that guarding against regression should be a separate, explicit test then (IDK if we have one already), where the assertion is written out.
if it is outside of the scope, the assertion is not necessary...

let's not get hung up on that though.

paullatzelsperger (approved):

ronjaquensel (approved):

PR #2348, opened 2022-12-14
feat(api): improve error message on conflict error

### LINKED ISSUE #2344: Misleading Error Message when deleting Asset
# Bug

## Describe the Bug
After some testing with the API of the edc, Creating Asset, Policy and Contract Definition and afterwards the negotiation and transfer part.  I wanted to delete the Asset again, because it has an old DataAdress. 
After doing everything backwards, the deletion of the asset fails with an `409 Conflict Error` and with following Message
```
[
    {
        "message": "Object of type Asset already exists with ID = demo-asset-1",
        "type": "ObjectExists",
        "path": null,
        "invalidValue": null
    }
]
```
W.r.t. my action "Delete the Asset", this message is misleading.  

### Expected Behavior

Delete the asset or provide a more detailed error message (e.g. cannot delete because of ...)

### Observed Behavior

I debugged the code and figured out, that it already should throw/provide a more specific error message. The execution returns this `ServiceResult` in the `AssetServiceImpl.java` , but instead it shows the message from an `ObjectExistsExecption` .

```
try (var negotiationsOnAsset = contractNegotiationStore.queryNegotiations(query)) {
    if (negotiationsOnAsset.findAny().isPresent()) {
        return ServiceResult.conflict(format("Asset %s cannot be deleted as it is referenced by at least one contract agreement", assetId));
    }
}
```

wolf4ood:
## What this PR changes/adds

Changes the message in the response when the error on service layer is `Conflict`

## Why it does that

The prev implementation sends a generic message `Object exists` on conflict error which is confusing
in request like delete.

## Further notes

The `ObjectExistsException` has been renamed to `ObjectConflictException`  and instead of generic message

`Object of type {} already exists with ` it forward the error messages coming from the ServiceFailure`

## Linked Issue(s)

Closes #2344 

## Checklist


extensions/common/http/jersey-core/src/main/java/org/eclipse/edc/web/jersey/mapper/EdcApiExceptionMapper.java
```diff
@@ -50,7 +50,7 @@ public EdcApiExceptionMapper() {
                 NotAuthorizedException.class, FORBIDDEN,
                 InvalidRequestException.class, BAD_REQUEST,
                 ObjectNotFoundException.class, NOT_FOUND,
-                ObjectExistsException.class, CONFLICT,
+                ObjectConflictException.class, CONFLICT,
                 ObjectNotModifiableException.class, CONFLICT,
```
ndr-brt, line 54: while you are at it, this `ObjectNotModifiableException` class could be removed, it's not used by anyone

ndr-brt (approved):
Ace!

extensions/common/http/jersey-core/src/main/java/org/eclipse/edc/web/jersey/mapper/EdcApiExceptionMapper.java
wolf4ood: :axe: :+1:

paullatzelsperger (approved):

PR #2354, opened 2022-12-15
feat: Support for private-key based client credential in `provision-oauth2`

### LINKED ISSUE #2353: Improve `provision-oauth2` to support all client credential types
# Feature Request

Today the `provision-oauth2` extension introduced [here](https://github.com/eclipse-edc/Connector/pull/2135) only supports the use-case where the client use a shared secret to authenticate to the Oauth2 server. This is one type of client credential. The other one is the private-key based client credential wherein the client generates a signed JWT (`client_assertion`). This type of client credential is generally preferred in production system due to increased security. 

This aims aims at extending the existing implementation to support both type of client credentials, i.e. shared secret and private-key based.

Control plane.

Most production oauth2 servers will require a private-key based client credential.

bscholtes1A:
## What this PR changes/adds

Add support for private-key based client credential in `provision-oauth2`.

## Why it does that

Most production oauth2 server use private-key based client credential as this type of client credential is more secured.

## Linked Issue(s)

Closes #2353 

## Checklist


extensions/common/iam/oauth2/oauth2-client/src/main/java/org/eclipse/edc/iam/oauth2/Oauth2ClientExtension.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Inject
+    private EdcHttpClient httpClient;
+
+    @Override
+    public String name() {
+        return NAME;
+    }
+
+    @Provider
```
ndr-brt, line 39: I have a proposal, because this is good as we provide extensibility, but the downside here is then when I add a `provision-oauth2` dependency then I'd have to explicitly add the `oauth2-client` as well. Couldn't the `oauth2-client` added in the build file of `provision-oauth2` (it is already on the `oauth2-core`) and this service provided as `isDefault = true`?
This way it won't be needed to have to explicitly define the client dependency (as it was done in the `control-plane` runtime of the end to end tests) but it would be extensible if needed.

ndr-brt (commented):
Great PR, love that "client" refactor.

In addiction to my comments, there's the "typical me" topic against `private static` methods:
i don't like them :D , but the real  problem with them occurs when someone has an automatic code formatter, in that case they are pushed to the top of the class, not really good. That's why I would keep them not static.

extensions/control-plane/provision/provision-oauth2/src/main/java/org/eclipse/edc/connector/provision/oauth2/Oauth2Provisioner.java
```diff
@@ -114,12 +110,38 @@ public CompletableFuture<StatusResult<DeprovisionedResource>> deprovision(Oauth2
...
+                .build());
+    }
+
+    private Result<TokenRepresentation> createAssertion(String pkSecret, Oauth2ResourceDefinition rd) {
+        var privateKey = privateKeyResolver.resolvePrivateKey(pkSecret, PrivateKey.class);
+        if (privateKey == null) {
+            return Result.failure("Failed to resolve private key with alias: " + pkSecret);
         }
+        var decorator = new Oauth2AssertionDecorator(rd.getTokenUrl(), rd.getClientId(), clock, rd.getValidity());
+        var service = new TokenGenerationServiceImpl(privateKey);
```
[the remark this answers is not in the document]
bscholtes1A: It can't, as the constructor of `TokenGenerationServiceImpl` takes the private key as argument, while here the private key is "dynamic" in the sense that it is fetched from the `PrivateKeyResolver` based on the data address properties.

extensions/common/iam/oauth2/oauth2-client/src/main/java/org/eclipse/edc/iam/oauth2/Oauth2ClientExtension.java
bscholtes1A: I am pretty sure that @paullatzelsperger would scream on it :D 
As default provider are by design not intended to be used for extensibility purposes as mentioned in the doc.

extensions/control-plane/provision/provision-oauth2/src/main/java/org/eclipse/edc/connector/provision/oauth2/Oauth2Provisioner.java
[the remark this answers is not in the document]
ndr-brt: Oh, you're right, my fault.

extensions/common/iam/oauth2/oauth2-client/src/main/java/org/eclipse/edc/iam/oauth2/Oauth2ClientExtension.java
paullatzelsperger: As @bscholtes1A indicated, the `Oauth2ClientImpl` doesn't really fit the bill for a default provider, because it is not a "fallback" implementation, as in: "if there is nothing else available, use this".
In most cases people won't write their own `Oauth2Client` because the one we provide is deemed production-ready.
That means we _could_ let the `provision-oauth2` module explicitly depend on `oauth2-client`, presumably as `runtimeOnly`. That will mean however, that if someone wants to indeed write their own `Oauth2Client` implementation, they'd have to explicitly exclude that (transitive) dependency onto `oauth2-client` from their launcher config, otherwise they'd end up with multiple `Oauth2Client` impls on the classpath, which is undefined behaviour.

How about instead of doing that we provide a BOM that pulls in all of the OAuth2 modules? That way it would satisfy the needs of 99%, while still providing flexibility for the wicked :)

ndr-brt: > How about instead of doing that we provide a BOM that pulls in all the OAuth2 modules? That way people could just use that, and it would satisfy the needs of 99%, while still providing flexibility for the wicked :)

What if `oauth2` is used for provisioning but not for authenticating on the network? that BOM could be misleading because it will register an `IdentityService` instance. 
An `oauth2-client` is something that is likely needed by every `oauth2` extension, but having a BOM that loads every `oauth2`-related extension doesn't seems to be something really usable.

paullatzelsperger: that could be a separate BOM then: we could have an "OauthAuthentication" and an "OAuthProvisioning" BOM. Using both these BOMs should not cause any problems.

I think we should not declare that dependency on the module/artifact level, if we foresee the slightest chance that anyone will write their own `Oauth2Client` implementation. 

If we don't, then declaring it on the module/artifact level is fine, but in that case I would go as far as to say (and document) that provisioning with `OAuth2` is only supported with the `OAuth2ClientImpl`, everything else is a workaround.

bscholtes1A: @paullatzelsperger where would you see these BOM located in the codebase?

paullatzelsperger: @bscholtes1A for our purposes here, a BOM is just a gradle build script that aggregates dependencies (i.e. the other modules), so they would probably have to be in separate modules e.g. `oauth-core-bom` and `provision-oauth-bom` that contain only the aforementioned build files.

jimmarino: Similar to Paul, I think coupling the modules via dependency is not the best way to handle this. If they are coupled, they should be contained in a common module. However, that's not the case here so BOMs seem the right mechanism to use.

jimmarino: To clarify, by dependency in the module, I meant two “code” modules. That is a smell. If one module depends on another implementation that is not a library, then there should be one module; otherwise use a BOM.

bscholtes1A: @jimmarino  @paullatzelsperger @ndr-brt 
I created the two BOMs `oauth2-service-bom` and `provision-oauth2-bom` in the last commit. 
Could you please check if that looks better now?

extensions/common/iam/oauth2/oauth2-core/src/main/java/org/eclipse/edc/iam/oauth2/jwt/X509CertificateDecorator.java
```diff
@@ -0,0 +1,47 @@
...
+import org.eclipse.edc.spi.EdcException;
+
+import java.security.cert.CertificateEncodingException;
+import java.security.cert.X509Certificate;
+import java.util.Collections;
+import java.util.Map;
+
+import static org.eclipse.edc.iam.oauth2.jwt.Fingerprint.sha1Base64Fingerprint;
+
+public class X509CertificateDecorator implements JwtDecorator {
```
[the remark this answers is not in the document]
bscholtes1A: Done

extensions/control-plane/provision/provision-oauth2/src/main/java/org/eclipse/edc/connector/provision/oauth2/Oauth2DataAddressValidator.java
```diff
@@ -30,9 +32,12 @@ public class Oauth2DataAddressValidator implements Predicate<DataAddress> {
     private final Predicate<DataAddress> isHttpDataType = dataAddress -> "HttpData".equals(dataAddress.getType());
     private final Predicate<DataAddress> hasClientId = dataAddress -> dataAddress.hasProperty(CLIENT_ID);
     private final Predicate<DataAddress> hasClientSecret = dataAddress -> dataAddress.hasProperty(CLIENT_SECRET);
+    private final Predicate<DataAddress> hasPrivateKeySecret = dataAddress -> dataAddress.hasProperty(PRIVATE_KEY_NAME);
+    private final Predicate<DataAddress> hasValidity = dataAddress -> dataAddress.hasProperty(VALIDITY);
     private final Predicate<DataAddress> hasTokenUrl = dataAddress -> dataAddress.hasProperty(TOKEN_URL);
 
-    private final Predicate<DataAddress> isValid = isHttpDataType.and(hasClientId).and(hasClientSecret).and(hasTokenUrl);
+    private final Predicate<DataAddress> isValid = isHttpDataType.and(hasClientId).and(hasTokenUrl)
```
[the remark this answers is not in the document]
bscholtes1A: done

ndr-brt (commented):
Good for me, I'd just remove the `bom` term from the module name

paullatzelsperger (approved):

paullatzelsperger:
> Good for me, I'd just remove the `bom` term from the module name

@ndr-brt wouldn't that create a module name conflict?

paullatzelsperger:
> Good for me, I'd just remove the `bom` term from the module name

@ndr-brt wouldn't that create a module name conflict?

bscholtes1A:
>

bscholtes1A:
> that create a module name conflict?

Was going to say the same

ndr-brt:
> > Good for me, I'd just remove the `bom` term from the module name
> 
> @ndr-brt wouldn't that create a module name conflict?

If so I would rename the base module. I mean, module names should be self-explainatory, `bom` is not telling a lot.
If the current `provision-oauth2` doesn't work alone because it needs a `oauth2-client`, we could call the first `provision-oauth2-base` or `provision-oauth2-core` and the bom simply `provision-oauth2`, because that's the one that will be used in the 99% of cases.

PR #2369, opened 2022-12-20
feat: Introduce `Fallback` mechanism to `EdcHttpClient`

### LINKED ISSUE #2368: Make `EdcHttpClient` retry when the response code is not as expected
# Feature Request

The `EdcHttpClient` was implemented to provide an http client with a retry mechanism.
Currently it retries a call only when it failed with an exception (`ConnectException`, `IOException`), it would be great to make it retry also when the status of the response is not as expected, to handle eventual `503 - Service unavailable` or similar, that could be flaky.

http client

improve resilience

## Solution Proposal
Use `Fallback` to set custom fallbacks to `Failsafe`.

ndr-brt:
## What this PR changes/adds

Adds two methods to `EdcHttpClient`:
```
<T> Result<T> execute(Request request, Function<Response, Result<T>> mappingFunction)
<T> Result<T> execute(Request request, List<FallbackFactory> fallbacks, Function<Response, Result<T>> mappingFunction)
```

First one is just a shortcut to call the second one without `fallbacks`.
In `Failsafe`, a `Fallback` is an object that permits to apply retry also after a `Response` is received from the `OkHttpClient`, in particular it can be used to retry when the status result is not "successful" (not contained between 200 and 299) or is not a specific state. This is useful to retry calls when flaky statuses as `503 - service unavailable` are returned.

## Why it does that

Improve resiliency

## Further notes

- the methods accept a `mappingFunction` that will apply on the `Response`, so it will be up to the client to close the `Response` and catch unexpected exceptions and map them to `Result.failure`
- made `Oauth2ClientImpl` use these methods as a showcase.

## Linked Issue(s)

Closes #2368

## Checklist

### DIFF
--- a/spi/common/http-spi/src/main/java/org/eclipse/edc/spi/http/FallbackFactory.java
+++ b/spi/common/http-spi/src/main/java/org/eclipse/edc/spi/http/FallbackFactory.java
@@ -12,22 +12,21 @@
  *
  */
 
-plugins {
-    `java-library`
-    `maven-publish`
-}
+package org.eclipse.edc.spi.http;
 
-dependencies {
-    api(project(":spi:common:core-spi"))
+import dev.failsafe.Fallback;
+import okhttp3.Request;
+import okhttp3.Response;
 
-    api(libs.okhttp)
-    api(libs.failsafe.okhttp)
-}
-
-publishing {
-    publications {
-        create<MavenPublication>(project.name) {
-            from(components["java"])
-        }
-    }
+/**
+ * Factory for {@link Fallback} instances given a {@link Request}
+ */
+public interface FallbackFactory {
+    /**
+     * Create a {@link Fallback} from a {@link Request}
+     *
+     * @param request the {@link Request}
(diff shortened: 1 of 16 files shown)


spi/common/http-spi/src/main/java/org/eclipse/edc/spi/http/FallbackFactory.java
```diff
@@ -0,0 +1,32 @@
...
+package org.eclipse.edc.spi.http;
+
+import dev.failsafe.Fallback;
+import okhttp3.Request;
+import okhttp3.Response;
+
+/**
+ * Factory for {@link Fallback} instances given a {@link Request}
+ */
+public interface FallbackFactory {
```
paullatzelsperger, line 24: i could see this also being useful for other things than just HTTP requests, e.g. when we use some client SDK like Azure or AWS, were we typically get SDK exceptions or some status enum rather than HTTP status codes.  
May be something to look at in the future.

paullatzelsperger (approved):

PR #2373, opened 2022-12-21
feat: add synthetic service provider

### LINKED ISSUE #2372: Create synthetic provider for `Monitor`, `Clock`, `Telemetry` and `TypeManager`
# Feature Request

The `Monitor` and the `TypeManager` can be `@Inject`-ed into extensions just like any other service, but in contrast to other services, they are not provided by an extension.
Rather, they are set on the context before the entire service registration process starts. 

That means, that even though they technically _always there_, it is the `CoreServicesExtension` that actually declares them in its `@Provides` list which makes it susceptible to a `CyclicDependencyException`.

bootstrap, core

robustness, dependency resolution

## Solution Proposal
Instead of (artifically) letting the `CoreServicesExtension` provide these services, we should create an extension, whos only job is to declare an outgoing dependency onto these services:
```java

@Provides({ Monitor.class, TypeManager.class, Telemetry.class, Clock.class })
public class SyntheticExtension implements org.eclipse.edc.spi.system.ServiceExtension {
   // nothing to do here
}
```

As an optimization, I would not register this through the normal service loader mechanism, but rather explicitly in the `DependencyGraph`:
```java
  //DependencyGraph.java, Line 64ff:
  public List<InjectionContainer<ServiceExtension>> of(List<ServiceExtension> loadedExtensions) {
        var extensions = sortByType(loadedExtensions);
        extensions.add(new SyntheticExtension());
        var dependencyMap = createDependencyMap(extensions);
        //...
  }
```

paullatzelsperger:
## What this PR changes/adds

Adds a "synthetic" extension, whos job is solely to declare a `@Provides` list for the `Monitor`, `TypeMAnager`, `Clock` and `Telemetry`, all of which are services that are instantiated in a special way.

## Why it does that

Avoid cyclic dependencies

## Further notes

- ~~I opted for adding the `SyntheticExtension` and then removing it again, because that seemed to have the least impact on the code base while at the same time satisfying the dependency resolution mechanism.~~
- ~~The `SyntheticExtension` is not needed outside of dependency loading, which is why I chose not to instantiate it through the service loader and I removed it again right after the topological sort went through.~~ 
-
- The `ServiceExtensionContext` is used as fallback: whenever a feature cannot be satisfied through normal dependency resolution, we check if the context already has that service registered, in which case no error is thrown.
- Another option would have been to add those services directly into the dependency map (see next snippet), but that seems less flexible and heavy-handed and error-prone:
  ```java
    public List<InjectionContainer<ServiceExtension>> of(List<ServiceExtension> loadedExtensions) {
        var extensions = sortByType(loadedExtensions);
        var dependencyMap = createDependencyMap(extensions);
        addStandardServices(dependencyMap);
        // ...
    }

    private void addStandardServices(Map<String, List<ServiceExtension>> dependencyMap) {
        dependencyMap.put(Monitor.class.getName(), emptyList());
        dependencyMap.put(TypeManager.class.getName(), emptyList());
        dependencyMap.put(Clock.class.getName(), emptyList());
        dependencyMap.put(Telemetry.class.getName(), emptyList());
    }

## Linked Issue(s)

Closes #2372 

## Checklist


core/common/boot/src/main/java/org/eclipse/edc/boot/system/SyntheticExtension.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.eclipse.edc.runtime.metamodel.annotation.Provides;
+import org.eclipse.edc.spi.monitor.Monitor;
+import org.eclipse.edc.spi.telemetry.Telemetry;
+import org.eclipse.edc.spi.types.TypeManager;
+
+import java.time.Clock;
+
+@Provides({ Monitor.class, TypeManager.class, Telemetry.class, Clock.class })
+class SyntheticExtension implements org.eclipse.edc.spi.system.ServiceExtension {
```
ndr-brt, line 25: What's the concept behind the `Synthetic` term? 
Could the name be something that explains its use? Like `BaseServicesExtension` or, I don't know, `FoundationalServicesExtension`

core/common/boot/src/main/java/org/eclipse/edc/boot/system/SyntheticExtension.java
```diff
@@ -0,0 +1,26 @@
...
+package org.eclipse.edc.boot.system;
+
+import org.eclipse.edc.runtime.metamodel.annotation.Provides;
+import org.eclipse.edc.spi.monitor.Monitor;
+import org.eclipse.edc.spi.telemetry.Telemetry;
+import org.eclipse.edc.spi.types.TypeManager;
+
+import java.time.Clock;
+
+@Provides({ Monitor.class, TypeManager.class, Telemetry.class, Clock.class })
```
ndr-brt, line 24: I get why `Monitor` and `Telemetry` should be "peculiar", as they are a specific loading mode in the `ExtensionLoader`, but `TypeManager` and `Clock` could be treated as standard services, or am I wrong?
This could be something for another issue maybe.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/SyntheticExtension.java
paullatzelsperger: The extension providing these services is not _really_ providing them, it merely declares the outgoing dependency (`@Provides`), thus acting as a ... well... synthetic provider for them :)
Other than that, it was the first and (to me) most intuitive term, but I have absolutely nothing agains renaming it.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/SyntheticExtension.java
paullatzelsperger: All of them are registered in the `DefaultServiceEXtensionContext`'s constructor, i.e. they are not provided by an extension. That is why I had intended to treat them equally. Also, the original problem @FlorianSeidelAWS reported dealt with a potential cyclic dependency, which would hit us just the same with the `TM` and the `Clock`

ndr-brt: Ok, but there's no actual need to have `TypeManager` and `Clock` registered before the service loading and passed to the context. Will create an issue about that, nothing to do here.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/SyntheticExtension.java
ndr-brt: Ok, `Dummy` could be a good term to be used as it does not really provide anything but it has to be there for "appearence". What about `BaseServicesDummyExtension`? So we could refer to those services as "base", they are provided out of the box with the `boot` module.

paullatzelsperger: Could we name it _just_ `BaseServicesExtension`? Not too keen on having the term `Dummy` in production code...

core/common/boot/src/main/java/org/eclipse/edc/boot/system/SyntheticExtension.java
paullatzelsperger: @ndr-brt the only think to be aware of is that in case they get provided conventionally by an extension, that particular extension should have as little incoming dependencies (`@Inject`) as possible to minimize the risk for dependency cycles

core/common/boot/src/main/java/org/eclipse/edc/boot/system/SyntheticExtension.java
ndr-brt: sure, sounds good to me

core/common/boot/src/main/java/org/eclipse/edc/boot/system/SyntheticExtension.java
ndr-brt: that extension should be located in `boot`, so at the root of everything

paullatzelsperger: > that extension should be located in `boot`, so at the root of everything

I meant "code" dependencies, not Gradle dependencies. Ideally it should not have any `@Inject`-ed fields at all, which would make it dependent on other extensions, which we should avoid at all cost - only then is it truly at the "root" (=beginning) of the topologically sorted list.

paullatzelsperger:
@ndr-brt after talking with @jimmarino I think there is a better way than instantiating an "artificial" extension: we could just use the `ServiceExtensionContext` as fallback in case a dependency is not satisfied. That way no special treatment is needed.
My last [commit](https://github.com/eclipse-edc/Connector/pull/2373/commits/9ef9e2e196661ba834b751652159c9b2146d4b1b) displays this.

ndr-brt:
> @ndr-brt after talking with @jimmarino I think there is a better way than instantiating an "artificial" extension: we could just use the `ServiceExtensionContext` as fallback in case a dependency is not satisfied. That way no special treatment is needed. My last [commit](https://github.com/eclipse-edc/Connector/pull/2373/commits/9ef9e2e196661ba834b751652159c9b2146d4b1b) displays this.

Looks good to me, so can we get rid of the `SyntheticExtension`?

paullatzelsperger:
> > @ndr-brt after talking with @jimmarino I think there is a better way than instantiating an "artificial" extension: we could just use the `ServiceExtensionContext` as fallback in case a dependency is not satisfied. That way no special treatment is needed. My last [commit](https://github.com/eclipse-edc/Connector/pull/2373/commits/9ef9e2e196661ba834b751652159c9b2146d4b1b) displays this.
> 
> Looks good to me, so can we get rid of the `SyntheticExtension`?

yes of course, forgot to delete, thanks.

ndr-brt (approved):

core/common/boot/src/main/java/org/eclipse/edc/boot/system/DependencyGraph.java
```diff
@@ -121,6 +130,22 @@ public List<InjectionContainer<ServiceExtension>> of(List<ServiceExtension> load
                 .collect(Collectors.toList());
     }
 
+    private boolean canResolve(Map<String, List<ServiceExtension>> dependencyMap, String featureName) {
```
ndr-brt, line 133: I already approved and I'm ok with the whole PR, but I noticed that `InjectionPoint.featureName` is always the same as `InjectionPoint.injectedField().getType().getName()` (see the only constructor call in `InjectionPointScanner`), so here the injection point could be called so the `Class.forName` at line 140 can be removed and replaced with `injectionPoint.getType()`

paullatzelsperger: yep, makes sense.

PR #2378, opened 2022-12-22
refactor: improve `Event` hierarchy

### LINKED ISSUE #2375: `Event` hierarchy refactor
# Feature Request

The `Event` hierarchy seems to be overcomplicated, bringing two different hierarchies: one for the `Event` and one for the `Payload`, it could be improved

event framework

cleanup

## Solution Proposal
-

ndr-brt:
## What this PR changes/adds

provides `<entity>Event` base classes with proper `Builder` and `Payload` related a specific entity.

## Why it does that

Tidy up hierarchy tree

## Further notes

- added some missing references in `verify.yaml`

## Linked Issue(s)

Closes #2375

## Checklist


paullatzelsperger (approved):
The improvement itself seems good to me, but I can't help but wonder whether we really need all the `XYZEvent.Payload` marker classes. 
As a further simplification we could reduce the amount of classes, so that we deal just with "entity-type payloads". I.e. instead of a specific payload type per event type, we'd have eg. _just_ a `PolicyDefinitionEvent.Payload` etc., and we use that in all the policy-definition-related events. From what I saw, none of the `Payload` adds any value. WDYT?

jimmarino:
> The improvement itself seems good to me, but I can't help but wonder whether we really need all the `XYZEvent.Payload` marker classes. As a further simplification we could reduce the amount of classes further, so that we deal just with "entity-type payloads". I.e. instead of a specific payload type per event type, we'd have eg. _just_ a `PolicyDefinitionEvent.Payload` etc., and we use that in all the policy-definition-related events. From what I saw, none of the `Payload` adds any value. WDYT?

I was going to say the same thing but @paullatzelsperger beat me to it.

ndr-brt:
@paullatzelsperger @jimmarino currently the payloads are the same across the events of the same entity, but this is not the optimal case, the event should describe something that happened, so it could bring other kind of data, e.g., a `ContractNegotiationFailed` event could have a `reason` field, that shouldn't be part of other `ContractNegotiationEvent`s. This kind of things will emerge as soon as we'll intensify the event work.
I can open an issue to add such fields to specific events, but to be honest I would wait for the new protocol to be in place, in the meantime I think this PR brings some cleanup that could be merged anyway.

paullatzelsperger:
> @paullatzelsperger @jimmarino currently the payloads are the same across the events of the same entity, but this is not the optimal case, the event should describe something that happened, so it could bring other kind of data, e.g., a `ContractNegotiationFailed` event could have a `reason` field, that shouldn't be part of other `ContractNegotiationEvent`s. This kind of things will emerge as soon as we'll intensify the event work.
> 
> I can open an issue to add such fields to specific events, but to be honest I would wait for the new protocol to be in place, in the meantime I think this PR brings some cleanup that could be merged anyway.

I agree but aren't these orthogonal issues then? I.e. simplifying the payload subclasses and introducing another field (s) into the event class?

PR #2381, opened 2022-12-22
fix: Fix `baseUrl` handling in `HttpRequestParams`

### LINKED ISSUE #2367: `HttpRequestParams` does not support base url containing path segments or query parameters
# Bug Report

## Describe the Bug
Current implementation of `HttpRequestParams` assumes that the `baseUrl` passed in input only contains the scheme + host. In some cases a data provider could be interested in also supplying some default path or query parameters into this `baseUrl` which are only for internal usage and should not be exposed to the provider.
 
### Expected Behavior
`HttpRequestParams` should support cases wherein `baseUrl` actually contains path segments and/or query parameters.

### Observed Behavior
`HttpRequestParams` does not support `baseUrl` containing path segments and/or query parameters.

bscholtes1A:
## What this PR changes/adds

Use `OkHttp` API to build the URL in `HttpRequestParams`

## Why it does that

URL was previously built manually, which could lead to issues in case wherein the provided `baseUrl` already contains path segments and/or request params.

## Linked Issue(s)

Closes #2367 

## Checklist


extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpDataSink.java
```diff
@@ -37,23 +38,36 @@ public class HttpDataSink extends ParallelSink {
     @Override
     protected StatusResult<Void> transferParts(List<DataSource.Part> parts) {
         for (DataSource.Part part : parts) {
-            var request = params.toRequest(part::openStream);
-            try (var response = httpClient.execute(request)) {
-                if (!response.isSuccessful()) {
-                    monitor.severe(format("Error {%s: %s} received writing HTTP data %s to endpoint %s for request: %s",
-                            response.code(), response.message(), part.name(), request.url().url(), request));
-                    return ERROR_WRITING_DATA;
+            try (var is = part.openStream()) {
```
bscholtes1A, line 41: Add previously missing try-with-resources here.

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParams.java
```diff
@@ -81,15 +75,29 @@ private RequestBody createRequestBody(@Nullable Supplier<InputStream> bodySuppli
...
+        if (queryParams != null) {
+            parseQueryParams(queryParams).forEach(builder::addQueryParameter);
         }
-        if (queryParams != null && !queryParams.isBlank()) {
-            url += "?" + queryParams;
+
+        return builder.build();
+    }
+
+    private static Map<String, String> parseQueryParams(String s) {
```
paullatzelsperger, line 92: instead of first encoding all query params in a single string outside of the HttpRequestParams, and then parsing them here, couldn't we modify the `HttpRequestParams.Builder` such that it takes the query params as a `Map<String, String>` and add a `queryParam(String key, String value)` method?

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpDataSink.java
```diff
@@ -37,23 +38,36 @@ public class HttpDataSink extends ParallelSink {
...
         return StatusResult.success();
     }
 
+    private StatusResult<Void> transferPart(String name, byte[] bytes) {
+        var request = params.toRequest(bytes);
+        try (var response = httpClient.execute(request)) {
+            if (!response.isSuccessful()) {
+                monitor.severe(format("Error {%s: %s} received writing HTTP data %s to endpoint %s for request: %s",
+                        response.code(), response.message(), name, request.url().url(), request));
+                return ERROR_WRITING_DATA;
```
paullatzelsperger, line 60: may want to attach the `response`'s message/error to the `ERROR_WRITING_DATA` result

paullatzelsperger (approved):

extensions/data-plane/data-plane-http/src/test/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParamsTests.java
```diff
@@ -77,7 +76,7 @@ void verifyHeaders() {
     @Test
     void verifyComplexUrl() {
         var path = "testpath";
-        var queryParams = "test-queryparams";
+        var queryParams = "foo=bar&hello=world";
```
paullatzelsperger, line 79: since we're actually parsing the query params, it may make sense to test the negative case here too, i.e. an invalid queryparams string, e.g. `foo-bar&hello`

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParams.java
bscholtes1A: done

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpDataSink.java
bscholtes1A: Don't know... Do we want to expose the response from the backend REST API targeted by the source?

extensions/data-plane/data-plane-http/src/test/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParamsTests.java
bscholtes1A: Done in `HttpSourceRequestParamsSupplier` in which I moved the query params parsing logic.

ndr-brt (commented):
I think the previous `InputStream` access design was pretty good, why passing to a "read everything in a in-memory bytearray and then write it to output stream"? It would have been just enough to close the stream in `NonChunkedTransferRequestBody` constructor to avoid leaks.

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParams.java
```diff
@@ -32,7 +33,7 @@ public class HttpRequestParams {
     private String method;
     private String baseUrl;
     private String path;
-    private String queryParams;
+    private final Map<String, String> queryParams = new HashMap<>();
```
ndr-brt, line 36: I think the correct signature for this field is `Map<String, List<String>>`, as every param could have multiple occurrences.

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParams.java
```diff
@@ -81,17 +81,18 @@ private RequestBody createRequestBody(@Nullable Supplier<InputStream> bodySuppli
...
+        var builder = parsed.newBuilder();
         if (path != null && !path.isBlank()) {
-            url += "/" + path;
+            builder.addPathSegments(path);
         }
-        if (queryParams != null && !queryParams.isBlank()) {
-            url += "?" + queryParams;
-        }
-        return url;
+        queryParams.forEach(builder::addQueryParameter);
```
ndr-brt, line 91: withouth parsing the query params and then passing them to the builder why don't use the `HttpUrl.Builder.encodedUrl`, that would avoid the parse work on our side.

bscholtes1A: Using `encodedQuery` would override the potential query parameters already provided in the baseUrl.

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParams.java
bscholtes1A: Not sure it's useful, as `HttpUrl.Builder.addQueryParameter` only supports unique value.

ndr-brt: no, it handles query params as a list, so there can be multiple values for the same key: https://github.com/1184893257/okhttp/blob/1149fa6233442eb686d5e8a363e6cf37deb52ed9/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java#L808

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpRequestParams.java
ndr-brt: I think the query params should be either on the `baseUrl` or in the `queryParams` , not in both. At least that's not currently supported.

bscholtes1A: We currently have a use-case in Eona-X wherein some query parameters are internal to the API (for example some API filters with some internal codes that should not be exposed to the consumer). 
On the other hand, the `queryParams` field convey the query parameters provided by the consumer through the proxy. 

So for me we should support both.

bscholtes1A: Could you please check if latest commit fits better? https://github.com/eclipse-edc/Connector/pull/2381/commits/977e06943da7b3cf975f7b6189f4311ba596ec2d

ndr-brt (approved):

PR #2494, opened 2023-02-17
fix: catalog pagination

### LINKED ISSUE #2493: Offers catalog pagination not working as expected
# Bug Report

## Describe the Bug
This bug relates to #2008.
The fix applied in #2018 didn't resolve the issue completely.
When requesting the catalog using `limit` and `offset` queryparams, the number of contractOffers returned does not correspond to the expected.

## Steps to Reproduce
Test scenario:
 - 2 Contract definitions with asset selector criteria refering to 50 assets each (Expected total of 100 offers).
 - Consumer requests the catalog from provider using different `limit` and `offset`.

### Expected Behavior
**Behaviour a)**
limit=52
offset=0
expected offers: 52

**Behaviour b)**
limit=2
offset=49
expected offers: 2

### Observed Behavior
**Behaviour a)**
observed offers: 100

**behaviour b)**
observed offers: 3

### Used version
0.0.1-20230213-SNAPSHOT

## Possible Implementation
On `ContractOfferResolverImpl.queryContractOffers()` when `skip` is a value under 0, `limit` for next iteration should be the result of the current `limit` minus the `skip`.

yurimssilva:
## What this PR changes/adds

- Adds test cases to evaluate catalog pagination across multiple contract definitions.
- Changes the logic which dynamically builds the offset and limit for each contract definition.

## Why it does that

  Previously the number of assets that should be skipped in the current iteration was dynamic but the limit was not, which caused the pagination to fail when across multiple contract definitions.
  I decided to use a different approach that for me was easier to understand.
  Based on the given range, the absolute offset and limit are static but could assume a different state on each iteration responding to the number of seen and fetched assets until that point. Therefore, I implemented the logic that sets the values of the dynamic offset and limit variables in the following expressions:

- The dynamic offset must be equal to the absolute offset minus the number of assets seen, but never less than zero.
- The dynamic limit must be equal to the absolute limit minus the number of assets already fetched, but never exceed the number of assets obtained in the current contract definition.

## Linked Issue(s)

Closes #2493

## Checklist

### DIFF
--- a/core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
+++ b/core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
@@ -36,9 +36,11 @@ import org.eclipse.edc.spi.types.domain.asset.Asset;
 import org.eclipse.edc.spi.types.domain.asset.AssetEntry;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
-import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+import org.junit.jupiter.params.provider.ArgumentsSource;
 
 import java.time.Clock;
 import java.time.Instant;
@@ -112,20 +114,9 @@ class ContractOfferResolverImplIntegrationTest {
         verify(policyStore).findById("contract");
     }
 
-    private static Stream<Arguments> ranges() {
-        return Stream.of(
-                Arguments.of(0, 12),
-                Arguments.of(8, 14),
-                Arguments.of(0, 999),
-                Arguments.of(4, 888),
-                Arguments.of(3, 20),
-                Arguments.of(23, 25)
-        );
-    }
-
     @ParameterizedTest
-    @MethodSource("ranges")
-    void shouldReturnOfferSubsetAcrossMultipleContractDefinitions(int from, int to) {
+    @ArgumentsSource(RangeProvider.class)
+    void should_return_offers_subset_when_across_multiple_contract_definitions(int from, int to) {
 
         var assets1 = range(0, 10).mapToObj(i -> createAsset("asset-" + i).build()).collect(Collectors.toList());
         var assets2 = range(10, 20).mapToObj(i -> createAsset("asset-" + i).build()).collect(Collectors.toList());
@@ -219,4 +210,17 @@ class ContractOfferResolverImplIntegrationTest {
         return Asset.Builder.newInstance().id(id).name("test asset " + id);
     }
 
+    static class RangeProvider implements ArgumentsProvider {
+        @Override
+        public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
+            return Stream.of(
+                    Arguments.of(0, 12),
+                    Arguments.of(8, 14),
+                    Arguments.of(0, 999),
+                    Arguments.of(4, 888),
+                    Arguments.of(3, 20),
+                    Arguments.of(23, 25)
+            );
+        }
+    }
 }
--- a/core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
+++ b/core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
@@ -36,6 +36,9 @@ import org.eclipse.edc.spi.types.domain.asset.Asset;
 import org.eclipse.edc.spi.types.domain.asset.AssetEntry;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 import java.time.Clock;
 import java.time.Instant;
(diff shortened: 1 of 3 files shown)


core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
```diff
@@ -107,6 +102,35 @@ void shouldLimitResult_withHeterogenousChunks() {
...
+
+        var from = 0;
+        var to = 12;
+        var query = ContractOfferQuery.builder().range(new Range(from, to)).claimToken(ClaimToken.Builder.newInstance().build()).build();
+        assertThat(contractOfferResolver.queryContractOffers(query)).hasSize(12);
+
+        from = 8;
+        to = 14;
+        query = ContractOfferQuery.builder().range(new Range(from, to)).claimToken(ClaimToken.Builder.newInstance().build()).build();
+        assertThat(contractOfferResolver.queryContractOffers(query)).hasSize(6);
```
paullatzelsperger, line 136: nit: may want to use variables here, i.e. `to - from`

core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
```diff
@@ -107,6 +107,35 @@ void shouldLimitResult_withHeterogenousChunks() {
...
+
+        var contractDefinition1 = getContractDefBuilder("contract-definition-")
+                .selectorExpression(selectorFrom(assets1)).build();
+        var contractDefinition2 = getContractDefBuilder("contract-definition-")
+                .selectorExpression(selectorFrom(assets2)).build();
+
+        when(agentService.createFor(isA(ClaimToken.class))).thenReturn(new ParticipantAgent(emptyMap(), emptyMap()));
+        when(contractDefinitionService.definitionsFor(isA(ParticipantAgent.class))).thenAnswer(i -> Stream.of(contractDefinition1, contractDefinition2));
+        when(policyStore.findById(any())).thenReturn(PolicyDefinition.Builder.newInstance().policy(Policy.Builder.newInstance().build()).build());
+
```
paullatzelsperger, line 127: those two assertions seem to test the same thing, i.e. getting contract offers that "span" across assets that are referred to by two contract definitions. Therefor one of them could be removed. However, we should actually test edge cases too:
- `0...12`
- `0...9999`: tests a range overshoot, expect 20 offers
- `4...9999`: test a range overshoot starting at an offset, expect 16 offers
- `3...20`: test exact upper bound with offset, expect 17 offers

At that point you may want to consider a `@ParameterizedTest` to avoid test code duplication.

ndr-brt (approved):

paullatzelsperger (changes requested):
Thanks for the contribution @yurimssilva. Please try to improve your PR description so that people can understand _what_ exactly you changed:
> Changes the logic which dynamically builds the offset and limit for each contract definition

yeah, but _how_ did you change the logic? what was the bug you fixed? Reviewers should not be expected to figure this out on their own. Some more comments inline.

Also Unit tests and checkstyle are failing.

core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
```diff
@@ -50,12 +50,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.isA;
-import static org.mockito.Mockito.atLeastOnce;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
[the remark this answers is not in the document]
yurimssilva: ty, fixed

core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
yurimssilva: ty, fixed.

core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
yurimssilva: thanks, @paullatzelsperger.. doing it.

core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
```diff
@@ -107,6 +112,42 @@ void shouldLimitResult_withHeterogenousChunks() {
...
+                Arguments.of(8, 14),
+                Arguments.of(0, 999),
+                Arguments.of(4, 888),
+                Arguments.of(3, 20),
+                Arguments.of(23, 25)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("ranges")
```
paullatzelsperger, line 127: I'd rather you use `@CsvSource`, because then at least the test values are immediately visible, or `@ArgumentSource`, because that uses a strongly typed provider rather thant relying solely on the method name.
Also, the method naming should follow the `should_when_then` pattern, separated by an underscore, which increases legibility quite a bit.

core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
paullatzelsperger: we're getting there :) check updated comments.

core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImplIntegrationTest.java
yurimssilva: thx, Paul.. it's fixed.

paullatzelsperger (approved):

PR #2511, opened 2023-02-22
feat(protocol): adapt `TransferProcess` states to Dataspace Protocol

### LINKED ISSUE #2491: map `Transfer Process` status to new IDS Specification
# Feature Request

The upcoming IDS specification defines slightly different status from the current ones.
Let's adapt to the new status without breaking up the current implementation.
More details in https://github.com/eclipse-edc/Connector/pull/2448

transfer process

new IDS specification

## Solution Proposal

ndr-brt:
## What this PR changes/adds

Adapt `TransferProcess` states to the new protocol as described in https://github.com/eclipse-edc/Connector/pull/2489

- `UNSAVED` got deprecated
- `IN_PROGRESS` and `STREAMING` got replaced by `STARTED`
- `CANCELLED`, `ENDED` and `ERROR` got replaced by `TERMINATED` 

## Why it does that

Dataspace Protocol implementation

## Further notes
- states, events and listener methods that are referring to old states have been not deleted but deprecated, to permit a smoother transition
- migrating all the error states to terminated brought in some doubts about the allowed states to be migrated to terminated. Will be addressed in another issue.

## Linked Issue(s)

Closes #2491 

## Checklist


paullatzelsperger:
removing myself, as @jimmarino and @juliapampus are probably more knowledgeable anyway

jimmarino (approved):
LGTM. Question: do we need to deprecate the states or can we just remove them? For example, if no "end-user" APIs reference the states, and argument could be made to remove them. Otherwise, we would have to map them to their Protocol Spec counterparts.

ndr-brt:
> LGTM. Question: do we need to deprecate the states or can we just remove them? For example, if no "end-user" APIs reference the states, and argument could be made to remove them. Otherwise, we would have to map them to their Protocol Spec counterparts.

I deprecated them to give them the possibility to handle the cleanup of the TP in those states after the update to the version with the new ones, but, in fact, nothing will be broken, at worst they would have "ghost" TP that they'd need to cleanup, I will go with removing them

PR #2516, opened 2023-02-23
feat: update asset and dataaddress

### LINKED ISSUE #2508: Update of `Asset` + `DataAddress`

paullatzelsperger:
## What this PR changes/adds

Adds the `UPDATE` endpoint to the `AssetApi`, in particular the following parts are added:
- add two `PUT` endpoints to the `AssetApiController`
- adds new `update()` methods to the `AssetIndex`
- adds implementations for `SqlAssetIndex`, `CosmosAssetIndex` and `InMemoryAssetIndex`
- adds tests for all these impls

## Why it does that

to enable update semantics on assets and dataaddresses

## Further notes

- I created a dedicated `AssetUpdateDto`, that doesn't have an ID, just to explicitly exclude the ID. The `assetId` is passed as path param.
- updating `Assets` will fail if the `Asset` isn't found, because it would be an `Asset` without a `DataAddress`, which is invalid.
- similarly, updating `DataAddresses` for `Assets` that don't exist will fail as well

_As an alternative, we could always update `Asset` and `DataAddress` together, but for that we'd also have to add the `DataAddress` to the `GET` endpoint_.

## Linked Issue(s)

Closes #2508 

## Checklist

### DIFF
--- a/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
+++ b/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
@@ -109,4 +109,32 @@ public class AssetServiceImpl implements AssetService {
             return ServiceResult.success(deleted);
         });
     }
+
+    @Override
+    public ServiceResult<Void> update(String assetId, Asset asset) {
+        return transactionContext.execute(() -> {
+            if (findById(assetId) == null) {
+                return ServiceResult.notFound(format("Asset %s cannot be updated because it does not exist", assetId));
+            }
+            var updatedAsset = index.updateAsset(assetId, asset);
+            observable.invokeForEach(l -> l.updated(updatedAsset));
+
+            return ServiceResult.success(null);
+        });
+    }
+
+    @Override
+    public ServiceResult<Void> update(String assetId, DataAddress dataAddress) {
+        return transactionContext.execute(() -> {
+            var asset = findById(assetId);
+            if (asset == null) {
+                return ServiceResult.notFound(format("DataAddress for Asset ID= %s cannot be updated because it does not exist", assetId));
+            }
+
+            index.updateDataAddress(assetId, dataAddress);
+            observable.invokeForEach(l -> l.updated(asset));
+
+            return ServiceResult.success(null);
+        });
+    }
 }
--- a/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
+++ b/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
@@ -28,6 +28,7 @@ import org.eclipse.edc.spi.types.domain.asset.Asset;
 import org.eclipse.edc.transaction.spi.TransactionContext;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.stream.Stream;
 
 import static java.lang.String.format;
@@ -112,6 +113,9 @@ public class AssetServiceImpl implements AssetService {
 
     @Override
     public ServiceResult<Void> update(String assetId, Asset asset) {
+        if (!Objects.equals(assetId, asset.getId())) {
+            return ServiceResult.badRequest("Asset.getId() must match assetId");
+        }
         return transactionContext.execute(() -> {
             if (findById(assetId) == null) {
                 return ServiceResult.notFound(format("Asset %s cannot be updated because it does not exist", assetId));
--- a/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
+++ b/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
@@ -119,7 +119,7 @@ public class AssetServiceImpl implements AssetService {
             var updatedAsset = index.updateAsset(assetId, asset);
             observable.invokeForEach(l -> l.updated(updatedAsset));
 
-            return ServiceResult.success(null);
+            return ServiceResult.success();
         });
     }
 
@@ -134,7 +134,7 @@ public class AssetServiceImpl implements AssetService {
             index.updateDataAddress(assetId, dataAddress);
             observable.invokeForEach(l -> l.updated(asset));
 
-            return ServiceResult.success(null);
+            return ServiceResult.success();
         });
     }
 }
--- a/extensions/control-plane/api/management-api/asset-api/src/main/java/org/eclipse/edc/connector/api/management/asset/model/AssetUpdateDto.java
+++ b/extensions/control-plane/api/management-api/asset-api/src/main/java/org/eclipse/edc/connector/api/management/asset/model/AssetUpdateDto.java
@@ -1,0 +1,69 @@
+/*
+ *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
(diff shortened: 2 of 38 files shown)


core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
```diff
@@ -109,4 +109,32 @@ public ServiceResult<Asset> delete(String assetId) {
...
+    @Override
+    public ServiceResult<Void> update(String assetId, Asset asset) {
+        return transactionContext.execute(() -> {
+            if (findById(assetId) == null) {
+                return ServiceResult.notFound(format("Asset %s cannot be updated because it does not exist", assetId));
+            }
+            var updatedAsset = index.updateAsset(assetId, asset);
+            observable.invokeForEach(l -> l.updated(updatedAsset));
+
+            return ServiceResult.success(null);
```
ndr-brt, line 122: nit: we could add a `success()` method on `ServiceResult` like it was done on `Result` to hide the `null` thing

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
```diff
@@ -109,4 +109,32 @@ public ServiceResult<Asset> delete(String assetId) {
             return ServiceResult.success(deleted);
         });
     }
+
+    @Override
+    public ServiceResult<Void> update(String assetId, Asset asset) {
+        return transactionContext.execute(() -> {
+            if (findById(assetId) == null) {
```
ndr-brt, line 116: I would add a validation here to ensure that the `assetId` parameter is equal as `asset.getId()`, to avoid the possibility to update the id.

extensions/control-plane/api/management-api/asset-api/src/main/java/org/eclipse/edc/connector/api/management/asset/model/AssetUpdateDto.java
```diff
@@ -0,0 +1,69 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import jakarta.validation.constraints.AssertTrue;
+import jakarta.validation.constraints.NotNull;
+
+import java.util.Map;
+
+@JsonDeserialize(builder = AssetUpdateDto.Builder.class)
+public class AssetUpdateDto {
```
ndr-brt, line 27: Couldn't this be used as superclass for `AssetRequestDto`? Maybe it's just useless hierarchy, but it would remove some duplication

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
paullatzelsperger: good point. will do

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
paullatzelsperger: that validation will likely fail, because the `AssetUpdateDto` doesn't (on purpose!) have an `id`, so it will be transformed into an `Asset` that does not have an `id` either. That's the reason I tunneled the `assetId` in separately. To me, having the ID separate makes it more explicit.

extensions/control-plane/api/management-api/asset-api/src/main/java/org/eclipse/edc/connector/api/management/asset/model/AssetUpdateDto.java
paullatzelsperger: so, we'd have to have an `AssetDto` (superclass, just has the `Map`), and derived from that the `AssetRequestDto`, that adds the `id`, and potentially also an `AssetUpdateDto`, that is just a marker and adds nothing. The latter is not needed, strictly speaking. WDYT?

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
ndr-brt: ok, but looking at the `Asset.Builder`:
```
 @Override
        public Asset build() {
            if (entity.getId() == null) {
                id(UUID.randomUUID().toString());
            }
            return super.build();
        }
```

so this Asset will have a randomly generated id, also if that doesn't count because it won't be updated I think it would be better to have it valued correctly and then checked anyway because the service shouldn't rely on controller and store behavior (e.g. in a document db I would expect just a "set the whole object") behavior, and if the id is not correct that will be an issue.

extensions/control-plane/api/management-api/asset-api/src/main/java/org/eclipse/edc/connector/api/management/asset/model/AssetUpdateDto.java
ndr-brt: `AssetDto` does not exist... otherwise it could have been the right object to do, in fact now there's no hierarchy on `AssetRequestDto`, in fact it represent a creation request, that at the end is an update request plus the id.
Could be weird but maybe it's just a naming issue...

`AssetCreateRequestDto` that's subclass of
`AssetUpdateRequestDto`

sounds better to me

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/asset/AssetServiceImpl.java
paullatzelsperger: ok, so we either need a way to set the `id` after the transformer has created the asset, e.g. with a `Asset.toBuilder` method, or by not using the transformer at all, and "manually" create the asset object.

extensions/control-plane/api/management-api/asset-api/src/main/java/org/eclipse/edc/connector/api/management/asset/model/AssetUpdateDto.java
paullatzelsperger: > AssetDto does not exist...
i know, i meant that would probably be the base class's name :)

Let me try and refactor these DTOs a bit, we'll see how it turns out

paullatzelsperger: check https://github.com/eclipse-edc/Connector/pull/2516/commits/9b352742ec353f9fd4b9d7c2ed062b2e5e16c710

ndr-brt: @paullatzelsperger looks good!

paullatzelsperger:
> The id check is still missing, currently with the cosmos db implementation the id could be changed

there is still an open question about that. I agree that having an Asset with a random ID is not acceptable. We could have a `toBuilder` method, that allows setting the ID after the transformer runs, or we don't use the transformer here at all. WDYT?

extensions/control-plane/api/management-api/asset-api/src/main/java/org/eclipse/edc/connector/api/management/asset/model/AssetUpdateRequestWrapperDto.java
```diff
@@ -0,0 +1,60 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.connector.api.management.asset.model;
+
+public class AssetUpdateRequestWrapperDto extends AssetRequestDto {
```
ndr-brt, line 17: does it need to extend the `AssetRequestDto`?

paullatzelsperger: no, copy/paste...

ndr-brt (approved):

PR #2518, opened 2023-02-23
feat: add update endpoint for ContractDefinition

### LINKED ISSUE #2510: Update of `ContractDefinition`

wolf4ood:
## What this PR changes/adds

Add the `UPDATE` endpoint to the `ContractDefinitionApi`

## Why it does that

to enable update semantics on contract definitions

## Further notes

- I created `ContractDefinitionUpdateDto` for updating the contract definition (same as ContractDefinitionRequestDto without the id)
- I created `ContractDefinitionUpdateDtoWrapper` which is a composition of `ContractDefinitionUpdateDto` + id to update
- The `ContractDefinitionUpdateDtoWrapper` is then transformed through the transformer registry into the final `ContractDefinition` to update

## Linked Issue(s)

Closes #2510 

## Checklist

### DIFF
--- a/extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/model/ContractDefinitionUpdateDtoWrapper.java
+++ b/extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/model/ContractDefinitionUpdateDtoWrapper.java
@@ -1,0 +1,65 @@
+/*
+ *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.connector.api.management.contractdefinition.model;
+
+import java.util.Objects;
+
+public class ContractDefinitionUpdateDtoWrapper {
+
+
+    private ContractDefinitionUpdateDto contractDefinition;
+    private String id;
+
+    private ContractDefinitionUpdateDtoWrapper() {
+    }
+
+    public String getId() {
+        return id;
+    }
+
+    public ContractDefinitionUpdateDto getContractDefinition() {
+        return contractDefinition;
+    }
+
+
+    public static final class Builder {
+        private final ContractDefinitionUpdateDtoWrapper wrapper;
+
+        private Builder() {
+            wrapper = new ContractDefinitionUpdateDtoWrapper();
+        }
+
+        public static Builder newInstance() {
+            return new Builder();
+        }
+
+        public Builder id(String id) {
+            wrapper.id = id;
+            return this;
+        }
+
+        public Builder contractDefinition(ContractDefinitionUpdateDto contractDefinition) {
+            wrapper.contractDefinition = contractDefinition;
+            return this;
+        }
+
+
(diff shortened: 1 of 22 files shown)


extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/model/ContractDefinitionUpdateDtoWrapper.java
```diff
@@ -0,0 +1,65 @@
...
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.connector.api.management.contractdefinition.model;
+
+import java.util.Objects;
+
+public class ContractDefinitionUpdateDtoWrapper {
```
paullatzelsperger, line 19: i like this idea, in fact I might even adopt this into the Asset/DataAddress subfeature, but since the sole purpose of this class is to wrap the dto and the ID, personally I would make it package private, or maybe even an inner class of the `ContractDefinitionApiController`, if possible. In case a smidge of documentation wouldn't hurt :)

extensions/control-plane/api/management-api/contract-definition-api/src/test/java/org/eclipse/edc/connector/api/management/contractdefinition/ContractDefinitionApiControllerIntegrationTest.java
```diff
@@ -220,15 +221,55 @@ void deleteContractDefinition_notExists() {
...
                 .accessPolicyId(UUID.randomUUID().toString())
                 .criteria(List.of(CriterionDto.Builder.newInstance().operandLeft("left").operator("=").operandRight("right").build()))
                 .build();
     }
 
+    private ContractDefinitionUpdateDto updateDto() {
+        return ContractDefinitionUpdateDto.Builder.newInstance()
+                .contractPolicyId(UUID.randomUUID().toString())
+                .accessPolicyId(UUID.randomUUID().toString())
+                .criteria(List.of(CriterionDto.Builder.newInstance().operandLeft("left").operator("=").operandRight("right").build()))
```
paullatzelsperger, line 269: i know it may be nit-picking, but updating the criteria to something entirely different might highlight the test case better.

extensions/control-plane/api/management-api/contract-definition-api/src/test/java/org/eclipse/edc/connector/api/management/contractdefinition/ContractDefinitionApiControllerIntegrationTest.java
```diff
@@ -220,15 +221,55 @@ void deleteContractDefinition_notExists() {
...
     }
 
-    private ContractDefinitionRequestDto createDto(String definitionId) {
-        return ContractDefinitionRequestDto.Builder.newInstance()
+    @Test
+    void updateContractDefinition_whenExists(ContractDefinitionStore store) {
+
+        store.accept(createContractDefinition("definitionId"));
+
+        var dto = updateDto();
```
paullatzelsperger, line 229: With the current test setup it's possible that the test is green, but no actual update has happened.
I'd suggest using the original DTO, then updating it (e.g. the `AssetSelectorExpression`), and then asserting that the updated DTO is indeed in the `store`.

paullatzelsperger (commented):
There should be tests for the DTOs as well, i.e. validation tests and SerDes tests. Other than that, and a few inline comments, it seems fine

extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/model/ContractDefinitionUpdateDtoWrapper.java
ndr-brt: I like this too

ndr-brt (approved):

paullatzelsperger (approved):

PR #2546, opened 2023-03-03
feat: make `CompleteTransferCommand` transit TP to `COMPLETING`

### LINKED ISSUE #2543: TransferProcess: make `complete` command follow the correct flow
# Feature Request

Currently the `complete` method of the `TransferProcessService` puts the TP in `COMPLETED` state, but doing like this it won't send the `COMPLETED` protocol message to the counter part.
To have this process in place, we'd need to put the TP in `COMPLETING` state.

transfer process

dataspace protocol

## Solution Proposal

ndr-brt:
## What this PR changes/adds

Makes `TransferProcessService.complete` pass through `COMPLETING` state

## Why it does that

To permit upcoming notification to the counter part with the `COMPLETED` protocol message

## Further notes

- Removed `FailTransferCommand` as it was a duplicate of `TerminateTransferCommand`
- make "notify started" logic pass through commands
- add `can*` methods on `TransferProcess` so that the service could avoid fake-transit the TP to a certain state and catching eventual exceptions to know if the transition is possible or not.

## Linked Issue(s)

Closes #2543 

## Checklist


core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/command/handlers/NotifyStartedTransferCommandHandler.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.eclipse.edc.connector.transfer.spi.observe.TransferProcessObservable;
+import org.eclipse.edc.connector.transfer.spi.store.TransferProcessStore;
+import org.eclipse.edc.connector.transfer.spi.types.TransferProcess;
+import org.eclipse.edc.connector.transfer.spi.types.command.NotifyStartedTransferCommand;
+
+import static org.eclipse.edc.connector.transfer.spi.types.TransferProcess.Type.CONSUMER;
+
+/**
+ * Puts a TransferProcess in the STARTED state as the counter-party actually started the transfer.
```
paullatzelsperger, line 25: small nit: To me the class name sounds a bit misleading: `NotifyStarted...` implies, that "I am notifying" someone. While, what actually happens, is that "I am notified, that something has happened". How about naming the class `CounterPartyStarted...` or `TransferHasStarted...`?

paullatzelsperger (approved):

core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/command/handlers/NotifyStartedTransferCommandHandler.java
ndr-brt: @paullatzelsperger the idea was to have the command conjugated in the present tense, like a direction coming from the outside that tells something that must happen but it's not yet happened, yes, the provider will emit the `TransferStartedEvent` and that will cause on the provider the creation of a command that would be end in an action, in this case I think it could be treated as a notification because the only thing the consumer can do is to accept it and change the status.
Putting it in the past tense would create confusion between commands and events.

PR #2547, opened 2023-03-03
refactor: rename and move ServiceResult

### LINKED ISSUE #2545: Refactor `ServiceResult`
In order to be usable in entity store implementations, the `ServiceResult` should be refactored in two major aspects:
- the name, something like `RequestResult` to be universally applicable
- the module: it needs to be moved into the `core-spi` module

Technically, we could leave the `ServiceResult` in place where it is, and have it extend the `RequestResult`, but it is questionable whether that offers much value, since the impact on the code base would likely be equally huge.

paullatzelsperger:
## What this PR changes/adds

- Renames `ServiceResult` -> `RequestResult` and `ServiceFailure` -> `RequestFailure`
- Moves the `RequestResult` and `RequestFailure` to the `core-spi` module
- deleted the now-empty `aggregate-service-spi` module and references to it

_this is just renaming and moving, no functional change was performed_

## Why it does that

Make these two classes accessible to more core features like entity stores

## Further notes

- I also uncommented a line in the root `build.gradle.kts`, which I assume slipped in by mistake

## Linked Issue(s)

Closes #2545 

## Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -46,7 +46,7 @@ allprojects {
 
     // configure which version of the annotation processor to use. defaults to the same version as the plugin
     configure<org.eclipse.edc.plugins.autodoc.AutodocExtension> {
-//        processorVersion.set(annotationProcessorVersion)
+        processorVersion.set(annotationProcessorVersion)
         outputDirectory.set(project.buildDir)
     }
 
--- a/spi/common/aggregate-service-spi/src/main/java/org/eclipse/edc/service/spi/result/ServiceResult.java
+++ b/spi/common/aggregate-service-spi/src/main/java/org/eclipse/edc/service/spi/result/ServiceResult.java
@@ -15,6 +15,7 @@
 package org.eclipse.edc.service.spi.result;
 
 import org.eclipse.edc.spi.result.AbstractResult;
+import org.eclipse.edc.spi.result.StoreResult;
 
 import java.util.List;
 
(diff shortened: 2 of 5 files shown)


wolf4ood (approved):

ronjaquensel (approved):

ndr-brt (commented):
I'm not really sure this would be the best thing to do.
Personally I would go with a `StoreResult` or `UpdateStoreResult` with its own `Failure`.

I would do that because it would lead to better separation of concerns, but also because the currently named `ServiceFailure` has 3 reasons to happen: `NOT FOUND`, `CONFLICT` and `BAD REQUEST` that are really strictly related to the Service layer, while, for example, in a `Store` a "Bad Request" in fact couldn't happen, while it could fail for a connection issue (that currently are handled by throwing runtime exceptions, but they could in fact being described by a dedicated `Failure`, specific to the `Store` layer),

paullatzelsperger:
> I'm not really sure this would be the best thing to do. Personally I would go with a `StoreResult` or `UpdateStoreResult` with its own `Failure`.
> 
> I would do that because it would lead to better separation of concerns, but also because the currently named `ServiceFailure` has 3 reasons to happen: `NOT FOUND`, `CONFLICT` and `BAD REQUEST` that are really strictly related to the Service layer, while, for example, in a `Store` a "Bad Request" in fact couldn't happen, while it could fail for a connection issue (that currently are handled by throwing runtime exceptions, but they could in fact being described by a dedicated `Failure`, specific to the `Store` layer),

Mmh, I see your point. I was a bit concerned about mapping different results into one another, because they probably should not have an inheritance relation to one another. This was really the first attempt at this, let me try and work in your suggestion.

paullatzelsperger:
@ndr-brt I changed the PR significantly, so now we have a `StoreResult` and `StoreFailure`. The latter has constants for `NOT_FOUND` and `ALREADY_EXISTS`. 
Actually using them in stores will be done in subsequent PRs, see parent issue. Also, technical errors, such as SQL connection errors, should continue to be communicated via exceptions.

ndr-brt (approved):

Issue #2539, opened 2023-03-02
Refactor `TransferProcessStore` (only save)


paullatzelsperger:
The `TransferProcessStore` does not user-generated entities, so for now it keeps its "upsert" semantics. 
However, due to changes to e.g. the `CosmosDbApi` (#2554) the upsert at the store level must now be done explicitly.

PR #2554, opened 2023-03-06
refactor: add create update to cosmos api

### LINKED ISSUE #2531: Make data stores return `Result` on modifying requests
# Feature Request

Currently, many of the stores return either the modified entity, or `null` if the update failed (e.g. object not found).
This should be changed so that the stores return a `Result` (or an appropriate subclass).

_this issue requires a decision-record beforehand_

various *stores and the calling services. 

Make success or failure more explicit than just nullity, clearly delineate `CREATE` and `UPDATE` requests

## Solution Proposal
We should do this together with the the "upsert" semantic issue (#2532), 

As part of this, we should also get rid of spurious `upsert` semantics that some store implementations have as described in the now-closed #2532. Because then we can clearly identify `save()` and `update()` operations with clearly defined success- and failure cases. 

So three change areas emerge:
- remove `upsert` semantics
- make modifying methods on stores return a `Result`
- remove the explicit object ID on `update()` methods

## Subtasks
- [x] #2535
- [x] #2545 **do this first!**
- [x] #2554 
- [x] #2536
- [x] #2537
- [x] #2538
- [x] #2539
- [x] #2556
- [x] #2573

paullatzelsperger:
## What this PR changes/adds

Replaces the `saveItem` method of the `CosmosDbApi` with a dedicated `create()` and `update()` method, both of which in turn call the `createItem` and `replaceItem` methods of the Cosmos SDK.

## Why it does that

Preparatory work for #2531 

## Further notes

- Any exceptions coming from the CosmosSDK are not wrapped in an `EdcException` anymore, the calling store implementations must handle a `NotFoundException` and a `ConflictException`
- Subsequent work (#2536 #2537 #2538 #2539) will add code to handle the exceptions and convert them into `StoreResult`s

## Linked Issue(s)

Closes #2531 

## Checklist

### DIFF
--- a/spi/common/aggregate-service-spi/src/main/java/org/eclipse/edc/service/spi/result/ServiceResult.java
+++ b/spi/common/aggregate-service-spi/src/main/java/org/eclipse/edc/service/spi/result/ServiceResult.java
@@ -57,7 +57,20 @@ public class ServiceResult<T> extends AbstractResult<T, ServiceFailure> {
         if (storeResult.succeeded()) {
             return success(storeResult.getContent());
         }
+        switch (storeResult.reason()) {
+            case NOT_FOUND:
+                return notFound(storeResult.getFailureDetail());
+            case ALREADY_EXISTS:
+                return conflict(storeResult.getFailureDetail());
+            default:
+                return badRequest(storeResult.getFailureDetail());
+        }
+    }
 
+    public static <T> ServiceResult<T> fromFailure(StoreResult<?> storeResult) {
+        if (storeResult.succeeded()) {
+            throw new IllegalArgumentException("Can only use fromFailure() when the argument is a failed result. Please use ServiceResult.from() to also convert success results!");
+        }
         switch (storeResult.reason()) {
             case NOT_FOUND:
                 return notFound(storeResult.getFailureDetail());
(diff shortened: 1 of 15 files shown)


wolf4ood (approved):

spi/common/aggregate-service-spi/src/main/java/org/eclipse/edc/service/spi/result/ServiceResult.java
```diff
@@ -57,7 +57,20 @@ public static <T> ServiceResult<T> from(StoreResult<T> storeResult) {
...
+            case ALREADY_EXISTS:
+                return conflict(storeResult.getFailureDetail());
+            default:
+                return badRequest(storeResult.getFailureDetail());
+        }
+    }
 
+    public static <T> ServiceResult<T> fromFailure(StoreResult<?> storeResult) {
+        if (storeResult.succeeded()) {
+            throw new IllegalArgumentException("Can only use fromFailure() when the argument is a failed result. Please use ServiceResult.from() to also convert success results!");
```
jimmarino, line 72: Can you not reference fromFailure() b/c of potential refactoring issues, e.g. spell it out or something else.

jimmarino (approved):

PR #2572, opened 2023-03-09
refactor: replace `provisionedResourceSet` with `provisionedResources`

### LINKED ISSUE #2565: Refactor `TransferProcess` `provisionedResourceSet`
# Feature Request

Currently the `provisionedResourceSet` is described by a `ProvisionedResourceSet` that is an object wrapping a list. For matter of simplicity it could be modeled as we did for `deprovisionedResources`, so a `List<ProvisionedResource>`

transfer process

cleanup

## Solution Proposal
add a new `provisionedResources` field and deprecate `provisionedResourceSet`, that will be removed in the future.

ndr-brt:
## What this PR changes/adds

Replaces `TransferProcess` `provisionedResourceSet` with `provisionedResources`

## Why it does that

Avoid unnecessary classes, tidy up domain model class.

## Further notes

- I left the deprecated field + getter + builder method to avoid breaking changes within store implementations. The way it's handled is:
  - sql: on select tries to deserialize `provisionedResources`, if it's not able, deserialize `provisionedResourceSet` but loads the content into `provisionedResources` so at the first update the `provisionedResources` value will be stored
  - cosmosdb: on select passes through `provisionedResourceSet()` builder method, and fill up the `provisionedResources` field as well, so it will be stored at the first update.  

## Linked Issue(s)

Closes #2565

## Checklist


jimmarino (approved):

paullatzelsperger (approved):

ndr-brt:
Actually I'm stuck with some failed tests for the sql implementation of `TransferProcess`, unfortunately the serialization of `List<*>` types does not work correctly because of type erasure, and that's by jackson design (https://stackoverflow.com/a/34194367), I will try to get around this, but otherwise I will likely close the issue, as likely the `ProvisionedResourceSet` was introduced for the same reason.

ndr-brt:
> Actually I'm stuck with some failed tests for the sql implementation of `TransferProcess`, unfortunately the serialization of `List<*>` types does not work correctly because of type erasure, and that's by jackson design (https://stackoverflow.com/a/34194367), I will try to get around this, but otherwise I will likely close the PR, as it could be that `ProvisionedResourceSet` was introduced for the same reason.

close as this seems not to be fixable currently

PR #2576, opened 2023-03-11
fix: repair cosmos tests and simplify GH action check

paullatzelsperger:
## What this PR changes/adds

This PR does two things:
- fixes some broken CosmosDB tests
- simplifies the check in `verify.yaml` whether or not to actually run the Cosmos tests.

## Why it does that

Some test were simply broken due to misconfiguration, e.g. the `EndToEndTransferCosmosDbTest` used the same Cosmos database and containers for both the `Provider` and the `Consumer`, causing them to overwrite each other's entries. 

Other tests were broken due to the recent distinction between `createItem()` and `updateItem()`, replacing the previous `upsert()`, which caused exceptions and assertion errors.

As for the check in the `verify.yaml`: it apparently didn't work (or stopped working at some point in time), causing the Cosmos tests to not have been executed.

Now, the check is simplified to `if: env.COSMOS_KEY`, checking if the `COSMOS_KEY` secret is present or not. 

## Further notes

- successful run in my fork: https://github.com/paullatzelsperger/DataSpaceConnector/actions/runs/4397174102/jobs/7700131928

## Linked Issue(s)

.

## Checklist

### DIFF
--- a/system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/EndToEndTransferCosmosDbTest.java
+++ b/system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/EndToEndTransferCosmosDbTest.java
@@ -33,7 +33,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 @AzureCosmosDbIntegrationTest
 class EndToEndTransferCosmosDbTest extends AbstractEndToEndTransfer {
 
-    public static final String E2E_TEST_NAME = "e2e-transfer-test-" + UUID.randomUUID();
+    private static final String E2E_TEST_NAME = "e2e-transfer-test-" + UUID.randomUUID();
     @RegisterExtension
     static EdcRuntimeExtension consumerControlPlane = new EdcRuntimeExtension(
             ":system-tests:e2e-transfer-test:control-plane-cosmosdb",
--- a/system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/EndToEndTransferCosmosDbTest.java
+++ b/system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/EndToEndTransferCosmosDbTest.java
@@ -25,28 +25,27 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
 import java.util.HashMap;
+import java.util.UUID;
 import java.util.stream.Stream;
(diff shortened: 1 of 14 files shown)


jimmarino (approved):

jimmarino (approved):

system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/Participant.java
```diff
@@ -43,8 +43,13 @@
 import static org.eclipse.edc.junit.testfixtures.TestUtils.getFreePort;
 
 public class Participant {
-
+    public static final String E2E_TEST_NAME;
```
ndr-brt, line 46: I'm not a big fan of these shared "static" variables initialized in a static block.
I think this "test name" variable should belong to the test class, and then it could be passed to the `controlPlaneCosmosDbConfiguration` to set up the configuration and used in the beforeAll to create the database

system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/Participant.java
```diff
@@ -358,27 +364,27 @@ public Map<String, String> controlPlanePostgresConfiguration() {
 
     public Map<String, String> controlPlaneCosmosDbConfiguration() {
         var baseConfiguration = controlPlaneConfiguration();
-
+        var prefix = name;
```
ndr-brt, line 367: can be inlined

system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/EndToEndTransferCosmosDbTest.java
```diff
@@ -25,28 +25,27 @@
...
+import java.util.UUID;
 import java.util.stream.Stream;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.eclipse.edc.test.e2e.Participant.E2E_TEST_NAME;
 
 @AzureCosmosDbIntegrationTest
 class EndToEndTransferCosmosDbTest extends AbstractEndToEndTransfer {
 
+    public static final String E2E_TEST_NAME = "e2e-transfer-test-" + UUID.randomUUID();
```
ndr-brt, line 36: super minor nit
```suggestion
    private static final String E2E_TEST_NAME = "e2e-transfer-test-" + UUID.randomUUID();
```

ndr-brt (approved):

PR #2577, opened 2023-03-12
docs: decision record about refactoring the test pipeline

paullatzelsperger:
## What this PR changes/adds

Creates a decision-record that outlines proposed changes to our testing and testing pipeline.

## Why it does that

Improve performance, convenience and maintainability.

## Further notes

.

## Linked Issue(s)

.
## Checklist

### DIFF
--- a/docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
+++ b/docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
@@ -1,0 +1,124 @@
+# Refactoring of testing
+
+The testing pipeline (in GitHub Actions) will be refactored as outlined in this decision record.
+
+## Rationale
+
+Currently the testing pipeline used in EDC is as follows:
+
+- `Unit-Tests`: runs all untagged tests
+- `Helm-Chart`: runs the minikube based Helm chart test. Is actually an End-2-End Test
+- `Azure-Storage-Integration-Test`: runs all tests in `extensions` and - `system-tests-azure-tests` that are tagged
+  with `@AzureStorageIntegrationTest` using the Azurite service. Uploads the Gatling report
+- `Azure-Cosmos-Integration-Test`: runs all tests in `extensions` tagged with `@AzureCosmosIntegrationTest`, if
+  the `COSMOS_KEY` secret is found. Does **not** run the `EndToEndTransferCosmosDbTest`!!
+- `Aws-Integration-Test`: runs all tests in `extensions` tagged with `@AwsS3IntegrationTest` using Minio
+- `Daps-Integration-Tests`: runs alls tests in `extensions/common/iam/oauth2/oauth2-daps` tagged
+  with `@DapsIntegrationTest`
+- `Postgresql-Integration-Tests`: runs all tests tagged with `@PostgresqlIntegrationTest` using the `postgres` service
+- `Hashicorp-Vault-Integration-Test`: runs all tests in `extensions` tagged with `@HashicorpVaultIntegrationTest` using
+  the `vault` service
+- `End-To-End-Test`: runs all tests tagged with `@EndToEndTest` using Minio
+- `Component-Test`: runs all tests tagged with `@ComponentTest`
+- `Api-Test`: runs all tests tagged with `@ApiTest`
+
+With this structure, there are some problems, which are:
+
+- Tests are run in wrong GH Actions jobs, e.g. `End-To-End` test only executes the inmem variant, the Postgres e2e
+  variant is run in the `Postgres-Integration-Test` job
+- Tests are incorrectly categorized:
+    - The concrete `EndToEnd___Test` instances are tagged with `@PostgresqlIntegrationTest`
+      or `@AzureCosmosIntegrationTEst`, but not with `@EndToEndTest`
+    - End-to-end tests tagged with `@PostgresqlIntegrationTest` are executed during the `Postgresql-Integration-Test`
+      job, and not during the End2End job.
+    - The `EndToEndTransferCosmosDbTest` is never executed due to incorrect/missing tags
+- Conditional execution based on the existence of secrets (i.e. CosmosDB) is cumbersome (fixed in my latest PR)
+
+## Approach / Solution proposals
+
+### Avoid running untagged tests
+
+Whenever a tag inclusion is specified (`-DincludeTags="..."`), we should _only_ run tests tagged as such. We must avoid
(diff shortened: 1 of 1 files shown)


docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
```diff
@@ -0,0 +1,124 @@
...
+### Refactoring the CI Pipeline
+
+We propose to refactor the current pipeline from its currently very linear form with a single workflow into several
+workflows:
+
+- `Verify.yaml`: all other jobs depend on this one
+    - `Checkstyle`: job that runs checkstyle
+    - `Dependency-Analysis`: runs the dependency analysis
+- `Unit-Test.yaml`: runs unit tests with coverage and uploads test reports
+- `Api-Tests.yaml`: contains a job that runs all API tests
```
ndr-brt, line 96: ApiTests are in fact integration tests (they test the interaction between our controllers and jetty+jersey), so can be included in the Integration-test workflow

docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Refactoring the CI Pipeline
+
+We propose to refactor the current pipeline from its currently very linear form with a single workflow into several
+workflows:
+
+- `Verify.yaml`: all other jobs depend on this one
+    - `Checkstyle`: job that runs checkstyle
+    - `Dependency-Analysis`: runs the dependency analysis
+- `Unit-Test.yaml`: runs unit tests with coverage and uploads test reports
```
ndr-brt, line 95: If I remember correctly unit & integration tests were put in the same workflow to upload coverage and test reports in the same run, if we separate them I guess we'll need to define a dedicated workflow that will run only when the dependent ones are finished

docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
```diff
@@ -0,0 +1,124 @@
...
+    - `Hashicorp-Vault`
+    - `Component-Test`
+- `EndToEndTest.yaml`: runs complex and potentially long running tests. may need more time.
+    - `In-Memory`: runs the `EndtoEndTransferInMemoryTest`
+    - `CosmosDb`: runs the `EndToEndTransferCosmosDbTest`, only if `COSMOS_KEY` is present
+    - `PostgreSQL`: runs the `EndToEndTransferPostgresTest`
+    - `Helm-Chart`: runs the minikube-based Helm chart test
+    - `OpenTelemetry`: runs the Opentelemetry tests
+
+Moving these jobs out into seperate workflows allows for a higher degree of parallelization, trigger configuration and
```
ndr-brt, line 111: I think parallelization doesn't change much, jobs are already using parallelization at the highest degree.

docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
paullatzelsperger: To be honest, I'm not sure if we really need those coverage reports and test reports. Is anyone every looking at them? AFAIK there is a novel way to incorporate Codecov, maybe that's all we need?

docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
paullatzelsperger: Currently the `EndToEnd` job only runs the in-memory variant. Once we also run the postgres and cosmos e-2-e tests, and if we kept the current format, they would run sequentially as steps, which would blow up test runtime significantly

docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
paullatzelsperger: That's correct, however, having them as separate workflow allows us to assign different triggers or execution guards. if we don't need that, we can put them into the `Integration-test` workflow.

ndr-brt: I think we created a different tag for them just to save some time on the unit tests but they don't need special attentions.

docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
ndr-brt: Postgres E2E are running in fact, because I broke time one time or two during the last weeks, they are running within the "Postgres-Integration-Test" job but I agree that we should improve the separation

ndr-brt (commented):
responded to some comments

docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
paullatzelsperger: ok, we can have them run in the i-test workflow.

docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
paullatzelsperger: ok, so what we'll probably want is an aggregated Jacoco report across all modules, right? Currently, we only have separate reports _per module_. 
That aggregate report can then be uploaded to Codecov.

The problem with that is though, since we don't run all the tests (and thus coverage repots) in one job or even workflow, we never have all reports in one place. Maybe we can upload all reports separately as GH artifacts and publish them to Codecov in a separate action?

docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
paullatzelsperger: yes they run, but not in the `End-To-End` test job, but in the `Postgres-Integration-Tests`, which is ... well... wrong :)

End-to-End tests are expected to take a long time, so parallelizing those more makes sense. Currently `End-To-End-TEsts` is a job, where all steps (PG, Az, ...) would run sequentially, but if we convert it to a workflow, PG, Az, etc could run as parallel jobs.

docs/developer/decision-records/2023-03-12_testing_pipeline/README.md
ndr-brt: I'm not really a codeconv expert, but if that's possible, it could be the way to go

paullatzelsperger: neither am I, but AFAIK Codecov accepts jacoco reports, and it probably doesn't care whether it's a specific one or the aggregated one. 
outputting them as artifacts, then downloading and aggregating them will likely be greater challenge.

PR #2600, opened 2023-03-16
fix: data plane public api error propagation

### LINKED ISSUE #2599: Inappropriate error propagation by Data Plane's Public API
# Bug Report

## Describe the Bug
The Data Plane transforms 40x-HTTP-error codes received from Data Source to code 500 with an unclear message.

### Expected Behavior
Data Plane transforms Data Source communication errors to 502 (Bad Gateway)

### Observed Behavior
Data Plane transforms Data Source communication errors to 500 (Internal Server error)

## Steps to Reproduce
1. Register an Asset with a non-existing backend endpoint.
```json
{
  "asset": {
    "properties": {
      "asset:prop:id": "test-404-not-found",
      "asset:prop:name": "Asset to test an service that is not available.",
      "asset:prop:contenttype": "application/json",
      "asset:prop:policy-id": "use-eu"
    }
  },
  "dataAddress": {
    "properties": {
      "baseUrl": "{{EXISTING_SERVER}}/non-existing-endpoint",
      "type": "HttpData",
      "proxyBody": false,
      "proxyMethod": false
    }
  }
}
```
2. Initiate transfer flow
3. Consumer Data Plane will respond with 500.

## Context Information

-  Provider EDC 0.1.2
- Consumer EDC 0.2.0

yurimssilva:
## What this PR changes/adds

- Enriches the error message propagated by the DataSink implementations with the message sent by the DataSource/DataSink processes and fixed tests
- Changed the status code sent by the DataPlanePublicApiController from 500 to 502 in cases of failure inside the transfer method, when external dependencies are called

## Why it does that

- Previously the Data Plane would return a status 500 with a generic error message, leading the user to believe it's a failure in the consumer's connector when it is not.

## Linked Issue(s)

Closes #2599

## Checklist

- [x ] added appropriate tests?

### DIFF
--- a/core/data-plane/data-plane-util/src/main/java/org/eclipse/edc/connector/dataplane/util/sink/OutputStreamDataSink.java
+++ b/core/data-plane/data-plane-util/src/main/java/org/eclipse/edc/connector/dataplane/util/sink/OutputStreamDataSink.java
@@ -57,7 +57,7 @@ public class OutputStreamDataSink implements DataSink {
                     });
         } catch (Exception e) {
             monitor.severe("Error processing data transfer request", e);
-            return CompletableFuture.completedFuture(StatusResult.failure(ERROR_RETRY, "Error processing data transfer request"));
+            return CompletableFuture.completedFuture(StatusResult.failure(ERROR_RETRY, String.format("Error processing data transfer request - %s", e.getMessage())));
         }
     }
 
--- a/core/data-plane/data-plane-util/src/main/java/org/eclipse/edc/connector/dataplane/util/sink/ParallelSink.java
+++ b/core/data-plane/data-plane-util/src/main/java/org/eclipse/edc/connector/dataplane/util/sink/ParallelSink.java
@@ -64,7 +64,7 @@ public abstract class ParallelSink implements DataSink {
                     .exceptionally(throwable -> StatusResult.failure(ERROR_RETRY, "Unhandled exception raised when transferring data: " + throwable.getMessage()));
         } catch (Exception e) {
             monitor.severe("Error processing data transfer request: " + requestId, e);
-            return CompletableFuture.completedFuture(StatusResult.failure(ERROR_RETRY, "Error processing data transfer request"));
+            return CompletableFuture.completedFuture(StatusResult.failure(ERROR_RETRY, String.format("Error processing data transfer request - %s", e.getMessage())));
         }
     }
 
--- a/extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/controller/DataPlanePublicApiController.java
+++ b/extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/controller/DataPlanePublicApiController.java
@@ -42,6 +42,7 @@ import java.util.concurrent.ExecutorService;
 
 import static java.lang.String.format;
 import static java.lang.String.join;
+import static org.eclipse.edc.connector.dataplane.api.response.ResponseFunctions.badGatewayErrors;
 import static org.eclipse.edc.connector.dataplane.api.response.ResponseFunctions.internalErrors;
 import static org.eclipse.edc.connector.dataplane.api.response.ResponseFunctions.validationError;
 
@@ -149,7 +150,7 @@ public class DataPlanePublicApiController implements DataPlanePublicApi {
                         if (result.succeeded()) {
                             response.resume(Response.ok(stream.toString()).build());
                         } else {
-                            response.resume(internalErrors(result.getFailureMessages()));
+                            response.resume(badGatewayErrors(result.getFailureMessages()));
                         }
                     } else {
                         response.resume(internalErrors(List.of("Unhandled exception occurred during data transfer: " + throwable.getMessage())));
--- a/core/data-plane/data-plane-util/src/test/java/org/eclipse/edc/connector/dataplane/util/sink/ParallelSinkTest.java
+++ b/core/data-plane/data-plane-util/src/test/java/org/eclipse/edc/connector/dataplane/util/sink/ParallelSinkTest.java
@@ -81,7 +81,7 @@ class ParallelSinkTest {
 
         assertThat(fakeSink.transfer(dataSourceMock)).succeedsWithin(500, TimeUnit.MILLISECONDS)
                 .satisfies(transferResult -> assertThat(transferResult.failed()).isTrue())
-                .satisfies(transferResult -> assertThat(transferResult.getFailureMessages()).containsExactly("Error processing data transfer request"));
+                .satisfies(transferResult -> assertThat(transferResult.getFailureMessages()).containsExactly("Error processing data transfer request - test-errormessage"));
         assertThat(fakeSink.complete).isEqualTo(0);
     }
 
--- a/extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/response/ResponseFunctions.java
+++ b/extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/response/ResponseFunctions.java
@@ -19,6 +19,7 @@ import org.eclipse.edc.connector.dataplane.spi.response.TransferErrorResponse;
 
 import java.util.List;
 
+import static jakarta.ws.rs.core.Response.Status.BAD_GATEWAY;
 import static jakarta.ws.rs.core.Response.Status.BAD_REQUEST;
 import static jakarta.ws.rs.core.Response.Status.FORBIDDEN;
 import static jakarta.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
@@ -47,6 +48,16 @@ public final class ResponseFunctions {
      */
     public static Response validationErrors(List<String> errors) {
         return createErrorResponse(BAD_REQUEST, errors);
+    }
+
+    /**
+     * Returns a response for a collection of bad gateway errors.
+     *
+     * @param errors List of errors.
+     * @return Error response.
+     */
+    public static Response badGatewayErrors(List<String> errors) {
+        return createErrorResponse(BAD_GATEWAY, errors);
     }
 
     /**


core/data-plane/data-plane-util/src/main/java/org/eclipse/edc/connector/dataplane/util/sink/OutputStreamDataSink.java
```diff
@@ -57,7 +57,7 @@ public CompletableFuture<StatusResult<Void>> transfer(DataSource source) {
                     });
         } catch (Exception e) {
             monitor.severe("Error processing data transfer request", e);
-            return CompletableFuture.completedFuture(StatusResult.failure(ERROR_RETRY, "Error processing data transfer request"));
+            return CompletableFuture.completedFuture(StatusResult.failure(ERROR_RETRY, String.format("Error processing data transfer request - %s", e.getMessage())));
```
ndr-brt, line 60: please static-import `format` (but also `completedFuture` while you are here :) )

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/controller/DataPlanePublicApiController.java
```diff
@@ -149,7 +150,7 @@ private void handle(ContainerRequestContext context, AsyncResponse response) {
                         if (result.succeeded()) {
                             response.resume(Response.ok(stream.toString()).build());
                         } else {
-                            response.resume(internalErrors(result.getFailureMessages()));
+                            response.resume(badGatewayErrors(result.getFailureMessages()));
```
ndr-brt, line 153: please add a test to verify this condition

core/data-plane/data-plane-util/src/main/java/org/eclipse/edc/connector/dataplane/util/sink/ParallelSink.java
```diff
@@ -64,7 +64,7 @@ public CompletableFuture<StatusResult<Void>> transfer(DataSource source) {
                     .exceptionally(throwable -> StatusResult.failure(ERROR_RETRY, "Unhandled exception raised when transferring data: " + throwable.getMessage()));
         } catch (Exception e) {
             monitor.severe("Error processing data transfer request: " + requestId, e);
-            return CompletableFuture.completedFuture(StatusResult.failure(ERROR_RETRY, "Error processing data transfer request"));
+            return CompletableFuture.completedFuture(StatusResult.failure(ERROR_RETRY, String.format("Error processing data transfer request - %s", e.getMessage())));
```
ndr-brt, line 67: This way the exception message will be propagated to the counter-part, I'm not sure this is something we really want.

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/controller/DataPlanePublicApiController.java
ndr-brt: plus, I'm not sure that only "bad gateway" errors will finish here, in the `ParallelSink` and in the `OutputStreamDataSink` there are multiple reasons why an `Exception` could be thrown and caught and transformed into a `StateResult.failure`

ndr-brt (changes requested):
.

core/data-plane/data-plane-util/src/main/java/org/eclipse/edc/connector/dataplane/util/sink/ParallelSink.java
jimmarino: Agreed, we don't want this. It's sufficient to send the exception to the monitor. We could discuss a correlation mechanism where a GUID is generated on an exception, sent to the monitor with the exception, and the GUID returned to the client.

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/controller/DataPlanePublicApiController.java
jimmarino: To step back, is the issue being addressed error correlation (see my comment below), or that 502 should be returned instead of 500?

yurimssilva: I agree with that, although that is the way both DataSink implementations (and their subclasses) deal with it, they wrap up the entire transfer method code into one single try/catch block which catches a generic Exception and treats anything that may go wrong down da stack as an internal error. Considering that the methods down the stack mostly rely on external dependencies (DataSource and DataSink) I figured it would make sense to treat errors thrown as a bad gateway. In order to improve error handling it would be necessary to change the DataSink and DataSource openPartStream and openStream implementations identifying and wrapping code blocks addressing accurately error causes throwing proper exceptions, because most of them (if not all of them) would also wrap the entire code block catching a generic exception. Would you guys think that is the case?

core/data-plane/data-plane-util/src/main/java/org/eclipse/edc/connector/dataplane/util/sink/ParallelSink.java
yurimssilva: I can understand the reason why we don't want to propagate the message retrieved by the DataSource but it would always cause the consumer counter-part end user to first believe it is their connector failure. I think a GUID on exceptions is a great approach we could discuss.

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/controller/DataPlanePublicApiController.java
jimmarino: Back-end exceptions should not be available to clients as this information is internal and potentially sensitive. The failure messages can provide correlation information but should not pass exception details. 

For me, 500 is the correct response because the fact that one data plane may be a proxy is an internal detail. For example, another data plane may not be a proxy and pull data from some storage device.

yurimssilva: > Back-end exceptions should not be available to clients as this information is internal and potentially sensitive. The failure messages can provide correlation information but should not pass exception details.
> 
- I agree.

> For me, 500 is the correct response because the fact that one data plane may be a proxy is an internal detail. For example, another data plane may not be a proxy and pull data from some storage device.
> 
- Then in the cases when it plays a proxy role, a failure caused by external dependencies shall be treated as internal, maybe to be more embracing I'd say, I got that.

> `DataSource` and `DataSink` should throw unchecked exceptions only to indicate non-recoverable conditions. If there is a recoverable condition, the implementations should deal with it.
> 
- That for sure, is the desired behavior, but the `DataSouce` and `DataSink` implementations, as mentioned above, only catch a generic `Exception` (which includes checked and unchecked) indistinctly making any condition irrecoverable.

jimmarino: > * That for sure, is the desired behavior, but the `DataSouce` and `DataSink` implementations, as mentioned above, only catch a generic `Exception` (which includes checked and unchecked) indistinctly making any condition irrecoverable.

I don't follow. The implementations correctly handle exceptions and convert unrecoverable ones into `EdcException`. What is the issue? Note that a checked exception can be unrecoverable, e.g. `FileNotFoundException` as in `FileTransferDataSource`.

core/data-plane/data-plane-util/src/main/java/org/eclipse/edc/connector/dataplane/util/sink/ParallelSink.java
yurimssilva: The client might/should not be aware of external dependencies and a status 500 combined with an undetailed message would certainly cause this misinterpretation, such as the one in the report that originated the issue.

I think we could improve the exception handling to retrieve correlation information as you suggested.

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/controller/DataPlanePublicApiController.java
yurimssilva: e.g `HttpDataSource` turns any non-2xx status returned into an `EdcException` which is eventually caught as an `Exception` in `DataSink `implementations which despises the response content.

core/data-plane/data-plane-util/src/main/java/org/eclipse/edc/connector/dataplane/util/sink/ParallelSink.java
jimmarino: The client *should* interpret the problem as an issue with the system it communicates with (the data plane), even if another system behind it caused the error. 

The issue is not exception handling; that is done correctly in the code. There are two issues as I see it. 

The first is providing a correlation mechanism for a *data plane operator* (not the client) to link unsuccessful responses returned to a client and internal diagnostic information. IMO, this can be handled by a Decision Record that calls for two minor changes. Specifically:

1. One or more correlation ids are sent to the client
2. Exceptional conditions sent to the monitor should contain a correlation id

A second issue is the HTTP DataSource could log an error response body (if one exists).

yurimssilva: > ...
> 
> 1. One or more correlation ids are sent to the client
> 2. Exceptional conditions sent to the monitor should contain a correlation id
> 
> A second issue is the HTTP DataSource could log an error response body (if one exists).

- `ParallelSink`'s implementation already provides something very similar to what you suggested. It uses the DataFlowRequest id to correlate the request to the exception thrown, it logs the id alongside the exception message and returns just the id to the client, and it also enables the operator to link unsuccessful responses to internal diagnostic information. Wouldn't it be the case just to reproduce these actions also in `OutputStreamDataSink`?

jimmarino: > > ...
> > 
> > 1. One or more correlation ids are sent to the client
> > 2. Exceptional conditions sent to the monitor should contain a correlation id
> > 
> > A second issue is the HTTP DataSource could log an error response body (if one exists).
> 
> * `ParallelSink`'s implementation already provides something very similar to what you suggested. It uses the `DataFlowRequest` id to correlate the request to the exception thrown, it logs the id alongside the exception message and returns just the id to the client, and it also enables the operator to link unsuccessful responses to internal diagnostic information. Wouldn't it be the case just to reproduce these actions also in `OutputStreamDataSink`?

Yes, I think using the request id would be fine.

yurimssilva:
A new issue has been opened to better describe and treat the case.

Issue #2606

PR #2623, opened 2023-03-22
feat: support virtual-hosted style addressing for overridden endpoint

### LINKED ISSUE #2426: Virtual-hosted style URL is not supported by S3Cient with endpoint override
# Bug Report

## Describe the Bug

If endpoint is overridden, virtual-hosted style URL can not be used due to https://github.com/aws/aws-sdk-java-v2/issues/2317.

## Context Information
Used version: main cc5b34833574be9b5f20d7c128f4e1c6a840e129
OS: Ubuntu 22.04 x86_64
openjdk-11-jdk (11.0.16+8-0ubuntu1~22.04)

## Possible Implementation

bump the version of AWS SDK for Java to 2.18.x or above.

iwasakims:
## What this PR changes/adds

Add configuration knob to switch addressing style of S3 between virtual-hosted style and path style addressing.

## Why it does that

To access S3 compatible data store using virtual-hosted style addressing of S3. The current AwsClientProviderImpl always enable path style addressing for overridden endpoint.

## Testing

I tested the patch against real S3 using endpoint override as following.

```
$ IT_AWS_ENDPOINT=https://s3.ap-northeast-1.amazonaws.com/ \
  IT_AWS_REGION=ap-northeast-1 \
  IT_AWS_PROFILE=iwasakims \
  IT_AWS_PATH_STYLE_ACCESS_ENABLED=false \
    ./gradlew clean test -p extensions/control-plane/provision/provision-aws-s3 -DincludeTags=AwsS3IntegrationTest -PverboseTest

$ IT_AWS_ENDPOINT=https://s3.ap-northeast-1.amazonaws.com/ \
  IT_AWS_REGION=ap-northeast-1 \
  IT_AWS_PROFILE=iwasakims \
  IT_AWS_PATH_STYLE_ACCESS_ENABLED=false \
    ./gradlew clean test -p extensions/data-plane/data-plane-aws-s3 -DincludeTags=AwsS3IntegrationTest -PverboseTest
```

While MinIO supports virtual-hosted style addressing if server is started with configuration like `MINIO_DOMAIN=example.com`, it returns 400 to requests from existing test without enabling path style. I will look into the cause and file another issue later.

## Linked Issue(s)

Closes #2426 

## Checklist

- [n/a] added appropriate tests?
- [n/a] added/updated relevant documentation?
- [] assigned appropriate label? (exclude from changelog with label `no-changelog`)

### DIFF
--- a/extensions/common/aws/aws-s3-core/src/main/java/org/eclipse/edc/aws/s3/AwsClientProviderImpl.java
+++ b/extensions/common/aws/aws-s3-core/src/main/java/org/eclipse/edc/aws/s3/AwsClientProviderImpl.java
@@ -107,6 +107,7 @@ public class AwsClientProviderImpl implements AwsClientProvider {
                 .region(Region.of(region));
 
         handleBaseEndpointOverride(builder);
+        handlePathStyleEnabled(builder);
 
         return builder.build();
     }
@@ -117,6 +118,7 @@ public class AwsClientProviderImpl implements AwsClientProvider {
                 .region(Region.of(region));
 
         handleBaseEndpointOverride(builder);
+        handlePathStyleEnabled(builder);
 
         return builder.build();
     }
@@ -128,6 +130,7 @@ public class AwsClientProviderImpl implements AwsClientProvider {
                 .region(Region.of(region));
 
         handleBaseEndpointOverride(builder);
+        handlePathStyleEnabled(builder);
 
         return builder.build();
     }
@@ -157,8 +160,7 @@ public class AwsClientProviderImpl implements AwsClientProvider {
     private void handleBaseEndpointOverride(S3BaseClientBuilder<?, ?> builder) {
         var endpointOverride = configuration.getEndpointOverride();
         if (endpointOverride != null) {
-            builder.serviceConfiguration(S3Configuration.builder().pathStyleAccessEnabled(true).build())
-                    .endpointOverride(endpointOverride);
+            builder.endpointOverride(endpointOverride);
         }
     }
 
@@ -167,5 +169,9 @@ public class AwsClientProviderImpl implements AwsClientProvider {
         if (endpointOverride != null) {
             builder.endpointOverride(endpointOverride);
         }
+    }
+
+    private void handlePathStyleEnabled(S3BaseClientBuilder<?, ?> builder) {
+        builder.serviceConfiguration(S3Configuration.builder().pathStyleAccessEnabled(configuration.getPathStyleAccessEnabled()).build());
     }
 }
--- a/extensions/common/aws/aws-s3-core/src/main/java/org/eclipse/edc/aws/s3/AwsClientProviderImpl.java
+++ b/extensions/common/aws/aws-s3-core/src/main/java/org/eclipse/edc/aws/s3/AwsClientProviderImpl.java
@@ -104,10 +104,10 @@ public class AwsClientProviderImpl implements AwsClientProvider {
         var credentialsProvider = StaticCredentialsProvider.create(credentials);
         var builder = S3Client.builder()
                 .credentialsProvider(credentialsProvider)
-                .region(Region.of(region));
+                .region(Region.of(region))
+                .serviceConfiguration(createS3Configuration());
 
         handleBaseEndpointOverride(builder);
-        handlePathStyleEnabled(builder);
 
         return builder.build();
     }
@@ -115,10 +115,10 @@ public class AwsClientProviderImpl implements AwsClientProvider {
     private S3Client createS3Client(String region) {
         S3ClientBuilder builder = S3Client.builder()
                 .credentialsProvider(credentialsProvider)
-                .region(Region.of(region));
+                .region(Region.of(region))
+                .serviceConfiguration(createS3Configuration());
 
         handleBaseEndpointOverride(builder);
-        handlePathStyleEnabled(builder);
 
         return builder.build();
     }
@@ -127,10 +127,10 @@ public class AwsClientProviderImpl implements AwsClientProvider {
         var builder = S3AsyncClient.builder()
                 .asyncConfiguration(b -> b.advancedOption(FUTURE_COMPLETION_EXECUTOR, executor))
                 .credentialsProvider(credentialsProvider)
-                .region(Region.of(region));
+                .region(Region.of(region))
+                .serviceConfiguration(createS3Configuration());
 
         handleBaseEndpointOverride(builder);
-        handlePathStyleEnabled(builder);
 
         return builder.build();
     }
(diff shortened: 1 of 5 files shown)


paullatzelsperger (commented):
do i understand correctly, that MinIO works either with path-style addressing or virtual-host style addressing, but not both?

If that is the case, what we could do is spin up _another_ MinIO instance with the `MINIO_DOMAIN` flag in the `verify.yaml` workflow and run the existing tests against the first instance, and the virtual-host-based tests against the second instance. We'd separate them by port, and make the port configurable in the `AbstractS3Test`. 
I'm hesitant to accept untested code.

iwasakims:
> do i understand correctly, that MinIO works either with path-style addressing or virtual-host style addressing, but not both?

No. MINIO accepts both style request if it was launched with `MINIO_DOMAIN=example.com`. I confirmed that by running existing tests (with path style addressing) and aws-cli.

While running tests against real S3 succeed regardless of addressing style (as expected), tests against MINIO with virtual-hosted style fail. I'm still digging and suspecting compatibility issue between MINIO and AWS SDK for Java 2.

Since we need local DNS resolving `*.example.com` to 127.0.0.1 for testing this against local MINIO with `MINIO_DOMAIN=example.com`, I need additional effort to make workflow of verify.yaml work.

extensions/common/aws/aws-s3-core/src/main/java/org/eclipse/edc/aws/s3/S3CoreExtension.java
```diff
@@ -45,6 +45,8 @@ public class S3CoreExtension implements ServiceExtension {
     private static final String AWS_ENDPOINT_OVERRIDE = "edc.aws.endpoint.override";
     @Setting(value = "The size of the thread pool used for the async clients")
     private static final String AWS_ASYNC_CLIENT_THREAD_POOL_SIZE = "edc.aws.client.async.thread-pool-size";
+    @Setting(value = "If true, path style addressing is used instead of virtual-hosted style")
```
ndr-brt, line 48: please add the "type" and the "default" attributes

extensions/common/aws/aws-s3-core/src/main/java/org/eclipse/edc/aws/s3/S3CoreExtension.java
```diff
@@ -65,10 +67,13 @@ public AwsClientProvider awsClientProvider(ServiceExtensionContext context) {
 
         var threadPoolSize = context.getSetting(AWS_ASYNC_CLIENT_THREAD_POOL_SIZE, DEFAULT_AWS_ASYNC_CLIENT_THREAD_POOL_SIZE);
 
+        var pathStyleAccessEnabled = context.getSetting(AWS_PATH_STYLE_ACCESS_ENABLED, true);
```
ndr-brt, line 70: is `true` as default the behavior we want? currently isn't `false`?
In any case, please extract a constant, that will be used also in the `@Setting` annotation and in the configuration class.

extensions/common/aws/aws-s3-test/src/testFixtures/java/org/eclipse/edc/aws/s3/testfixtures/AbstractS3Test.java
```diff
@@ -68,12 +68,14 @@ public abstract class AbstractS3Test {
     // [see http://stackoverflow.com/questions/13898057/aws-error-message-a-conflicting-conditional-operation-is-currently-in-progress]
     protected static final String MINIO_ENDPOINT = "http://localhost:9000";
     protected static final URI S3_ENDPOINT = URI.create(propOrEnv("it.aws.endpoint", MINIO_ENDPOINT));
+    protected static final Boolean PATH_STYLE_ACCESS_ENABLED = Boolean.valueOf(propOrEnv("it.aws.path.style.access.enabled", "true"));
```
ndr-brt, line 71: this propOrEnv is not set by any test, I wold hardcode the `true` in the configuration.

extensions/common/aws/aws-s3-core/src/main/java/org/eclipse/edc/aws/s3/AwsClientProviderImpl.java
```diff
@@ -168,4 +170,8 @@ private void handleEndpointOverride(SdkClientBuilder<?, ?> builder) {
             builder.endpointOverride(endpointOverride);
         }
     }
+
+    private void handlePathStyleEnabled(S3BaseClientBuilder<?, ?> builder) {
```
ndr-brt, line 174: why not having a `createS3Configuration` method that returns the `S3Configuration` instance and then it can be set on every builder? it would avoid the side effect

ndr-brt (changes requested):

extensions/common/aws/aws-s3-test/src/testFixtures/java/org/eclipse/edc/aws/s3/testfixtures/AbstractS3Test.java
iwasakims: This is for manual testing with endpoint override. We can run test against real S3 with virtual-hosted style addressing
by following command line. Same for another S3 compatible storage supporting.

```
$ IT_AWS_ENDPOINT=https://s3.ap-northeast-1.amazonaws.com/ \
  IT_AWS_REGION=ap-northeast-1 \
  IT_AWS_PROFILE=iwasakims \
  IT_AWS_PATH_STYLE_ACCESS_ENABLED=false \
    ./gradlew clean test -p extensions/control-plane/provision/provision-aws-s3 -DincludeTags=AwsS3IntegrationTest -PverboseTest
```

ndr-brt: if so, please document it within the `README` file of the module

extensions/common/aws/aws-s3-core/src/main/java/org/eclipse/edc/aws/s3/S3CoreExtension.java
iwasakims: fixed the inconsistent default value and added constant.

extensions/common/aws/aws-s3-core/src/main/java/org/eclipse/edc/aws/s3/S3CoreExtension.java
iwasakims: added type and defaultValue.

extensions/common/aws/aws-s3-core/src/main/java/org/eclipse/edc/aws/s3/AwsClientProviderImpl.java
iwasakims: It makes sense. I refactored the code based on your suggestion.

iwasakims:
> I'm hesitant to accept untested code.

I tested the feature against AWS S3 supporting both addressing style instead of MinIO. Let me address the issue of virtual-hosted style addressing with MinIO in another PR. @paullatzelsperger

ndr-brt (approved):

Issue #2626, opened 2023-03-23
Story: Processing callbacks


wolf4ood:
# Feature Request

This story covers the implementation of  the new callback mechanism decided in the DR https://github.com/eclipse-edc/Connector/pull/2549

- #2637
- #2625
- #2624
- #2627
- #2819
- #2926

wolf4ood:
Closing this since all the sub issues are done

PR #2549, opened 2023-03-03
docs: provide a Decision Record for processing callbacks

jimmarino:
## What this PR changes/adds

Adds a Decision Record for processing callbacks.

## Why it does that

Processing callbacks will provide low-level features to build additional capabilities such as automated API flows.

## Linked Issue(s)

Closes #2527

## Checklist

- [ ] added appropriate tests?
- [ ] performed checkstyle check locally?
- [ ] added/updated copyright headers?
- [ ] documented public classes/methods?
- [ ] added/updated relevant documentation?
- [ ] assigned appropriate label? (exclude from changelog with label `no-changelog`)
- [ ] formatted title correctly? (_take a look at the [CONTRIBUTING](https://github.com/eclipse-edc/Connector/blob/main/CONTRIBUTING.md#submit-a-pull-request) and [Etiquette for pull requests](https://github.com/eclipse-edc/Connector/blob/main/pr_etiquette.md) for details_)

PR #2628, opened 2023-03-23
feat(Callbacks): adds list of callback addresses in CN and TP

### LINKED ISSUE #2625: Callbacks in store
# Feature Request

This feature is part of the story #2626.

As stated in the DR #2549  a new fields containing the list of the dynamic callbacks should be added in:

- ContractNegotiation
- TransferProcess

core, stores

Ref DR #2549

wolf4ood:
## What this PR changes/adds

Add the support for storing list of callbacks in:

- ContractNegotiation
- TransferProcess

## Why it does that

Initial support for callbacks Ref DR #2549 

## Further notes

The initial implementation support storing custom callbacks via DataManagement API in

- ContractNegotiation
- TransferProcess

As the first PR of the story #2626, it doesn't had any behavior yet behind those callbacks.

 
## Linked Issue(s)

Closes #2625 

## Checklist

### DIFF
--- a/extensions/common/api/api-core/src/main/java/org/eclipse/edc/api/model/CallbackAddressDto.java
+++ b/extensions/common/api/api-core/src/main/java/org/eclipse/edc/api/model/CallbackAddressDto.java
@@ -1,0 +1,93 @@
+/*
+ *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.api.model;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import org.eclipse.edc.spi.types.domain.callback.CallbackAddress;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * DTO for {@link CallbackAddress}
+ */
+@JsonDeserialize(builder = CallbackAddressDto.Builder.class)
+public class CallbackAddressDto {
+
+    @NotBlank(message = "uri is mandatory")
+    private String uri;
+
+    @NotNull(message = "events cannot be null")
+    private Set<String> events = new HashSet<>();
+
+    private boolean transactional;
+
+    private CallbackAddressDto() {
+
+    }
+    
+    public boolean isTransactional() {
+        return transactional;
+    }
+
+    public Set<String> getEvents() {
+        return events;
+    }
+
+    public String getUri() {
+        return uri;
+    }
+
+
+    @JsonPOJOBuilder(withPrefix = "")
+    public static final class Builder {
+        private final CallbackAddressDto dto;
+
+        private Builder() {
+            this.dto = new CallbackAddressDto();
+        }
+
+        @JsonCreator
+        public static Builder newInstance() {
+            return new Builder();
+        }
+
+
+        public Builder uri(String url) {
+            dto.uri = url;
+            return this;
+        }
+
+        public Builder events(Set<String> events) {
+            dto.events = events;
+            return this;
+        }
+
+        public Builder transactional(boolean transactional) {
+            dto.transactional = transactional;
+            return this;
+        }
+
+
+        public CallbackAddressDto build() {
+            return dto;
+        }
+
+    }
+}
--- a/extensions/control-plane/api/management-api/contract-negotiation-api/src/test/java/org/eclipse/edc/connector/api/management/contractnegotiation/ContractNegotiationApiControllerIntegrationTest.java
+++ b/extensions/control-plane/api/management-api/contract-negotiation-api/src/test/java/org/eclipse/edc/connector/api/management/contractnegotiation/ContractNegotiationApiControllerIntegrationTest.java
@@ -29,6 +29,7 @@ import org.eclipse.edc.spi.entity.StatefulEntity;
 import org.eclipse.edc.spi.message.RemoteMessageDispatcher;
 import org.eclipse.edc.spi.message.RemoteMessageDispatcherRegistry;
 import org.eclipse.edc.spi.query.SortOrder;
+import org.eclipse.edc.spi.types.domain.callback.CallbackAddress;
 import org.eclipse.edc.spi.types.domain.message.RemoteMessage;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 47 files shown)


spi/common/core-spi/src/main/java/org/eclipse/edc/spi/types/domain/callback/CallbackAddress.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        public Builder transactional(boolean transactional) {
+            callbackAddress.transactional = transactional;
+            return this;
+        }
+
+
+        public CallbackAddress build() {
+            Objects.requireNonNull(callbackAddress.uri, "URI should not be null");
```
paullatzelsperger, line 79: if we impose these requirements, I would say there should be a separate DTO for the `CallbackAddress` so we can leverage `jakarta.validation` annotations there.

extensions/control-plane/api/management-api/contract-negotiation-api/src/test/java/org/eclipse/edc/connector/api/management/contractnegotiation/transform/ContractNegotiationToContractNegotiationDtoTransformerTest.java
```diff
@@ -63,6 +69,7 @@ void transform() {
         assertThat(dto.getErrorDetail()).isEqualTo("errorDetail");
         assertThat(dto.getUpdatedAt()).isEqualTo(contractNegotiation.getUpdatedAt());
         assertThat(dto.getCreatedAt()).isEqualTo(contractNegotiation.getCreatedAt());
+        assertThat(dto.getCallbackAddresses()).usingRecursiveFieldByFieldElementComparator().contains(callback);
```
paullatzelsperger, line 72: nit: may want to verify that it contains _only_ the `callback` object.

extensions/control-plane/api/management-api/contract-negotiation-api/src/test/java/org/eclipse/edc/connector/api/management/contractnegotiation/ContractNegotiationApiControllerIntegrationTest.java
```diff
@@ -29,6 +29,7 @@
 import org.eclipse.edc.spi.message.RemoteMessageDispatcher;
 import org.eclipse.edc.spi.message.RemoteMessageDispatcherRegistry;
 import org.eclipse.edc.spi.query.SortOrder;
+import org.eclipse.edc.spi.types.domain.callback.CallbackAddress;
```
paullatzelsperger, line 32: IMO this test and possibly also the `ContractNegotiationApiControllerTest` should explicitly verify these cases:
- with _one_ callback address
- with _no_ callback address
- with _multple_ callback addresses. 

could be a `ParameterizedTest`

paullatzelsperger (changes requested):
- there should be tests for saving, updating, etc in the ContractNegotiationStoreTestBase`
- SQL persistence for `TransferProcess` + tests is missing
- IMO a `CallbackAddressDto` could make sense, esp. regarding validation

wolf4ood:
@paullatzelsperger  what do you mean  by  `SQL persistence for TransferProcess + tests is missing` ?

wolf4ood:
No I did do a specific test for callbacks, I just added them always in  the helper function 

https://github.com/eclipse-edc/Connector/pull/2628/files#diff-e7ff2dfbf9ddaf64970e29e69fc091f81725a6173288f971deb9eaf2953432daR93

wolf4ood:
@paullatzelsperger 

same for the `ContractNegotiationStoreTestBase` i changed the builder function here.

https://github.com/eclipse-edc/Connector/pull/2628/files#diff-b4431c5bbfb371c6bd64384e1414ca16bd5811180f51ec2f66c9cf2466ab454dR77

extensions/control-plane/api/management-api/contract-negotiation-api/src/test/java/org/eclipse/edc/connector/api/management/contractnegotiation/transform/ContractNegotiationToContractNegotiationDtoTransformerTest.java
wolf4ood: Changed the assertion after introducing the `CallbackAddressDto`

wolf4ood:
@paullatzelsperger added tests for callbacks checks on stores

jimmarino:
> Indeed we could store a lot of callbacks, but i got that's was the point from the DR. Which contains either the static one (with extension) and the dynamic ones
> 
> https://github.com/eclipse-edc/Connector/tree/main/docs/developer/decision-records/2023-02-28-processing-callbacks#dynamic-endpoints
> 
> In case we can discuss if we need the dynamic one ? or we can do all with the static one? cc @jimmarino

We need the dynamic ones, for example, when a client wants to automatically fire off a transfer process after a contract negotiation completes, but not in all cases. It's much easier to achieve this scenario if the callback is scoped to the TP.

ndr-brt:
> We need the dynamic ones, for example, when a client wants to automatically fire off a transfer process after a contract negotiation completes, but not in all cases. It's much easier to achieve this scenario if the callback is scoped to the TP.

It could be discriminated at the client level, like, it receives a callback for every negotiation completion and if it's required, it fires the transfer.
It seems to be a pretty specialized feature that we are trying to reach before actually getting the most simpler one (handled by the extension).

Sorry I completely lost the "Dynamic Endpoints" paragraph, but to be fair it seems that we're bringing in unneeded complexity.

jimmarino:
> > We need the dynamic ones, for example, when a client wants to automatically fire off a transfer process after a contract negotiation completes, but not in all cases. It's much easier to achieve this scenario if the callback is scoped to the TP.
> 
> It could be discriminated at the client level, like, it receives a callback for every negotiation completion and if it's required, it fires the transfer. It seems to be a pretty specialized feature that we are trying to reach before actually getting the most simpler one (handled by the extension).
> 
> Sorry I completely lost the "Dynamic Endpoints" paragraph, but to be fair it seems that we're bringing in unneeded complexity.

I think this is pretty straightforward to implement and greatly simplifies clients which otherwise would need to create and handle their own storage mechanism. We also need it for C-X.

extensions/common/api/api-core/src/main/java/org/eclipse/edc/api/model/CallbackAddressDto.java
```diff
@@ -0,0 +1,90 @@
...
+    @NotNull(message = "events cannot be null")
+    private Set<String> events = new HashSet<>();
+
+    private boolean transactional;
+
+    private CallbackAddressDto() {
+
+    }
+
+
```
jimmarino, line 41: nit: spacing

extensions/common/api/api-core/src/main/java/org/eclipse/edc/api/model/CallbackAddressDto.java
```diff
@@ -0,0 +1,90 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+
+import java.util.HashSet;
+import java.util.Set;
+
+@JsonDeserialize(builder = CallbackAddressDto.Builder.class)
```
jimmarino, line 26: Documentation

extensions/control-plane/store/sql/transfer-process-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/transferprocess/store/schema/TransferProcessStoreStatements.java
```diff
@@ -154,6 +154,11 @@ default String getDeprovisionedResourcesColumn() {
         return "deprovisioned_resources";
     }
 
+    default String getCallbackAddressesColumn() {
+        return "callback_addresses";
+    }
+
```
jimmarino, line 160: nit: spacing

jimmarino (approved):
Looks good, just a few small nits.

ndr-brt (approved):

PR #2630, opened 2023-03-23
feat: Adds support for participant identities and checks

jimmarino:
## What this PR changes/adds

This PR adds support for configuring participant identities and verifications against those identities when contract negotiation and transfer process operations are performed.

## Why it does that

See the included documentation.

### Requested change

This PR makes an important change in `ProviderContractNegotiationManagerImpl.requested(ClaimToken token, ContractOfferRequest request)`. Previously, contract negotiations were persisted and transitioned to terminated if the request was invalid. To avoid potential issues, the correct behavior should be to reject the request outright and not store the negotiation. This results in a double state transition we should look at cleaning up in a subsequent PR:

```
transitToRequested(negotiation);

negotiation.addContractOffer(offer);

monitor.debug("[Provider] Contract offer received. Will be approved.");
transitToProviderAgreeing(negotiation);
``` 

### Declined change

Similarly,  `ProviderContractNegotiationManagerImpl.declined(ClaimToken token, String correlationId)` and `ConsumerContractNegotiationManagerImpl. declined(ClaimToken token, String negotiationId)` reject invalid requests outright instead of transitioning the negotiation to terminated.

## Linked Issue(s)

Closes #2631 

## Checklist

- [ ] added appropriate tests?
- [ ] performed checkstyle check locally?
- [ ] added/updated copyright headers?
- [ ] documented public classes/methods?
- [ ] added/updated relevant documentation?
- [ ] assigned appropriate label? (exclude from changelog with label `no-changelog`)
- [ ] formatted title correctly? (_take a look at the [CONTRIBUTING](https://github.com/eclipse-edc/Connector/blob/main/CONTRIBUTING.md#submit-a-pull-request) and [Etiquette for pull requests](https://github.com/eclipse-edc/Connector/blob/main/pr_etiquette.md) for details_)

PR #2652, opened 2023-03-29
feat(dsp): add missing notify methods on `ContractNegotiationService`

### LINKED ISSUE #2612: dataspace protocol: Add missing notify methods on `ContractNegotiation` services
# Feature Request

Add methods on `ContractNegotiationService` that will permit the (currently called) message handlers to being able to apply the needed state transition.

e.g. currently in `TransferProcessService` there's the `notifyStarted` method, that will be called when a `TransferStartMessage` is received from a provider, and the referred TransferProcess state need to be changed accordingly.

Adding the `notify` prefix will help to distinguish methods where actually an action is performed on the connector to the ones that represent an action taken on the counter-part and to keep the present tense (typical for an action/command).
e.g. `terminate`/`notifyTerminated` .

contract negotiation

cleanup

ndr-brt:
## What this PR changes/adds

Adds missing methods on `ContractNegotiationService` that will permit the DSP controller to ingest an incoming `RemoteMessage` to the service

## Why it does that

refactoring

## Further notes

- at this point is pretty clear the `*ed` public methods on the managers could be inlined, or their logic being put into  command/command handlers (to follow a pattern that we're currently following). I left this refactoring for the future to avoid conflicts with #2630 
- set `correlationId` also on consumer side, this will simplify the "fetch" logic a bit, because the correlationId will become the key used both by consumer and provider to retrieve the negotiation from the store.
- pulled up `Result.flatMap` to `AbstractResult`, so every other `*Result` could use it (really handy to convert from an `AbstractResult` extension to another (e.g. `ServiceResult` to `StatusResult`)

## Linked Issue(s)

Closes #2612 

## Checklist

### DIFF
--- a/core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
+++ b/core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
@@ -88,8 +88,10 @@ public class ConsumerContractNegotiationManagerImpl extends AbstractContractNego
     @WithSpan
     @Override
     public StatusResult<ContractNegotiation> initiate(ContractOfferRequest contractOffer) {
+        var id = UUID.randomUUID().toString();
         var negotiation = ContractNegotiation.Builder.newInstance()
-                .id(UUID.randomUUID().toString())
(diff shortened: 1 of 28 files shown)


ndr-brt:
> You may want to edit the title smiley

whops! :rofl:

jimmarino (approved):

Issue #2474, opened 2023-02-08
Implement new protocol module


juliapampus:
# Feature Request

The current IDS multipart should be replaced by an implementation of the new IDS protocol's https binding. This includes
- new controllers for catalog requests and negotiation and transfer processes
- new services for handling requests: mapping to EDC model, fowarding information to core implementations (e.g., state machines)
- new transformes for mapping EDC to DCAT/ODRL and vice versa

## Which Areas Would Be Affected?
`:data-protocols`

## Why Is the Feature Desired?
clean up, support new protocol as one step towards technical interoperability with other data space connector implementations

## Solution Proposal

...

juliapampus:
basic implementations are done

PR #2653, opened 2023-03-29
feat: add json-ld module

ronjaquensel:
## What this PR changes/adds

Adds an extension that provides an ObjectMapper for working with JSON-LD and utility functions for expanding and compacting JSON-LD structures.

## Why it does that

JSON-LD support is required for the dataspace protocol implementation.

## Further notes

Transformers that handle the mapping between EDC model and JSON-LD will be added in a separate PR.

## Linked Issue(s)

Relates #2474

## Checklist

### DIFF
--- a/extensions/common/json-ld/src/main/java/org/eclipse/edc/jsonld/JsonLdExtension.java
+++ b/extensions/common/json-ld/src/main/java/org/eclipse/edc/jsonld/JsonLdExtension.java
@@ -1,0 +1,62 @@
+/*
+ *  Copyright (c) 2023 Fraunhofer Institute for Software and Systems Engineering
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Fraunhofer Institute for Software and Systems Engineering - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.jsonld;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.datatype.jsonp.JSONPModule;
+import org.eclipse.edc.policy.model.AtomicConstraint;
+import org.eclipse.edc.policy.model.LiteralExpression;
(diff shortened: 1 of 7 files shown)


extensions/common/json-ld/src/main/java/org/eclipse/edc/jsonld/JsonLdExtension.java
```diff
@@ -0,0 +1,62 @@
...
+        var mapper = new ObjectMapper();
+        mapper.registerModule(new JSONPModule());
+        var module = new SimpleModule() {
+            @Override
+            public void setupModule(SetupContext context) {
+                super.setupModule(context);
+            }
+        };
+        mapper.registerModule(module);
+        mapper.registerSubtypes(AtomicConstraint.class, LiteralExpression.class);
```
ndr-brt, line 58: this subtypes registration shouldn't happen in this module

ronjaquensel: where should it happen then? I think the only other option would be for every extension using this mapper to register them itself.

jimmarino: They can be registered in a module lower or higher, depending on where we want to deserialize them. For example, in the IDS transformer module for JSON-LD or in core module if we want them registered against the default context.

extensions/common/json-ld/src/main/java/org/eclipse/edc/jsonld/JsonLdExtension.java
```diff
@@ -0,0 +1,62 @@
...
+import org.eclipse.edc.policy.model.AtomicConstraint;
+import org.eclipse.edc.policy.model.LiteralExpression;
+import org.eclipse.edc.runtime.metamodel.annotation.Extension;
+import org.eclipse.edc.runtime.metamodel.annotation.Inject;
+import org.eclipse.edc.spi.system.ServiceExtension;
+import org.eclipse.edc.spi.system.ServiceExtensionContext;
+import org.eclipse.edc.spi.types.TypeManager;
+
+@Extension(value = JsonLdExtension.NAME)
+public class JsonLdExtension implements ServiceExtension {
```
jimmarino, line 29: Can you add documentation here on what this extension does?

jimmarino (approved):
I left two comments. After those are resolved, LGTM.

ndr-brt (approved):

PR #2670, opened 2023-04-04
feat: dataspace protocol API configuration

ronjaquensel:
## What this PR changes/adds

Introduces the module `dsp-api-configuration`. The extension creates an API context for dataspace protocol endpoints, if no context is explicitly defined via configuration. It provides the `json-ld` ObjectMapper to the API context, allowing to use Jakarta JSON types as controller parameters. 

## Why it does that

To have an API context shared by all dataspace protocol endpoints.

## Linked Issue(s)

Relates #2474

## Checklist


data-protocols/dsp/dsp-api-configuration/src/main/java/org/eclipse/edc/protocol/dsp/api/configuration/DspApiConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Fraunhofer Institute for Software and Systems Engineering - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.protocol.dsp.api.configuration;
+
+public class DspApiConfiguration {
```
jimmarino, line 17: documentation

data-protocols/dsp/dsp-api-configuration/src/test/java/org/eclipse/edc/protocol/dsp/api/configuration/DspApiConfigurationExtensionTest.java
```diff
@@ -0,0 +1,101 @@
...
+import static org.eclipse.edc.protocol.dsp.api.configuration.DspApiConfigurationExtension.DEFAULT_DSP_WEBHOOK_ADDRESS;
+import static org.eclipse.edc.protocol.dsp.api.configuration.DspApiConfigurationExtension.DSP_WEBHOOK_ADDRESS;
+import static org.eclipse.edc.protocol.dsp.api.configuration.DspApiConfigurationExtension.SETTINGS;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(DependencyInjectionExtension.class)
```
jimmarino, line 45: Can we convert this to unit tests?

jimmarino (commented):
I know this is in draft, but I wanted to add some early comments

data-protocols/dsp/dsp-api-configuration/src/main/java/org/eclipse/edc/protocol/dsp/api/configuration/DspApiConfigurationExtension.java
```diff
@@ -0,0 +1,92 @@
...
+@Provides({ DspApiConfiguration.class })
+//@Requires({ JsonLdTransformerRegistry.class })
+public class DspApiConfigurationExtension implements ServiceExtension {
+    
+    public static final String NAME = "Dataspace Protocol API Configuration Extension";
+    
+    public static final String CONTEXT_ALIAS = "protocol";
+    
+    public static final String DEFAULT_DSP_WEBHOOK_ADDRESS = "http://localhost:8282/api/v1/dsp";
+    public static final String DSP_WEBHOOK_ADDRESS = "dsp.webhook.address";
```
ndr-brt, line 52: `edc.` prefix. 
one is enough, or should we have one for every context (catalog, negotiation, transfer)?

data-protocols/dsp/dsp-api-configuration/src/main/java/org/eclipse/edc/protocol/dsp/api/configuration/auth/ClaimTokenRequestFilter.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Intercepts all incoming requests and authenticates them using the IdentityService. As a ClaimToken
+ * instance is required for further processing of the claims, the ClaimToken returned by the
+ * IdentityService is serialized and added as an additional header.
+ */
+public class ClaimTokenRequestFilter implements ContainerRequestFilter {
```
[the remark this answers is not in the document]
ndr-brt: I would also change the approach, personally I'm more for the number 2, it will be just simpler to call the `IdentityService` directly in the controller, throwing the authentication failed exception when the result is failed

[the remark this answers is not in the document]
ronjaquensel: Removed the request filter. Authentication will then be handled in the controllers.

data-protocols/dsp/dsp-api-configuration/src/test/java/org/eclipse/edc/protocol/dsp/api/configuration/DspApiConfigurationExtensionTest.java
ronjaquensel: When converting this to a unit test, the injected services would have to be set on the extension by e.g. using reflection, not sure we want that. Or is there another way to construct the extension class that I'm missing?

data-protocols/dsp/dsp-api-configuration/src/main/java/org/eclipse/edc/protocol/dsp/api/configuration/DspApiConfigurationExtension.java
ronjaquensel: I would leave it at one for the time being. We can still switch to using multiple later on if necessary.

data-protocols/dsp/dsp-api-configuration/src/test/java/org/eclipse/edc/protocol/dsp/api/configuration/DspApiConfigurationExtensionTest.java
jimmarino: You could use `@ExtendWith(DependencyInjectionExtension.class)`. 

However, my opinion is we should be injecting on protected fields, not private, as that is common practice in Java (which would solve this issue). I realize private field injection is used throughout the codebase but I would be in favor of changing that.

ronjaquensel: If it's something that would need to be changed throughout the code base, there should be a separate PR anyway. I would then leave the test as is for the time being, as it is set up in the same way as other api configuration tests (ids, management). Refactoring could then be done on all of those at once. Wdyt?

jimmarino: I think there are two issues here:

1. Refactor the codebase to cleanup injection. We would need to discuss this, and I agree this is best done later after this release.

2. Clean this test up and break it into targetted unit tests. This could be done using actors or the dependency injection extension.

I think we should do #2 as I am currently trying to get rid of a lot of these "unit tests masquerading as integration tests" things.

jimmarino (approved):

ndr-brt (approved):

PR #2685, opened 2023-04-05
feat: dataspace protocol catalog API

ronjaquensel:
## What this PR changes/adds

Adds the dataspace protocol catalog request API. This includes the controller as well as required message classes and transformers. For now, the controller returns an empty catalog. Creation of the DCAT catalog will follow in a separate PR.

## Why it does that

To support the dataspace protocol specification.

## Further notes

This PR moves the plugin file for the `json-ld` extension, as it was in the wrong directory.

## Linked Issue(s)

Relates #2474 

## Checklist


data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/test/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogControllerIntegrationTest.java
```diff
@@ -0,0 +1,131 @@
...
+import static org.eclipse.edc.protocol.dsp.transform.transformer.Namespaces.DCAT_PREFIX;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@ApiTest
+@ExtendWith(EdcExtension.class)
+class CatalogControllerIntegrationTest {
```
jimmarino, line 51: Can we replace this with a unit test because it is testing the controller? We can address end-to-end IDS integration tests later.

data-protocols/dsp/dsp-catalog/dsp-catalog-spi/src/main/java/org/eclipse/edc/protocol/dsp/catalog/spi/types/CatalogRequestMessage.java
```diff
@@ -0,0 +1,57 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import org.eclipse.edc.spi.query.QuerySpec;
+
+/**
+ * Models a catalog request message as defined in the dataspace protocol specification.
+ */
+@JsonDeserialize(builder = CatalogRequestMessage.Builder.class)
+public class CatalogRequestMessage {
```
jimmarino, line 26: Serialization/deserialization test?

jimmarino (changes requested):
A couple of minor things as explained in the comments

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/test/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogControllerIntegrationTest.java
ndr-brt: I'm not against these API integration tests, they are covering the "hidden" jersey stuff like the  path annotations, the json deserialization and so on, every controller class in the whole codebase is tested also like this. I would keep implementing them unless we'll change idea

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
```diff
@@ -0,0 +1,119 @@
...
+        this.mapper = mapper;
+        this.identityService = identityService;
+        this.transformerRegistry = transformerRegistry;
+        this.dspCallbackAddress = dspCallbackAddress;
+    }
+    
+    @POST
+    @Path(CATALOG_REQUEST)
+    public Map<String, Object> getCatalog(JsonObject jsonObject, @HeaderParam(HttpHeaders.AUTHORIZATION) String token) {
+        var expanded = expand(jsonObject).getJsonObject(0); //expanding returns a JsonArray of size 1
```
ndr-brt, line 78: Is there a specific reason why this "expansion" operation is not in the transformer?

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
```diff
@@ -0,0 +1,119 @@
...
+        var messageResult = transformerRegistry.transform(expanded, CatalogRequestMessage.class);
+        if (messageResult.failed()) {
+            throw new InvalidRequestException("Request body was malformed.");
+        }
+    
+        checkAuthToken(token);
```
ndr-brt, line 84: this could be done before the transformation

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
```diff
@@ -0,0 +1,119 @@
...
+                .datasets(new ArrayList<>())
+                .dataServices(new ArrayList<>())
+                .build();
+        
+        var catalogResult = transformerRegistry.transform(catalog, JsonObject.class);
+        if (catalogResult.failed()) {
+            throw new EdcException(format("Failed to build response: %s", join(", ", catalogResult.getFailureMessages())));
+        }
+        
+        return mapper.convertValue(compact(catalogResult.getContent(), jsonLdContext()), Map.class);
```
ndr-brt, line 97: same as before, could this `compact` operation be done into the transformer?

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/test/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogControllerIntegrationTest.java
ronjaquensel: There is also a unit test already in place. But I would also vote for keeping this integration test to ensure integration with Jersey is working correctly.

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
ronjaquensel: Could probably also be done in the transformers, if we want them to return the compacted form. As I built this up on Jim's initial proposal for the transformer structure, @jimmarino what's your opinion on this?

But as this would also entail refactoring all the transformer tests, I would vote to change this in a separate PR, would then open a respective issue.

jimmarino: I think the expansion should be done outside the transformers because it needs to be performed at the root of the object graph, and the catalog is not necessarily the root in all foreseeable cases. For example, it may be contained in another structure or its elements may be top-level items in different contexts.

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
jimmarino: Not sure about this. Is the token being verified or its structure checked? If it is the former, I think that should be done outside the transformer (they only convert from one structure to another), otherwise, putting it in the transformed would make sense.

ronjaquensel: It's about token verification, but I understood the comment as simply moving the checking of the token to be the first operation in the controller method. So if the token is not valid, the transformation of the request body will not be performed. And I think that makes sense.

ndr-brt (approved):

PR #2687, opened 2023-04-06
feat(Callbacks): add callbacks http dispatcher via EventListener

### LINKED ISSUE #2624: Callback implementation : http dispatcher
# Feature Request

This feature is part of the story https://github.com/eclipse-edc/Connector/issues/2626.

The first implementation will support only http dynamic endpoints

stores, management api, spi

Ref DR #2549

wolf4ood:
## What this PR changes/adds

Add a dispatcher of the callbacks via Http and `RemoteMessageDispatcherRegistry`

## Why it does that

To provide a flexible mechanism for users to plug into the eventing system via http hooks

## Further notes

This PR adds two modules:

- `callback-event-dispatcher`
- `callback-http-dispatcher`

The `callback-event-dispatcher` contains an EventSubscriber that listens on all events and based on the dynamic callbacks configured , invoke the callbacks using the `RemoteMessageDispatcherRegistry`

 The `callback-http-dispatcher` implement a generic http dispatcher for RemoteMessage. It's based of the work on `dps` protocol and it may in the future refactored to be used in more places. It also implement a delegate for sending the `CallbackEventRemoteMessage` 

## Linked Issue(s)

Closes #2624 

## Checklist


extensions/control-plane/callback/callback-event-dispatcher/src/test/java/org/eclipse/edc/connector/callback/dispatcher/CallbackEventDispatcherExtensionTest.java
```diff
@@ -0,0 +1,63 @@
...
+    private CallbackEventDispatcherExtension extension;
+
+    @BeforeEach
+    void setUp(ServiceExtensionContext context, ObjectFactory factory) {
+        context.registerService(EventRouter.class, router);
+        context.registerService(RemoteMessageDispatcherRegistry.class, mock(RemoteMessageDispatcherRegistry.class));
+
+        extension = factory.constructInstance(CallbackEventDispatcherExtension.class);
+    }
+
```
jimmarino, line 48: nit: spacing

extensions/control-plane/callback/callback-event-dispatcher/src/main/java/org/eclipse/edc/connector/callback/dispatcher/CallbackEventDispatcher.java
```diff
@@ -0,0 +1,74 @@
...
+    private final boolean transactional;
+    private final Monitor monitor;
+
+    public CallbackEventDispatcher(RemoteMessageDispatcherRegistry dispatcher, boolean transactional, Monitor monitor) {
+        this.dispatcher = dispatcher;
+        this.transactional = transactional;
+        this.monitor = monitor;
+    }
+
+
```
jimmarino, line 46: nit: spacing

extensions/control-plane/callback/callback-http-dispatcher/src/main/java/org/eclipse/edc/connector/callback/dispatcher/http/CallbackEventDispatcherHttpExtension.java
```diff
@@ -0,0 +1,51 @@
...
+import org.eclipse.edc.spi.system.ServiceExtension;
+import org.eclipse.edc.spi.system.ServiceExtensionContext;
+import org.eclipse.edc.spi.types.TypeManager;
+
+@Extension(value = CallbackEventDispatcherHttpExtension.NAME)
+public class CallbackEventDispatcherHttpExtension implements ServiceExtension {
+
+
+    public static final String NAME = "Callback dispatcher http extension";
+    @Inject
```
jimmarino, line 30: nit: spacing

extensions/control-plane/callback/callback-http-dispatcher/src/main/java/org/eclipse/edc/connector/callback/dispatcher/http/CallbackEventDispatcherHttpExtension.java
```diff
@@ -0,0 +1,51 @@
...
+import org.eclipse.edc.runtime.metamodel.annotation.Inject;
+import org.eclipse.edc.spi.http.EdcHttpClient;
+import org.eclipse.edc.spi.message.RemoteMessageDispatcherRegistry;
+import org.eclipse.edc.spi.system.ServiceExtension;
+import org.eclipse.edc.spi.system.ServiceExtensionContext;
+import org.eclipse.edc.spi.types.TypeManager;
+
+@Extension(value = CallbackEventDispatcherHttpExtension.NAME)
+public class CallbackEventDispatcherHttpExtension implements ServiceExtension {
+
```
jimmarino, line 27: nit: spacing

extensions/control-plane/callback/callback-http-dispatcher/src/main/java/org/eclipse/edc/connector/callback/dispatcher/http/GenericHttpDispatcherDelegate.java
```diff
@@ -0,0 +1,51 @@
...
+
+import okhttp3.Request;
+import okhttp3.Response;
+import org.eclipse.edc.spi.types.domain.message.RemoteMessage;
+
+import java.util.function.Function;
+
+/**
+ * Generic Http dispatcher delegate that handle a single type of message
+ */
```
jimmarino, line 25: Doc could be more descriptive: A delegate that dispatches HTTP requests for a message type.

jimmarino (approved):
I only had a few nits on spacing/doc

extensions/control-plane/callback/callback-http-dispatcher/src/main/java/org/eclipse/edc/connector/callback/dispatcher/http/CallbackEventDispatcherHttpExtension.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+
+        var baseDispatcher = new GenericHttpRemoteDispatcherImpl(client);
+        baseDispatcher.registerDelegate(new CallbackEventRemoteMessageDispatcher(typeManager.getMapper()));
+
+        registry.register(new GenericHttpRemoteDispatcherWrapper(baseDispatcher, "http"));
+        registry.register(new GenericHttpRemoteDispatcherWrapper(baseDispatcher, "https"));
```
ndr-brt, line 48: I would avoid to have http/https specific dispatcher registered, in fact the "protocol" thing speaks the domain language instead of the actual low-level protocol. 
As the current ids protocol is described by `ids-multipart`, and for the upcoming dsp protocol is `dataspace-protocol-http`, I would suggest to define a `callback-event-http` protocol string that will work for both http and https, removing the wrapper thing.

extensions/control-plane/callback/callback-event-dispatcher/src/main/java/org/eclipse/edc/connector/callback/dispatcher/CallbackEventDispatcher.java
```diff
@@ -0,0 +1,73 @@
...
+        var callbacks = eventEnvelope.getPayload().getCallbackAddresses()
+                .stream().filter(cb -> cb.isTransactional() == transactional)
+                .collect(Collectors.toList());
+
+        var eventName = eventEnvelope.getPayload().name();
+
+        for (var callback : callbacks) {
+            if (matches(eventName, callback)) {
+                try {
+                    dispatcher.send(Object.class, new CallbackEventRemoteMessage<>(callback, eventEnvelope)).get();
```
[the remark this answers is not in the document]
jimmarino: This should be tied to the configured transactional behavior of the callback, i.e. block if it is transactional.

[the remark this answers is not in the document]
wolf4ood: Don't we have a default timeout on the `EdcHttpClient` ? 

It's tied to the transactional. If you see the extensions registers two listeners:

- async -> handles notx callbacks
- sync -> handles tx callbacks

and the base logic always `get`, in case we can introduce a  timeout if there isn't one in the `EdcHttpClient`

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/event/Event.java
```diff
@@ -14,9 +14,26 @@
...
+        return new ArrayList<>();
+    }
+
+
+    /**
+     * The name of the event in dot notation.
+     *
+     * @return the event name.
+     */
+    public abstract String name();
```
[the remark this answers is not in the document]
wolf4ood: In this case it's not really needed to be serialized. 
This one it's only to match events in callbacks like `contract.negotiation.initiated`. For the serialized version we have the `type`. If we also need the name we can change that of course

extensions/control-plane/callback/callback-event-dispatcher/src/main/java/org/eclipse/edc/connector/callback/dispatcher/CallbackEventDispatcher.java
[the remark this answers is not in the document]
ndr-brt: I don't think this `CallbackEventDispatcher` could assume that underneath there's an already configured `EdcHttpClient`, because there should be a completely different dispatching protocol (mq...) so it should have it's own timeout.

regarding the sync/async, I saw that we register two listeners, but I don't see the condition to avoid the thread blocking in the async case, I mean, we're filtering the callbacks but we're treating them in the same way, while the async ones shouldn't block any thread

extensions/control-plane/callback/callback-http-dispatcher/src/main/java/org/eclipse/edc/connector/callback/dispatcher/http/CallbackEventDispatcherHttpExtension.java
ndr-brt: I'm not sure we need that, in the same way we don't need such a distinguish in the `ids-multipart` or in the `dataspace-protocol-http`. they will be dispatched in the same way, no differences between `http` and `https`

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/event/Event.java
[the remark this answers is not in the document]
ndr-brt: ok, got it, makes sense

extensions/control-plane/callback/callback-event-dispatcher/src/main/java/org/eclipse/edc/connector/callback/dispatcher/CallbackEventDispatcher.java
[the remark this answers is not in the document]
wolf4ood: Yeah it will block but only the thread in the executor service when running async. I didn't want to make the code complex since the the listener will not block the current thread but only one Thread of the `EventRouter`.  But in case it should be clear on the `EventSubscriber` API when you can block or not.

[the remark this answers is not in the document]
ndr-brt: oh, right, it will be already on another thread, this can be resolved

extensions/control-plane/callback/callback-http-dispatcher/src/main/java/org/eclipse/edc/connector/callback/dispatcher/http/CallbackEventDispatcherHttpExtension.java
wolf4ood: yeah but in this case what drives the protocol is the scheme part of the URL. If we want to change the protocol name to `callback-event-http` probably we would need a mapping between the schema part and `callback-event-http` so in the future we could have another dispatcher different from http/https

PR #2693, opened 2023-04-07
feat: implement `CatalogProtocolService`

### LINKED ISSUE #2492: Support new IDS specification's `Catalog`
# Feature Request

Make the EDC able to request and receive Catalogs following the new IDS specifications.
More details in https://github.com/eclipse-edc/Connector/pull/2448

catalog

new IDS specification

- [x] #2691 
- [x] #2697 
- [x] #2698 
- [x] #2708
- [x] #2726

ndr-brt:
## What this PR changes/adds

Implements `CatalogProtocolService` and `DatasetResolver`.

**Concept**
- the `CatalogController`:
  - calls the `CatalogProtocolService` passing the message and the claim token (as done in the other `ProtocolService`s)
- the `CatalogProtocolService`: 
  - creates the `DataService` related to the connector
  - extracts the `ParticipantAgent` from the `ClaimToken`
  - pass the `ParticipantAgent`, the `QuerySpec` and the `DataService` to the `DatasetResolver`
  - compose the `Catalog` with the resulting datasets 
- the `DatasetResolver`:
  - queries the `ContractDefinition` for the specified `ParticipantAgent`
  - for every definition fetches the `Asset`s and build an `Offer` (is an internal type)
  - the `Offer`s are grouped by `Asset`s
  - for every `Asset` build the `Dataset`

Note: this PR **DOES NOT** implement filtering and pagination, they will come in separated PRs (isses about to come)  

**Open questions**:
- who (and how) should build the `DataService` object? For what I understood in this case it should point to the connector itself, I left a fake url and a TODO to be solved.
- About the `Distribution` `format`, should they come from the registered data planes? e.g. get all the dataplanes and gather all the "allowedDestTypes"? I left a fake type and a TODO to be solved.

## Why it does that

dataspace protocol

## Further notes

- Renamed `ContractDefinitionService` in `ContractDefinitionResolver`, this will avoid the current homonymy with the actual service.

## Linked Issue(s)

Closes #2492 

## Checklist


core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/catalog/CatalogProtocolServiceImpl.java
```diff
@@ -0,0 +1,62 @@
...
+        this.datasetResolver = datasetResolver;
+        this.participantAgentService = participantAgentService;
+    }
+
+    @Override
+    @NotNull
+    public ServiceResult<Catalog> getCatalog(CatalogRequestMessage message, ClaimToken token) {
+        var dataService = DataService.Builder.newInstance()
+                .id(randomUUID().toString())
+                .terms("dspace:connector")
```
jimmarino, line 47: thenamespace shoouldn't be hardcoded. This should be a single constant used by all modules that create dspace messages

spi/control-plane/control-plane-spi/src/main/java/org/eclipse/edc/connector/spi/catalog/DatasetResolver.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Resolves {@link Dataset}s given the {@link ParticipantAgent}, a {@link QuerySpec} and a {@link DataService}
+     *
+     * @param agent the participant agent that requested the dataset.
+     * @param querySpec the query spec for filtering and pagination.
+     * @param dataService the data service to be associated to the datasets.
+     * @return a stream of datasets.
+     */
+    @NotNull
+    Stream<Dataset> query(ParticipantAgent agent, QuerySpec querySpec, DataService dataService);
```
jimmarino, line 39: Passing in the DataService seems a bit strange. Shouldn't the resolver delegate to another service internally to resolve the DataService? For example, different Datasets may be provided by different DataServices.

jimmarino (commented):
I had one question and a small comment

spi/control-plane/control-plane-spi/src/main/java/org/eclipse/edc/connector/spi/catalog/DatasetResolver.java
ndr-brt: The `DataService` objects are needed at the `Catalog` level, and for every `Distribution` we'd need the `DataService` id. To avoid this there could be a `DataServiceResolver` service that could be injected both in the `CatalogService` and in the `DatasetResolver` and retrieve all the `DataService` (that, for a connector will be the connector itself).

I would prefer to address this to a separate issue/PR to provide the functionality incrementally.

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/catalog/CatalogProtocolServiceImpl.java
ndr-brt: removed the `dspace` namespace as it should be added in the json-ld transformation.

jimmarino (approved):

PR #2731, opened 2023-04-15
refactor: adapt RemoteMessage naming to DSP

### LINKED ISSUE #2690: dataspace-protocol: adapt `RemoteMessage` naming to DSP
# Feature Request

Currently, the `RemoteMessage` subclasses have a naming convention that does not adhere with the [dsp specification](https://docs.internationaldataspaces.org/dataspace-protocol/overview/readme), it's time to rename them

core

dataspace protocol

## Solution Proposal
Rename classes, fields, add missing fields, remove unneeded ones.

juliapampus:
## What this PR changes/adds

See commit messages for a list of changes. Values that are only used to set a property that is never used are marked as deprecated & optional and can be removed with the IDS modules.

## Why it does that

Aligns properties and naming of `RemoteMessage` classes to spec documents.

## Further notes

Properties that are still missing will be added later when supporting new protocol functionalities.

## Linked Issue(s)

Closes #2690

## Checklist

### DIFF
--- a/spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/transfer/spi/types/protocol/TransferRequestMessage.java
+++ b/spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/transfer/spi/types/protocol/TransferRequestMessage.java
@@ -31,9 +31,11 @@ public class TransferRequestMessage implements TransferRemoteMessage {
     private String protocol;
     private String id;
     private String contractId;
-    private String assetId; // TODO remove when removing ids module
+    @Deprecated(forRemoval = true)
+    private String assetId;
     private DataAddress dataDestination;
-    private String connectorId; // TODO remove when removing ids module
+    @Deprecated(forRemoval = true)
+    private String connectorId;
     private Map<String, String> properties = new HashMap<>();
 
     @Override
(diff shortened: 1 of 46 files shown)


spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/transfer/spi/types/protocol/TransferRequestMessage.java
```diff
@@ -31,9 +31,9 @@ public class TransferRequestMessage implements TransferRemoteMessage {
     private String protocol;
     private String id;
     private String contractId;
-    private String assetId;
+    private String assetId; // TODO remove when removing ids module
```
ndr-brt, line 34: deprecation will work better than TODO in cases like these.

ndr-brt (approved):
I would only use deprecations instead of TODO's, all good for the rest

spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/transfer/spi/types/protocol/TransferRequestMessage.java
juliapampus: Added 
```java
@Deprecated(forRemoval = true)
```

PR #2752, opened 2023-04-18
refactor: clean up callbackAddress in remote messages

### LINKED ISSUE #2732: refactor: merge callback addresses in ContractRequestMessage
# Feature Request

Unify `connectorAddress` and `callbackAddress` in `ContractRequestMessage`.

spi

We should only use one callback property and decide on its type. 

## Solution Proposal

Rename `connectorAddress` in `RemoteMessage` to align with the DSP.

Decide whether this should be a `List<CallbackAddress>`, `List<String>`, or `String`.

Although currently not included in all message examples of the DSP, this should be an optional property for each message type, as we are supporting async communication.

juliapampus:
## What this PR changes/adds

Rename `connectorAddress` to `callbackAddress`, remove null check where not needed. Add `dataSet` to `ContractRequestMessage`.

## Why it does that

Align with spec.

## Further notes

Renaming has to be applied to the management API as well.

## Linked Issue(s)

Closes #2732

## Checklist


core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -86,13 +89,17 @@ public void stop() {
...
+                .build();
         var negotiation = ContractNegotiation.Builder.newInstance()
                 .id(id)
                 .correlationId(id)
                 .protocol(contractOffer.getProtocol())
                 .counterPartyId(contractOffer.getConnectorId())
-                .counterPartyAddress(contractOffer.getConnectorAddress())
-                .callbackAddresses(contractOffer.getCallbackAddress())
+                .counterPartyAddress(contractOffer.getCallbackAddress())
+                .callbackAddresses(List.of(callbackAddress))
```
ndr-brt, line 102: I think this `callbackAddresses` field shouldn't be filled here, because they are used for the "connector - client" interaction, not for the "connector - connector" one.

ndr-brt (changes requested):
all good for the rest

core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
juliapampus: I applied your feedback. However, this changes previous behavior as the `ContractNegotiationIntegrationTest` was explicitely checking for the callback addresses. Was this wrong before? @ndr-brt

ndr-brt: I'd say so, the `callbackAddresses` are set by the part that initiate the negotiation only in its side, because they don't make sense on the other side, so they shouldn't be passed over the protocol.

ndr-brt (approved):

juliapampus:
Had to fix one still failing unit test, will merge now.

PR #2771, opened 2023-04-19
feat: dataspace protocol negotiation transformer

juliapampus:
## What this PR changes/adds

Implements contract negotiation spi and transformers:
- `:data-protocols:dsp:dsp-negotiation:dsp-negotiation-spi`
- `:data-protocols:dsp:dsp-negotiation:dsp-negotiation-transform`

spec alignments:
- remove null check for `callback` in `ContractAgreementMessage`
- add checksum to `ContractNegotiationEventMessage`
- add `code` in `ContractNegotiationTerminationMessage`
- remove null check for `rejectionReason` in `ContractNegotiationTerminationMessage`
- deprecate `contractStart` and `contractEnd` in `ContractOffer`, see #2758
- add `checksum` to `ContractNegotiation`

## Why it does that

Support the dataspace protocol specification.

## Further notes

Derived issue #2775.

## Linked Issue(s)

Relates #2474
Closes #2764

## Checklist

- [x] added appropriate tests?
- [x] performed checkstyle check locally?
- [x] added/updated copyright headers?
- [x] documented public classes/methods?
- [ ] added/updated relevant documentation?
- [x] assigned appropriate label? (exclude from changelog with label `no-changelog`)
- [x] formatted title correctly? (_take a look at the [CONTRIBUTING](https://github.com/eclipse-edc/Connector/blob/main/CONTRIBUTING.md#submit-a-pull-request) and [Etiquette for pull requests](https://github.com/eclipse-edc/Connector/blob/main/pr_etiquette.md) for details_)

PR #2760, opened 2023-04-19
feat: dataspace protocol transfer-process-dispatcher|spi

janpmeyer:
## What this PR changes/adds

Implements transfer-process spi and dispatcher following :dsp:dsp-catalog.

`:data-protocols:dsp:dsp-transfer-process:dsp-transfer-process-dispatcher`
`:data-protocols:dsp:dsp-transfer-process:dsp-transfer-process-spi`

## Why it does that

Support the dataspace protocol specification.

## Further notes

Waits for PR #2759 to be merged

## Linked Issue(s)

Relates #2474

## Checklist

- [x] added appropriate tests?
- [x] performed checkstyle check locally?
- [x] added/updated copyright headers?
- [ ] documented public classes/methods?
- [ ] added/updated relevant documentation?
- [x] assigned appropriate label? (exclude from changelog with label `no-changelog`)
- [x] formatted title correctly? (_take a look at the [CONTRIBUTING](https://github.com/eclipse-edc/Connector/blob/main/CONTRIBUTING.md#submit-a-pull-request) and [Etiquette for pull requests](https://github.com/eclipse-edc/Connector/blob/main/pr_etiquette.md) for details_)

PR #2780, opened 2023-04-20
feat: dataspace protocol negotiation dispatcher

juliapampus:
## What this PR changes/adds

Adds delegates for dsp negotiation messages.

## Why it does that

Implement dataspace protocol

## Further notes

Adds dependencies:
- `:data-protocols:dsp:dsp-negotiation:dsp-negotiation-spi`
- `:data-protocols:dsp:dsp-negotiation:dsp-negotiation-transform` (classes are contributed with #2771)

## Linked Issue(s)

Relates #2474

## Checklist


jimmarino (commented):
Couldn't some of the boilerplate code around creating a JSON structure, HTTP request and dispatching it be pulled up into an abstract class?

juliapampus:
> Couldn't some of the boilerplate code around creating a JSON structure, HTTP request and dispatching it be pulled up into an abstract class?

Agree. When you are okay with it, I would open an issue for this refactoring. I guess we could also merge this with the transformer dispatchers and add the abstract class to the `dsp-http-core` module.

jimmarino:
I think we should refactor this PR to include the abstract class because it will simplify the code, which contains a lot of duplication. Later we can move the abstract class out in a subsequent PR.

ndr-brt (commented):
> Couldn't some of the boilerplate code around creating a JSON structure, HTTP request and dispatching it be pulled up into an abstract class?

I would prefer a collaborator, its behavior is pretty clear, serialize a `RemoteMessage`, it could be an interface called `RemoteMessageSerializer` with the `JsonLdRemoteMessageSerializer` implementation.

I agree that this should be done in this PR, it won't take much and it will simplify tests a lot

juliapampus:
PR is rebased. @ronjaquensel will take care of refactoring. The refactored structure will then also be used in #2760.

ronjaquensel:
@ndr-brt @jimmarino I refactored the delegates as follows:

* Introcuded `JsonLdRemoteMessageSerializer` interface and implementation in `dsp-http-spi` and `dsp-http-core` respectively (for now, can be moved further up when required). This serializer is now used by the delegates to transform, compact and serialize a message.
* Made `DspHttpDispatcherDelegate` an abstract class and moved the code for creating the HTTP request there. As this is specific to HTTP dispatchers, I think it makes more sense in the abstract class than as an additional helper class. But as this is just a single method, it should not make the code harder to read/understand.
* Introduced a `testFixture` for delegates, as pretty much the same code was duplicated between many of the delegate tests. The `testFixture` itself does not contain `@Test` annotated methods, but instead provides methods for testing the common behaviour that can be used in the delegate tests. First, I wanted to move the actual tests to the `testFixture` class, but as the delegate tests use these methods in different combinations (e.g. some parse the response body, others don't), that quickly became rather confusing and unreadable.

jimmarino (approved):

ndr-brt (approved):

PR #2825, opened 2023-04-27
feat(Callbacks): propagate the EDR in dataAddress form with events

### LINKED ISSUE #2819: Callback implementation : propagate the EDR
# Feature Request

As part of the story [#2626](https://github.com/eclipse-edc/Connector/issues/2626) and the DR about callbacks, 
the EDR should be integrated in callbacks  in the `TransferProcessStarted` event

callbacks, events, tpm

Use callbacks for receiving the EDR

wolf4ood:
## What this PR changes/adds

Adds the `dataAddress` field on `TransferProcessStarted` event which will be populated on consumer side upon receiving the dsp message about the transfer process started

## Why it does that

Support for EDR in callbacks

## Linked Issue(s)

Closes #2819 

## Checklist

### DIFF
--- a/spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/transfer/spi/observe/TransferProcessStartedData.java
+++ b/spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/transfer/spi/observe/TransferProcessStartedData.java
@@ -1,0 +1,55 @@
+/*
+ *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.connector.transfer.spi.observe;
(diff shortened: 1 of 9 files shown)


jimmarino (approved):

spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/transfer/spi/observe/TransferProcessStartedData.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.eclipse.edc.connector.transfer.spi.observe;
+
+import org.eclipse.edc.connector.transfer.spi.types.TransferProcess;
+import org.eclipse.edc.spi.types.domain.DataAddress;
+
+/**
+ * Additional data when calling {@link TransferProcessListener#started} not stored in the {@link TransferProcess}
+ */
+public class TransferProcessStartedData {
```
paullatzelsperger, line 23: Are you anticipating more items than just the `DataAddress` to be communicated through the event listener? Because if not, we could just pass the plain `DataAddress` into the listener method instead of this wrapper.

paullatzelsperger (approved):

spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/transfer/spi/observe/TransferProcessStartedData.java
wolf4ood: Not for now, since I'm not sure if the `TransferStartMessage` is complete and the listener it's an interface i didn't want to hardcode only the `DataAddress`

PR #2858, opened 2023-04-28
fix(gradle.properties): increases maximum heap size during gradle build

saschaisele-zf:
## What this PR changes/adds

This PR adds a line in the `gradle.properties` file that increses the heap size during the build to 2048MB.

## Why it does that

Without this change, the automatic and manual building in the IDE will fail with an error message, that there is not enough space on the heap.

## Checklist

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,3 +10,4 @@ metaModelVersion=0.0.1-SNAPSHOT
 edcScmConnection=scm:git:git@github.com:eclipse-edc/Connector.git
 edcWebsiteUrl=https://github.com/eclipse-edc/Connector.git
 edcScmUrl=https://github.com/eclipse-edc/Connector.git
+org.gradle.jvmargs=-Xmx2048m


jimmarino:
This doesn't need to be added here and can be set in the user's ~/.gradle/gradle.properties directory.

saschaisele-zf:
I would mention it in `the Checkout and build from source` section of the README then.
As things stand building just automatically fails for new users which makes for a bad first experience.

ndr-brt:
> I would mention it in the `Checkout and build from source` section of the README then. As things stand building just automatically fails for new users which makes for a bad first experience.

We could do that, but it's a gradle issue, not an EDC one, and I don't think we should explain people how to configure dev tools like gradle or their IDE. Maybe a link to the gradle documentation page about that will be enough:
https://docs.gradle.org/current/userguide/build_environment.html

jimmarino:
> > I would mention it in the `Checkout and build from source` section of the README then. As things stand building just automatically fails for new users which makes for a bad first experience.
> 
> We could do that, but it's a gradle issue, not an EDC one, and I don't think we should explain people how to configure dev tools like gradle or their IDE. Maybe a link to the gradle documentation page about that will be enough: https://docs.gradle.org/current/userguide/build_environment.html

Agreed. I also don't think we should include links as this becomes difficult to maintain as their number increases.

saschaisele-zf:
Instead of telling people how to configure their tools, i would then simply mention a minimum heap size as a requirement for building.

ndr-brt:
> Instead of telling people how to configure their tools, i would then simply mention a minimum heap size as a requirement for building.

I don't think it's a requirement... my setting is `-Xmx1024M` and it works

PR #2863, opened 2023-04-29
feat: add `JsonObject` from/to `PolicyDefinition` transformers

### LINKED ISSUE #2769: Create `JsonObject[To|From]PolicyDefinitionTransformer`

ndr-brt:
## What this PR changes/adds

- Implements `JsonObject` from/to  `PolicyDefinition`
- the process implemented in the new controller is 
  - input: `JsonObject` -> expansion -> transform to dto -> transform to entity
  - output: entity -> transform to dto -> compaction -> `JsonObject`

## Why it does that

dataspace protocol

## Further notes

- registered the json-ld object mapper once in `ManagementApiConfigurationExtension`
- introduced `EDC_PREFIX` constant
- got rid of the `*New*Dto` classes, as they became useless
- I'm not sure why we are keeping two DTOs, one for creation (`RequestDto`) and one for modification (`UpdateDto`): they could be the same (this would save us two transformers and a wrapper class). Maybe something for a subsequent PR
- I think some additional effort needs to be made for the swagger documentation

## Linked Issue(s)

Closes #2769 

## Checklist


extensions/common/api/management-api-configuration/src/main/java/org/eclipse/edc/connector/api/management/configuration/ManagementApiConfigurationExtension.java
```diff
@@ -105,5 +110,8 @@ public void initialize(ServiceExtensionContext context) {
 
         context.registerService(ManagementApiConfiguration.class, new ManagementApiConfiguration(webServiceConfiguration));
         webService.registerResource(webServiceConfiguration.getContextAlias(), new AuthenticationRequestFilter(authenticationService));
+
+        var jsonLdMapper = typeManager.getMapper("json-ld");
```
paullatzelsperger, line 114: nit: there is a constant for this in `JsonLdExtension.java`, which - if moved out to maybe `CoreConstants.java` or something, could be re-used.

paullatzelsperger (commented):
regarding multiple DTOs for request/response: if we don't need them we could get rid of them, but AFAIR when we devised them it made sense for multiple reasons e.g. different validation, or the fact that only the `ResponseDto` has a `createdAt` field

paullatzelsperger:
Just a quick question: we are now converting twice, once from `JsonObject` to `Dto`, and then from `Dto` to the entity. Technically, one single conversion would suffice. 
Are we doing that just to have the DTO in the Swagger doc?

ndr-brt:
> regarding multiple DTOs for request/response: if we don't need them we could get rid of them, but AFAIR when we devised them it made sense for multiple reasons e.g. different validation, or the fact that only the `ResponseDto` has a `createdAt` field

`ResponseDto` is different because, as you said, it has the `createdAt`, I was referring to `RequestDto` and `UpdateDto`, the only difference is the latter doesn't have the `id` but, to be fair, it would make sense to have the id both in the path and in the `Dto`, this would save us some transformers and the wrapper.



> Just a quick question: we are now converting twice, once from `JsonObject` to `Dto`, and then from `Dto` to the entity. Technically, one single conversion would suffice. Are we doing that just to have the DTO in the Swagger doc?

The main reason is to have a convention because, in this case the `Dto` and the entity map pretty much 1:1, but in other entities the `Dto` is a stripped down version of the entity. In these case that middle conversion will be mandatory.

Regarding swagger documentation, I think the `Dto` won't be enough because we'd like to list Json-LD specific properties as `@context`, `@type`, `@id`..., things that for the moment are not listed, I would discuss this topic separately.

extensions/common/api/management-api-configuration/src/main/java/org/eclipse/edc/connector/api/management/configuration/ManagementApiConfigurationExtension.java
ndr-brt: Not a huge fan of global generic purpose constants, however for the moment I put it there as suggested

paullatzelsperger:
> `ResponseDto` is different because, as you said, it has the `createdAt`, I was referring to `RequestDto` and `UpdateDto`, the only difference is the latter doesn't have the `id` but, to be fair, it would make sense to have the id both in the path and in the `Dto`, this would save us some transformers and the wrapper.
> 
got it. yep, generally I think that would be fine.

> The main reason is to have a convention because, in this case the `Dto` and the entity map pretty much 1:1, but in other entities the `Dto` is a stripped down version of the entity. In these case that middle conversion will be mandatory.

But couldn't we just do the "Stripping" in the `JsonObjectFrom<ENTITY>Transformer` then? Much the same as we are doing with DTOs anyway?

> 
> Regarding swagger documentation, I think the `Dto` won't be enough because we'd like to list Json-LD specific properties as `@context`, `@type`, `@id`..., things that for the moment are not listed, I would discuss this topic separately.

Maybe there is a way to solve this generically, e.g. through something like a DTO base class, that declares those JSON-LD props?

paullatzelsperger (approved):

PR #2877, opened 2023-05-03
fix: rationalize provider and consumer id handling

### LINKED ISSUE #2841: Fix handling of consumer and provider ids in `ContractNegotiation,` `ContractOffer,` and `ContractAgreement`
The provider and consumer ids must be rationalized across the runtime to resolve several subtle bugs and to drive consistency across different architectural layers.

The consumer and provider ids are of type URI on `ContactOffer` and type String on `ContractAgreement.` This was to accommodate the old IDS protocol, which has now been changed to String. When creating a `ContractAgreeement` in `ProviderContractNegotiationManagerImpl` the conversion from URI to string is performed as follows: 

```
.providerAgentId(String.valueOf(lastOffer.getProvider()))
.consumerAgentId(String.valueOf(lastOffer.getConsumer()))   
```

The conversion masked a bug where the last offer had null values for the ids, which were converted to "null." In addition, the last offer should not be used to set the ids for the contract agreement, as this may be susceptible to manipulation. Rather, the consumer id should be taken from the contract negotiation and the provider id from the connector configuration.
  

1. Rationalize consumer and provider ids in ContractOffer and ContractAgreement:
1. `ContractOffer.getConsumer` and `ContractOffer.getProvider` should be renamed to `getXXXXId` and return a `String`

3. `ContractAgreement.getConsumerAgentId` and `ContractAgreement.getProviderAgentId` should be renamed to `get[Consumer|Provider]Id`

jimmarino:
## What this PR changes/adds

As explained in #2841 

The provider and consumer ids must be rationalized across the runtime to resolve several subtle bugs and to drive consistency across different architectural layers:

1. Migrate the consumer and provider ids are of type URI on ContactOffer to type String to align with ContractAgreement. 
2. Rename ContractAgreement.getConsumerAgentId and ContractAgreement.getProviderAgentId to get[Consumer|Provider]Id
3. Fix bugs where null URI ids were set as "null" strings
4. Fix a potential exploit where the consumer and provider ids were taken from the last offer. The consumer id is taken from the contract negotiation (which is derived from a token claim) and the provider id is a configured value (see below).
5. Add a configuration value for the provider id `edc.participant.id`. If not configured, the default will use `anonymous` as a value and issue a warning. Note that the default connector id where a random value is generated cannot be used since the participant identity must remain stable across runtime restarts.

Note the following cleanup will be done in subsequent PRs:

1. `QuerySpec` can be cleaned up to not include provider and consumer ids
2. The contract offer does not need to include the consumer id
3. The consumer and provider ids need to be added to the new DSP contract agreement message. The DSP specification also needs to be updated to include these properties (there is an open issue in the IDS repo for this)

## Why it does that

Fixes bugs and makes the codebase consistent.

## Linked Issue(s)

Closes #2841 

## Checklist


spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/types/agreement/ContractAgreement.java
```diff
@@ -68,29 +68,25 @@ public String getId() {
...
-     * The id of the data providing agent.
+     * The id of the data providing participant.
      * Please note that id should be taken from the corresponding data ecosystem.
-     * For example: In IDS the connector uses a URI from the IDS Information Model as ID. If the contract was
-     * negotiated inside the IDS ecosystem, this URI should be used here.
      *
      * @return provider id
      */
     @NotNull
-    public String getProviderAgentId() {
```
ndr-brt, line 79: wouldn't this be a breaking change? Maybe it would be better to rename the fields, adding new getters and builder methods and deprecating the old ones, it should be enough to avoid problems with the already persisted agreements (in the cosmos store, because in the sql one there will be no issue)

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/system/ServiceExtensionContext.java
```diff
@@ -26,6 +27,11 @@
  */
 public interface ServiceExtensionContext extends SettingResolver {
 
+    @Setting(value = "Configures the participant id this runtime is operating on behalf of")
+    String PARTICIPANT_ID = "edc.participant.id";
+
+    String ANONYMOUS_PARTICIPANT = "anonymous";
```
ndr-brt, line 33: this can be set as `defaultValue` of the `PARTICIPANT_ID` `@Setting` annotation

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/types/offer/ContractOffer.java
```diff
@@ -81,13 +80,13 @@ public String getId() {
     }
 
     @Nullable
-    public URI getProvider() {
```
ndr-brt, line 84: breaking change: the old getters and builder methods should stay and been deprecated

ndr-brt (commented):
all good, just a pair of breaking changes notes

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/types/offer/ContractOffer.java
jimmarino: But we're not supporting (and can't support) backward compatibility. This feature also never worked. So what's the point?

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/types/agreement/ContractAgreement.java
jimmarino: Unfortunately, there's no way to avoid this due to the validity checks. Since we are not supporting backward compatibility, people will need to migrate their databases

wolf4ood: The SQL should not be a problem, for Cosmos, that we serialize/deserialize the `ContractAgreement` might.  
We could leave the old builder method in place depracated or if we choose to breake the changes here we need to remember this and put it in the release notes probably.

jimmarino: Yeah, I think we should not keep the old builder methods because there is no way to get this to work without migrating the serialized JSON form anyway and having the methods can lead to people using them by accident in the future. This is basically a clean break along with the protocol changes.

wolf4ood (approved):

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/types/agreement/ContractAgreement.java
jimmarino: We definitely need to put this in the documentation along with how to configure the participant id. I've opened an issue for that here: #2887.

ndr-brt (approved):

PR #2893, opened 2023-05-07
feat: use consumer id from claim token and remove from contract offer

### LINKED ISSUE #2892: Remove consumerId id from contract offer
Contract offer should not contain the consumer id as this should be based on a validated security token. This field will be removed in `ContractOffer` and the consumer id will be taken from the `ParticipantAgent` derived from the client `ClaimToken`.

jimmarino:
## What this PR changes/adds

This PR uses the consumer id from `ParticipantAgent` derived from the client's `ClaimToken` so it is rooted in a trust anchor. The `consumerId` field is removed from the `ContractOffer` as it represents a potential path to introduce security exploits.

One additional fix was added to `ContractValidationServiceImpl` to ensure that the target of the policy matches the asset referenced by the offer :

```
        // if policy target is null, default to the asset, otherwise validate it
        var policyTarget = offer.getPolicy().getTarget() != null ? offer.getPolicy().getTarget() : targetAsset.getId();
        if (!targetAsset.getId().equals(policyTarget)) {
            return failure(format("Contract offer asset '%s' does not match policy target: %s", offer.getAsset().getId(), policyTarget));
        }
```
This technically should be done in another PR, but it is a small additional check in an area of the code that is already being re-worked.

## Why it does that

Security and code clarity.

## Linked Issue(s)

Closes #2892

## Checklist


core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImpl.java
```diff
@@ -38,25 +38,33 @@
...
 import static java.lang.Math.max;
 import static java.lang.Math.min;
+import static java.util.stream.Collectors.toList;
 import static java.util.stream.Stream.concat;
 
 /**
  * Implementation of the {@link ContractOfferResolver}.
  */
 public class ContractOfferResolverImpl implements ContractOfferResolver {
+    private final String participantId;
```
paullatzelsperger, line 52: just as an idea: maybe there could be a method `getOwnParticipantId()` or similar on the `ParticipantService` instead of the explicit `participantId` argument?

paullatzelsperger (approved):

core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/offer/ContractOfferResolverImpl.java
jimmarino: There is one on `ServiceExtensionContext`, but `ParticipantAgentService` is for foreign participants, not the current runtime.

wolf4ood (approved):
LGTM

PR #2917, opened 2023-05-09
feat(dsp): add Error Responses to Catalog|TransferProcess|ContractNegotiation-API

### LINKED ISSUE #2796: dataspace protocol: add error support
# Feature Request

The protocol defines `CatalogError`, `NegotiationError`, and `TransferError`. These should be supported.

dsp, api

dataspace protocol support

## Solution Proposal
Return in case of failure. Could be something that an `ExceptionMapper` could take care of.

janpmeyer:
## What this PR changes/adds

Adding `TransferError`, `CatalogError` and `ContractNegotiationError` to the Dsp-APIs.

Add `JsonObjectFromTransferErrorTransformer`, `JsonObjectFromCatalogErrorTransformer` and `JsonObjectFromContractNegotiationErrorTransformer` to the Dsp-Transformer.

Add `DspHttpStatusCodeMapper` to the Dsp-Spi.

Add `DspHttpStatusCodeMapper` to the Dsp-Core.

## Why it does that

To get align with the dataspace protocol spec.

## Linked Issue(s)

Closes #2796 

## Checklist


data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
```diff
@@ -74,39 +78,46 @@ public CatalogController(Monitor monitor, ObjectMapper mapper, IdentityService i
...
-        }
-        var expandedJson = expanded.getContent();
-        if (!isOfExpectedType(expandedJson, DSPACE_CATALOG_REQUEST_TYPE)) {
-            throw new InvalidRequestException(format("Request body was not of expected type: %s", DSPACE_CATALOG_REQUEST_TYPE));
-        }
+            var claimToken = identityService.verifyJwtToken(tokenRepresentation, dspCallbackAddress)
+                    .orElseThrow(failure -> new AuthenticationFailedException());
+
+            var expanded = jsonLdService.expand(jsonObject);
+            if (expanded.failed()) {
```
ndr-brt, line 93: `orElseThrow(InvalidRequestException::new)`

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
```diff
@@ -74,39 +78,46 @@ public CatalogController(Monitor monitor, ObjectMapper mapper, IdentityService i
...
-        var catalogJson = transformerRegistry.transform(catalog, JsonObject.class)
-                .orElseThrow(failure -> new EdcException(format("Failed to build response: %s", failure.getFailureDetail())));
+            var catalogJson = transformerRegistry.transform(catalog, JsonObject.class)
+                    .orElseThrow(failure -> new EdcException(format("Failed to build response: %s", failure.getFailureDetail())));
 
-        var compacted = jsonLdService.compact(catalogJson);
-        if (compacted.failed()) {
-            throw new InvalidRequestException(compacted.getFailureDetail());
+            var compacted = jsonLdService.compact(catalogJson);
+            if (compacted.failed()) {
```
ndr-brt, line 111: ditto

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
```diff
@@ -74,39 +78,46 @@ public CatalogController(Monitor monitor, ObjectMapper mapper, IdentityService i
...
+                throw new InvalidRequestException(compacted.getFailureDetail());
```
ndr-brt, line 112: if it fails in compaction is not a request exception, better go with an EdcException (500)

data-protocols/dsp/dsp-negotiation/dsp-negotiation-api/src/test/java/org/eclipse/edc/protocol/dsp/negotiation/api/controller/DspNegotiationControllerTest.java
```diff
@@ -193,11 +196,19 @@ void getNegotiation_shouldReturnNotImplemented_whenOperationNotSupported() {
...
         //operation not yet supported
-        baseRequest()
+        var result = baseRequest()
                 .get(BASE_PATH + "testId")
                 .then()
-                .statusCode(501);
+                .statusCode(501)
+                .extract().as(Map.class);
+
+        monitor.debug(result.keySet().toString());
```
ndr-brt, line 208: please remove log

data-protocols/dsp/dsp-http-core/src/main/java/org/eclipse/edc/protocol/dsp/util/ErrorUtil.java
```diff
@@ -0,0 +1,90 @@
...
+public class ErrorUtil {
+
+    /** Creating ErrorResponse
+     *
+     * @param type dspace:ContractNegotiationError/dspace:TransferProcessError/dspace:CatalogError depending on executing Controller
+     * @param processId referenced processId Error is occurring
+     * @param throwable describing ErrorMessage
+     * @return {@link Response} ErrorResponse
+     */
+    public static Response createErrorResponse(String type, Optional<String> processId, Throwable throwable) {
```
ndr-brt, line 56: to follow an approach more similar to the existing, we could introduce three pojos:
- `CatalogErrorMessage`
- `ContractNegotiationErrorMessage`
- `TransferProcessErrorMessage`

each one containing a field `processId` and a field with the throwable.
and for each one implement a `*ErrorToResponse` that will take care of creating the `Response` with the `JsonObject`. 
this will simplify tests, separate responsibilities, avoid the util class and remove unneeded direct dependencies to the `dsp-http-core` module.

ndr-brt (commented):
I will suggest to use transformers to create the error response

data-protocols/dsp/dsp-catalog/dsp-catalog-api/build.gradle.kts
```diff
@@ -20,6 +20,7 @@ dependencies {
     api(project(":data-protocols:dsp:dsp-api-configuration"))
     api(project(":data-protocols:dsp:dsp-catalog:dsp-catalog-transform"))
     api(project(":data-protocols:dsp:dsp-transform"))
+    api(project(":data-protocols:dsp:dsp-http-core"))
```
jimmarino, line 23: Why is this dependency needed?

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
```diff
@@ -74,39 +76,41 @@ public CatalogController(Monitor monitor, ObjectMapper mapper, IdentityService i
...
 
-        var compacted = jsonLdService.compact(catalogJson);
-        if (compacted.failed()) {
-            throw new InvalidRequestException(compacted.getFailureDetail());
+            var compacted = jsonLdService.compact(catalogJson).orElseThrow(failure -> new EdcException(failure.getFailureDetail()));
+
+            var result = mapper.convertValue(compacted, Map.class);
+
+            return Response.status(Response.Status.OK).entity(result).build();
+        } catch (Throwable throwable) {
```
jimmarino, line 110: Catching `Throwable` is generally not a good thing since that will also trap `Error` and its subclasses which should not be done. Better to catch `EdcException` or `RuntimeException` (preferably the former).

data-protocols/dsp/dsp-catalog/dsp-catalog-transform/src/main/java/org/eclipse/edc/protocol/dsp/catalog/transform/CatalogError.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ *  Contributors:
+ *       Fraunhofer Institute for Software and Systems Engineering - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.protocol.dsp.catalog.transform;
+
+public class CatalogError {
+    private Throwable throwable;
```
jimmarino, line 18: Ditto comment about Throwable

data-protocols/dsp/dsp-catalog/dsp-catalog-transform/build.gradle.kts
```diff
@@ -21,4 +21,7 @@ dependencies {
     api(project(":spi:common:json-ld-spi"))
     api(project(":extensions:common:json-ld"))
     api(project(":data-protocols:dsp:dsp-http-spi"))
+    api(project(":spi:common:web-spi"))
+
+    implementation(libs.bundles.jersey.core)
```
jimmarino, line 26: Why does this rely on Jersey?

data-protocols/dsp/dsp-catalog/dsp-catalog-transform/src/main/java/org/eclipse/edc/protocol/dsp/catalog/transform/CatalogErrorToResponseTransformer.java
```diff
@@ -0,0 +1,85 @@
...
+import static jakarta.ws.rs.core.Response.Status.CONFLICT;
+import static jakarta.ws.rs.core.Response.Status.FORBIDDEN;
+import static jakarta.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static jakarta.ws.rs.core.Response.Status.NOT_FOUND;
+import static jakarta.ws.rs.core.Response.Status.NOT_IMPLEMENTED;
+import static jakarta.ws.rs.core.Response.Status.UNAUTHORIZED;
+import static org.eclipse.edc.jsonld.spi.Namespaces.DSPACE_SCHEMA;
+import static org.eclipse.edc.protocol.dsp.catalog.transform.DspCatalogPropertyAndTypeNames.DSPACE_CATALOG_ERROR;
+
+public class CatalogErrorToResponseTransformer extends AbstractJsonLdTransformer<CatalogError, Response> {
```
jimmarino, line 45: Transformers should not know about transport-specific classes. Please re-think this area of your PR

jimmarino (changes requested):
There are a few specific issues in the comments. However, this PR has a broader issue: the transformers should not know about transport-specific classes. Please introduce a proper decoupling between these layers.

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
ronjaquensel: But for protocol communication, shouldn't an error response in the correct format be returned for any problem that occurs? If we only catch e.g. `EdcException` here, for any other exception/error that may occur, a Jersey default error message would be returned to the requesting connector. That doesn't seem right.

data-protocols/dsp/dsp-catalog/dsp-catalog-transform/src/main/java/org/eclipse/edc/protocol/dsp/catalog/transform/CatalogErrorToResponseTransformer.java
ndr-brt: this was a suggestion of mine, because the status of the `Response` comes out from the `Error` type, otherwise we'd need to transform the error in the response body and separately map the exception to statusCode.
If we don't want that, the alternative would be to convert the `Error` to JsonObject and then another component (like `DspHttpStatusCodeMapper`) that maps an exception to a status code.

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
jimmarino: Not if it is an `Error` - those should not be caught.

data-protocols/dsp/dsp-catalog/dsp-catalog-transform/src/main/java/org/eclipse/edc/protocol/dsp/catalog/transform/CatalogErrorToResponseTransformer.java
janpmeyer: This solution would result in having the status code mapping duplicated, between the Transformer and the DspHttpStatusCodeMapper as the transformer also need to put an Error code into the Json.

Is this duplication of code for you ok?

ndr-brt: nope, because the mapper will be a component registered once and used between the different modules.
With the dependency injection pattern we're following, shared implementations should be avoided (shared static methods as well), only spi should be shared.

data-protocols/dsp/dsp-catalog/dsp-catalog-api/build.gradle.kts
janpmeyer: Dependency removed

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/CatalogController.java
janpmeyer: Changed from Throwable to Exception.

data-protocols/dsp/dsp-catalog/dsp-catalog-transform/src/main/java/org/eclipse/edc/protocol/dsp/catalog/transform/CatalogError.java
janpmeyer: Changed from Throwable to Exception.

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/DspCatalogApiController.java
```diff
@@ -60,58 +63,78 @@ public class DspCatalogApiController {
...
-        if (compacted.failed()) {
-            throw new InvalidRequestException(compacted.getFailureDetail());
+    private Response createResponse(JsonObject jsonEntity, Exception exception) {
+        JsonObject compacted = null;
+        try {
+            compacted = jsonLdService.compact(jsonEntity)
+                    .orElseThrow(failure -> new EdcException("Failed to compact JSON-LD."));
+            return Response.status(statusCodeMapper.mapErrorToStatusCode(exception)).entity(compacted).build();
+        } catch (EdcException e) {
+            return Response.status(500).entity("{\"message\": \"Failed to create response\"").build();
```
ndr-brt, line 137: log this unexpected exception, same in the other controllers

data-protocols/dsp/dsp-catalog/dsp-catalog-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/api/controller/DspCatalogApiController.java
```diff
@@ -60,58 +63,78 @@ public class DspCatalogApiController {
...
+
+            var catalogJson = transformerRegistry.transform(catalog, JsonObject.class)
+                    .orElseThrow(failure -> new EdcException(format("Failed to build response: %s", failure.getFailureDetail())));
+
+            var compacted = jsonLdService.compact(catalogJson);
+            if (compacted.failed()) {
+                throw new InvalidRequestException(compacted.getFailureDetail());
+            }
+            //noinspection unchecked
+            var result = mapper.convertValue(compacted.getContent(), Map.class);
```
ndr-brt, line 120: I always wondered... do we need this conversion? jersey should be able to serialize `JsonObject` correctly without converting it to a map before. Maybe something for another issue/PR

data-protocols/dsp/dsp-catalog/dsp-catalog-transform/src/main/java/org/eclipse/edc/protocol/dsp/catalog/transform/from/JsonObjectFromCatalogErrorTransformer.java
```diff
@@ -0,0 +1,56 @@
...
+import org.eclipse.edc.transform.spi.TransformerContext;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import static org.eclipse.edc.jsonld.spi.Namespaces.DSPACE_SCHEMA;
+import static org.eclipse.edc.protocol.dsp.catalog.transform.DspCatalogPropertyAndTypeNames.DSPACE_CATALOG_ERROR;
+
+public class JsonObjectFromCatalogErrorTransformer extends AbstractJsonLdTransformer<CatalogError, JsonObject> {
+
+    private DspHttpStatusCodeMapper statusCodeMapper;
```
ndr-brt, line 32: final

ndr-brt (commented):
I think it's pretty much ready.
I guess on the long run the error code will be different from the http status, but for the moment it will be enough

ndr-brt (approved):

PR #2927, opened 2023-05-09
feat(core): validate that requested asset is contained in ContractDefinition

### LINKED ISSUE #2894: Asset Id on initial contract offer must be validated
`ContractValidationServiceImpl.validateInitialOffer(ClaimToken token, ContractOffer offer)` must validate that the asset id specified by the contract offer is selected by the contract definition the offer is derived from.

paullatzelsperger:
## What this PR changes/adds

This PR adds a validation to the `ContractValidationServiceImpl#validateInitialOffer`, which ensures that the `assetId` that is contained in the incoming `ContractOffer`, is indeed referenced by the `AssetSelectorExpression` of the `ContractDefinition`', based on which the offer was supposedly created.

## Why it does that

Not verifying this could be a potential avenue for malicious actors/attackers to negotiate a contract for an asset, that was not offered to them.

## Further notes

- The algorithm is as follows: a new `Criterion` ("assetId = xxx") is added to the selector expression, which is then used in a `count` query against the `AssetIndex`. If the count comes back non-zero, the check succeeds, otherwise fails. This works because by convention multiple criteria are conjoined with a logical `and` at the DB level.

- I added assertions on the failure method to a couple of tests, since asserting that the validation failed is not sufficient.

## Alternative approaches that were ultimately discarded:
- using the `AssetPredicateConverter` to convert the selector expression into a `Predicate`, and testing the `Asset` against it. 
  _Discarded because of potential disparity in query capabilities of the `AssetPredicateConverter` and the actual persistence impl._
- adding a new method to the `AssetIndex` that performs the check on DB level. 
  _Discarded because disproportionate impact on code base_
- performing a query based on the selector expression, and checking in-mem if the Asset matches any. 
  _Discarded because that would have potentially materialized a huge stream in memory._

## Linked Issue(s)

Closes #2894 

## Checklist

### DIFF
--- a/core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/validation/ContractValidationServiceImpl.java
+++ b/core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/validation/ContractValidationServiceImpl.java
@@ -29,11 +29,14 @@ import org.eclipse.edc.policy.engine.spi.PolicyEngine;
 import org.eclipse.edc.spi.agent.ParticipantAgentService;
 import org.eclipse.edc.spi.asset.AssetIndex;
 import org.eclipse.edc.spi.iam.ClaimToken;
+import org.eclipse.edc.spi.query.Criterion;
 import org.eclipse.edc.spi.result.Result;
+import org.eclipse.edc.spi.types.domain.asset.Asset;
 import org.jetbrains.annotations.NotNull;
 
 import java.time.Clock;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
 import java.util.HashMap;
 
 import static java.lang.String.format;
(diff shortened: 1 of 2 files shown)


core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/validation/ContractValidationServiceImpl.java
```diff
@@ -102,6 +105,13 @@ public Result<ValidatedConsumerOffer> validateInitialOffer(ClaimToken token, Con
             return failure("Invalid target: " + offer.getAssetId());
         }
 
+        // verify that the asset in the offer is actually in the contract definition
+        var testCriteria = new ArrayList<>(contractDefinition.getSelectorExpression().getCriteria());
+        testCriteria.add(new Criterion(Asset.PROPERTY_ID, "=", offer.getAssetId()));
+        if (assetIndex.countAssets(testCriteria) <= 0) {
+            return failure("Asset ID from the ContractOffer is not included in the ContractDefinition!");
```
jimmarino, line 112: minor nit, can we remove the "!"?

paullatzelsperger: done

PR #2935, opened 2023-05-10
fix: make `ContractNegotiation` managers fetch negotiation with correct type

### LINKED ISSUE #2923: `ContractNegotiation` managers are fetching the wrong entities
# Bug Report

## Describe the Bug
`ConsumerContractNegotiationManager` is fetching `PROVIDER` negotiations and other way around.

### Expected Behavior
this shouldn't happen

## Possible Implementation
add the type as filter to the `nextForState` method

ndr-brt:
## What this PR changes/adds

Introducing a new `nextNotLeased` method in `StateEntityStore` that will permit to filter not leased entity by a custom filter and not only by state as the `nextForState` permitted.

## Why it does that

fix bug

## Further notes

- **BREAKING CHANGE** in SQL ContractNegotiation schema. for some reason, the `type` field was not a string containing "CONSUMER" or "PROVIDER" as done for TransferProcess, but an int containing the `Type` enum ordinal value. This was not only blocking the bugfix, but also dangerous because changing the enum entries order would have been caused a data inconsistency.
  - migration suggestion: to avoid data loss the advice would be to implement a sql migration that will do:
    - create a new column called `type_temp` of type vararg
    - fill the `type_temp` with `CONSUMER` where `type` is 0 and with `PROVIDER` where `type` is 1
    - remove the `type` column
    - rename `type_temp` to `type`
- `nextForState` has been deprecated and all its usages switched to `nextNotLeased`
- `SqlConditionExpression.toStatementParameter` has been stopped from converting `Object`s to `String`s

## Linked Issue(s)

Closes #2923

## Checklist

### DIFF
--- a/core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
+++ b/core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
@@ -30,6 +30,7 @@ import org.eclipse.edc.connector.contract.spi.types.negotiation.ContractNegotiat
 import org.eclipse.edc.connector.contract.spi.types.negotiation.ContractRequest;
 import org.eclipse.edc.connector.contract.spi.types.negotiation.ContractRequestMessage;
 import org.eclipse.edc.connector.contract.spi.types.negotiation.command.ContractNegotiationCommand;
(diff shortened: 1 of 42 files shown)


ndr-brt:
for some reasons checks failed, but they worked on my fork: https://github.com/Think-iT-Labs/EDC-Connector/actions/runs/4937321037

the PR is ready for review

jimmarino (approved):

PR #2986, opened 2023-05-12
fix: ODRL logical operators transformation

### LINKED ISSUE #2985: dsp: treat ODRL `operand` as an abstract property
# Feature Request

Currently, in the transformation from/to policy we're treating `operand` as an actual property, but it is abstract. 
See https://www.w3.org/TR/odrl-model/#infoModel:
> The ODRL Information Model includes property relationships between the classes. Most are explicitly named properties and some are abstract properties (specifically, relation, function, operand, and failure). The abstract properties are generic parent properties that are designed to be represented by child properties (sub-types) with explicit semantics.

> For example, the two properties relation and function in Figure 1 are designed to represent the conceptual relation between the Rule and the Asset and Party classes. The figure shows the relation property with subtype target to express that the Asset is the primary subject of the Rule. The function property has subtype assigner to express the Party issuing the Rule, and subtype assignee to express the recipient Party of the Rule.

json transformers

odrl compliance

ndr-brt:
## What this PR changes/adds

Fix the LogicalConstraint transformation.

## Why it does that

dsp

## Further notes

I merged the MultiplicityConstraint and the AtomicConstraint transformers because at the end they act on the same object and the discriminant to choose one or another is given by their properties (so now the ConstraintTransformer tries to transform Logical first, if it didn't find the right properties, it transforms Atomic as fallback).

## Linked Issue(s)

Closes #2985

## Checklist


ndr-brt:
> Can we merge my PR first and then look at this, otherwise, it will mess with my changes which are more extensive?

@jimmarino I rebased against main, now it's ready for review

extensions/common/json-ld/src/test/java/org/eclipse/edc/jsonld/transformer/to/JsonObjectToConstraintTransformerTest.java
```diff
@@ -57,49 +70,157 @@ void setUp() {
...
+        var jsonNumber = Json.createValue(1);
+
+        var constraint = jsonFactory.createObjectBuilder()
+                .add(ODRL_LEFT_OPERAND_ATTRIBUTE, jsonNumber)
+                .add(ODRL_OPERATOR_ATTRIBUTE, jsonNumber)
+                .add(ODRL_RIGHT_OPERAND_ATTRIBUTE, jsonNumber)
                 .build();
 
-        var result = transformer.transform(getExpanded(invalidConstraintJson), context);
+        var result = transformer.transform(constraint, context);
```
jimmarino, line 159: shouldn't this be expanded first to always ensure the input?

jimmarino (approved):

extensions/common/json-ld/src/test/java/org/eclipse/edc/jsonld/transformer/to/JsonObjectToConstraintTransformerTest.java
ndr-brt: I wasn't aware that this need to be done on all the tests. Did it

PR #2988, opened 2023-05-12
fix: make DSP dispatcher fails if response code is not 200

### LINKED ISSUE #2987: `DspHttpRemoteMessageDispatcherImpl` never fails
# Feature Request

Currently `DspHttpRemoteMessageDispatcherImpl` always returns a completed future, also when the counter party returns 4xx or 5xx. 
This is because by default when `OkHttpClient` receives a failed response is still considered valid.
We should use `Fallback`s to solve that. 

_Are there any requirements?_

## Solution Proposal

ndr-brt:
## What this PR changes/adds

Make `DspHttpRemoteMessageDispatcherImpl` fails if response status is not 200.

## Why it does that

dsp

## Linked Issue(s)

Closes #2987 

## Checklist

### DIFF
--- a/spi/common/http-spi/src/main/java/org/eclipse/edc/spi/http/EdcHttpClient.java
+++ b/spi/common/http-spi/src/main/java/org/eclipse/edc/spi/http/EdcHttpClient.java
@@ -51,7 +51,8 @@ public interface EdcHttpClient {
      * Accepts a list of {@link FallbackFactories} that could apply retry in particular occasions.
      *
      * @param request the {@link Request}.
-     *                @param mappingFunction the function that will be applied to the {@link Response}.
+     * @param fallbacks a list of fallbacks to be applied.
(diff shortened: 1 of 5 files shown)


spi/common/http-spi/src/main/java/org/eclipse/edc/spi/http/EdcHttpClient.java
```diff
@@ -66,6 +67,18 @@ public interface EdcHttpClient {
      */
     <T> CompletableFuture<T> executeAsync(Request request, Function<Response, T> mappingFunction);
 
+    /**
+     * Executes the specified request asynchronously, maps the response with the mappingFunction.
+     * ccepts a list of {@link FallbackFactories} that could apply retry in particular occasions.
```
jimmarino, line 72: Looks like the `A` was chopped

jimmarino (approved):

PR #2990, opened 2023-05-12
fix: policy transformation processing expanded Json-Ld

### LINKED ISSUE #2989: Policy transormers don't always handle expanded JSON
The policy transformers don't always handle expanded JSON properly.

jimmarino:
## What this PR changes/adds

Fixes bugs in the policy transformers, mostly dealing with handling expanded Json-Ld. It also cleans up transform constant naming.

## Linked Issue(s)

Closes #2989

## Checklist

### DIFF
--- a/extensions/common/json-ld/src/test/java/org/eclipse/edc/jsonld/transformer/to/TestInput.java
+++ b/extensions/common/json-ld/src/test/java/org/eclipse/edc/jsonld/transformer/to/TestInput.java
@@ -1,0 +1,40 @@
+/*
+ *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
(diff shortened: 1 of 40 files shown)


extensions/common/json-ld/src/test/java/org/eclipse/edc/jsonld/transformer/to/TestInput.java
```diff
@@ -0,0 +1,40 @@
...
+ * Functions for shaping test input.
+ */
+public class TestInput {
+
+    /**
+     * Expands test input as Json-ld is required to be in this form
+     */
+    public static JsonObject getExpanded(JsonObject message) {
+        try {
+            return expand(JsonDocument.of(message)).get().asJsonArray().getJsonObject(0);
```
ndr-brt, line 32: why don't use our `TitaniumJsonLd` service instead?

ndr-brt (approved):

extensions/common/json-ld/src/test/java/org/eclipse/edc/jsonld/transformer/to/TestInput.java
jimmarino: It's a lot less verbose, and we would have to inject a mock monitor. I figured we only need to do a simple expansion, and we use a similar method in other modules that don't have a direct dependency on the Json-ld module. We could look into switching this, but it would result in a lot more test code.

wolf4ood (approved):

PR #3064, opened 2023-05-22
fix: feed `TransferProcess#properties` from the new field 

### LINKED ISSUE #3063: TransferProcess#properties wrongly populated
# Bug Report

## Describe the Bug
Currently `TransferProcess#properties` are populated by the` DataRequest#properties` but since the DataRequest it's stored it's a duplication of data. Originally `TransferProcess#properties` were introduced for storing the URL of dynamic edr receiver.  But those properties are transmitted over the dsp protocol. We should differentiate those two set of properties

## Possible Implementation
Take in input a new field in `TransferRequestDto` JsonObject that represent the "privateProperties" of a transfer process. Those properties will stay on the consumer side and they will not be transmitted to the counter party.

wolf4ood:
## What this PR changes/adds

Adds a new field `TransferProcessRequestDto#privateProperties` that will be stored in the `TransferProcess#properties` field

## Why it does that

Consistency

## Further notes

Also in this PR the `HttpDynamicEndpointDataReferenceReceiver` has been fixed

To make the change more consistent `TransferProcess#properties` has been renamed to `TransferProcess#privateProperties`

## Linked Issue(s)

Closes #3063 

### DIFF
--- a/core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/process/TransferProcessManagerImpl.java
+++ b/core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/process/TransferProcessManagerImpl.java
@@ -184,7 +184,7 @@ public class TransferProcessManagerImpl implements TransferProcessManager, Provi
                 .dataRequest(dataRequest)
                 .type(CONSUMER)
                 .clock(clock)
-                .properties(dataRequest.getProperties())
+                .properties(transferRequest.getPrivateProperties())
                 .callbackAddresses(transferRequest.getCallbackAddresses())
                 .traceContext(telemetry.getCurrentTraceContext())
                 .build();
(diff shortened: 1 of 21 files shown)


core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/process/TransferProcessManagerImpl.java
```diff
@@ -184,7 +184,7 @@ public StatusResult<TransferProcess> initiateConsumerRequest(TransferRequest tra
                 .dataRequest(dataRequest)
                 .type(CONSUMER)
                 .clock(clock)
-                .properties(dataRequest.getProperties())
+                .properties(transferRequest.getPrivateProperties())
```
ndr-brt, line 187: why don't have a `privateProperties` field on `TransferProcess`? I'm thinking this in the process of getting rid of `DataRequest` (in fact `DataRequest` properties are `TransferProcess` properties)

ndr-brt: I think we wanted to get rid of `DataRequest` as a whole, because at the end it's just an unneeded type, so the `TransferProcess` `properties` are the `DataRequest` properties (as it is currently), 
while the private properties should be a completely new field.

TL;DR: `TransferProcess` should have a `properties` field and a `privateProperties` one.

maybe something for a subsequent PR, just wanted to check if we're aligned on this

wolf4ood: :+1:  in this case in this PR i would just rename `TransferProcess#properties` to `TransferProcess#privateProperties` then, so in subsequent PRs we can reintroduce `TransferProcess#properties` that will be the one from `DataRequest#properties`

ndr-brt: ok, fine for me

paullatzelsperger (approved):
LGTM, optionally we could rename the column in the Postgres schema. Pro: consistency, Contra: not strictly speaking needed, and it will cause migrations downstream.

ndr-brt (approved):
I'd rename the column as well, likely not a lot of people migrated to milestone-9 so they'll likely migrate to the new version directly, but this is already good for merge anyway

PR #3081, opened 2023-05-24
fix: Set missing `client_id` in `DecentralizedIdentityService`

bscholtes1A:
## What this PR changes/adds

Set missing `client_id` in `DecentralizedIdentityService`.

## Why it does that

`client_id` is request by `ContractValidationServiceImpl`.

### DIFF
--- a/extensions/common/iam/decentralized-identity/identity-did-service/src/main/java/org/eclipse/edc/iam/did/service/DecentralizedIdentityService.java
+++ b/extensions/common/iam/decentralized-identity/identity-did-service/src/main/java/org/eclipse/edc/iam/did/service/DecentralizedIdentityService.java
@@ -37,7 +37,10 @@ import java.text.ParseException;
 import java.time.Clock;
 import java.util.Optional;
 
+import static org.eclipse.edc.spi.agent.ParticipantAgentService.DEFAULT_IDENTITY_CLAIM_KEY;
+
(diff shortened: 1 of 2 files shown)


ndr-brt (commented):
couldn't this be solved by setting the `edc.agent.identity.key` setting with "iss"?

bscholtes1A:
> couldn't this be solved by setting the `edc.agent.identity.key` setting with "iss"?

I tried but no. 
`ParticipantAgent.getIdentity()` uses the claim token returned by `IdentityService.verifyJwtToken()`, which in the case of decentralized identity contains only the VCs from the caller.

extensions/common/iam/decentralized-identity/identity-did-service/src/main/java/org/eclipse/edc/iam/did/service/DecentralizedIdentityService.java
```diff
@@ -38,6 +38,9 @@
 import java.util.Optional;
 
 public class DecentralizedIdentityService implements IdentityService {
+
+    private static final String CLIENT_ID = "client_id";
```
ndr-brt, line 42: the `DEFAULT_IDENTITY_CLAIM_KEY` constant can be used

ndr-brt (approved):

PR #3090, opened 2023-05-24
fix: provide correct id for lease deletion

### LINKED ISSUE #3088: Lease not deleted properly for transfer process
# Bug Report

## Describe the Bug
`SQLLeaseContext` is passing wrong ID to the delete query. Due to this, the old leases are not cleaned up and piles up in the database.

### Expected Behavior
The expectation is that the old lease is deleted from the database and only new lease is present in the database after `SQLLeaseContext::acquireLease()`.

### Observed Behavior
As long as the transfer process is in `IN_PROGRESS` or `REQUESTED`, a new lease is created every minute and the old one still remains in the table.

## Steps to Reproduce
1. Initiate a transfer process
2. Check the contents of the `edc_lease` table

## Context Information
None

## Detailed Description
image
There are a lot of leases without any transfer processes associated with it.

## Possible Implementation
https://github.com/eclipse-edc/Connector/blob/main/extensions/common/sql/sql-lease/src/main/java/org/eclipse/edc/sql/lease/SqlLeaseContext.java#L87 This seems to be the problem. Instead of `entityId`, `lease.getLeaseId()` should be passed.

lanthoor:
## What this PR changes/adds

- pass the correct id for deleting the outdated lease
- test cases

## Why it does that

Bug fix for the issue where `edc_lease` table was growing infinitely

## Further notes

N/A

## Linked Issue(s)

Closes #3088 


extensions/common/sql/sql-lease/src/test/java/org/eclipse/edc/sql/lease/SqlLeaseContextTest.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.when;
+
+class SqlLeaseContextTest {
```
paullatzelsperger, line 42: generally I  tend to agree that unit tests should be preferred, but since we already have the `PostgresLeaseContextTest`, please just move the test methods over there. 
As a neat side effect, will make the tests much clearer

extensions/common/sql/sql-lease/src/test/java/org/eclipse/edc/sql/lease/SqlLeaseContextTest.java
```diff
@@ -0,0 +1,121 @@
...
+        connection = Mockito.mock(Connection.class);
+        statement = Mockito.mock(PreparedStatement.class);
+        resultSet = Mockito.mock(ResultSet.class);
+
+        sqlLeaseContext = new SqlLeaseContext(transactionContext, leaseStatements, leaseHolder, clock, leaseDuration,
+                connection);
+    }
+
+    @Test
+    void givenNoLeasePresent_WhenAcquiringLease_ShouldAcquireLease() throws SQLException {
```
paullatzelsperger, line 67: please follow our convention for test methods outlined in `docs/developer/architecture/coding-principles.md` specifically regarding test method naming. 
also, don't use upper case letters after underscores.

extensions/common/sql/sql-lease/src/test/java/org/eclipse/edc/sql/lease/TestEntityLeaseStatements.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ *  Contributors:
+ *       Microsoft Corporation - initial API and implementation
+ *       SAP SE - refactoring
+ *
+ */
+
+package org.eclipse.edc.sql.lease;
+
+class TestEntityLeaseStatements implements LeaseStatements {
```
paullatzelsperger, line 18: this should be moved back as internal class to `PostgresLeaseContextTest`

extensions/common/sql/sql-lease/src/test/java/org/eclipse/edc/sql/lease/SqlLeaseContextTest.java
```diff
@@ -0,0 +1,121 @@
...
+        when(statement.executeQuery()).thenReturn(resultSet);
+        doNothing().when(statement).setString(anyInt(), paramCaptor.capture());
+        when(resultSet.next()).thenReturn(true, false);
+        String existingLeaseId = UUID.randomUUID().toString();
+        when(resultSet.getString(anyString())).thenReturn(leaseHolder, existingLeaseId);
+        when(resultSet.getLong(anyString())).thenReturn(clock.millis() - 60000L, 60000L);
+
+        assertDoesNotThrow(() -> sqlLeaseContext.acquireLease("test-entity"));
+        List<String> capturedQueries = queryCaptor.getAllValues();
+        assertEquals(4, capturedQueries.size());
```
paullatzelsperger, line 102: please use assertJ assertions, e.g. `assertThat(something).isTrue()`

paullatzelsperger (changes requested):
Good catch, thanks! 
some remarks inline

extensions/common/sql/sql-lease/src/test/java/org/eclipse/edc/sql/lease/SqlLeaseContextTest.java
lanthoor: 1. Are the integration test results also part of code coverage? Will it reduce the code coverage if I don't have a unit test?
2. Should I move the tests to `PostgresLeaseContextTest` (there are currently no tests here) or to `LeaseContextText` (abstract class that has all the tests)?

extensions/common/sql/sql-lease/src/test/java/org/eclipse/edc/sql/lease/TestEntityLeaseStatements.java
lanthoor: May I know why? The class is already package private and inside a test package. It has no visibility elsewhere.

extensions/common/sql/sql-lease/src/test/java/org/eclipse/edc/sql/lease/SqlLeaseContextTest.java
ndr-brt: Coverage it's just a number that's not that important in term of code quality, so I wouldn't bother about it, things needs to be tested properly.

`LeaseContextTest` is the right place to put such tests

extensions/common/sql/sql-lease/src/test/java/org/eclipse/edc/sql/lease/SqlLeaseContextTest.java
```diff
@@ -0,0 +1,121 @@
...
+        statement = Mockito.mock(PreparedStatement.class);
+        resultSet = Mockito.mock(ResultSet.class);
+
+        sqlLeaseContext = new SqlLeaseContext(transactionContext, leaseStatements, leaseHolder, clock, leaseDuration,
+                connection);
+    }
+
+    @Test
+    void givenNoLeasePresent_WhenAcquiringLease_ShouldAcquireLease() throws SQLException {
+        ArgumentCaptor<String> queryCaptor = ArgumentCaptor.forClass(String.class);
```
ndr-brt, line 68: please use `var` (same in the rest of the code)

extensions/common/sql/sql-lease/src/test/java/org/eclipse/edc/sql/lease/TestEntityLeaseStatements.java
ndr-brt: unless it needed to be used by some other tests I would let it as an internal class.

ndr-brt (changes requested):

extensions/common/sql/sql-lease/src/test/java/org/eclipse/edc/sql/lease/SqlLeaseContextTest.java
lanthoor: @ndr-brt I will no longer have access to the `connection` object as it is defined in the child class (`PostgresLeaseContextTest`). Is it okay to add an abstract method for the query/param verification in the `LeaseContextTest` and implement the actual check in the child class?

paullatzelsperger:
@lanthoor generally speaking, once you have adopted all PR remarks from a reviewer, pls resolve the conversation and re-request the review, so they have it in their todo list. thanks

extensions/common/sql/sql-lease/src/test/java/org/eclipse/edc/sql/lease/SqlLeaseContextTest.java
lanthoor: Agreed. But the leases are orphaned and not associated with transfer process or negotiations anymore as the update statement will associate it with the latest lease. The `getLease` call will always return only 1. I will move the test. Thank you.

ndr-brt: What "update statement"? The error is in the "delete", right? If this mechanism won't permit the existence of two leases on the same entity (because it does not make sense) then it would be easier to test, without having to check in the table with a specific query.

lanthoor: Tests moved to `PostgresLeaseContextTest` as agreed.

lanthoor:
> @lanthoor generally speaking, once you have adopted all PR remarks from a reviewer, pls resolve the conversations and re-request the review, so they have it in their todo list. thanks

@paullatzelsperger @ndr-brt All requested changes are incorporated, conversations are resolved and requested reviews again. Could you also add the `bug` label please?

paullatzelsperger (approved):

ndr-brt (approved):

PR #3103, opened 2023-05-25
fix(TransferProcessController): fix NPE on empty request

### LINKED ISSUE #3102: NPE when calling `transferprocesses/request`
# Bug Report

## Describe the Bug

when doing a request `POST|transferprocesses/request` without a body an NPE is thrown

### Expected Behavior
Should not NPE and use the default `QuerySpec`

wolf4ood:
## What this PR changes/adds

fix NPE when calling `POST|/transferprocesses/request` with without body

## Why it does that

Fix the InternalServerError returned

## Linked Issue(s)

Closes #3102

### DIFF
--- a/extensions/control-plane/api/management-api/transfer-process-api/src/main/java/org/eclipse/edc/connector/api/management/transferprocess/TransferProcessApiController.java
+++ b/extensions/control-plane/api/management-api/transfer-process-api/src/main/java/org/eclipse/edc/connector/api/management/transferprocess/TransferProcessApiController.java
@@ -81,7 +81,7 @@ public class TransferProcessApiController implements TransferProcessApi {
         var querySpec = ofNullable(querySpecDto)
                 .map(jsonLd::expand)
                 .map(expandedMapper)
-                .orElse(Result.success(QuerySpec.Builder.newInstance().build()))
+                .orElse(Result.success(QuerySpec.none()))
                 .orElseThrow(InvalidRequestException::new);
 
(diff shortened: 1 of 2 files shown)


extensions/control-plane/api/management-api/transfer-process-api/src/main/java/org/eclipse/edc/connector/api/management/transferprocess/TransferProcessApiController.java
```diff
@@ -70,16 +72,23 @@ public TransferProcessApiController(Monitor monitor, TransferProcessService serv
     @Path("request")
     @Override
     public List<JsonObject> queryTransferProcesses(JsonObject querySpecDto) {
-        var querySpec = jsonLd.expand(querySpecDto)
-                .compose(expanded -> transformerRegistry.transform(expanded, QuerySpecDto.class))
-                .compose(dto -> transformerRegistry.transform(dto, QuerySpec.class))
+
+        Function<Result<JsonObject>, Result<QuerySpec>> expandedMapper =
```
paullatzelsperger, line 76: why not just use `querSpecDto = ofNullable(querySpecDto).orElseGet(() -> QuerySpec.none())`? that would be a bit easier to read IMO

wolf4ood: Do you mean instead of using the full empty builder? `QuerySpec.Builder.newInstance().build()`?

ndr-brt (approved):

PR #3187, opened 2023-06-16
feat(policy): Implement validator filtering including changing request scope names 

### LINKED ISSUE #3186: feat(policy): Implement validator filtering including
Pre- and post- validator scope filtering is not yet implemented in the `PolicyEngine`. This needs to be done.

jimmarino:
## What this PR changes/adds

This PR implements pre- and post- policy validator filtering (which was never done). 

Also, this PR renames the following recently added scopes:

`contract.negotiation.request` --> `request.contract.negotiation`
`transfer.process.request` --> `request.transfer.process`

This renaming is necessary due to the hierarchical behavior of policy scope filtering. Under the previous naming scheme, any function and rule registered to the `contract.negotiation` and `transfer.process` scopes would be engaged when their `.request` analogs are in scope. Prefixing instead of post-fixing `request.` solves this problem and also avoids breaking changes if we were to introduce a post-fix such as `.receive` to the existing scope names (since in that case existing end-user functions would be egaged when either the the request or receive scopes are fired)

## Linked Issue(s)

Closes #3186

### DIFF
--- a/core/common/policy-engine/src/main/java/org/eclipse/edc/policy/engine/PolicyEngineImpl.java
+++ b/core/common/policy-engine/src/main/java/org/eclipse/edc/policy/engine/PolicyEngineImpl.java
@@ -73,8 +73,7 @@ public class PolicyEngineImpl implements PolicyEngine {
     public Result<Policy> evaluate(String scope, Policy policy, ParticipantAgent agent, Map<Class<?>, Object> contextInformation) {
         var context = new PolicyContextImpl(agent, contextInformation);
 
-        final var delimitedScope = scope + ".";
-
+        var delimitedScope = scope + ".";
 
         var scopedPreValidators = preValidators.entrySet().stream().filter(entry -> scopeFilter(entry.getKey(), delimitedScope)).flatMap(l -> l.getValue().stream()).toList();
         for (var validator : scopedPreValidators) {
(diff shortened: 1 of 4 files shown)


core/common/policy-engine/src/main/java/org/eclipse/edc/policy/engine/PolicyEngineImpl.java
```diff
@@ -69,16 +73,18 @@ public Result<Policy> evaluate(String scope, Policy policy, ParticipantAgent age
     public Result<Policy> evaluate(String scope, Policy policy, ParticipantAgent agent, Map<Class<?>, Object> contextInformation) {
         var context = new PolicyContextImpl(agent, contextInformation);
 
-        for (BiFunction<Policy, PolicyContext, Boolean> validator : preValidators) {
+        final var delimitedScope = scope + ".";
```
paullatzelsperger, line 76:
```suggestion
       var delimitedScope = scope + ".";
```

ronjaquensel (approved):

paullatzelsperger (approved):

core/common/policy-engine/src/main/java/org/eclipse/edc/policy/engine/PolicyEngineImpl.java
jimmarino: Snipped that final!

wolf4ood (approved):

PR #3203, opened 2023-06-20
fix: handle fatal errors in dsp http dispatcher delegate

### LINKED ISSUE #3202: A failed Negotiation request results in a termination message
# Bug Report

## Describe the Bug
Currently, when a consumer negotiation request fails, the manager tries to terminate the negotiation by sending the termination message to the provider, but this is not needed as on the provider side the negotiation has not been created, so on the consumer side it should be terminated directly.

ndr-brt:
## What this PR changes/adds

Add a new `dispatch` method on both `RemoteMessageDispatcher` and `RemoteMessageDispatcherRegistry` that return `StatusResult`, doing this it has been possible catch `FATAL_ERROR`s and avoid retry

## Why it does that

avoid network pollution, improve error messaging

## Further notes
From bottom to top:
- changed the fallback used to call the http client, now it enable retry mechanism only if the status is not 2xx or 4xx (because client errors usually cannot be fixed by retry)
- add the `handleResponse` method in `DspHttpDispatcherDelegate` that takes the place of `parseResponse`. It takes care to look at the status code and create the `StatusResult` accordingly, if response is successful, call `parseResponse` (that became protected)
- on the state machine level, added the `AsyncStatusResultRetryProcess` class, that extends `CompletableFutureRetryProcess` and uses `StatusResultRetryProcess` to handle the `StatusResult` when the future succeeds.
- the managers on fatal error will terminate the negotiation/transfer without sending anything to the counter-part.
- the deprecated `send` methods on `RemoteMessageDispatcher` and `RemoteMessageDispatcherRegistry` are now unused, but I left them there to avoid compilation errors on downstream projects.

## Linked Issue(s)

Closes #3202

### DIFF
--- a/core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/base/RemoteMessageDispatcherRegistryImpl.java
+++ b/core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/base/RemoteMessageDispatcherRegistryImpl.java
@@ -18,8 +18,8 @@ package org.eclipse.edc.connector.core.base;
 import org.eclipse.edc.spi.EdcException;
 import org.eclipse.edc.spi.message.RemoteMessageDispatcher;
 import org.eclipse.edc.spi.message.RemoteMessageDispatcherRegistry;
+import org.eclipse.edc.spi.response.StatusResult;
 import org.eclipse.edc.spi.types.domain.message.RemoteMessage;
-import org.jetbrains.annotations.Nullable;
 
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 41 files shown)


core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/base/RemoteMessageDispatcherRegistryImpl.java
```diff
@@ -38,22 +38,14 @@ public void register(RemoteMessageDispatcher dispatcher) {
...
         if (dispatcher == null) {
             return failedFuture(new EdcException("No provider dispatcher registered for protocol: " + protocol));
         }
-        return dispatcher.send(responseType, message);
+        return dispatcher.dispatch(responseType, message);
     }
 
-    @Nullable
-    private RemoteMessageDispatcher getDispatcher(@Nullable String protocol) {
-        if (protocol == null) {
```
ndr-brt, line 53: I removed this fallback dispatcher because I can't see how it could be useful

jimmarino (approved):
Great stuff!

wolf4ood (approved):
:rocket:

paullatzelsperger (approved):
🧨

PR #3206, opened 2023-06-20
fix: Add support for RSA keys in Decentralized Identity modules

### LINKED ISSUE #3205: `DecentralizedIdentityServiceExtension` does not support RSA keys
# Bug Report

## Describe the Bug
Current implementation of the decentralized identity modules assumes that the keys (private used to sign the token, and public key given in the DID document verification methods) are of type Elliptic Curve (EC). 

Even if EC keys are to be preferred, RSA keys must also be supported are they are widely used as well.

### Expected Behavior
Both EC and RSA should be supported.

### Observed Behavior
Exception is thrown at startup used keys are not EC keys.

bscholtes1A:
## What this PR changes/adds

Add support for RSA keys in Decentralized Identity modules.

## Why it does that

Both EC and RSA keys should be supported as they are both widely used.

## Linked Issue(s)

Closes #3205


paullatzelsperger (approved):
I'm generally fine with this, but I think in a future effort me may want to refactor/simplify our crypto framework. I understand why we have library-agnostic classes in SPI (`JwkPublicKey`, etc.), but realistically we will be tied to nimbus anyway, and on top of that, most keys can be converted to `java.security` types.
Similarly, we may want to generalize our crypto framework to support other key types as well, eg. `OctetKeyPair`.

bscholtes1A:
> I'm generally fine with this, but I think in a future effort me may want to refactor/simplify our crypto framework. I understand why we have library-agnostic classes in SPI (`JwkPublicKey`, etc.), but realistically we will be tied to nimbus anyway, and on top of that, most keys can be converted to `java.security` types. That would save us all the conversion classes. Similarly, we may want to generalize our crypto framework to support other key types as well, eg. `OctetKeyPair`.

I did some simplifications after your comment, if you could have a second look ;)

PR #3207, opened 2023-06-21
feat(build): use release version in bump-job

### LINKED ISSUE #3198: Build: feed the release version into the version-bump job
# Feature Request

Currently, when we release a version, the version-bump job takes the "old version" from `gradle.properties` and then bumps the "patch" version and appends `-SNAPSHOT`. Since the release version is never committed back to `gradle.properties`, the bump job would bump an old version.

Release, the bump action in every repo

In [decision record "2023-05-17-release-process"](https://github.com/eclipse-edc/Connector/tree/main/docs/developer/decision-records/2023-05-17-release-process) it was decided that the semVer "minor" version gets bumped everytime, and only bugfixes will bump the "patch" version.

## Solution Proposal
Instead of taking the "oldVersion" from `gradle.properties`, just use the workflow input `edc_version` and forward it to the `bump-version/action.yml`. This has to be done in every repo.

paullatzelsperger:
## What this PR changes/adds

This PR modifies the `bump-version` action and takes the release version (base version) from an action input and only falls back to the `gradle.properties` if that input was `null`.

## Why it does that

when releasing e.g. `0.2.0`, the next snapshot is supposed to be `0.2.1-SNAPSHOT`.

## Further notes

- I also improved the snapshot patch version increment, so that when releasing versions with metadata, e.g. `0.3.6-rc2`, the next snapshot version will still be `0.3.6-SNAPSHOT`.
- once this PR is approved, i'll do the same thing in all other repos as well.

## Linked Issue(s)

Closes #3198 

### DIFF
--- a/.github/actions/bump-version/action.yml
+++ b/.github/actions/bump-version/action.yml
@@ -5,7 +5,7 @@ inputs:
     default: 'main'
     description: "Branch on which the version bump is to be done."
     required: false
-  from_version:
+  base_version:
     description: "The current version, which is to be bumped to the next snapshot"
     required: false
 
@@ -25,10 +25,10 @@ runs:
         git checkout ${{ inputs.target_branch }}
 
         # use current version from input
-        oldVersion=${{ inputs.from_version }}
+        oldVersion=${{ inputs.base_version }}
         if [ -z $oldVersion ]; then
           #... or read it from gradle.properties
-          echo "No from_version input was given, will read base version from gradle.properties"
+          echo "No base_version input was given, will read base version from gradle.properties"
           oldVersion=$(grep "version" gradle.properties  | awk -F= '{print $2}')
         fi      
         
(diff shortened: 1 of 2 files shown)


.github/actions/bump-version/action.yml
```diff
@@ -21,11 +24,24 @@ runs:
...
+        # Compute new snapshot version, do not increment snapshot on non-final releases, e.g. -rc1
+        if [ -z $SNAPSHOT ]; then
+          echo "$oldVersion is a final release version, increase patch for next snapshot"
+          INC=1
+        else
+          echo "$oldVersion is not a final release version (contains \"$SNAPSHOT\"), will not increase patch"
+        fi
 
         # construct the new version
         newVersion="$RELEASE_VERSION_MAJOR.$RELEASE_VERSION_MINOR.$((RELEASE_VERSION_PATCH+1))"-SNAPSHOT
```
ronjaquensel, line 47: shouldn't this be `$((RELEASE_VERSION_PATCH+INC))`? as it is, this would always increase the path version, right?

paullatzelsperger: you are absolutely right. copy-paste error...

ronjaquensel (approved):

wolf4ood (approved):

PR #3212, opened 2023-06-22
feat(dsp): implement GET /id negotiation endpoint

### LINKED ISSUE #2795: dataspace protocol: add method findById to negotiation protocol service
# Feature Request

Add method for getting a `ContractNegotiation` to the `ContractNegotiationProtocolService`. The method should take a `ClaimToken` as parameter and verify that the requesting participant is involved in the requested `ContractNegotiation`.

Using this method, the GET endpoint on the `DspNegotiationController` can be implemented.

spi, core, dsp

Support the protocol specification

ronjaquensel:
## What this PR changes/adds

Adds a method `findById` to the `ContractNegotiationProtocolService`, that takes a `ClaimToken` as parameter and returns a negotiation only if the requesting party is authorized. Using this method, implements the `GET /id` endpoint in the `DspNegotiationApiController`.

## Why it does that

To support the protocol specification

## Linked Issue(s)

Closes #2795

### DIFF
--- a/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/contractnegotiation/ContractNegotiationProtocolServiceImpl.java
+++ b/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/contractnegotiation/ContractNegotiationProtocolServiceImpl.java
@@ -149,7 +149,8 @@ public class ContractNegotiationProtocolServiceImpl implements ContractNegotiati
     @NotNull
     public ServiceResult<ContractNegotiation> findById(String id, ClaimToken claimToken) {
         return transactionContext.execute(() -> Optional.ofNullable(store.findById(id))
-                .map(negotiation -> validateGetRequest(claimToken, id, negotiation))
+                .map(negotiation -> validateGetRequest(claimToken, negotiation))
+                .map(ServiceResult::success)
                 .orElse(ServiceResult.notFound(format("No negotiation with id %s found", id))));
     }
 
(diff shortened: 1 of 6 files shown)


jimmarino (approved):

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/contractnegotiation/ContractNegotiationProtocolServiceImpl.java
```diff
@@ -194,6 +204,17 @@ private ServiceResult<ContractNegotiation> validateRequest(ClaimToken claimToken
             return ServiceResult.success(negotiation);
         }
     }
+    
+    @NotNull
+    private ServiceResult<ContractNegotiation> validateGetRequest(ClaimToken claimToken, String id, ContractNegotiation negotiation) {
+        var result = validationService.validateRequest(claimToken, negotiation);
+        if (result.failed()) {
+            // should return not found if counter-party not authorized
```
ndr-brt, line 212: I know that this is stated in the specs, but shouldn't be more http-like to return 403?

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/contractnegotiation/ContractNegotiationProtocolServiceImpl.java
```diff
@@ -194,6 +204,17 @@ private ServiceResult<ContractNegotiation> validateRequest(ClaimToken claimToken
...
+            return ServiceResult.notFound(format("No negotiation with id %s found", id));
```
ndr-brt, line 213: if this failure is the same as the one 153, to avoid the duplication this method could just return the `ContractNegotiation` (so `null` in this branch and `negotiation` in the `else`), then in the mail optional flow a `.map(StatusResult::success)` can be added,

ronjaquensel: good point, refactored it

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/contractnegotiation/ContractNegotiationProtocolServiceImpl.java
ronjaquensel: I think the idea behind returning the 404 is to not give out information about the existence of processes to unauthorized parties.

ndr-brt (approved):

PR #3235, opened 2023-06-26
feat: encode `ContractId` parts

### LINKED ISSUE #3211: Problem when constructing `ContractId`s containing colons ":"
# Bug Report

## Describe the Bug
Having Assets where the `id` contains the ":" colon character causes problems during contract negotiation, causing the negotiation instance not to be found.

### Expected Behavior
(see link below)

### Observed Behavior
(see link below)

## Steps to Reproduce, Context Information
https://github.com/eclipse-tractusx/tractusx-edc/issues/484

## Detailed Description
`ContractId` instances are constructed in the form `<CONTRACT_DEF_ID>:<ASSET_ID>:<UUID>` using ":" as delimiter character and are parsed accordingly. If either the contract definition id or the asset id contains a colon ":" character, that would throw off the parsing algorithm.

## Possible Implementations
- make the delimiter configurable, default to ":"
- make `ContractIds` opaque, e.g. base64 encode the parts (<-- probably the best solution)

ndr-brt:
## What this PR changes/adds

Encodes the three parts of a `ContractId` with base64

## Why it does that

 avoid issues with id containing `:`.

## Further notes

- ~~this is a BREAKING CHANGE, so already stored ongoing contracts could stop working~~ it will work also with not encoded old ids
- refactored the `ContractId` class to adhere our convention, adding the `parseId` method returning a `Result<ContractId>` and adding a `create` that returns a `ContractId`
- deprecating old methods to avoid compilation errors in downstream projects. 

## Linked Issue(s)

Closes #3211 

### DIFF
--- a/spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/ContractId.java
+++ b/spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/ContractId.java
@@ -56,7 +56,7 @@ public final class ContractId {
      * @deprecated please use {@link #create(String, String)}
      */
     @NotNull
-    @Deprecated(since = "0.1.2")
+    @Deprecated(since = "0.1.2", forRemoval = true)
     public static String createContractId(String definitionPart, String assetId) {
         return create(definitionPart, assetId).toString();
     }
@@ -100,7 +100,7 @@ public final class ContractId {
      * @return the {@link ContractId} instance that represent the id
      * @deprecated please use {@link #parseId(String)}
      */
-    @Deprecated(since = "0.1.2")
+    @Deprecated(since = "0.1.2", forRemoval = true)
     public static ContractId parse(String id) {
         return parseId(id).getContent();
     }
@@ -115,7 +115,7 @@ public final class ContractId {
      * @return true if it is valid, false otherwise
      * @deprecated an instantiated {@link ContractId} object is always valid
      */
-    @Deprecated(since = "0.1.2")
+    @Deprecated(since = "0.1.2", forRemoval = true)
     public boolean isValid() {
         return true;
     }
@@ -148,7 +148,12 @@ public final class ContractId {
                 encoder.encodeToString(uuid.getBytes());
     }
 
-    public ContractId spawn() {
+    /**
+     * Create a new {@link ContractId} with the same definitionId and assetId but a new random UUID part.
+     *
+     * @return new {@link ContractId} instance.
+     */
+    public ContractId derive() {
         return create(definitionId, assetId);
     }
 }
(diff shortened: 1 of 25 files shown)


spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/ContractId.java
```diff
@@ -83,16 +135,20 @@ public String definitionPart() {
...
-        return value;
+        var encoder = Base64.getEncoder();
+        return encoder.encodeToString(definitionId.getBytes()) +
+                DELIMITER +
+                encoder.encodeToString(assetId.getBytes()) +
+                DELIMITER +
+                encoder.encodeToString(uuid.getBytes());
+    }
+
+    public ContractId spawn() {
```
paullatzelsperger, line 151: this deserves some documentation, since its purpose is not readily deducible from the name. on that note, maybe a name like `derive` or `copy` would be better?

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/ContractId.java
```diff
@@ -40,10 +53,43 @@ private ContractId(String value) {
      * @param definitionPart the part that will be used as prefix of the id
      * @param assetId        The ID of the asset that is contained in the offer
      * @return a {@link String} that represent the contract id
+     * @deprecated please use {@link #create(String, String)}
      */
     @NotNull
+    @Deprecated(since = "0.1.2")
```
paullatzelsperger, line 59: nit: adding `forRemoval=true` will cause IDEs to issue a "stronger" warning

paullatzelsperger (approved):
minor nits, all good otherwise

wolf4ood (approved):

jimmarino:
@ndr-brt Can we discuss tomorrow during the committer's call? I'd like to see if we can avoid the breaking change.

ndr-brt:
> @ndr-brt Can we discuss tomorrow during the committer's call? I'd like to see if we can avoid the breaking change.

sure, just for the record I also tried do, doing a "try to decode, if it does not work just use the plain string", but it happened that strings that aren't base64 representation can be detected as base64 representation of weird characters, e.g. "definitionId" decodes correctly into a bunch of unreadable data.

We can, though state that is practically impossible to have an UUID that can result as a decodable base64, so if not of all the three parts decode correctly then we can assume that's an "old fashioned" id and we can build a valid object as it is.

ndr-brt:
@jimmarino now the `ContractId` will also being able to handle not encoded ids, so this PR does not introduce any breaking change

jimmarino (approved):

PR #3244, opened 2023-06-29
refactor: put DataAddress as Asset field

### LINKED ISSUE #3151: management api: Create asset payload inconsistent with the rest of the API entities
I didn't know whether this counted as an bug or not, so I am opening a blank issue.

While working with the `POST` `/management/v2/assets` endpoints,
I tried creating an **asset** with a **pre-configured** `id`,
this is the accepted payload
```json
{
  "@context": {
    "@vocab": "https://w3id.org/edc/v0.0.1/ns/"
  },
  "@type": "AssetEntryDto",
  "asset": {
    "@id": "some-id",
    "properties": {
      "name": "asset-name"
    }
  },
  "dataAddress": {
    "type": "asset-address-type",
    "keyName": "asset-key-name"
  }
}
```
Notice how the `@id` is nested under `asset` and not at top level.

This is inconsistent with the other management entities (example, policies) and with the swagger documentation.

So I am just wondering whether this was intentional or an omission during the move to json-ld.

ndr-brt:
## What this PR changes/adds

Put `DataAddress` as a field of `Asset`

## Why it does that

At first they were kept separated because `DataAddress` was like "private data for asset", but now we have also `privateProperties` in `Asset`, so there aren't good reasons to keep them split. Doing this we'll end up with a way more clean api. 

## Further notes

- added a `/v3/assets` api endpoint that would avoid breaking changes, though the advice is to switch to the new api as soon as possible.
- cleaned up AssetIndex tests on both in memory and postgresql, likely they could broke in the cosmosdb, eventually I will take care of fixing them

## Linked Issue(s)

Closes #3151 

### DIFF
--- a/core/control-plane/catalog-core/src/test/java/org/eclipse/edc/connector/catalog/DatasetResolverImplIntegrationTest.java
+++ b/core/control-plane/catalog-core/src/test/java/org/eclipse/edc/connector/catalog/DatasetResolverImplIntegrationTest.java
@@ -29,7 +29,6 @@ import org.eclipse.edc.spi.query.Criterion;
 import org.eclipse.edc.spi.query.QuerySpec;
 import org.eclipse.edc.spi.types.domain.DataAddress;
 import org.eclipse.edc.spi.types.domain.asset.Asset;
-import org.eclipse.edc.spi.types.domain.asset.AssetEntry;
 import org.jetbrains.annotations.NotNull;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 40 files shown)


paullatzelsperger (approved):
Since we are not exposing raw `Assets` (and thus: `DataAddress`es) to the public, this change is good. It'll simplifies our management API.

However, since the `DataAddress` itself is only a wrapper around a `Map` (i.e. `properties`) with named accessors, we could probably simplify things further by completely collapsing `DataAddress` into `Asset`.  In another PR...

wolf4ood (approved):

PR #3285, opened 2023-07-07
feat: Add fixtures for end-to-end transfer tests

### LINKED ISSUE #3282: Expose the `Participant` in test fixtures
# Feature Request

The `Participant` class, which is a convenient way to communicate with an in-JVM runtime, should be exposed in a `testFixtures` package.

test fixtures, e2e tests

downstream projects and other components may want to reuse that functionality

## Solution Proposal

- Move the `Participant` to the `:system-tests:tests:testFixtures` module

bscholtes1A:
## What this PR changes/adds

Introduce helpers for testing end-to-end transfers.

## Why it does that

Ease end-to-end testing in downstream repositories.

## Linked Issue(s)

Closes #3282 


system-tests/e2e-test-fixtures/src/testFixtures/java/org/eclipse/edc/test/system/utils/TransferTestRunner.java
```diff
@@ -0,0 +1,73 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.ID;
+import static org.eclipse.edc.jsonld.spi.PropertyAndTypeNames.ODRL_POLICY_ATTRIBUTE;
+
+/**
+ * Orchestrate a data transfer between the consumer and the provider and returns the transfer process id.
+ */
+public class TransferTestRunner implements UnaryOperator<String> {
```
ndr-brt, line 32: I'd tend to define custom interfaces instead of using the Java ones, this because with them it's actually impossible to get the usages of the `apply` method.

system-tests/e2e-test-fixtures/src/testFixtures/java/org/eclipse/edc/test/system/utils/TransferTestRunner.java
```diff
@@ -0,0 +1,73 @@
...
+        var contractDefinitionId = ContractId.parseId(policy.getString(ID))
+                .orElseThrow(failure -> new RuntimeException(failure.getFailureDetail()));
+        var contractAgreementId = consumer.negotiateContract(provider, contractDefinitionId.toString(), assetId, policy);
+        var transferProcessId = consumer.initiateTransfer(provider, contractAgreementId, assetId, privateProperties, destination);
+        assertThat(transferProcessId).isNotNull();
+
+        await().atMost(TIMEOUT).untilAsserted(() -> {
+            var state = consumer.getTransferProcessState(transferProcessId);
+            assertThat(state).isEqualTo(expectedState.name());
+        });
```
ndr-brt, line 69: I'd let the state assertion out of this, otherwise there would be tests like `httpPull_withExpiredContract_fixedInForcePeriod` without any visible assertion

system-tests/e2e-test-fixtures/src/testFixtures/java/org/eclipse/edc/test/system/utils/TransferTestRunner.java
ndr-brt: looking better at the usages of this I think it would be better have this as a method of `Participant` with a signature like:
```java
public String requestAsset(String assetId, Participant provider, JsonObject destination, JsonObject privateProperties)
```

because the `apply` is currently always called after object instantiation, a single method would be simpler.

system-tests/e2e-test-fixtures/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
+/*
+ *  Copyright (c) 2022 Microsoft Corporation
```
paullatzelsperger, line 2: is this correct?

paullatzelsperger (approved):

system-tests/e2e-test-fixtures/build.gradle.kts
bscholtes1A: Yes I think this is. This file was initially in the `testFixtures/tests` folder.

ndr-brt (approved):
LGTM!

PR #3296, opened 2023-07-13
feat(ci): publish OpenAPI specs on releases

### LINKED ISSUE #2834: Version the APIs that are published to SwaggerHub
# Feature Request

Currently we publish every iteration of our API spec as `0.0.1-SNAPSHOT`, which should be improved. Every release of EDC should also produce a versioned API spec in SwaggerHub. 

We should not do this on nightlies, because that would absolutely SPAM SwaggerHub.

build, CI

Downstream projects use pinned versions of EDC, and they currently have no way of referencing _their_ version of the EDC APIs.

## Solution Proposal
- improve the generation of the API to make the version input configurable:
 ```kotlin
 // root build.gradle.kts
 swagger {
        title.set((project.findProperty("apiTitle") ?: "EDC REST API") as String)
        description =
            (project.findProperty("apiDescription") ?: "EDC REST APIs - merged by OpenApiMerger") as String
        version = project.version // <-- this is new, ties it to the project version, override with -Pversion=...
        outputFilename.set(project.name)
        outputDirectory.set(file("${rootProject.projectDir.path}/resources/openapi/yaml"))
    }
  ```
- make the `apidoc.yaml` workflow user-triggerable (`workflow_dispatch`)
- avoid publishing API specs on nightlies
- for manual triggering: add an optional `version` input to it and use that `version` input to set the API version

paullatzelsperger:
## What this PR changes/adds

This PR lets the `release-edc` workflow call the `apidoc` workflow, which publishes the OpenAPI specs to Swaggerhub.

If the `apidoc` workflow is invoked passing in a non-snapshot version, the API is set to `"published"` on SwaggerHub, which means it cannot be altered afterwards.

## Why it does that

We should publish immutable OpenAPI documents of our release versions.

## Further notes

## Linked Issue(s)

Closes #2834 

### DIFF
--- a/.github/workflows/apidoc.yaml
+++ b/.github/workflows/apidoc.yaml
@@ -2,6 +2,12 @@ name: Verify and Publish OpenAPI Specs
 
 on:
   workflow_call:
+    inputs:
+      version:
+        required: true
+        description: "The version under which the API should be published"
+        type: string
+
   push:
   pull_request:
     branches: [ main ]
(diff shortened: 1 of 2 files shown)


paullatzelsperger:
\cc @bcronin90 you may be interested as well

ndr-brt (approved):

PR #3313, opened 2023-07-18
feat: make state transitions sync

### LINKED ISSUE #3303: Story: make state transitions synchronous
# Feature Request

Currently, some state changes of contract negotiations and transfer processes are handled through a `CommandQueue`, which allows for asynchronous processing of commands.

This should be changed such that all state changes are handled _synchronously_. 

state machines, command handlers, mgmt api 

Async processing of commands has a major drawback: it cannot communicate the result back to the sender, because the information about the sender is not carried over. The only information that _can_ be communicated back is, whether or not the command was inserted in the queue.

This becomes even more apparent when thinking about a situation where a human is requesting the state change through an API call (e.g. cancelling a contract negotiation).

Furthermore, moving this over to a sync invocation, and getting rid of the command queue code, will greatly simplify things. 

## Solution Proposal

Every state transition request (e.g. through the management api) is a synchronous request to the service layer. If the `Lease` for the entity can be acquired, and the state transition is possible, then the state transition is performed. If the entity is already `leased`, or the state transition is impossible, the state transition is aborted. Either way, the result is communicated back to the requestor.

ndr-brt:
## What this PR changes/adds

Ensures that every change state is made on an appropriately leased entity.
To do that:
 - command queue has been removed, from now on every command will be executed synchronously
 - every command and generally state changes that are caused by incoming protocol messages will use `findByIdAndLease` and `findByCorrelationIdAndLease` methods from `StateEntityStore`. If the lease cannot be obtained, the method will return a failure that can be handled accordingly.
 - all the command handlers now return a `CommandResult` object, that can be used to notify the result to the caller.

## Why it does that

avoid state conflicts

## Further notes

- had to do an huge refactoring where all the logic involved after the provisioning/deprovisioning operations are completed has been moved to `ProvisionResponsesHandler` and `DeprovisionResponsesHandler`, this permitted to clean up `TransferProcessManager` a lot and avoid duplication between the manager and the command. 

## Linked Issue(s)

Closes #3303


spi/common/core-spi/src/main/java/org/eclipse/edc/spi/command/Command.java
```diff
@@ -29,43 +27,5 @@
  * </ul>
  */
 public abstract class Command {
```
paullatzelsperger, line 29: could this be an `interface`? or rather: is this needed at all? it seems to only have one direct inheritor in production code (`SingleEntityCommand`) and another one in test code

ndr-brt: yes, I thought the same, I'll remove it and leave the `SingleEntityCommand` as the root class

ndr-brt:
DR added

paullatzelsperger (approved):

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/command/Command.java
```diff
@@ -14,8 +14,6 @@
 
 package org.eclipse.edc.spi.command;
 
-import java.util.UUID;
```
jimmarino, line 17: Can we remove Command, it's not referenced?

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/command/SingleEntityCommand.java
```diff
@@ -0,0 +1,30 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.spi.command;
+
+public abstract class SingleEntityCommand {
```
jimmarino, line 17: Can we call this Command or EntityCommand?

jimmarino (approved):
LGTM, two comments.

PR #3316, opened 2023-07-20
chore: handle missing destination in transfer protocol service

### LINKED ISSUE #2728: dataspace protocol: transfer process should handle missing data destinations
# Feature Request

`TransferProcessProtocolServiceImpl::notifyRequested` should be able to handle missing `dataDestination`/`dataAddress` in `TransferRequestMessage`. Same goes for start actions, see #2727.

transfer process (protocol) service

The [spec](https://github.com/International-Data-Spaces-Association/ids-specification/blob/main/transfer/transfer.process.protocol.md#1-transferrequestmessage) states that `dataAddress` is optional, citing
> must only be provided if the dct:format requires a push transfer.

`DataAddressValidatorImpl::validate` will fail with an empty data destination.

## Solution Proposal

Either, add null check to protocol service method, e.g.:
```java
    @Override
    @WithSpan
    @NotNull
    public ServiceResult<TransferProcess> notifyRequested(TransferRequestMessage message, ClaimToken claimToken) {
        var dataAddress = message.getDataDestination();
        if (dataAddress != null) { // NOTE: optional property according to spec
            var validDestination = dataAddressValidator.validate(message.getDataDestination());
            if (validDestination.failed()) {
                return ServiceResult.badRequest(validDestination.getFailureMessages());
            }
        }

        ...
    }
```

Or, add null check to `DataAddressValidatorImpl::validate`.

I think the first approach would be cleaner than allowing null values in a validator.

ronjaquensel:
## What this PR changes/adds

Adds a null check on the data destination to `TransferProcessProtocolServiceImpl::notifyRequested` before validating the destination. In `requestedAction` creates a destination `DataAddress` of type `HTTP_PROXY` (= pull) for the `DataRequest`, if no destination was specified.

## Why it does that

The destination is optional according to the protocol spec, as it is only required for push transfers.

## Linked Issue(s)

Closes #2728

### DIFF
--- a/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/transferprocess/TransferProcessProtocolServiceImpl.java
+++ b/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/transferprocess/TransferProcessProtocolServiceImpl.java
@@ -37,6 +37,7 @@ import org.eclipse.edc.spi.iam.ClaimToken;
 import org.eclipse.edc.spi.monitor.Monitor;
 import org.eclipse.edc.spi.result.Result;
 import org.eclipse.edc.spi.telemetry.Telemetry;
+import org.eclipse.edc.spi.types.domain.DataAddress;
 import org.eclipse.edc.transaction.spi.TransactionContext;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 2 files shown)


paullatzelsperger (approved):

ndr-brt (approved):
I'd tend to avoid comments unless they are strictly necessary because they tend to rot quickly,  in my opinion the unit test already describes the need for this condition.
all good otherwise.

Issue #3308, opened 2023-07-17
Add manual interaction to state machines


paullatzelsperger:
# Feature Request

Our state machines should have a way for humans to manually advance the state. For instance, this will be needed when manual sign-off on a contract negotiation is required. 

## Which Areas Would Be Affected?

core: transfer and negotiation state machine

## Why Is the Feature Desired?

In certain scenarios human intervention will be required, for example signing off on contract negotiations.

## Solution Proposal

To provide maximum flexibility, the implementation of this feature should be able to dynamically make a decision (eg. through a `Function`) whether or not an entity's state can be modified automatically. In combination with adding an extension point here, implementors can easily add their own logic. 
For example, one use case might only want to add manual interaction to a certain set of states, but might also want to decide based on a certain property of an asset.

PR #3318, opened 2023-07-20
docs: add decision record about state machine manual interactions

ndr-brt:
## What this PR changes/adds

Adds a decision record about state machine manual interactions

## Why it does that

## Further notes

## Linked Issue(s)

Part of #3308 

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -116,10 +116,10 @@ maven/mavencentral/io.netty/netty-tcnative-boringssl-static/2.0.56.Final, Apache
 maven/mavencentral/io.netty/netty-tcnative-classes/2.0.56.Final, Apache-2.0, approved, clearlydefined
 maven/mavencentral/io.netty/netty-transport-native-unix-common/4.1.86.Final, Apache-2.0 AND BSD-3-Clause AND MIT, approved, CQ20926
 maven/mavencentral/io.netty/netty-transport/4.1.86.Final, Apache-2.0 AND BSD-3-Clause AND MIT, approved, CQ20926
-maven/mavencentral/io.opentelemetry.instrumentation/opentelemetry-instrumentation-annotations/1.28.0, , restricted, clearlydefined
+maven/mavencentral/io.opentelemetry.instrumentation/opentelemetry-instrumentation-annotations/1.28.0, Apache-2.0, approved, #9662
 maven/mavencentral/io.opentelemetry.proto/opentelemetry-proto/0.20.0-alpha, Apache-2.0, approved, #9274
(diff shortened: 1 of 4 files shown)


jimmarino (approved):
I have two wording nits:

1. Can we call hook `TransitionGuard` or just `Guard` since it is a bit more descriptive?
2. instead of `setWaitingForManualInteraction` can we name it `setWaitingOnGuard` or a more generic term? This mechanism could be used for things other than manual interaction.

Otherwise LGTM.

ndr-brt:
> I have two wording nits:
> 
> 1. Can we call hook `TransitionGuard` or just `Guard` since it is a bit more descriptive?
> 2. instead of `setWaitingForManualInteraction` can we name it `setWaitingOnGuard` or a more generic term? This mechanism could be used for things other than manual interaction.
> 
> Otherwise LGTM.

I agree on the `Guard` term, actually I like it more than hook.

Talking about the manual interaction, that's only a particular kind of guard that will only serve that precise scope, so in the entity I would still talking about "manual interaction" or something that reminds that.

jimmarino:
> > I have two wording nits:
> > 
> > 1. Can we call hook `TransitionGuard` or just `Guard` since it is a bit more descriptive?
> > 2. instead of `setWaitingForManualInteraction` can we name it `setWaitingOnGuard` or a more generic term? This mechanism could be used for things other than manual interaction.
> > 
> > Otherwise LGTM.
> 
> I agree on the `Guard` term, actually I like it more than hook.
> 
> Talking about the manual interaction, that's only a particular kind of guard that will only serve that precise scope, so in the entity I would still talking about "manual interaction" or something that reminds that.

Would "external interaction" work? I'm looking for a more general term since it could be an external system that transitions the sm and not necessarily a manual interaction.

ndr-brt:
> > > I have two wording nits:
> > > 
> > > 1. Can we call hook `TransitionGuard` or just `Guard` since it is a bit more descriptive?
> > > 2. instead of `setWaitingForManualInteraction` can we name it `setWaitingOnGuard` or a more generic term? This mechanism could be used for things other than manual interaction.
> > > 
> > > Otherwise LGTM.
> > 
> > 
> > I agree on the `Guard` term, actually I like it more than hook.
> > Talking about the manual interaction, that's only a particular kind of guard that will only serve that precise scope, so in the entity I would still talking about "manual interaction" or something that reminds that.
> 
> Would "external interaction" work? I'm looking for a more general term since it could be an external system that transitions the sm and not necessarily a manual interaction.

Good point, I'd say yes, also, it could also be called just "waiting", or "pending", like, a "pending contract negotiation" is a negotiation that's waiting for an external interaction, and having a simple term to call it will make everything easier to understand. what do you think?

PR #3320, opened 2023-07-20
fix: the Criterion Transformer correctly handles the rightOperand

### LINKED ISSUE #3319: Querying transferprocess with certain state not possible
# Bug Report

## Describe the Bug

Queriying all transfer process in certain state by using 'filterExpression' is not possible.

### Expected Behavior

It should be possible to query all transferprocesses with certain sate.

### Observed Behavior

If you try to query all transferprocess in certain by using 'filterExpression' fails with a HTTP 500 error.

## Steps to Reproduce

Use following json to query all `transferprocesses` with status `COMPLETED`

```json
{
    "@context": {
        "edc": "https://w3id.org/edc/v0.0.1/ns/"
    },
    "@type": "QuerySpec",
    "filterExpression": [
        {
            "operandLeft": "state",
            "operandRight": "COMPLETED",
            "operator": "="
        }
    ],
    "limit": 100,
    "offset": 0
}
```

## Context Information

- Used version [e.g. EDC v0.1.3]

## Detailed Description

Error message from Controlplane::
```
Caused by: org.postgresql.util.PSQLException: FEHLER: Operator does not exist: integer = character varying
  Hint: No operator matches the specified name and argument types. You may need to add explicit casts.
  Position: 188
    at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2713)
    at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2401)
    at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:368)
    at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:498)
    at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:415)
    at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:190)
    at org.postgresql.jdbc.PgPreparedStatement.executeQuery(PgPreparedStatement.java:134)
    at org.eclipse.edc.sql.SqlQueryExecutor.query(SqlQueryExecutor.java:93)
```

The field `state` in database is defined as int but the API is using string and produces a SQL query with `...WHERE  state = 'COMPLETED'` which fails because the database expects a int value for `state`.

## Possible Implementation

The API should convert the string states into corresponding int values when querying the database.

paullatzelsperger:
## What this PR changes/adds

This PR changes the `JsonObjectToCriterionTransformer` so that the `rightOperand` is transformed to `Object`, rather than `String`. 

## Why it does that

Right-hand operands other than lists and strings caused a failure in the transformer.

## Further notes

- I added another end-to-end test to test API calls in conjunction with two connectors and Postgres.
- 

## Linked Issue(s)

Closes #3319 


extensions/control-plane/store/sql/transfer-process-store-sql/src/test/java/org/eclipse/edc/connector/store/sql/transferprocess/PostgresTransferProcessStoreTest.java
```diff
@@ -171,6 +171,21 @@ void find_queryByLease() {
 
     }
 
+    @Test
+    void find_queryByState() {
```
ndr-brt, line 175: this should stay in the `TestBase` class

system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/EndToEndApiWithPostgresqlTest.java
```diff
@@ -0,0 +1,156 @@
...
+import static org.eclipse.edc.connector.transfer.spi.types.TransferProcessStates.TERMINATED;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.TYPE;
+import static org.eclipse.edc.spi.CoreConstants.EDC_NAMESPACE;
+import static org.eclipse.edc.test.e2e.PostgresConstants.JDBC_URL_PREFIX;
+import static org.eclipse.edc.test.e2e.PostgresConstants.PASSWORD;
+import static org.eclipse.edc.test.e2e.PostgresConstants.USER;
+import static org.eclipse.edc.test.system.utils.PolicyFixtures.noConstraintPolicy;
+
+@PostgresqlDbIntegrationTest
+class EndToEndApiWithPostgresqlTest {
```
ndr-brt, line 51: I'm not sure about this test scope, seems to be too broad, I mean, one we ensured that the store filters out the entities by a numeric rightOperand, we're good to go, because all the rest of the interaction is already tested.

TL;DR: I don't think we need this test

extensions/control-plane/store/sql/transfer-process-store-sql/src/test/java/org/eclipse/edc/connector/store/sql/transferprocess/PostgresTransferProcessStoreTest.java
paullatzelsperger: the test is going to be different for inmem and postgres though, because of the way how the `state` get stored. 
In-mem stores the entire `TransferProcessStates` object, while Postgres persists it as int. Thus, `state = 800` would work in the Postgres case, but not in the inmem case. And yes, we could make it work generically, but that's beyond the scope of this PR.

system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/EndToEndApiWithPostgresqlTest.java
paullatzelsperger: we don't have any other test, were we actually run a JSON-LD REST request through the connector, until it hits the database.

Now, we could create a separate runtime for management-api-e2e-tests (presumably `management-api-test-runtime-pg`), but I would do it in another PR. It's probably a good idea anyway, even beyond this particular bug.

ndr-brt: PG implementation should behave the same as the InMemory (see my other comment), to test the transformation thing I would use the `TransferProcessApiEndToEndTest` with the in memory store, given that the two implementations should behave the same (and this will be ensured by putting the `find_queryByState` in the `TestBase` class.

extensions/control-plane/store/sql/transfer-process-store-sql/src/test/java/org/eclipse/edc/connector/store/sql/transferprocess/PostgresTransferProcessStoreTest.java
paullatzelsperger: huh... I don't know how I overlooked that. yes, then it makes sense 🤦

system-tests/e2e-transfer-test/runner/src/test/java/org/eclipse/edc/test/e2e/EndToEndApiWithPostgresqlTest.java
paullatzelsperger: yeah, should be the same. creating a PG variant of the management api tests would still be a cheap (in terms of dev cost) and valuable addition i think. 
Here, the particular test case can be moved 👍

ndr-brt (approved):

PR #3321, opened 2023-07-21
feat: add pending state to `ContractNegotiation` and `TransferProcess`

### LINKED ISSUE #3308: Add manual interaction to state machines

ndr-brt:
## What this PR changes/adds

Adds a `pending` field to the `StateEntity` extension classes (`ContractNegotiation` and `TransferProcess`), and the possibility to enable it based on ax extensible `PendingGuard`. (see `state_machine.md` docs for details)

## Why it does that

This is the first step to enable "external interactions", that is the foundation for features such as the "counter offer" flow. 

## Further notes
- introduced a `Builder` based `ProcessorImpl` that will supersed `StateProcessorImpl`
- removed some longtime deprecated stuff related to CN and TP
- added `TransferProcessDefaultServicesExtension` and `ContractNegotiationDefaultServicesExtension`, to tidy up a little
- no e2e test was implemented because the feature is still not end to end complete (e.g. we're missing a way to do the state change manually)
- extracted a `SqlExecuteStatement` that helps in building up `INSERT` and `UPDATE` sql statements without having to fight against `String.format` stuff. Ref: #3322

## Linked Issue(s)

Closes #3308


core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/ContractNegotiationDefaultServicesExtension.java
```diff
@@ -0,0 +1,70 @@
...
+import org.eclipse.edc.connector.policy.spi.store.PolicyDefinitionStore;
+import org.eclipse.edc.policy.engine.spi.PolicyEngine;
+import org.eclipse.edc.runtime.metamodel.annotation.Extension;
+import org.eclipse.edc.runtime.metamodel.annotation.Inject;
+import org.eclipse.edc.runtime.metamodel.annotation.Provider;
+import org.eclipse.edc.spi.system.ServiceExtension;
+import org.eclipse.edc.spi.system.ServiceExtensionContext;
+
+@Extension(value = ContractNegotiationDefaultServicesExtension.NAME)
+public class ContractNegotiationDefaultServicesExtension implements ServiceExtension {
```
paullatzelsperger, line 35: bit of documentation?

core/common/state-machine/src/main/java/org/eclipse/edc/statemachine/ProcessorImpl.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private ProcessorImpl(Supplier<Collection<E>> entitiesSupplier) {
+        entities = entitiesSupplier;
+    }
+
+    @Override
+    public Long process() {
+        return entities.get().stream()
+                .map(entity -> {
+                    if (guard.predicate().test(entity)) {
```
paullatzelsperger, line 50: inner monk: ternary operator?

extensions/common/sql/sql-core/src/main/java/org/eclipse/edc/sql/statement/SqlExecuteStatement.java
```diff
@@ -0,0 +1,65 @@
...
+package org.eclipse.edc.sql.statement;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.lang.String.format;
+import static java.util.stream.Collectors.joining;
+import static org.eclipse.edc.sql.statement.ColumnEntry.standardColumn;
+
+public class SqlExecuteStatement {
```
paullatzelsperger, line 24: doc? Maybe `SqlModificationStatement` would be a more suitable name, to mirror the `SqlQueryStatement`?

paullatzelsperger (approved):
documentation should be added here and there, LGTM otherwise.
Love the `SqlExecuteStatement` BTW 🚀

core/common/state-machine/src/main/java/org/eclipse/edc/statemachine/StateProcessorImpl.java
```diff
@@ -29,8 +29,10 @@
  * to apply the wait strategy or not.
  *
  * @param <T> the entity that is processed
+ * @deprecated please use {@link ProcessorImpl}
  */
-public class StateProcessorImpl<T> implements StateProcessor {
+@Deprecated(since = "0.1.3")
+public class StateProcessorImpl<T> implements Processor {
```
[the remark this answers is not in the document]
ndr-brt: At least tractus-x is using this for the EDR state machine, I think this deprecation work helps a lot migrating from version to version (because the `@deprecated` tag in the javadoc text)

extensions/common/sql/sql-core/src/main/java/org/eclipse/edc/sql/statement/SqlExecuteStatement.java
ndr-brt: I took the name from the methods in `SqlQueryExecutor`: `execute` (to modify) and `query` (to get), I think it make sense to stay consistent, I'm open to modifications but on both sides eventually

jimmarino (approved):

core/common/state-machine/src/main/java/org/eclipse/edc/statemachine/StateProcessorImpl.java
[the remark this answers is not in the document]
paullatzelsperger: It seems this begs a broader discussion about SPIs, what we should and should not deprecate etc.

PR #3369, opened 2023-08-16
fix: Add PathParam

bcronin90:
## What this PR changes/adds

Adds the parameter specified in the path to the methods under said path.

## Why it does that

OpenAPI considers this an error otherwise.

## Further notes

There's a few ways we could go about this and I'd like some input:
1) Remove the path parameter, as it is used only to forward a call to any path to this API. This would be my preferred approach.
2) Use the path parameter to resolve the source data address if this cannot be done via the token (not sure if this makes sense though).
3) Leave the parameter, but don't use it. (Current state of this PR).

Ref. #3371


ndr-brt:
@bcronin90 could you open a discussion first to shown the issue and propose a solution?

bcronin90:
> @bcronin90 could you open a discussion first to shown the issue and propose a solution?

I opened a discussion. After much googling, there might be a simple and clean solution. I'll try to test that in the meanwhile.

PR #3387, opened 2023-08-28
fix: send correlation id as process id in the transfer messages

### LINKED ISSUE #3334: consumer `TransferProcesses` not Transitioning to Completed State
# Bug Report

## Describe the Bug

With the latest 0.2.0 version our tests are failing, because the consumer `transferprocess` is not transitioning to `Completed` state. This is caused by a missing `StatusChecker` implementation for `HttpData` type.

### Expected Behavior

Providing a default implementation for the `HttpData` `StatusChecker` or transition to `Completed` when there is no `StatusChecker` registered.

### Observed Behavior

Data successfully transfered, but consumer `transferprocesses` stuck in `Started` state.

## Context Information

- Used version EDC v0.2.0

## Detailed Description

Log:
No checker found for process d0fcaa06-2e81-466f-a95a-3debf49a0885. The process will not advance to the COMPLETED state.

https://github.com/eclipse-edc/Connector/blob/de80081d547c66f127dd32e1a8b27d9fca2d4410/core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/process/TransferProcessManagerImpl.java#L422-L433

### LINKED ISSUE #3377: TPM does not always use correlation ID in messages
# Bug Report

## Describe the Bug

When sending a message regarding a `TransferProcess` (e.g. for termination), the `TransferProcessManager` does not always use the same ID. For outgoing messages regarding a process, the process's correlation ID should be used, as this is shared between the two parties. For e.g. sending a  `TransferStartMessage`, the correct ID is used:

```java
var message = TransferStartMessage.Builder.newInstance()
                 .processId(process.getCorrelationId())
``` 

but in case of termination and completion, the process ID is used (which, if the message is sent by the provider, is not known to the consumer):

```java
var message = TransferTerminationMessage.Builder.newInstance()
                ...
                .processId(process.getId())
``` 

This causes an error on the counter-party side, as the message can not be correlated to an existing process.

### Expected Behavior
Each message regarding a `TransferProcess` contains an ID known to the counter-party.

### Observed Behavior

Some messages contain an ID that is not known to the counter-party.

## Possible Implementation
Set the `process.getCorrelationId()` for the `processId` on all messages.

ndr-brt:
## What this PR changes/adds

Send correlationId and not transfer process id in the protocol messages

## Why it does that

fix bug

## Further notes

- I enhanced the e2e transfer tests by making the TP going to COMPLETED, this was done adding the `control-plane-api` and `control-plane-api-client` to the runtimes
- **question 1**: given that we have this control-plane to data-plane communication I wonder if we need the `statusChecker` at all. From an hi level perspective it is correct that the data plane has the knowledge about the transfer completion, this will download some responsibilities from the control-plane
- **question 2** Currently as specified in #3334 not everyone is aware that there's this communication layer between CP and DP so I was wondering if it would be better to get rid of the `NoopTransferProcessClient` to clearly state that a DP needs to communicate with the CP, otherwise it should not start.

thoughts on these 2? 

## Linked Issue(s)

Closes #3377 
Closes #3334 

### DIFF
--- a/core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/process/TransferProcessManagerImpl.java
+++ b/core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/process/TransferProcessManagerImpl.java
@@ -455,7 +455,7 @@ public class TransferProcessManagerImpl implements TransferProcessManager {
         var message = TransferCompletionMessage.Builder.newInstance()
                 .protocol(process.getProtocol())
                 .counterPartyAddress(process.getConnectorAddress())
-                .processId(process.getId())
+                .processId(process.getCorrelationId())
                 .policy(policyArchive.findPolicyForContract(process.getContractId()))
                 .build();
 
@@ -487,7 +487,7 @@ public class TransferProcessManagerImpl implements TransferProcessManager {
         var message = TransferTerminationMessage.Builder.newInstance()
                 .counterPartyAddress(process.getConnectorAddress())
                 .protocol(process.getProtocol())
-                .processId(process.getId())
+                .processId(process.getCorrelationId())
                 .policy(policyArchive.findPolicyForContract(process.getContractId()))
                 .build();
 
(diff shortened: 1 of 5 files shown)


wolf4ood:
@ndr-brt

IMHO

1) I would be inclined to remove the status checker since it require client side polling, which can slowdown the client statemachine. Since the provider dataplane knows when the transfer is finished it make sense to propagate it directly via `TransferCompletionMessage` 

 2) Probably it make sense to enforce this communication in the runtime without providing a noop one.

wolf4ood (approved):

Issue #3398, opened 2023-08-31
Add optional logging of HTTP requests and responses in tests


iwasakims:
# Feature Request

Recording HTTP requests and responses in unit tests which can be enabled by knobs like system properties

## Which Areas Would Be Affected?

tests

## Why Is the Feature Desired?

Logging HTTP requests and responses is useful for debugging and analysing the interaction between components.

## Solution Proposal

For okhttp, logging can be enabled by adding [okhttp-logging-interceptor](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/README.md).

For rest-assured, logging can be enabled by adding [RequestLoggingFilter](https://github.com/rest-assured/rest-assured/blob/rest-assured-5.3.1/rest-assured/src/main/java/io/restassured/filter/log/RequestLoggingFilter.java) an\
d [ResponseLoggingFilter](https://github.com/rest-assured/rest-assured/blob/rest-assured-5.3.1/rest-assured/src/main/java/io/restassured/filter/log/ResponseLoggingFilter.java).

PR #3399, opened 2023-08-31
feat: logging http requests and responses handled by rest-assured

iwasakims:
## What this PR changes/adds

add utility method to enable log HTTP requests and responses handled by rest-assured in tests. 

## Why it does that

Logging HTTP requests and responses is useful for debugging and analysing the interaction between components.

## Further notes

This PR enabled the logging in e2e-transfer-test as a first place to poc. The logging can be shown by running the integration test like:

```
$ ./gradlew clean test -p system-tests/e2e-transfer-test/runner -PverboseTest -DincludeTags="EndToEndTest" --tests "*EndToEndTransferInMemoryTest.httpPushDataTransfer"
```

## Linked Issue(s)

This is addressing the part of #3398. I would like to address logging of okhttp in another PR.

### DIFF
--- a/core/common/junit/build.gradle.kts
+++ b/core/common/junit/build.gradle.kts
@@ -31,6 +31,7 @@ dependencies {
 
     implementation(libs.junit.pioneer)
     implementation(libs.testcontainers.junit)
+    implementation(libs.restAssured)
 }
 
 
--- a/core/common/junit/src/main/java/org/eclipse/edc/junit/testfixtures/TestUtils.java
+++ b/core/common/junit/src/main/java/org/eclipse/edc/junit/testfixtures/TestUtils.java
@@ -15,6 +15,9 @@
 package org.eclipse.edc.junit.testfixtures;
 
 import dev.failsafe.RetryPolicy;
+import io.restassured.RestAssured;
+import io.restassured.filter.log.RequestLoggingFilter;
+import io.restassured.filter.log.ResponseLoggingFilter;
 import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
 import org.eclipse.edc.connector.core.base.EdcHttpClientImpl;
@@ -211,4 +214,10 @@ public class TestUtils {
         return null;
     }
 
+    /**
+     * Utility method to add filters to log http requests and respoinses handled by rest-assured.
+     */
+    public static void addRestAssuredLoggingFilters() {
+        RestAssured.filters(new RequestLoggingFilter(), new ResponseLoggingFilter());
+    }
 }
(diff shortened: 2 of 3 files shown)


jimmarino:
I'm not a fan of adding additional logging in the tests, particularly since a debugger is more efficient for analyzing interactions. In most cases, the debugger will be needed anyway, so is there something that logging provides that the former doesn't?

ndr-brt (commented):
I'm not sure what this should be useful for, logging every http call would result in a bloated console output.

Tests should work without having the need to look at what the SUT is doing. 
If that's for debugging purpose, I suggest to use at need the `RestAssured` `.log().ifError()` call to log only calls that didn't return a successful response.

iwasakims:
Debugger is useful if you already know where to set watches/breakpoints. It is difficult to trace the interaction in wider scope especially for tests like e2e in which many modules asynchronously interact. Attaching to a thread and seeing around internals tends to affect timing and cause timeout. In addition, seeing whole headers and bodies of HTTP requests/responses by debugger is not so easy since it is handled inside code path of okhttp/rest-assured rather than EDC itself.

iwasakims:
If successful response is not returned, we can leverage debugger since we already know where to see. HTTP log is useful for investigating the effect of code changes even in successfully working code path.

I do not intend to always enable logging in all tests here.

jimmarino:
> Debugger is useful if you already know where to set watches/breakpoints. It is difficult to trace the interaction in wider scope especially for tests like e2e in which many modules asynchronously interact. Attaching to a thread and seeing around internals tends to affect timing and cause timeout. In addition, seeing whole headers and bodies of HTTP requests/responses by debugger is not so easy since it is handled inside code path of okhttp/rest-assured rather than EDC itself.

I'm not sure what use case is being solved here, particularly since this is being added to the tests. The EDC already supports OKHttp interceptors for people who want to monitor raw HTTP messages in their environment.

Breakpoints don't need to block the JVM and for most classes of bugs that involve the end-to-end tests, a debugger is the most efficient tool. HTTP messages generally won't provide enough data to understand most end-to-end issues and logging tends to pollute the console (we already have way too much logging in the test cases as it is).

ndr-brt:
> Debugger is useful if you already know where to set watches/breakpoints. It is difficult to trace the interaction in wider scope especially for tests like e2e in which many modules asynchronously interact. Attaching to a thread and seeing around internals tends to affect timing and cause timeout. In addition, seeing whole headers and bodies of HTTP requests/responses by debugger is not so easy since it is handled inside code path of okhttp/rest-assured rather than EDC itself.

restassured throws a failed assertion error in case the response is not as expected, from the stack trace you'll be able to go up to the failed call, you can then add the `.log().onError()` and you're ready to go.

PR #3402, opened 2023-08-31
feat: adds counter party validation on all DSP messages

### LINKED ISSUE #3397: dsp: add counter party validation to all the `TransferProcess` DSP incoming messages
# Feature Request

Currently we're validating only the `TransferRequestMessage` because it's the only that really matters, but for consistency we should validate also the counter party id on every other message (start, completion, termination)

protocol service

consistency

## Solution Proposal

wolf4ood:
## What this PR changes/adds

Add counter party validation on all DSP messages

## Why it does that
consistency

## Further notes
Shouldn't we also break lease in case of conflict or other error case? e.g.

https://github.com/eclipse-edc/Connector/blob/73d53586a6bb82e083de3dbd6b469e006111b39a/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/transferprocess/TransferProcessProtocolServiceImpl.java#L172-L184

## Linked Issue(s)

Closes #3397 


jimmarino (approved):

ndr-brt (approved):
Yes, we should break the lease for every "failed" operation, this slipped off the PR when we introduced leasing in the services

PR #3428, opened 2023-09-08
feat: add priority on `DataFlowController` registration

### LINKED ISSUE #3414: `DataFlowController` registration is not fully controllable
# Feature Request

Currently the choice of the `DataFlowController` to be used by the `DataFlowManagerImpl` depends on the order they get registered, but this could cause some `DataFlowController` to be not usable, e.g.
in a control-plane where there's the `transfer-data-plane` module registered, any other `DataFlowController` could be registered after the  `ProviderPushTransferDataFlowController` and `ConsumerPullTransferDataFlowController` and never be invoked (because the conditions of their "canHandle" methods cover all the possibilities)
It would be better to have a more fine grained way to handle that choice.

transfer

extensibility

## Solution Proposal
one easy solution could be to have a fallback registration (that in our case could be the `ProviderPushTransferDataFlowController`) that will be invoked as a fallback, then leaving the other be chosen based on the result of the `canHandle` method.

ndr-brt:
## What this PR changes/adds

Add priority on `DataFlowController` registration

## Why it does that

to permit having multiple `DataFlowController`s that could be chosen for a particular transfer, and having other used as fallback.
This is useful when there's, for example, the `ProviderPushDataFlowController` that pretty much can transfer anything but there's a specific `DataFlowController` for a particular `source`-to`destination` transfer that will be chosen first if the priority set is bigger. 

## Further notes

- simplified the interface methods on `DataFlowManager` and `DataFlowController`, now they accept the whole `TransferProcess` object.

## Linked Issue(s)

Closes #3414

### DIFF
--- a/core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/flow/DataFlowManagerImpl.java
+++ b/core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/flow/DataFlowManagerImpl.java
@@ -18,13 +18,13 @@ import io.opentelemetry.instrumentation.annotations.WithSpan;
 import org.eclipse.edc.connector.transfer.spi.flow.DataFlowController;
 import org.eclipse.edc.connector.transfer.spi.flow.DataFlowManager;
 import org.eclipse.edc.connector.transfer.spi.types.DataFlowResponse;
-import org.eclipse.edc.connector.transfer.spi.types.DataRequest;
+import org.eclipse.edc.connector.transfer.spi.types.TransferProcess;
 import org.eclipse.edc.policy.model.Policy;
 import org.eclipse.edc.spi.response.StatusResult;
-import org.eclipse.edc.spi.types.domain.DataAddress;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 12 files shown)


core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/flow/DataFlowManagerImpl.java
```diff
@@ -34,33 +34,43 @@
...
     }
 
     private String runtimeException(String id, String message) {
-        return format("Unable to process data request %s. Data flow controller throws an exception: %s", id, message);
+        return format("Unable to process transfer %s. Data flow controller throws an exception: %s", id, message);
     }
 
     private String controllerNotFound(String id) {
-        return format("Unable to process data request %s. No data flow controller found", id);
+        return format("Unable to process transfer process %s. No data flow controller found", id);
```
jimmarino, line 71:
```suggestion
        return format("Unable to process transfer %s. No data flow controller found", id);
```

jimmarino (approved):

wolf4ood (approved):

PR #3443, opened 2023-09-12
build: speed up e2e transfer tests

### LINKED ISSUE #3442: speed up E2E transfer tests
# Feature Request

Currently these tests took a lot because for every test the whole set of runtimes need to be started and stopped

system-tests

CI speed

## Solution Proposal

start the runtimes once per class

ndr-brt:
## What this PR changes/adds

Speed up E2E transfer test, by implementing a `EdcClassRuntimesExtension` that takes care of starting the runtimes once per every test classes.

## Why it does that

CI

## Further notes

- had to add a check to force retry on the `TransferProcessHttpClient` because it's too quick in completing the transfer and the data-plane is not able to notify the completion because the entity is still leased. This implementation is useful also on production because it prevents eventual control-plane downs.

## Linked Issue(s)

Closes #3442


paullatzelsperger (approved):
I like the idea, we just need to careful when to use it an when not to, because if the runtime "survives" multiple test runs, we could potentially have leftover data in the (inmem) databases.

PR #3463, opened 2023-09-20
fix: remove SLF4j-Monitor bridge

### LINKED ISSUE #3451: Make log level of slf4j logger configurable
# Feature Request

boot, monitor

We can not get log of underlying libraries using slf4j if the log level is equal to or smaller than INFO [since the log level of slf4j logger is hard coded](https://github.com/eclipse-edc/Connector/blob/7c109a8696404759700bb43e20b61c9bb9afb188/core/common/boot/src/main/java/org/eclipse/edc/boot/monitor/MonitorProvider.java#L67C26-L142).

It is inconvenient for developing modules depending on libraries using slf4j. [Technology-Aws](https://github.com/eclipse-edc/Technology-Aws) depending on AWS SDK is the example.

We can not temporalily use another binding like slf4j-simple if [MonitorProvider](https://github.com/eclipse-edc/Connector/blob/7c109a8696404759700bb43e20b61c9bb9afb188/core/common/boot/src/main/java/org/eclipse/edc/boot/monitor/MonitorProvider.java) is in the classpath. Since the MonitorProvider is provided by org.eclipse.edc:boot, most of the code depending on EDC could be affected.

```
S3DataPlaneIntegrationTest STANDARD_ERROR
    SLF4J: Class path contains multiple SLF4J providers.
    SLF4J: Found provider [org.eclipse.edc.boot.monitor.MonitorProvider@47c64cfe]
    SLF4J: Found provider [org.slf4j.simple.SimpleServiceProvider@6ce90bc5]
    SLF4J: See https://www.slf4j.org/codes.html#multiple_bindings for an explanation.
    SLF4J: Actual provider is of type [org.eclipse.edc.boot.monitor.MonitorProvider@47c64cfe]
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
```

## Solution Proposal

Making the log level of slf4j logger configurable [as done for slf4j-simple](https://www.slf4j.org/apidocs/org/slf4j/simple/SimpleLogger.html).

Making the feature opt-in could be an option. If the feature is enabled by loading the extension, EDC Config is available on `initialize`. Modules not loading BaseRuntime (e.g. Technology-AWS) can use alternative slf4j binding during development.

iwasakims:
## What this PR changes/adds

Remove MonitorProvider from the code path of Base Runtime and make it opt-in feature provided by ServiceExtension.

## Why it does that

MonitorProvider is a SLF4J binding which forward logs emitted by SLF4J API to Monitor. Downstreams of org.eclipse.edc:boot can not use another SLF4J bindings it the MonitorProvider is in the classpath by default. Since current implementation of MonitorProvider#MonitorLogger has hard-coded log level, most of logs of underlying libraries using SLF4J can not be seen.

## Further notes

This PR moved MockMonitorExtension to testFixtures of :core:common:junit in order to make it accessible from added test code.

## Linked Issue(s)

Closes #3451 by making another SLF4J binding available. This also is prerequisite to make the MonitorLogger configurable via EDC configuration.


paullatzelsperger:
@iwasakims thanks for the contribution. Be aware though, that we first need to triage the linked issue before we merge this.

iwasakims:
I'm ok to submit another PR to remove the added slf4j-bridge shortly after merging this.

paullatzelsperger:
> I'm ok to submit another PR to remove the added slf4j-bridge shortly after merging this.

why add something that's getting removed shortly thereafter? i think it can get removed right away. That's the reason why PRs should only raised once issues are out of `triage`: getting clarity on the desired implementation.

iwasakims:
I update this PR to remove the SLF4J bridge.

> That's the reason why PRs should only raised once issues are out of triage: getting clarity on the desired implementation.

Seeing the code change should be effective to make the issue clear and affected scope clear in this case. I needed some action before the issue is auto-closed without update for a week.

core/common/junit/build.gradle.kts
```diff
@@ -31,6 +31,8 @@ dependencies {
 
     implementation(libs.junit.pioneer)
     implementation(libs.testcontainers.junit)
+
+    testFixturesImplementation(libs.mockito.core)
```
jimmarino, line 35: Hi, why is this added?

iwasakims: This was for moving MockMonitorExtension from src/test to  src/testFixtures in order to use it from added test in [the first commit](https://github.com/eclipse-edc/Connector/pull/3463/commits/a6a0dcccb216eabee434c7b58d750c707f6564f3). It can be reverted if we add no test here.

paullatzelsperger (approved):
LGTM

paullatzelsperger:
> The diff of DEPENDENCIES looks not related to this PR this time. Should I update DEPENDENCIES again?
> 
> ```
> Run if diff DEPENDENCIES DEPENDENCIES-gen ; then
> 166,167c166,167
> < maven/mavencentral/jakarta.validation/jakarta.validation-api/2.0.2, Apache-2.0, approved, clearlydefined
> < maven/mavencentral/jakarta.validation/jakarta.validation-api/3.0.2, Apache-2.0, approved, clearlydefined
> ---
> > maven/mavencentral/jakarta.validation/jakarta.validation-api/2.0.2, Apache-2.0, approved, ee4j.validation
> > maven/mavencentral/jakarta.validation/jakarta.validation-api/3.0.2, Apache-2.0, approved, ee4j.validation
> 166,167c166,167
> < maven/mavencentral/jakarta.validation/jakarta.validation-api/2.0.2, Apache-2.0, approved, clearlydefined
> < maven/mavencentral/jakarta.validation/jakarta.validation-api/3.0.2, Apache-2.0, approved, clearlydefined
> ---
> > maven/mavencentral/jakarta.validation/jakarta.validation-api/2.0.2, Apache-2.0, approved, ee4j.validation
> > maven/mavencentral/jakarta.validation/jakarta.validation-api/3.0.2, Apache-2.0, approved, ee4j.validation
> ------------------------------------------------------------
> ```

The DEPENDENCIES file is required by the Eclipse Foundatation, and it must always be kept up-to-date. Unfortunately it can become outdated even if you didn't even change a thing, for example when an IPLab review request gets processed and updated. However, removing the SLF4J dependency will definitely outdate the DEPENDENCIES file.

core/common/junit/build.gradle.kts
jimmarino: Sorry, I'm a bit confused. That dependency no longer appears to be required based on the state of the current PR. Am I missing something? Where is the added test in the current PR state (not previous commits that appear to be partially reverted)?

iwasakims: This change was removed by https://github.com/eclipse-edc/Connector/pull/3463/commits/1cb54eebb92bf6076691c15deac40064c49a83ab.

jimmarino (approved):

PR #3471, opened 2023-09-25
fix: retry on 204 instead of 200 

### LINKED ISSUE #3468: `TransferProcessHttpClient` should retry on 204
# Bug Report

## Describe the Bug

The current retry when notifying the CP about a data plane transfer completed or failed is bound to `200` but it should be `204` since no content is returned from the `TransferProcessControlApi`

wolf4ood:
## What this PR changes/adds

Change the status code of the retry from `200` to `204` 

## Why it does that

## Further notes

Refactor also the `TransferProcessApiClient` to use Result on the return  so the `DataPlaneManagerImpl` can act accordingly

## Linked Issue(s)

Closes #3468


wolf4ood:
yeah in this case the controller return void in both cases. But i guess we could also support `200`

spi/common/http-spi/src/main/java/org/eclipse/edc/spi/http/FallbackFactories.java
```diff
@@ -67,4 +70,26 @@ static FallbackFactory retryWhenStatusIsNot(int status) {
                     .build();
         };
     }
+
+    /**
+     * Verifies that the response has a specific statuses, otherwise it should be retried
+     *
+     * @return the {@link FallbackFactory}
+     */
+    static FallbackFactory retryWhenStatusIsNotIn(int... status) {
```
ndr-brt, line 79: At this point `retryWhenStatusIsNot` could be inlined and all its callers could call this new method.

ndr-brt (approved):
just a duplication nit, all good otherwise

PR #3481, opened 2023-09-28
feat: terminate data flow on transfer completion

### LINKED ISSUE #3453: policy-monitor: data flow should be stopped on transfer completion
# Feature Request

When the `TransferProcess` gets completed by the policy monitor, it should terminate the data flow accordingly.

This should be done either on the `DataFlowControllerRegistry` that need to find the correct `DataFlowController` and terminate the flow, or in the "deprovisioning" phase.

The only implementation we need to put in place would be for the `ProviderPushTransferDataFlowController`, where the terminate call should be done either:
- on the previous choosed dataplane 
- on every dataplane that can handle that dataflow.
likely the second one is the correct one, because the dataplane that started the dataflow could have been stopped and replaced with another one,

The consumer-pull flow does not need a terminate method because the token expiration will already do the job.

_Are there any requirements?_

## Solution Proposal

ndr-brt:
## What this PR changes/adds

Ensures that the data flow gets terminated when the PolicyMonitor detects a no more valid policy.

## Why it does that

policy monitor

## Further notes

- now a `DataFlowController` can be terminated through the `terminate` method
- in the data plane, a flow termination can be done by calling `close` on the `DataSource`

## Linked Issue(s)

Closes #3453


jimmarino:
@ndr-brt It would be good to first add a DR outlining this new state and its uses.

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/transferprocess/TransferProcessServiceImpl.java
```diff
@@ -91,6 +92,11 @@ public ServiceResult<Stream<TransferProcess>> query(QuerySpec query) {
         });
     }
 
+    @Override
+    public ServiceResult<Void> stop(StopTransferCommand command) {
```
paullatzelsperger, line 96: to keep consistency with the other methods, this should probably also be `@NotNull`

paullatzelsperger (commented):
I had a few questions regarding "terminating" vs "stopping".

- when we `terminate`, we send a `TransferTerminationMessage`, but when we `stop`, we don't send anything. Is that on purpose? Shouldn't we notify the counter-party, that a transfer has been stopped?
- conversely, when we "terminate", we _don't_ invoke the dataflow manager, so theoretically there could still be an ongoing transfer, right?
- isn't "stopping" almost the same as "terminating", but with a "configurable" next state? Or rather: isn't terminating just stopping with a preconfigured next state?

If those questions are scoped too broadly, we can just merge the PR and discuss elsewhere

ndr-brt:
@paullatzelsperger , yes, sorry, I discussed with @jimmarino after then and decided to avoid that STOPPING status, instead the data flow termination will be invoked before the transfer process termination. I updated the PR accordingly, now it's ready to be reviewed.

bscholtes1A (approved):
Great stuff!

wolf4ood (approved):

jimmarino (approved):

PR #3484, opened 2023-09-29
docs: add control api endpoint 

wolf4ood:
## What this PR changes/adds

Adds docs for `edc.control.endpoint` which is needed for notify the CP once the transfer has finished at the DP level.

## Why it does that

documentation

## Further notes

Removes also `@Provides(TransferProcessApiClient.class)` not needed since there is a `Provider` annotation.

### DIFF
--- a/extensions/common/api/control-api-configuration/README.md
+++ b/extensions/common/api/control-api-configuration/README.md
@@ -10,4 +10,8 @@ Exemplary configuration:
 ```properties
 web.http.control.port=9191
 web.http.control.path=/api/v1/control
+edc.control.endpoint=<control-api-endpoint>
 ```
+
+The `edc.control.endpoint` will be used by the data plane to notify the control plane when data transfer completes or
+fails.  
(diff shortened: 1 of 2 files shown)


ndr-brt (commented):
I think the configuration documentation should be done through `@Setting` annotation and rendered by the autodoc, generally we should avoid documenting configuration in the README.md because they tend to get out of date.

wolf4ood:
I Agree but until we have in place an actual final rendering MD or whatever i would need i pointer for users in case right? :)

paullatzelsperger:
> I agree but until we have in place an actual final rendering MD or whatever i would need i pointer for users in case right? :)

rendering autodoc manifests to markdown is [already implemented](https://github.com/eclipse-edc/GradlePlugins/pull/166), so everyone could already generate md files

paullatzelsperger (approved):
i'd remove the README too, but it can be done later as well

PR #3486, opened 2023-09-29
refactor: make `DataSink` return a content

### LINKED ISSUE #3474: data-plane: make `DataSink` result contain a content object.
# Feature Request

Currently `DataSink.transfer` method returns a `CompletableFuture<StreamResult<Void>>`:
https://github.com/eclipse-edc/Connector/blob/7dc9dded2f1abd35eab5cbdb38154e2991cf37d7/spi/data-plane/data-plane-spi/src/main/java/org/eclipse/edc/connector/dataplane/spi/pipeline/DataSink.java#L27

that does not permit a transfer operation to have an output result, that's why in the `DataPlanePublicApiController` there's this workaround to permit to get an output.
https://github.com/eclipse-edc/Connector/blob/7dc9dded2f1abd35eab5cbdb38154e2991cf37d7/extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/controller/DataPlanePublicApiController.java#L143-L157

It would be nice to have an optional output content.

This will make the `TransferService.validate` and, consequently, the `DataPlaneManager.validate` eligible for being included in the `transfer` method logic, as they are always called one after the other.

data-plane

avoid duplication

ndr-brt:
## What this PR changes/adds

Change the return type of the `DataSink` `transfer` method in `CompletableFuture<StreamResult<Object>>` so then optionally a `DataSink` can return a content, that can be used in cases as the pull transfer, this will avoid the need to pass a custom `DataSink` (see `DataPlanePublicApiController`).
This permitted some cleanup here and there.

## Why it does that

refactor

## Further notes

- this PR will be ready for review after #3481 will be merged.

## Linked Issue(s)

Closes #3474 


spi/data-plane/data-plane-spi/src/main/java/org/eclipse/edc/connector/dataplane/spi/pipeline/PipelineService.java
```diff
@@ -15,28 +15,14 @@
...
-
-import java.util.concurrent.CompletableFuture;
 
 /**
  * Transfers data from a source to a sink.
+ *
+ * @deprecated this interface will be removed from the hierarchy, please use {@link TransferService} instead.
  */
 @ExtensionPoint
+@Deprecated(since = "0.3.0")
```
bscholtes1A, line 25: Do we really want to completely get rid of PipelineService? 

In my opinion, it should stay with only the `registerFactory` methods in it. As the `DataSourceFactory` and `DataSinkFactory` will only be used within the PipelineServiceImpl, don't they?

ndr-brt: I found out that `PipelineService` at this point is exactly the same as `TransferService`, I don't see a reason to keep both, we should either keep one or the other.

ndr-brt: `TransferService` is also (quoting the javadoc) 
> can satisfy a DataFlowRequest by transferring data from a source to a destination.

so I cannot see the need for having such distinction, also the api is pretty much the same.
What could be a `TransferService` that does not rely on `DataSource` or `DataSink`? Currently there isn't any implementation available.

bscholtes1A: Here is an example: https://github.com/eclipse-edc/Technology-Azure/blob/main/extensions/data-plane/data-plane-azure-data-factory/src/main/java/org/eclipse/edc/connector/dataplane/azure/datafactory/AzureDataFactoryTransferService.java 

`TransferService` was initially introduced when the source/destination pair enables to transfer the data without streaming them through the Data Plane (which is what the `PipelineService` is used for).

extensions/data-plane/data-plane-http/src/test/java/org/eclipse/edc/connector/dataplane/http/DataPlaneHttpExtensionTest.java
```diff
@@ -83,7 +84,7 @@ void transferSourceToDestination(PipelineService pipeline) {
     }
 
     @Test
-    void transferSourceToDestinationAddHeaders(PipelineService pipeline, HttpRequestParamsProvider paramsProvider) {
+    void transferSourceToDestinationAddHeaders(TransferService transferService, HttpRequestParamsProvider paramsProvider) {
```
bscholtes1A, line 87: Should be reverted to `PipelineService` after last change?

bscholtes1A (approved):
Unit-test to be fixed but apart from that looks great to me!

Issue #3508, opened 2023-10-04
JSONLD: Context Scopes


jimmarino:
We need to add the ability to define Json-Ld contexts by processing context. 

Currently, all namespaces are included when performing compaction, regardless of whether they are referenced in the message body. Having a scoping mechanism will allow namespaces to be registered for specific egress (compaction) points. In addition, we need to add the ability to specify context references that should be included in a scope, not just namespaces.

PR #3516, opened 2023-10-05
feat: add configurable JSON-LD namespaces/context for each scope

wolf4ood:
## What this PR changes/adds

Add configurable JSON-LD namespaces/prefix according to the DR #3509 

## Why it does that

Flexibility and clean API

## Further notes

Still not sure about the management API scopes. For now only the base `edc` and the `odrl` are registered there,
bit the mgmt catalog API might speak also `dcat` and `dct`. As today we just forward the JSON-LD from the provider without touching it, so at code level it works and we might not care for now. 

## Linked Issue(s)

Implements #3508

### DIFF
--- a/core/common/jersey-providers/src/test/java/org/eclipse/edc/web/jersey/jsonld/JerseyJsonLdInterceptorTest.java
+++ b/core/common/jersey-providers/src/test/java/org/eclipse/edc/web/jersey/jsonld/JerseyJsonLdInterceptorTest.java
@@ -33,6 +33,7 @@ import static io.restassured.RestAssured.given;
 import static io.restassured.http.ContentType.JSON;
 import static org.hamcrest.CoreMatchers.is;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoInteractions;
@@ -41,8 +42,9 @@ import static org.mockito.Mockito.when;
 @ApiTest
 class JerseyJsonLdInterceptorTest extends RestControllerTestBase {
 
+    private static final String SCOPE = "scope";
     private final JsonLd jsonLd = mock();
-    private final JerseyJsonLdInterceptor interceptor = new JerseyJsonLdInterceptor(jsonLd, objectMapper);
+    private final JerseyJsonLdInterceptor interceptor = new JerseyJsonLdInterceptor(jsonLd, objectMapper, SCOPE);
 
     @Test
     void expansion_shouldSucceed_whenInputIsJsonObject() {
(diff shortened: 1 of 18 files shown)


core/common/jersey-providers/src/test/java/org/eclipse/edc/web/jersey/jsonld/JerseyJsonLdInterceptorTest.java
```diff
@@ -188,6 +190,18 @@ protected Object additionalResource() {
         return interceptor;
     }
 
+    private JsonObject expandedJson() {
+        return Json.createObjectBuilder().add("expanded-key", "expanded-value").add("bau", 3).build();
```
ndr-brt, line 194: I like "bau" I'm also using it a lot :rofl:

spi/common/json-ld-spi/src/main/java/org/eclipse/edc/jsonld/spi/JsonLd.java
```diff
@@ -38,15 +39,54 @@ public interface JsonLd {
...
+    /**
+     * Register a JsonLD namespace in a provided scope
      *
-     * @param prefix the prefix
+     * @param prefix     the prefix
      * @param contextIri the string representing the IRI where the context is located
+     * @param scope      the scope where the prefix will be applied
      */
-    void registerNamespace(String prefix, String contextIri);
+    void registerNamespace(String prefix, String contextIri, String scope);
```
ndr-brt, line 72: Maybe I would put "scope" as first parameter, like it was done in the policy engine

extensions/common/json-ld/src/main/java/org/eclipse/edc/jsonld/TitaniumJsonLd.java
```diff
@@ -118,10 +140,39 @@ private JsonObject injectVocab(JsonObject json) {
...
+        // Compute the additional context IRI defined for * and the input scope
+        var contexts = Stream.concat(contextsForScope(DEFAULT_SCOPE), contextsForScope(scope))
+                .collect(Collectors.toSet());
+
+        // if not empty we build a JsonArray
+        if (!contexts.isEmpty()) {
+            var arrayBuilder = createArrayBuilder();
+            contexts.forEach(arrayBuilder::add);
+
+            var namespaces = builder.build();
```
ndr-brt, line 158: the namespace object built can be done before the if statement

ndr-brt (approved):
just some nits, but LGTM otherwise

spi/common/json-ld-spi/src/main/java/org/eclipse/edc/jsonld/spi/JsonLd.java
wolf4ood: We have the other signature without the scope and it will lead to different
arg positions which could confuse people?

```Java
 void registerNamespace(String prefix, String contextIri)
 void registerNamespace(String scope. String prefix, String contextIri)
```

ndr-brt: mmh.. yes, it could be as they are all strings... ok for me to leave as it is

jimmarino (approved):

PR #3526, opened 2023-10-10
feat: register @vocab context definition

### LINKED ISSUE #3522: openapi: PolicyDefinition examples are not expanding correctly
# Feature Request

The responseBody of the [`GET /policydefinitions` endpoint](https://app.swaggerhub.com/apis/eclipse-edc-bot/management-api/0.3.1#/Policy%20Definition/getPolicyDefinition) is not expanding correctly on the [jsonld playground](https://json-ld.org/playground/)

api

documentation

## Solution Proposal

add namespaces to the example

### LINKED ISSUE #3525: json-ld: register `@vocab` on the management api scope
# Feature Request

Currently the output json-ld created by the management api lack of the `@vocab` node.
That's why when we compact the json-ld, every "edc" namespaced property is shown as `edc:propertyName` and not only `propertyName`. Not an huge issue, but this could create pitfalls as in #3522 .

management api

api

## Solution Proposal

the `@vocab` should be registered, at least on the management api scope, but it could done also for all the scopes

ndr-brt:
## What this PR changes/adds

Registers `@vocab` context definition (with the `https://w3id.org/edc/v0.0.1/ns/` value) at the global scope

## Why it does that

To have consistent output jsonld responses

## Further notes

- this way, every json-ld produced by the edc won't have the "edc:" prefix in the edc namespaced properties
- as this could be considered as a breaking change (but technically it is not because the compacted json-ld responses will expand in the same way as the old ones) I added a `edc.jsonld.vocab.disable` that could be used to switch to the old behavior.

## Linked Issue(s)

Closes #3525 
Closes #3522 

### DIFF
--- a/extensions/control-plane/api/management-api/management-api-test-fixtures/src/testFixtures/java/org/eclipse/edc/test/system/utils/Participant.java
+++ b/extensions/control-plane/api/management-api/management-api-test-fixtures/src/testFixtures/java/org/eclipse/edc/test/system/utils/Participant.java
@@ -36,7 +36,6 @@ import java.util.concurrent.atomic.AtomicReference;
 import static io.restassured.RestAssured.given;
 import static io.restassured.http.ContentType.JSON;
 import static jakarta.json.Json.createObjectBuilder;
-import static java.lang.String.format;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
 import static org.eclipse.edc.connector.contract.spi.types.negotiation.ContractNegotiationStates.FINALIZED;
@@ -385,7 +384,7 @@ public class Participant {
                 .then()
                 .statusCode(200)
                 .extract().body().jsonPath()
-                .getString(format("%s", fieldName));
+                .getString(fieldName);
     }
 
     /**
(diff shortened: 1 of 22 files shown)


paullatzelsperger:
This is about adding the `@vocab` on "egress" (i.e. compaction), right? because on ingress (i.e. on expansion) we already have that...

ndr-brt:
@paullatzelsperger yes, in the compaction side of things

extensions/control-plane/api/management-api/management-api-test-fixtures/src/testFixtures/java/org/eclipse/edc/test/system/utils/Participant.java
```diff
@@ -385,7 +385,7 @@ private String getContractNegotiationField(String negotiationId, String fieldNam
                 .then()
                 .statusCode(200)
                 .extract().body().jsonPath()
-                .getString(format("'edc:%s'", fieldName));
+                .getString(format("%s", fieldName));
```
paullatzelsperger, line 388:
```suggestion
                .getString(fieldname);
```

paullatzelsperger (approved):
LGTM

wolf4ood (approved):

Issue #3499, opened 2023-10-02
IATP: implement self-contained STS


paullatzelsperger:
## WHAT

EDC will implement a standalone runtime exposing the STS API (OAuth2), that can create tokens [as per IATP spec](https://github.com/eclipse-tractusx/identity-trust/blob/8215853dddad222b1cdb81d2d16d202265ea698a/specifications/M1/identity.protocol.base.md#4-self-issued-id-tokens).

The functionality of creating tokens should be the same as in the embedded variant (#3500), but this time it's made available through a REST API (#3498).

In this issue we need:
- a api + controller
- a launcher

PR #3529, opened 2023-10-10
feat: embedded STS

### LINKED ISSUE #3500: IATP: implement embedded STS
## WHAT

EDC will implement a module, that can create tokens [as per IATP spec](https://github.com/eclipse-tractusx/identity-trust/blob/8215853dddad222b1cdb81d2d16d202265ea698a/specifications/M1/identity.protocol.base.md#4-self-issued-id-tokens).

Requirements for the EDC implementation of the STS:

- STS has its own keypair
- create a self-issued ID JWT token, that can contain an `access_token` within (see below)
- `sub` claim contains the DID of the intended recipient
- `iss` contains the DID of the creator (= "myself")
- `iss` == `sub` 

### Adding the `access_token` claim

If the token request is made specifying a `bearer_access_scope` parameter (#3498), then the STS shall add an `access_token` claim to the SI token. 

That `access_token` is signed with the STS's private key, and encodes the requested scopes. To that end, the `access_token` could just be another JWT.

If the `bearer_access_scope` is *not* specified by the caller, then the resulting SI token will **not** contain an `access_token` claim. This is useful when generating the token-in-token (see #3495), where the access token is already there.

## IMPLEMENTATION PROPOSAL

Possible implementation for the STS interface:
```java
/**
 * if bearerAccessScope != null -> creates an access_token claim, which is another JWT containing the scope as claims.
 * if bearerAccessScope == null -> creates a normal JWT using all the claims in the map
 */
public interface SecureTokenService {
   Result<TokenRepresentation> createToken(Map<String, String> claims, @Nullable String bearerAccessScope);
}
```

The `IdentityAndTrustService` will by default call out to a (remote) STS (#3499). However, if the `edc.oauth.token.url` config parameter isn't present, it can short-circuit the remote invocation and call out to an embedded STS as fallback.

The STS code would be the same, with one little difference:
- remote invocation: `IdentityAndTrustService -> RemoteStsClient -> STS REST API -> STS`
- embedded invocation: `IdentityAndTrustService -> EmbeddedStsClient -> STS`

wolf4ood:
## What this PR changes/adds

Implements STS (Secure Token Service) embedded

The`EmbeddesSecureTokenService` creates the Self-Issued ID token from the input claims by using a `TokenGenerationService`. If the param `bearerAccessScope` is provide it will also automatically attach the `access_token` claim which is another `Jwt` token containing the provided scopes (Credential Service Access).

The Self-Issued ID format is compliant with the spec [here](https://github.com/eclipse-tractusx/identity-trust/blob/main/specifications/M1/identity.protocol.base.md#41-self-issued-id-token-contents).

For the `access_token` the format is still a `JWT` similar to the Self-Issued ID one but the:

- `scope` claim that contains the `bearerAccessScope` for the `Credential Service` access.
- `iss` claim is the same as the Self-Issued
- `aud` is the `iss` of the Self-Issued token
- `sub` is the `aud` of the Self-Issued token

The the same private key it's used for signing the Self-Issued ID token and the Access token.

## Why it does that

IATP adoption

## Further notes

The `ConsumerPullKeyPairFactory` as been refactored  by extracting the `KeyPairFactory` spi and moving  `KeyPairFactoryImpl` into `connector-core`, so it can be reused by the `iatp` and `TransferDataPlaneCoreExtension`.

## Linked Issue(s)

Closes #3500


extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IatpDefaultServicesExtension.java
```diff
@@ -14,27 +14,68 @@
...
+    private KeyPairFactory keyPairFactory;
+
+    @Inject
+    private Clock clock;
 
     @Provider(isDefault = true)
     public SecureTokenService createDefaultTokenService(ServiceExtensionContext context) {
         context.getMonitor().info("Using the Embedded STS client, as no other implementation was provided.");
+        var keyPair = keyPairFromConfig(context);
+        var tokenExpiration = context.getSetting(STS_TOKEN_EXPIRATION, DEFAULT_STS_TOKEN_EXPIRATION);
```
bscholtes1A, line 58: Not sure it is correct to have injected fields within a `@Provider(isDefault = true)` method. Might be ok tough as the injected services are provided by Core extension... @paullatzelsperger WDYT?

extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IatpDefaultServicesExtension.java
```diff
@@ -14,27 +14,68 @@
...
         if (context.getSetting(OAUTH_TOKENURL_PROPERTY, null) != null) {
             context.getMonitor().warning("The property '%s' was configured, but no remote SecureTokenService was found on the classpath. ".formatted(OAUTH_TOKENURL_PROPERTY) +
                     "This could be an indicator of a configuration problem.");
         }
 
-        return new EmbeddedSecureTokenService();
+        return new EmbeddedSecureTokenService(new TokenGenerationServiceImpl(keyPair.getPrivate()), clock, TimeUnit.MINUTES.toSeconds(tokenExpiration));
+    }
+
+    private KeyPair keyPairFromConfig(ServiceExtensionContext context) {
```
bscholtes1A, line 69: Maybe the logic of this block should be in `KeyPairFactoryImpl` and thus have the `KeyPairFactory` exposing only a single method, WDYT ?

```java
@ExtensionPoint
@FunctionalInterface
public interface KeyPairFactory {
    Result<KeyPair> generate(ServiceExtensionContext context);
}
```

extensions/common/iam/identity-trust/identity-trust-sts-embedded/src/main/java/org/eclipse/edc/iam/identitytrust/sts/embedded/EmbeddedSecureTokenService.java
```diff
@@ -0,0 +1,82 @@
...
+                .map(deco -> null);
+    }
+
+    private Result<TokenRepresentation> createAccessToken(Map<String, Object> claims, String bearerAccessScope) {
+        return tokenGenerationService.generate(new SelfIssuedTokenDecorator(accessTokenClaims(claims), clock, validity), new BearerAccessScopeTokenDecorator(bearerAccessScope));
+    }
+
+    private Map<String, Object> accessTokenClaims(Map<String, Object> claims) {
+        return claims;
+    }
```
bscholtes1A, line 81: Why this method as it only returns the input?

extensions/common/iam/identity-trust/identity-trust-sts-embedded/src/main/java/org/eclipse/edc/iam/identitytrust/sts/embedded/EmbeddedSecureTokenService.java
```diff
@@ -0,0 +1,82 @@
...
+    public Result<TokenRepresentation> createToken(Map<String, Object> claims, @Nullable String bearerAccessScope) {
+        Collection<JwtDecorator> decorators = new ArrayList<>();
+        decorators.add(new SelfIssuedTokenDecorator(claims, clock, validity));
+
+        var result = withAccessToken(claims, bearerAccessScope, decorators::add);
+
+        if (result.failed()) {
+            return result.mapTo();
+        }
+        return tokenGenerationService.generate(decorators.toArray(JwtDecorator[]::new));
```
bscholtes1A, line 59: can be done in a more compact way no?
```java
return withAccessToken(claims, bearerAccessScope, decorators::add)
   .compose(v -> tokenGenerationService.generate(decorators.toArray(JwtDecorator[]::new)));
```

extensions/common/iam/identity-trust/identity-trust-sts-embedded/src/main/java/org/eclipse/edc/iam/identitytrust/sts/embedded/EmbeddedSecureTokenService.java
wolf4ood: Because the PR still in draft, i needed feedback from the CI :)

extensions/common/iam/identity-trust/identity-trust-sts-embedded/src/main/java/org/eclipse/edc/iam/identitytrust/sts/embedded/EmbeddedSecureTokenService.java
wolf4ood: Probably, still need to rework the PR before it's ready for review :)

extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IatpDefaultServicesExtension.java
wolf4ood: The `KeyPairFactory` should be able to generate a key pair from aliases right? not sure passing the `ServiceExtensionContext` solve this. In this way we can reuse the `KeyPairFactory` here and in other place like `TransferDataPlaneCoreExtension`

extensions/common/iam/identity-trust/identity-trust-sts-embedded/src/main/java/org/eclipse/edc/iam/identitytrust/sts/embedded/EmbeddedSecureTokenService.java
wolf4ood: done

extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IatpDefaultServicesExtension.java
paullatzelsperger: i think that is correct, as there will be another implementation later (#3499), where  instead of calling a java class, the call goes out over HTTP/REST to a standalone STS. The embedded one would be the "fallback" in that case, or when an `edc.oauth.token.url` is not specified.

extensions/common/iam/identity-trust/identity-trust-sts-embedded/src/main/java/org/eclipse/edc/iam/identitytrust/sts/embedded/EmbeddedSecureTokenService.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+import static java.lang.String.format;
+import static java.util.Optional.ofNullable;
+import static org.eclipse.edc.jwt.spi.JwtRegisteredClaimNames.AUDIENCE;
```
paullatzelsperger, line 33: minor nit: those constants should probably reference `com.nimbus.jwt.JWTClaimNames`. not related here though.

extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IatpDefaultServicesExtension.java
bscholtes1A: My comment was more about dependency injection in order to be sure the injected fields are properly initialized before the default provider method is invoked :)

extensions/common/iam/identity-trust/identity-trust-sts-embedded/src/main/java/org/eclipse/edc/iam/identitytrust/sts/embedded/SelfIssuedTokenDecorator.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import static java.util.Collections.emptyMap;
+import static org.eclipse.edc.jwt.spi.JwtRegisteredClaimNames.EXPIRATION_TIME;
+import static org.eclipse.edc.jwt.spi.JwtRegisteredClaimNames.ISSUED_AT;
+import static org.eclipse.edc.jwt.spi.JwtRegisteredClaimNames.JWT_ID;
+
+public class SelfIssuedTokenDecorator implements JwtDecorator {
```
paullatzelsperger, line 30: some doc pls

paullatzelsperger, line 30: could be package-private

paullatzelsperger (approved):
LGTM, mostly minor nits

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/security/KeyPairFactoryImpl.java
```diff
@@ -64,7 +81,7 @@ public KeyPair defaultKeyPair() {
     @NotNull
     private Result<PublicKey> publicKey(String alias) {
         return Optional.ofNullable(vault.resolveSecret(alias))
```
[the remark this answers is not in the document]
wolf4ood: I would do this clean up in a separated PR in case. Here i've just moved and renamed the impl

bscholtes1A (approved):

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/security/KeyPairFactoryImpl.java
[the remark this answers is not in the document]
paullatzelsperger: yep, alright.

Issue #3490, opened 2023-10-02
Adoption Request: Extend Contract Definition entity with Private Properties


mohannad-ezzo:
# Adoption Request

Adding private properties to contract definition entity.

## General information

Extending contract definition entity with the map privateProperties. Additionally, adjust the affected transformers and contract definition API managements accordingly. The functionality would support in-memory and postgres as a store.

## Adoption level

Next, please tell us what level of adoption you intend to provide. _(pick only one)_

- [ ] Reference a feature as "friend"
- [X ] Incorporate a feature as core EDC component

## Adoption in EDC core

_If you chose to add your feature as a core EDC component, please answer the following questions._

### Why should this contribution be adopted?
This feature is similar to what is currently available for Asset entity. Therefore, the required contribution for the contract definition to enable the private properties looks very similar.

### Could it be achieved with existing functionality? If not, why?
No. This is a completely new field withing the contract definition entity. 

### Are there multiple use cases or applications who will benefit from the contribution?
Yes, same as Asset entity.

### Can it be achieved without introducing third-party dependencies? If not, which ones?
Yes

### Would this feature limit platform independence in any way? If so, how and why?
No

### Is it going to be a self-contained feature, or would it cut across the entire code base?
a self-contained feature.

Reference items:
https://github.com/eclipse-edc/Connector/discussions/2592#discussioncomment-7162109

mohannad-ezzo:
Is there any update for this issue?
Changes are introduced in the branch https://github.com/mohannad-ezzo/Connector/tree/feat-contract-def-private-prop and ready to create a pull request.

ndr-brt:
Ok, let's open the PR, we will review that. thanks.

Issue #3424, opened 2023-09-07
QuerySpecValidator fails to catch errors when filterExpression has more than one objects


DimitrisGk-iel:
# Bug Report

## Describe the Bug
When we issue a EDC_QUERY_SPEC_TYPE query that has a filterExpresssion with more than one objects, QuerySpecValidator fails to catch it.
![image](https://github.com/eclipse-edc/Connector/assets/30075838/6116e613-e202-498d-a708-ba5129f309a7)
### Expected Behavior
A message that the operandLeft is missing e.g.
[{"message":"mandatory value '[https://w3id.org/edc/v0.0.1/ns/filterExpression/https://w3id.org/edc/v0.0.1/ns/operandLeft'](https://w3id.org/edc/v0.0.1/ns/filterExpression/https://w3id.org/edc/v0.0.1/ns/operandLeft%27) is missing or it is blank","type":"ValidationFailure","path":"[https://w3id.org/edc/v0.0.1/ns/filterExpression/https://w3id.org/edc/v0.0.1/ns/operandLeft","invalidValue":null}]](https://w3id.org/edc/v0.0.1/ns/filterExpression/https://w3id.org/edc/v0.0.1/ns/operandLeft%22,%22invalidValue%22:null%7D])
### Observed Behavior

SEVERE 2023-09-06T00:01:58.1307352 JerseyExtension: Unexpected exception caught
java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:208)
	at org.eclipse.edc.spi.query.Criterion$Builder.build(Criterion.java:132)
	at org.eclipse.edc.core.transform.transformer.to.JsonObjectToCriterionTransformer.transform(JsonObjectToCriterionTransformer.java:51)
	at org.eclipse.edc.core.transform.transformer.to.JsonObjectToCriterionTransformer.transform(JsonObjectToCriterionTransformer.java:29)
	at org.eclipse.edc.core.transform.Transfo

## Steps to Reproduce

1. Issue the above query to the /v3/assets/request endpoint
2. See error

## Context Information
- Used version EDC v0.2.1

## Detailed Description

_If applicable, add screenshots and logs to help explain your problem._

## Possible Implementation

CriterionValidator should be run on all filterExpression objects and not just the first one

ndr-brt:
At a glance, the issue should be caused by the wrong definition for the `filterExpression` validator, it should be `verifyArrayItem` instead of `verifyObject`.

[...]

PR #3534, opened 2023-10-12
feat: add private properties to contract definition entity

mohannad-ezzo:
## What this PR changes/adds

Add private properties to contract definition entity. The implementation supports in-memory and postgres as a store.

## Why it does that

The private properties are required to provide the owner with the possibility to add extra properties which should not be exposed outside. This requirement was initiated when we want to add administrative properties like (createdBy, createdAt, chnagedBy, chnagedAt). We assume that the private properties can be used to store such a requirement after registering a listener on the contract definition. 
More information is in this discussion:
https://github.com/eclipse-edc/Connector/discussions/2592#discussioncomment-7162109 

## Further notes

## Linked Issue(s)

Closes https://github.com/eclipse-edc/Connector/issues/3490


extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
```diff
@@ -178,9 +208,69 @@ private long mapCount(ResultSet resultSet) throws SQLException {
...
+    }
+
     private ContractDefinition findById(Connection connection, String id) {
-        var sql = statements.getFindByTemplate();
-        return queryExecutor.single(connection, false, this::mapResultSet, sql, id);
+        return transactionContext.execute(() -> {
+            var query = QuerySpec.Builder.newInstance().filter(List.of(new Criterion("id", "=", id))).build();
+            var queryStatement = statements.createQuery(query);
+
+            var contractDefinitionStream = queryExecutor.query(connection, false,
```
paullatzelsperger, line 232: could use `queryExecutor.single(...)` since findById can return at most one entity

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
```diff
@@ -178,9 +208,69 @@ private long mapCount(ResultSet resultSet) throws SQLException {
...
+                            this::mapResultSet, queryStatement.getQueryAsString(), queryStatement.getParameters());
+            var allPropertiesStream = queryExecutor.query(connection, false,
+                            this::mapProperties, statements.getFindPropertyByIdTemplate(), id);
+
+            Map<Boolean, List<SqlPropertyWrapper>> groupedProperties =
```
paullatzelsperger, line 237: `var`

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
```diff
@@ -178,9 +208,69 @@ private long mapCount(ResultSet resultSet) throws SQLException {
...
+                        allPropertiesStream.collect(partitioningBy(SqlPropertyWrapper::isPrivate));
+            var contractDefinitionPrivateProperties = groupedProperties.get(true).stream().collect(
```
paullatzelsperger, line 239: this grouping is not needed, because properties can only be private. don't just copy-paste code!

spi/control-plane/contract-spi/src/test/java/org/eclipse/edc/connector/contract/spi/types/offer/ContractDefinitionTest.java
```diff
@@ -29,11 +32,14 @@ class ContractDefinitionTest {
     @Test
     void verifySerializeDeserialize() throws JsonProcessingException {
         ObjectMapper mapper = new TypeManager().getMapper();
+        Map<String, Object> privateProperties = new HashMap<>();
```
paullatzelsperger, line 35: please use `var`

paullatzelsperger (changes requested):

paullatzelsperger:
@mohannad-ezzo two more things: in a PR description, it's fine to reference a discussion, but giving a brief reason is still needed, so that reviewers don't have to read long threads of discussions.
Also in the PR description, pls use [closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) when linking the issue, so it gets auto-closed when your PR is merged

extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/ContractDefinitionApiExtension.java
```diff
@@ -61,7 +69,9 @@ public String name() {
     @Override
     public void initialize(ServiceExtensionContext context) {
         var jsonFactory = Json.createBuilderFactory(Map.of());
-        transformerRegistry.register(new JsonObjectFromContractDefinitionTransformer(jsonFactory));
+        var mapper = typeManager.getMapper(JSON_LD);
+        mapper.registerSubtypes(AtomicConstraint.class, LiteralExpression.class);
```
ndr-brt, line 73: Is this registration needed? in this context (`ContractDefinition`) there shouldn't be any relationship with the `policy` model

extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/transform/JsonObjectFromContractDefinitionTransformer.java
```diff
@@ -44,12 +48,19 @@ public JsonObjectFromContractDefinitionTransformer(JsonBuilderFactory jsonFactor
...
+        var builder = jsonFactory.createObjectBuilder()
                 .add(ID, contractDefinition.getId())
                 .add(TYPE, CONTRACT_DEFINITION_TYPE)
                 .add(CONTRACT_DEFINITION_ACCESSPOLICY_ID, contractDefinition.getAccessPolicyId())
                 .add(CONTRACT_DEFINITION_CONTRACTPOLICY_ID, contractDefinition.getContractPolicyId())
-                .add(CONTRACT_DEFINITION_ASSETS_SELECTOR, criteria)
-                .build();
+                .add(CONTRACT_DEFINITION_ASSETS_SELECTOR, criteria);
+
+        if (contractDefinition.getPrivateProperties() != null && !contractDefinition.getPrivateProperties().isEmpty()) {
```
ndr-brt, line 58: private properties cannot be null, this check can be simplified

extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/transform/JsonObjectToContractDefinitionTransformer.java
```diff
@@ -36,19 +37,23 @@ public JsonObjectToContractDefinitionTransformer() {
...
-                default -> { }
-            }
-        });
-
+        visitProperties(object, (s, jsonValue) -> transformProperties(s, jsonValue, builder, context));
         return builderResult(builder::build, context);
     }
 
+    private void transformProperties(String key, JsonValue jsonValue, ContractDefinition.Builder builder, TransformerContext context) {
+        if (CONTRACT_DEFINITION_ACCESSPOLICY_ID.equals(key) ) {
```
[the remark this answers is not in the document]
ndr-brt: I think it could be motivated by the additional condition `jsonValue instanceof JsonArray` but that's useless: in an expanded jsonld document every property is either an array or null.

extensions/control-plane/api/management-api/contract-definition-api/src/test/java/org/eclipse/edc/connector/api/management/contractdefinition/transform/JsonObjectFromContractDefinitionTransformerTest.java
```diff
@@ -72,4 +71,24 @@ void transform() {
         verify(context, never()).reportProblem(anyString());
     }
 
+
+    @Test
+    void transform_withPrivateProperties_simpleTypes() {
+        var criterionJson = jsonFactory.createObjectBuilder().build();
+        when(context.transform(isA(Criterion.class), eq(JsonObject.class))).thenReturn(criterionJson);
+        var criterion = criterion("left", "=", "right");
+        var contractDefinition =ContractDefinition.Builder.newInstance()
```
ndr-brt, line 80:
```suggestion
        var contractDefinition = ContractDefinition.Builder.newInstance()
```

extensions/control-plane/api/management-api/contract-definition-api/src/test/java/org/eclipse/edc/connector/api/management/contractdefinition/transform/JsonObjectToContractDefinitionTransformerTest.java
```diff
@@ -84,4 +89,36 @@ void transform_whenNoAssetSelectorItShouldBeAnEmptyList() {
...
+                .add(TYPE, CONTRACT_DEFINITION_TYPE)
+                .add(CONTRACT_DEFINITION_ACCESSPOLICY_ID, "accessPolicyId")
+                .add(CONTRACT_DEFINITION_CONTRACTPOLICY_ID, "contractPolicyId")
+                .add(CONTRACT_DEFINITION_PRIVATE_PROPERTIES, createArrayBuilder().add(createObjectBuilder().add("test-prop", "test-val").build()).build())
+                .build();
+
+        jsonObj = expand(jsonObj);
+        var contractDefinition = transformer.transform(jsonObj, context);
+
+        //AbstractResultAssert.assertThat(contractDefinition).withFailMessage(contractDefinition::getFailureDetail).isSucceeded();
```
ndr-brt, line 108: please remove commented code

extensions/control-plane/store/sql/contract-definition-store-sql/docs/schema.sql
```diff
@@ -23,3 +24,30 @@ CREATE TABLE IF NOT EXISTS edc_contract_definitions
...
+
+
+-- table: edc_contract_definition_property
+CREATE TABLE IF NOT EXISTS edc_contract_definition_property
+(
+    contract_definition_id_fk       VARCHAR(255) NOT NULL,
+    property_name  VARCHAR(255) NOT NULL,
+    property_value TEXT         NOT NULL,
+    property_type  VARCHAR(255) NOT NULL,
+    property_is_private BOOLEAN,
```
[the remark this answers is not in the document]
ndr-brt: I think this was copied by the sql asset index implementation, but to be fair I'd prefer to follow the transfer process one, where the private properties are serialized into a json field.

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
```diff
@@ -153,9 +163,25 @@ private void insertInternal(Connection connection, ContractDefinition definition
                     definition.getContractPolicyId(),
                     toJson(definition.getAssetsSelector()),
                     definition.getCreatedAt());
+
+            insertProperties(definition, connection);
         });
     }
 
+    private void insertProperties(ContractDefinition definition, Connection connection) {
+        if (Objects.nonNull(definition.getPrivateProperties())) {
```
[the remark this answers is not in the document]
ndr-brt: as I wrote before, privateProperties can never be null (because they are initialized to an empty map and there's a check on the builder method, the getter is also marked with `@NotNull`)

spi/control-plane/contract-spi/src/test/java/org/eclipse/edc/connector/contract/spi/types/offer/ContractDefinitionTest.java
ndr-brt: but also `Map.of("key", "value")`.

ndr-brt (changes requested):
generally, I'd follow the `SqlTransferProcessStore` approach, putting the properties into a JSON column into the same entity table, it will save a lot of code lines and unneeded overcomplication.

extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/transform/JsonObjectToContractDefinitionTransformer.java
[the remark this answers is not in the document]
mohannad-ezzo: This is addresses

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
```diff
@@ -135,6 +141,10 @@ public StoreResult<ContractDefinition> deleteById(String id) {
 
     }
 
+    private String mapContractDefinitionIds(ResultSet resultSet) throws SQLException {
```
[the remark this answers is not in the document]
mohannad-ezzo: This is addresses

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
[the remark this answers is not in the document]
mohannad-ezzo: This is addresses

extensions/control-plane/api/management-api/contract-definition-api/src/main/java/org/eclipse/edc/connector/api/management/contractdefinition/ContractDefinitionApiExtension.java
mohannad-ezzo: Addressed

extensions/control-plane/api/management-api/contract-definition-api/src/test/java/org/eclipse/edc/connector/api/management/contractdefinition/transform/JsonObjectFromContractDefinitionTransformerTest.java
mohannad-ezzo: addresses

extensions/control-plane/api/management-api/contract-definition-api/src/test/java/org/eclipse/edc/connector/api/management/contractdefinition/transform/JsonObjectToContractDefinitionTransformerTest.java
mohannad-ezzo: Addressed

extensions/control-plane/store/sql/contract-definition-store-sql/docs/schema.sql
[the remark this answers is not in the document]
mohannad-ezzo: I followed the same pattern which is used for Asset. Therefore, I followed this approach. In addition, I intended to add `property_is_private` as a preparation for adding non-private properties for the contract definition. To be honest, our usecase for contract definition is just to have private properties, but not for non-private properties. So, as requested, the property `property_is_private` is removed and the corresponding code is adjusted.

spi/control-plane/contract-spi/src/testFixtures/java/org/eclipse/edc/connector/contract/spi/testfixtures/offer/store/ContractDefinitionStoreTestBase.java
```diff
@@ -159,6 +161,65 @@ void exists() {
                 assertThat(definition.getContractPolicyId()).isEqualTo(definition2.getContractPolicyId());
             });
         }
+
+        @Test
+        @DisplayName("Update contract definition that exists, adding a property")
+        void exists_addsProperty() {
```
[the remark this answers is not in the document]
mohannad-ezzo: Additional usecases are added.

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
mohannad-ezzo: Done

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
mohannad-ezzo: Done

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
mohannad-ezzo: Grouping is removed.

spi/control-plane/contract-spi/src/test/java/org/eclipse/edc/connector/contract/spi/types/offer/ContractDefinitionTest.java
mohannad-ezzo: Done

paullatzelsperger:
> generally, I'd follow the `SqlTransferProcessStore` approach, putting the properties into a JSON column into the same entity table, it will save a lot of code lines and unneeded overcomplication.

I agree, that is an even easier way way to do it. @mohannad-ezzo can you adapt your PR accordingly? 
@ndr-brt in a future issue, maybe we should also change the SQL-persistence of Assets so that is follows the JSON approach.

mohannad-ezzo:
> > generally, I'd follow the `SqlTransferProcessStore` approach, putting the properties into a JSON column into the same entity table, it will save a lot of code lines and unneeded overcomplication.
> 
> I agree, that is an even easier way way to do it. @mohannad-ezzo can you adapt your PR accordingly? @ndr-brt in a future issue, maybe we should also change the SQL-persistence of Assets so that is follows the JSON approach, wdyt?

@paullatzelsperger @ndr-brt By following the JSON approach, we may loose the flexibility to support search within the private properties across all contract definitions. For sure, for a small set of contract definitions, JSON approach would still work. However, whenever the set becomes bigger and bigger, cross search would be difficult. I still see that the current approach may fit better our current and future requirement. What do you think ?

paullatzelsperger:
> > > generally, I'd follow the `SqlTransferProcessStore` approach, putting the properties into a JSON column into the same entity table, it will save a lot of code lines and unneeded overcomplication.
> > 
> > 
> > I agree, that is an even easier way way to do it. @mohannad-ezzo can you adapt your PR accordingly? @ndr-brt in a future issue, maybe we should also change the SQL-persistence of Assets so that is follows the JSON approach, wdyt?
> 
> @paullatzelsperger @ndr-brt By following the JSON approach, we may loose the flexibility to support search within the private properties across all contract definitions. For sure, for a small set of contract definitions, JSON approach would still work. However, whenever the set becomes bigger and bigger, cross search would be difficult. I still see that the current approach may fit better our current and future requirement. What do you think ?

Postgres supports [JSON operators](https://www.postgresql.org/docs/9.5/functions-json.html), and EDC has a way of converting a `QuerySpec` into such an operator.

mohannad-ezzo:
> > @paullatzelsperger @ndr-brt By following the JSON approach, we may loose the flexibility to support search within the private properties across all contract definitions. For sure, for a small set of contract definitions, JSON approach would still work. However, whenever the set becomes bigger and bigger, cross search would be difficult. I still see that the current approach may fit better our current and future requirement. What do you think ?
> 
> @mohannad-ezzo Postgres supports [JSON operators](https://www.postgresql.org/docs/9.5/functions-json.html), and EDC has a way of converting a `QuerySpec` into such an operator.

@paullatzelsperger @ndr-brt ok, I am going to follow the JSON approach of storing the private properties (instead of a having a separate table) and provide you with an update asap.

ndr-brt:
> > > generally, I'd follow the `SqlTransferProcessStore` approach, putting the properties into a JSON column into the same entity table, it will save a lot of code lines and unneeded overcomplication.
> > 
> > 
> > I agree, that is an even easier way way to do it. @mohannad-ezzo can you adapt your PR accordingly? @ndr-brt in a future issue, maybe we should also change the SQL-persistence of Assets so that is follows the JSON approach, wdyt?
> 
> @paullatzelsperger @ndr-brt By following the JSON approach, we may loose the flexibility to support search within the private properties across all contract definitions. For sure, for a small set of contract definitions, JSON approach would still work. However, whenever the set becomes bigger and bigger, cross search would be difficult. I still see that the current approach may fit better our current and future requirement. What do you think ?

I agree with paul, plus, using the "separate table" approach will lead to issues in trying to query nested properties (see #3424 )

mohannad-ezzo:
As discussed I adapted the PR so that JSON is used to store private properties of contract definition instead of storing them in a separate tables.

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
```diff
@@ -62,7 +67,8 @@ public SqlContractDefinitionStore(DataSourceRegistry dataSourceRegistry, String
 
             try {
                 var queryStmt = statements.createQuery(spec);
-                return queryExecutor.query(getConnection(), true, this::mapResultSet, queryStmt.getQueryAsString(), queryStmt.getParameters());
+                return queryExecutor.query(getConnection(), true, this::mapContractDefinitionId, queryStmt.getQueryAsString(), queryStmt.getParameters())
+                        .map(this::findById);
```
ndr-brt, line 71: Doing two queries to get the contract definition seems overkill, the old behavior was ok

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
```diff
@@ -179,8 +192,24 @@ private long mapCount(ResultSet resultSet) throws SQLException {
...
+
+            return ContractDefinition.Builder.newInstance()
+                        .id(contractDefinition.getId())
+                        .createdAt(contractDefinition.getCreatedAt())
+                        .accessPolicyId(contractDefinition.getAccessPolicyId())
+                        .contractPolicyId(contractDefinition.getContractPolicyId())
+                        .assetsSelector(contractDefinition.getAssetsSelector())
+                        .privateProperties(contractDefinition.getPrivateProperties())
+                        .build();
+        });
```
ndr-brt, line 213: the old behavior was ok

ndr-brt (commented):
Just avoid those two not necessary changes to the sql store and it's ready to be merged

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
mohannad-ezzo: This is addresses

extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/contractdefinition/SqlContractDefinitionStore.java
mohannad-ezzo: This is addressed

mohannad-ezzo:
> Just avoid those two not necessary changes to the sql store and it's ready to be merged

Thank you for mentioning this. This actually was inherited from the old logic when I started to have another table for private properties. Old logic is back now :)

ndr-brt (approved):

ndr-brt:
@mohannad-ezzo you need to fix the broken tests and the checkstyle as well

mohannad-ezzo:
> @mohannad-ezzo you need to fix the broken tests and the checkstyle as well

I just fixed the checksytle issues. Regarding the broken tests, they ran fine this time without any change from me. 
As of now, ALL checks are green and have passed :)

paullatzelsperger (approved):
LGTM!

PR #3535, opened 2023-10-12
feat: Add warning log to Hashicorp Vault resolveSecret function

### LINKED ISSUE #3462: The HashiCorp Vault Secret resolution only returns a Secret or Null
As a developer, it's easy to get stuck here because if the access token to the Vault lacks permissions or is incorrect, it either returns a Secret or nothing.
https://github.com/eclipse-edc/Connector/blob/258a31190a9c837606c2e327fe4dd5e54f3f1921/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java#L42

In the case of a 403 error, it would also be helpful to pass along this message so that you can handle the respective errors afterward.

What are your thoughts on this?

awellnitz-materna:
## What this PR changes/adds

Add warning log to Hashicorp Vault `resolveSecret` function.

## Why it does that

If the secret cannot be resolved, it is currently difficult to find out why. 
The log should help the developer to find the bug faster.

## Linked Issue(s)

Closes #3462

### DIFF
--- a/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
+++ b/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
@@ -9,6 +9,7 @@
  *
  *  Contributors:
  *       Mercedes-Benz Tech Innovation GmbH - Initial API and Implementation
+ *       Materna Information & Communications SE - Refactoring
  *
  */
 
@@ -38,6 +39,10 @@ public class HashicorpVault implements Vault {
     @Override
     public @Nullable String resolveSecret(String key) {
         var result = hashicorpVaultClient.getSecretValue(key);
+
+        if (result.failed()) {
+            monitor.warning("Failed to resolve secret '%s': %s".formatted(key, result.getFailureMessages()));
+        }
 
         return result.succeeded() ? result.getContent() : null;
     }
--- a/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
+++ b/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
@@ -41,7 +41,7 @@ public class HashicorpVault implements Vault {
         var result = hashicorpVaultClient.getSecretValue(key);
 
         if (result.failed()) {
-            monitor.warning("Failed to resolve secret '%s': %s".formatted(key, result.getFailureMessages()));
+            monitor.debug("Failed to resolve secret '%s': %s".formatted(key, result.getFailureMessages()));
             return null;
         } 
         
--- a/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
+++ b/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
@@ -42,9 +42,10 @@ public class HashicorpVault implements Vault {
 
         if (result.failed()) {
             monitor.warning("Failed to resolve secret '%s': %s".formatted(key, result.getFailureMessages()));
-        }
-
-        return result.succeeded() ? result.getContent() : null;
+            return null;
+        } 
+        
+        return result.getContent();
     }
 
     @Override
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -14,28 +14,33 @@ maven/mavencentral/com.fasterxml.jackson.core/jackson-annotations/2.14.0, Apache
 maven/mavencentral/com.fasterxml.jackson.core/jackson-annotations/2.14.1, Apache-2.0, approved, #5303
 maven/mavencentral/com.fasterxml.jackson.core/jackson-annotations/2.15.1, Apache-2.0, approved, #7947
 maven/mavencentral/com.fasterxml.jackson.core/jackson-annotations/2.15.2, Apache-2.0, approved, #7947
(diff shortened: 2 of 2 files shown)


extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
```diff
@@ -39,6 +40,10 @@ public HashicorpVault(@NotNull HashicorpVaultClient hashicorpVaultClient, @NotNu
     public @Nullable String resolveSecret(String key) {
         var result = hashicorpVaultClient.getSecretValue(key);
 
+        if (result.failed()) {
+            monitor.warning("Failed to resolve secret: " + result.getFailureMessages());
```
jimmarino, line 44: Can you include the key name, e.g. warning(format(Failed to resolve secret '%s':%s", key, result.getFailureMessage()));

jimmarino (changes requested):

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
awellnitz-materna: @jimmarino You are right, I added the key to the message.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
```diff
@@ -39,6 +40,10 @@ public HashicorpVault(@NotNull HashicorpVaultClient hashicorpVaultClient, @NotNu
     public @Nullable String resolveSecret(String key) {
         var result = hashicorpVaultClient.getSecretValue(key);
 
+        if (result.failed()) {
```
jimmarino, line 43: Sorry, one other thing: there is no need to check for failure/success twice. Please combine the return statement with the failure check above.

paullatzelsperger:
took myself off as a reviewer, because the changeset isn't so huge :)

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
awellnitz-materna: @jimmarino Now i check if the result failed and then return null + log or the result.

jimmarino (approved):
Looks good!

bscholtes1A (approved):

jimmarino:
@awellnitz-materna please fix the failing check and I will merge the PR

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
```diff
@@ -39,7 +40,12 @@ public HashicorpVault(@NotNull HashicorpVaultClient hashicorpVaultClient, @NotNu
     public @Nullable String resolveSecret(String key) {
         var result = hashicorpVaultClient.getSecretValue(key);
 
-        return result.succeeded() ? result.getContent() : null;
+        if (result.failed()) {
+            monitor.warning("Failed to resolve secret '%s': %s".formatted(key, result.getFailureMessages()));
```
ndr-brt, line 44: I know that I suggested to have a `warning` log, but thinking about that we would result having the console log bloated with warnings that cannot be resolved because we're often using the pattern of "check into the vault, if it's not present there then look elsewhere". E.g.:
https://github.com/eclipse-edc/Connector/blob/12f1275e791252df14cf02e9fa2c5b884167942e/core/control-plane/transfer-core/src/main/java/org/eclipse/edc/connector/transfer/process/TransferProcessManagerImpl.java#L278-L281

maybe a debug would fit better

awellnitz-materna: @ndr-brt I would already stay here with warning. Because as admin you don't have the EDC running in debug mode. And here the message can help, if e.g. the access token is no longer valid or if the rights have changed.

jimmarino: @ndr-brt yes, that's a good point. Debug would be my preference.

awellnitz-materna:
@jimmarino I have not customized any dependency. Can you tell me how to regenerate the dependency file?

wolf4ood:
Hi @awellnitz-materna 

for the dependencies you can check the failing run of the license check. It suggests which the updated content of the `DEPENDENCIES` file should be.

Check

https://github.com/eclipse-edc/Connector/actions/runs/6506078793/job/17671039514?pr=3535#step:6:81

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVault.java
ndr-brt: > @ndr-brt I would already stay here with warning. Because as admin you don't have the EDC running in debug mode. And here the message can help, if e.g. the access token is no longer valid or if the rights have changed.

logs should not be used for such activity, the health check service is a better choice, and if I'm correct the hashicorp extension declares some provider on it, so it should be enough.

awellnitz-materna: @ndr-brt @jimmarino You both say debug so I would change that to debug, yes?

PR #3540, opened 2023-10-13
feat: extensible trusted-issuer-registry

### LINKED ISSUE #3531: IATP: create extensible way to contribute allowed issuers

paullatzelsperger:
## What this PR changes/adds

This PR adds an extensible registry of trusted issuers plus a default in-mem implementation. In order to do that, the following additions/changes were necessary:
- introduce new `Issuer` class
- introduce a new `JsonObjectToIssuerTransformer`
- change `VerifiableCredential#issuer` from `Object` to `Issuer`

## Why it does that

We need a possibility to contribute multiple trusted issuers in a dataspace, e.g. multiple govt. agencies and private companies.

## Further notes

- changed the wording from "allowed" to "trusted"
- other implementations could download that list of trusted issuers from some remote location

## Linked Issue(s)

Closes #3531

### DIFF
--- a/extensions/common/iam/identity-trust/identity-trust-transform/src/main/java/org/eclipse/edc/iam/identitytrust/transform/to/JsonObjectToIssuerTransformer.java
+++ b/extensions/common/iam/identity-trust/identity-trust-transform/src/main/java/org/eclipse/edc/iam/identitytrust/transform/to/JsonObjectToIssuerTransformer.java
@@ -16,6 +16,7 @@ package org.eclipse.edc.iam.identitytrust.transform.to;
 
 import jakarta.json.JsonObject;
 import org.eclipse.edc.identitytrust.model.Issuer;
+import org.eclipse.edc.jsonld.spi.JsonLdKeywords;
 import org.eclipse.edc.jsonld.spi.transformer.AbstractJsonLdTransformer;
 import org.eclipse.edc.transform.spi.TransformerContext;
 import org.jetbrains.annotations.NotNull;
@@ -33,8 +34,10 @@ public class JsonObjectToIssuerTransformer extends AbstractJsonLdTransformer<Jso
         var id = nodeId(jsonObject);
         var props = new HashMap<String, Object>();
         visitProperties(jsonObject, (key, jsonValue) -> {
-            var res = transformGenericProperty(jsonValue, context);
-            props.put(key, res);
+            if (!JsonLdKeywords.ID.equals(key)) {
+                var res = transformGenericProperty(jsonValue, context);
+                props.put(key, res);
+            }
         });
         return new Issuer(id, props);
     }
--- a/extensions/common/iam/identity-trust/identity-trust-transform/src/main/java/org/eclipse/edc/iam/identitytrust/transform/to/JsonObjectToIssuerTransformer.java
+++ b/extensions/common/iam/identity-trust/identity-trust-transform/src/main/java/org/eclipse/edc/iam/identitytrust/transform/to/JsonObjectToIssuerTransformer.java
@@ -1,0 +1,41 @@
+/*
+ *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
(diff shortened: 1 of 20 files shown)


extensions/common/iam/identity-trust/identity-trust-transform/src/main/java/org/eclipse/edc/iam/identitytrust/transform/to/JsonObjectToIssuerTransformer.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public @Nullable Issuer transform(@NotNull JsonObject jsonObject, @NotNull TransformerContext context) {
+        var id = nodeId(jsonObject);
+        var props = new HashMap<String, Object>();
+        visitProperties(jsonObject, (key, jsonValue) -> {
+            var res = transformGenericProperty(jsonValue, context);
+            props.put(key, res);
+        });
+        return new Issuer(id, props);
+    }
```
bscholtes1A, line 40: Might deserve few unit-tests. 

Also do we want the id to also be put in the `additionalProperties` map? I would say no but seems this is what current implem is doing

extensions/common/iam/identity-trust/identity-trust-transform/src/main/java/org/eclipse/edc/iam/identitytrust/transform/to/JsonObjectToVerifiableCredentialTransformer.java
```diff
@@ -84,7 +86,21 @@ private Instant parseDate(JsonValue jsonValue, TransformerContext context) {
...
-        return transformString(jsonValue, context); //todo: handle the case where the issuer is an object
+    private Issuer parseIssuer(JsonValue jsonValue, TransformerContext context) {
+        if (jsonValue.getValueType() == JsonValue.ValueType.STRING) {
+            return new Issuer(transformString(jsonValue, context), Map.of());
+        } else {
+            // issuers can be objects, that MUST contain an ID, and optional other properties
+            // an issuer is never an array with >1 elements
+            JsonObject issuer;
+            if (jsonValue.getValueType() == JsonValue.ValueType.ARRAY) {
+                issuer = jsonValue.asJsonArray().get(0).asJsonObject();
```
bscholtes1A, line 97: Don't we need to check the array is not empty before doing that? 
Might be a sneaky (and probably very unlikely) case in practice, but to be sure :)

extensions/common/iam/identity-trust/identity-trust-transform/src/main/java/org/eclipse/edc/iam/identitytrust/transform/to/JsonObjectToIssuerTransformer.java
paullatzelsperger: Yes, and I just remembered that the transformer also needs a few tests...

paullatzelsperger: Assuming we have validation on the ingress, we can assume that the issuer (mandatory field) is there, and since the jsonld is expanded, the array can't be empty

extensions/common/iam/identity-trust/identity-trust-transform/src/main/java/org/eclipse/edc/iam/identitytrust/transform/to/JsonObjectToVerifiableCredentialTransformer.java
paullatzelsperger: Assuming we have validation on the ingress, we can assume that the issuer (mandatory field) is there, and since the jsonld is expanded, the array can't be empty

bscholtes1A: fair enough

bscholtes1A (changes requested):
`JsonObjectToIssuerTransformer` to be unit-tested

Apart from that looks good to me!

bscholtes1A (approved):

PR #3541, opened 2023-10-13
feat: get rid of `cid` claim in `EndpointDataReference` auth token

### LINKED ISSUE #3478: EDR: Remove usage of the `cid` field
# Feature Request

Stop using the `cid` field to hold the contract ID.

EDR

`cid` is used by some token generators to hold the client ID, thus potentially causing conflicts/problems.

## Solution Proposal

Only use the `contractId` field

 ---
## CAUTION

this issue should be implemented close to the release of Milestone 12, as it constitutes a BREAKING CHANGE!

Depends on #3477

bscholtes1A:
## What this PR changes/adds

Remove `cid` claim from auth token generated for `EndpointDataReference` 

## Why it does that

Avoid conflict when `cid` claim is already used.

## Linked Issue(s)

Closes #3478 


paullatzelsperger (approved):
all good, but as issue #3478 states, it will be a breaking change, and thus we should merge this close to the next milestone release (0.4.0), which is scheduled for Nov 13.
The reason being is, that if we have to release a bugfix version in the meantime, that would already contain that breaking change.

bscholtes1A:
> all good, but as issue #3478 states, it will be a breaking change, and thus we should merge this close to the next milestone release (0.4.0), which is scheduled for Nov 13. The reason being is, that if we have to release a bugfix version in the meantime, that would already contain that breaking change.

No problem I'll keep it up-to-date until the release

ndr-brt (approved):

PR #3542, opened 2023-10-14
fix: Disable jetty server version

### LINKED ISSUE #3415: Avoid leaking version information about jetty runtime
# Bug Report

## Describe the Bug

We run a pentest on the EDC (specially the Tractus-X EDC 0.5.0) and discovered, that the jetty version is been exposed on 404 pages of the management api.

Detailed information about used technology and software products allow attackers to identify known exploits against these systems.

### Expected Behavior

Do not leak software version information to users in any way (headers or content) and configure systems to not send such information.

### Observed Behavior

Visiting the `/management` endpoint reveals the use of Jetty version (11.0.15). Accessing this information does not require any form of authentication:
image

## Steps to Reproduce

- Call the `/management` path or any other path which response will be 404.

## Context Information

- Used version [e.g. EDC v1.0.0]: Tractus-X EDC 0.5.0

awellnitz-materna:
## What this PR changes/adds

In the `JettyService` class I have adjusted the `HttpConnectionFactory` and disabled the Jetty server version via the `HttpConfiguration`.

## Why it does that

As mentioned in #3415, this issue can lead to a potential attacker finding a vulnerability in Jetty more quickly.

## Linked Issue(s)

Closes #3415

### DIFF
--- a/extensions/common/http/jetty-core/src/main/java/org/eclipse/edc/web/jetty/JettyService.java
+++ b/extensions/common/http/jetty-core/src/main/java/org/eclipse/edc/web/jetty/JettyService.java
@@ -11,6 +11,7 @@
  *       Microsoft Corporation - initial API and implementation
  *       Fraunhofer Institute for Software and Systems Engineering - added method
  *       ZF Friedrichshafen AG - Set connector name
+ *       Materna Information & Communications SE - disable Jetty send server version
  *
  */
 
@@ -203,6 +204,7 @@ public class JettyService implements WebServer {
     @NotNull
     private HttpConnectionFactory httpConnectionFactory() {
         HttpConfiguration https = new HttpConfiguration();
+        https.setSendServerVersion(false);
         return new HttpConnectionFactory(https);
     }
 


extensions/common/http/jetty-core/src/main/java/org/eclipse/edc/web/jetty/JettyService.java
```diff
@@ -102,7 +104,7 @@ public void start() {
 
                 var handler = createHandler(mapping);
                 handlers.put(mapping.getPath(), handler);
-            });
+            });
```
jimmarino, line 107:
```suggestion
            });
```

extensions/common/http/jetty-core/src/main/java/org/eclipse/edc/web/jetty/JettyService.java
```diff
@@ -203,6 +205,7 @@ private void configure(ServerConnector connector) {
     @NotNull
     private HttpConnectionFactory httpConnectionFactory() {
         HttpConfiguration https = new HttpConfiguration();
+        https.setSendServerVersion(SEND_SERVER_VERSION);
```
jimmarino, line 208: Please set this to false inline instead of using a static constant, as it is easier to read. Also, prefer primitives.

jimmarino (changes requested):

extensions/common/http/jetty-core/src/main/java/org/eclipse/edc/web/jetty/JettyService.java
awellnitz-materna: @jimmarino I have adjusted it.

jimmarino (approved):

PR #3543, opened 2023-10-16
fix: data-plane-http - HttpSink loosing messages #3417 

### LINKED ISSUE #3417: Multiple partitions are not being transfered in Http Datapalne
# Bug Report

## Describe the Bug
The HttpDataSink component only transfers the first partition of data when there are multiple partitions.

### Expected Behavior

The HttpDataSink should be able to transfer all partitions of data.

### Observed Behavior

Currently, when multiple partitions are sent to the HttpDataSink, only the first partition is transferred. The others are not transferred.

## Steps to Reproduce

- Use Kafka as the source for an asset.
- Configure HttpData as the target.
   - Note that the default partition size is hard-coded to 5.
- After the Kafka topic receives 5 messages, the transferPart function is triggered, but it only sends the first message to the designated endpoint.

## Possible Implementation

One possible solution to this bug is to remove the "Return Early" condition from the [TransferParts](https://github.com/eclipse-edc/Connector/blob/main/extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpDataSink.java#L51C20-L51C20) function. This will allow the HttpDataSink to process and transfer all partitions of data.

drcgjung:
## What this PR changes/adds

This PR introduces a test case to verify the loss of messages by HttpSink under various partitioning conditions.
This PR removes an early return statement in the partition transfer loop of HttpSink which leads to the loss of messages.

## Why it does that

Currently, HttpSink looses messages even in completely deterministic settings. It may only be safely applied with partition.size=1

## Linked Issue(s)

Closes #3417

### DIFF
--- a/extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpDataSink.java
+++ b/extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/pipeline/HttpDataSink.java
@@ -47,8 +47,6 @@ public class HttpDataSink extends ParallelSink {
                             response.code(), response.message(), part.name(), request.url().url(), request));
                     return ERROR_WRITING_DATA;
                 }
-
-                return StreamResult.success();
             } catch (Exception e) {
                 monitor.severe(format("Error writing HTTP data %s to endpoint %s for request: %s", part.name(), request.url().url(), request), e);
                 return ERROR_WRITING_DATA;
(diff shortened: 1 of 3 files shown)


ndr-brt (commented):
Is there some additional work to be done? If it's ready for review please mark it accordingly.

drcgjung:
> I'm waiting for @git-masoud to check before I undraft it.

One point of discussion was the location of [MultipleBinaryPartsDataSource.java](https://github.com/eclipse-edc/Connector/pull/3543/commits/460e8e977aadeeadd2c78477769f3a8dcf7f9d35#diff-6a3e4135b61a5e9275108694fc7cd845015272e1785b844c83deedc50b6b3f2b)

I placed it under the testFixtures of data-plane-spi because it is not specific to just data-plane-http and could be used to test all kinds of pipeline sinks ...

ndr-brt (approved):
looks good, thanks!

PR #3545, opened 2023-10-16
feat: verification of (consumer) VP (VC-JWT)

### LINKED ISSUE #3533: IATP: Verify consumer VP
## WHAT
Implement a verifier structure to perform the cryptographic verification of VPs and VCs.

We have to distinguish between two basic formats: JSON-LD and JWT. Every VP and every VC can be in either format, ~~and heterogenous~~ but only homogenous compositions are possible: A JWT-VP contains ~~a JSON-LD-VC and~~ only a JWT-VC, etc. An example of this can be found in the [Verifiable Credential Model v1.1](https://www.w3.org/TR/vc-data-model/#example-verifiable-presentation-using-jwt-compact-serialization-non-normative).

The procedure for verification should be as follows:
- take the raw VP input as received from the holder, and verify it according to the `format` property.
- for every VC, take the raw VC input as received from the holder, and verify it.

It should be emphasized, that the _raw input_ must be used, so as to allow for both JSON-LD literals and JWT strings.

[edit]: restrict VP/VC-formats.

paullatzelsperger:
## What this PR changes/adds

This PR adds the cryptographic verification of VerifiablePresentations for the following formats:
- JSON-LD: verification is done based on LinkedDataProofs (**edit: will come in another PR**)
- JWT: verification is done based on JWS

## Why it does that

Verifiers (e.g. a provider connector) must be able to ensure the cryptographic integrity of a presenter's VP

## Further notes

VPs must be homogenous, i.e. if a VP is presented as JWT, it may only contain VC's that are also encoded as JWT. 
**This is a restriction imposed by the [Verfiable Credentials Data Model v1.1](https://www.w3.org/TR/vc-data-model/#json-web-token-extensions).**

## Linked Issue(s)

Closes #3533

### DIFF
--- a/extensions/common/iam/identity-trust/identity-trust-service/src/main/java/org/eclipse/edc/iam/identitytrust/verification/JwtPresentationVerifier.java
+++ b/extensions/common/iam/identity-trust/identity-trust-service/src/main/java/org/eclipse/edc/iam/identitytrust/verification/JwtPresentationVerifier.java
@@ -26,7 +26,27 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * Verifies VerifiablePresentations, which are present in JWT format. Only the cryptographic integrity is asserted
+ * Computes the cryptographic integrity of a VerifiablePresentation when it's represented as JWT.
+ * In order to be successfully verified, a VP-JWT must contain a "vp" claim, that contains a JSON structure containing a
+ * "verifiableCredentials" object.
+ * This object contains an array of strings, each representing one VerifiableCredential, represented in JWT format.
+ * <p><br/>
+ * Both VP-JWTs and VC-JWTs must fulfill the following requirements:
+ *
+ * <ul>
+ *     <li>contain a JWS, that can be verified using the issuer's public key</li>
+ *     <li>have in its header a key-id ("kid"), which references the public key in the DID document of the VP-issuer</li>
+ *     <li>contain the following claims:
+ *     <ul>
+ *         <li>aud: must be equal to this connector's DID</li>
+ *         <li>sub: only presence is asserted</li>
+ *         <li>iss: is used to resolve the DID, that contains the key-id</li>
+ *     </ul>
+ *     <li>A VP is only verified, if it and all VCs it contains are verified</li>
+ *     </li>
+ * </ul>
+ *
+ * <em>Note: VP-JWTs may only contain VCs also represented in JWT format. Mixing formats is not allowed.</em>
  */
 class JwtPresentationVerifier {
     public static final String VERIFIABLE_CREDENTIAL_JSON_KEY = "verifiableCredential";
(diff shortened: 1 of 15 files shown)


paullatzelsperger:
adding glue code, testing, etc. for LDP verification has turned out to be quite significant in terms of changed files, so I decided to do it in another PR.

**This one only contains the JWT verifier!!**

extensions/common/iam/identity-trust/identity-trust-service/src/main/java/org/eclipse/edc/iam/identitytrust/verification/JwtPresentationVerifier.java
```diff
@@ -14,18 +14,112 @@
...
+                return Result.failure("No credential found in presentation.");
+            }
+
+            // every VC is represented as another JWT, so we verify all of them
+            for (String token : vcTokens) {
+                vpResult = vpResult.merge(jwtVerifier.verify(token, signedVpJwt.getJWTClaimsSet().getIssuer()));
+            }
+
+        } catch (ParseException | JsonProcessingException e) {
+            throw new RuntimeException(e);
```
bscholtes1A, line 113: Do we really want to throw? Shouldn't we return a failed result instead?

extensions/common/iam/identity-trust/identity-trust-service/src/main/java/org/eclipse/edc/iam/identitytrust/verification/MultiFormatPresentationVerifier.java
```diff
@@ -14,19 +14,27 @@
...
+import org.eclipse.edc.identitytrust.model.VerifiablePresentationContainer;
+import org.eclipse.edc.identitytrust.verification.JwtVerifier;
 import org.eclipse.edc.identitytrust.verification.PresentationVerifier;
 import org.eclipse.edc.spi.result.Result;
 
 public class MultiFormatPresentationVerifier implements PresentationVerifier {
-    private final JwtPresentationVerifier jwtVerifier = new JwtPresentationVerifier();
-    private final JsonLdPresentationVerifier jsonLdVerifier = new JsonLdPresentationVerifier();
+    private final JwtPresentationVerifier jwtPresentationVerifier;
+    private final JsonLdPresentationVerifier jsonLdVerifier;
```
bscholtes1A, line 25:
```suggestion
    private final JsonLdPresentationVerifier jsonLdPresentationVerifier;
```

to be consistent with the `JwtPresentationVerifier`

bscholtes1A (approved):
Few minor comments, but LGTM

extensions/common/iam/identity-trust/identity-trust-service/src/main/java/org/eclipse/edc/iam/identitytrust/verification/JwtPresentationVerifier.java
paullatzelsperger: I was sort of undecided, but this can only happen, if the key is in an invalid format, or we are not dealing with valid JSON, at which point any further processing would be pointless.

wolf4ood (approved):
LGTM!

PR #3549, opened 2023-10-18
feat: Remove duplicate information in the initiate negotiation api request

### LINKED ISSUE #3395: duplicate information in the initiate negotiation api request
# Feature Request

Currently in the `POST /v2/contractnegotiations` request the `ContractRequest` body has this structure:
```json
{
  "@context": {
    "edc": "https://w3id.org/edc/v0.0.1/ns/",
    "odrl": "http://www.w3.org/ns/odrl/2/"
  },
  "@type": "ContractRequest",
  "connectorId": "provider",
  "connectorAddress": "http://localhost:19194/protocol",
  "consumerId": "consumer",
  "providerId": "provider",
  "protocol": "dataspace-protocol-http",
  "offer": {
    "offerId": "MQ==:YXNzZXRJZA==:YTc4OGEwYjMtODRlZi00NWYwLTgwOWQtMGZjZTMwMGM3Y2Ey",
    "assetId": "assetId",
    "policy": {
      "@id": "MQ==:YXNzZXRJZA==:YTc4OGEwYjMtODRlZi00NWYwLTgwOWQtMGZjZTMwMGM3Y2Ey",
      "@type": "use",
      "odrl:permission": [],
      "odrl:prohibition": [],
      "odrl:obligation": [],
      "odrl:target": "assetId"
    }
  }
}
```

As you can see, the `offerId` and the `policy/@id` have the same value, because they're representing the same information, so there's no point in `offerId`, as the `policy` is mandatory anyway.

The proposal would be to remove `offerId`, and maybe inline the `offer` object and have `assetId` and `policy` at the root level, to have something like:
```json
{
  "@context": {
    "edc": "https://w3id.org/edc/v0.0.1/ns/",
    "odrl": "http://www.w3.org/ns/odrl/2/"
  },
  "@type": "ContractRequest",
  "connectorId": "provider",
  "connectorAddress": "http://localhost:19194/protocol",
  "consumerId": "consumer",
  "providerId": "provider",
  "protocol": "dataspace-protocol-http",
  "assetId": "assetId",
  "policy": {
    "@id": "MQ==:YXNzZXRJZA==:YTc4OGEwYjMtODRlZi00NWYwLTgwOWQtMGZjZTMwMGM3Y2Ey",
    "@type": "use",
    "odrl:permission": [],
    "odrl:prohibition": [],
    "odrl:obligation": [],
    "odrl:target": "assetId"
  }
}
```

api

simplify

## Solution Proposal

this shouldn't be a breaking change, deprecation model should be used.

tuncaytunc-zf:
## What this PR changes/adds

Removes the duplicate information in the initiate negotiation api request

## Why it does that

The offerId and the policy/@id have the same value, because they're representing the same information, so there's no point in offerId, as the policy is mandatory anyway.

## Further notes

This introduces new api version of negotiation to avoid any braking changes

## Linked Issue(s)

Closes #3395 

### DIFF
--- a/extensions/control-plane/api/management-api/contract-negotiation-api/src/main/java/org/eclipse/edc/connector/api/management/contractnegotiation/ContractNegotiationApiExtension.java
+++ b/extensions/control-plane/api/management-api/contract-negotiation-api/src/main/java/org/eclipse/edc/connector/api/management/contractnegotiation/ContractNegotiationApiExtension.java
@@ -38,6 +38,7 @@ import org.eclipse.edc.web.spi.WebService;
 
 import java.util.Map;
 
+import static org.eclipse.edc.connector.api.management.contractnegotiation.v3.model.ContractRequestDto.CONTRACT_REQUEST_DTO_TYPE;
 import static org.eclipse.edc.connector.contract.spi.types.command.TerminateNegotiationCommand.TERMINATE_NEGOTIATION_TYPE;
 import static org.eclipse.edc.connector.contract.spi.types.negotiation.ContractRequest.CONTRACT_REQUEST_TYPE;
 
@@ -80,7 +81,7 @@ public class ContractNegotiationApiExtension implements ServiceExtension {
 
         validatorRegistry.register(CONTRACT_REQUEST_TYPE, ContractRequestValidator.instance());
         validatorRegistry.register(TERMINATE_NEGOTIATION_TYPE, TerminateNegotiationValidator.instance());
-        validatorRegistry.register(CONTRACT_REQUEST_TYPE, ContractRequestDtoValidator.instance());
+        validatorRegistry.register(CONTRACT_REQUEST_DTO_TYPE, ContractRequestDtoValidator.instance());
 
         var monitor = context.getMonitor();
 
--- a/extensions/control-plane/api/management-api/contract-negotiation-api/src/main/java/org/eclipse/edc/connector/api/management/contractnegotiation/ContractNegotiationApiExtension.java
+++ b/extensions/control-plane/api/management-api/contract-negotiation-api/src/main/java/org/eclipse/edc/connector/api/management/contractnegotiation/ContractNegotiationApiExtension.java
@@ -23,6 +23,9 @@ import org.eclipse.edc.connector.api.management.contractnegotiation.transform.Js
 import org.eclipse.edc.connector.api.management.contractnegotiation.transform.JsonObjectToContractOfferDescriptionTransformer;
 import org.eclipse.edc.connector.api.management.contractnegotiation.transform.JsonObjectToContractRequestTransformer;
 import org.eclipse.edc.connector.api.management.contractnegotiation.transform.JsonObjectToTerminateNegotiationCommandTransformer;
+import org.eclipse.edc.connector.api.management.contractnegotiation.v3.transform.ContractRequestDtoToContractRequestTransformer;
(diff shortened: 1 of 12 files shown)


ndr-brt (changes requested):
I think there are two ways of doing this.
The way you followed brings in a Dto class, that we decided to remove some weeks ago.
Plus, a new Dto requires a two steps transformation (json -> dto -> model) that we don't really want.

I see that in this case they are needed because there is no way to define different validators and transformers for the same entity based on the controller they are used.

A new api version should be motivated for substantial changes, and I think the changes here are really little and they can be handled in the same version, as we did in the past in other APIs following this approach:
- deprecate the properties on the schema
- broaden the validator to accept the deprecated field as optional
- in the transformer, check if the deprecated properties are used, in that case, print a `warning` log

to get a grasp, we did the same thing when we renamed `providerUrl` to `counterPartyAddress` in the `CatalogRequest`, please look at `JsonObjectToCatalogRequestTransformer`.

tuncaytunc-zf:
> I think there are two ways of doing this. The way you followed brings in a Dto class, that we decided to remove some weeks ago. Plus, a new Dto requires a two steps transformation (json -> dto -> model) that we don't really want.
> 
> I see that in this case they are needed because there is no way to define different validators and transformers for the same entity based on the controller they are used.
> 
> A new api version should be motivated for substantial changes, and I think the changes here are really little and they can be handled in the same version, as we did in the past in other APIs following this approach:
> 
> * deprecate the properties on the schema
> * broaden the validator to accept the deprecated field as optional
> * in the transformer, check if the deprecated properties are used, in that case, print a `warning` log
> 
> to get a grasp, we did the same thing when we renamed `providerUrl` to `counterPartyAddress` in the `CatalogRequest`, please look at `JsonObjectToCatalogRequestTransformer`.

After a short discussion with @wolf4ood we decided to go for a new Dto that's why I have implemented it like this. Unfortunately I was not aware about such a decision probably Enrico neither. But indeed introducing a new api version for such a small change would cause inconvenience. I will close this PR and implement the second way you described.

wolf4ood:
My two cents, it's fine to supports different fields on the same Object, 
but in this case IMHO the shape of the input object changes `substantially` and the single transformer would basically handle two different shapes in one . 
Having the DTO in the middle it's something that it's not wanted but in this case would be only a shim for the time being until the v2 is deprecated.  

We did the same with Asset AP, by changing the shape cause the v3 :)

ndr-brt:
I'd propose then a compromise: how do you see to add the Dto for the deprecated endpoint and have the new one without it? In that case the Dto will be removed when the deprecation will expire

wolf4ood:
For me it's fine also to have it in a deprecation way, as long as it will be deprecated soon :joy: 
and we should at least document well the behaviour of the combination of assetId + policy vs offer.

But i guess at this point as you wrote we could remove also the `assetId` and just use the `policy`?

In that case I'm fine with supporting only the `policy` field  in addition and mark for deprecation `offer` :)

tuncaytunc-zf:
If dto comes into consideration what would help to not frequently change the core model whenever an api schema changes, why then not keep it like implemented here and remove it when it gets deprecated instead of changing it now and remove it afterward 😂

wolf4ood:
> So, having an optional policy (or contractOffer or whathever) field instead of the deprecated offer would already solve the issue, without v3, dtos, additional validators and transformers.

:+1:

PR #3553, opened 2023-10-19
feat: add multi-format VP/VC verifier

### LINKED ISSUE #3533: IATP: Verify consumer VP

paullatzelsperger:
## What this PR changes/adds

This PR adds a verifier structure, which can  verify `ldp_vp`, `ldp_vc`, `jwt_vp` and `jwt_vc` objects by using a flexible delegate mechanism. It is possible to mix-and-match them, for example one could have:
- a JWT-VP, that contains one JWT-VC and one LDP-VC
- an LDP-VP, that contains JWT-VCs
- etc.

_NB: according to their [schema](https://www.w3.org/2018/credentials/v1), LDP-VPs can only contain LDP-VCs, even if the [spec](https://www.w3.org/TR/vc-data-model/) doesn't really mandate that._

## Why it does that

Being able to handle different formats of VPs, and different formats of VCs.

## Further notes

Due to shortcomings in its implementation I abandoned the `Verifier` API of the `iron-verifiable-credentials` library (for LDP), for example, i doesn't properly handle multiple VCs in a presentation. Instead, I re-implemented the same behaviour in the EDC code base. 

Ultimately, I want to get rid of the `iron-verifiable-credentials` lib altogether, but that will require some more work in the following areas:
- issuance
- test compliance: there is significant testing to be done using the [official test suite](https://github.com/w3c/vc-test-suite).

These endeavors will happen in subsequent PRs, once the basic functionality is there.

## Linked Issue(s)

Closes #3533


extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IdentityAndTrustExtension.java
```diff
@@ -79,22 +87,22 @@
...
     @Provider
-    private JwtVerifier getJwtVerifier() {
+    public JwtVerifier getJwtVerifier() {
         if (jwtVerifier == null) {
             jwtVerifier = new SelfIssuedIdTokenVerifier(resolverRegistry);
         }
         return jwtVerifier;
     }
 
+    private String getOwnDid(ServiceExtensionContext context) {
```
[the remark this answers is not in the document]
paullatzelsperger: implementation will come later

extensions/common/crypto/jwt-verifiable-credentials/src/main/java/org/eclipse/edc/verifiablecredentials/jwt/JwtPresentationVerifier.java
```diff
@@ -101,12 +120,12 @@ public Result<Void> verifyPresentation(String serializedJwt) {
 
             if (vcTokens.isEmpty()) {
                 // todo: this is allowed by the spec, but it is semantic nonsense. Should we return failure or not?
-                return Result.failure("No credential found in presentation.");
+                return Result.success();
             }
 
             // every VC is represented as another JWT, so we verify all of them
             for (String token : vcTokens) {
```
bscholtes1A, line 127: According to the PR title, we now support a JWT presentation containing VC either as JWT of LDP so the notation `vcTokens` is misleading here, as it tends to indicate that only JWT VC can exist here.

spi/common/identity-trust-spi/src/main/java/org/eclipse/edc/identitytrust/model/VerifiableCredential.java
```diff
@@ -85,6 +88,10 @@ public String getName() {
         return name;
     }
 
+    public boolean isExpired() {
+        return issuanceDate.isAfter(Instant.now());
```
bscholtes1A, line 92: Usage of `Clock`?

spi/common/identity-trust-spi/src/main/java/org/eclipse/edc/identitytrust/verification/CredentialVerifier.java
```diff
@@ -0,0 +1,23 @@
...
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.identitytrust.verification;
+
+import org.eclipse.edc.spi.result.Result;
+
+public interface CredentialVerifier {
```
bscholtes1A, line 19: doc?

bscholtes1A (commented):
Huge work!
Few comments and questions

spi/common/identity-trust-spi/src/main/java/org/eclipse/edc/identitytrust/model/VerifiableCredential.java
paullatzelsperger: we could even remove this method altogether. Some implementors may allow for a small epsilon (couple of minutes) to account for clock skew

extensions/common/crypto/jwt-verifiable-credentials/src/main/java/org/eclipse/edc/verifiablecredentials/jwt/JwtPresentationVerifier.java
paullatzelsperger: you're right.

extensions/common/iam/identity-trust/identity-trust-service/src/main/java/org/eclipse/edc/iam/identitytrust/validation/rules/IsNotExpired.java
```diff
@@ -0,0 +1,31 @@
...
+import org.eclipse.edc.identitytrust.validation.CredentialValidationRule;
+import org.eclipse.edc.spi.result.Result;
+
+import java.time.Instant;
+
+public class IsNotExpired implements CredentialValidationRule {
+    @Override
+    public Result<Void> apply(VerifiableCredential credential) {
+        // issuance date can not be null, due to builder validation
+        return credential.getIssuanceDate().isAfter(Instant.now()) ?
```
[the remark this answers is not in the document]
paullatzelsperger: good idea!

extensions/common/crypto/jwt-verifiable-credentials/src/main/java/org/eclipse/edc/verifiablecredentials/jwt/SelfIssuedIdTokenVerifier.java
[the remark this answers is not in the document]
paullatzelsperger: I put the class here, because the `jwt-verifiable-credentials` module is supposed to have very few dependencies, so it was convenient.

How about I move it to `crypto-core`?



I think the issue at hand actually a bit more complex, because across our code base, whenever we do JWT verification, we actually do multiple steps:
- resolve the public key (e.g. via DID, or through some config value, through a PKI, etc.)
- perform cryptographic verification (that's the easy part, just delegate to Nimbus)
- perform additional _validation_, i.e. make sure some claims are there, that they have a certain values, not expired, etc.

@jimmarino will take inventory of the code base and propose a solution.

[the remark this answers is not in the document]
bscholtes1A: ok for me to put it in `crypto-core`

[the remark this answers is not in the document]
paullatzelsperger: i took another look, and that'll introduce a dependency onto the did module into `crypto-core`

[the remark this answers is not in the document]
paullatzelsperger: done in [eeb4c49](https://github.com/eclipse-edc/Connector/pull/3553/commits/eeb4c49cb89709fa770c0f6428b54a25c7777b74)

extensions/common/crypto/jwt-verifiable-credentials/src/main/java/org/eclipse/edc/verifiablecredentials/jwt/JwtPresentationVerifier.java
```diff
@@ -43,82 +48,110 @@
...
+            if (credentialTokens.isEmpty()) {
                 // todo: this is allowed by the spec, but it is semantic nonsense. Should we return failure or not?
-                return Result.failure("No credential found in presentation.");
+                return Result.success();
             }
 
             // every VC is represented as another JWT, so we verify all of them
-            for (String token : vcTokens) {
-                vpResult = vpResult.merge(jwtVerifier.verify(token, signedVpJwt.getJWTClaimsSet().getIssuer()));
+            for (String token : credentialTokens) {
```
bscholtes1A, line 127: I would have call it `rawCredentials` instead of `credentialTokens` to be more explicit but it's just detail :)

bscholtes1A (approved):
LGTM

extensions/common/crypto/jwt-verifiable-credentials/src/main/java/org/eclipse/edc/verifiablecredentials/jwt/JwtPresentationVerifier.java
paullatzelsperger: yup, sounds good also.

wolf4ood (approved):
LGTM! :rocket:

PR #3559, opened 2023-10-23
fix: permit query `Asset` json properties

### LINKED ISSUE #3423: Asset query service - JSON
# Bug Report

## Describe the Bug

Asset query service does not work in all cases when the property_value is a JSON.
See example in https://github.com/eclipse-edc/Connector/discussions/3416

### Expected Behavior

See https://github.com/eclipse-edc/Connector/discussions/3416

### Observed Behavior

See https://github.com/eclipse-edc/Connector/discussions/3416

## Steps to Reproduce

See https://github.com/eclipse-edc/Connector/discussions/3416

## Context Information

- Used version -> EDC v0.2.1

## Detailed Description

See https://github.com/eclipse-edc/Connector/discussions/3416

ndr-brt:
## What this PR changes/adds

Permits to query over nested properties in Asset.
To do that, the notation should be, e.g.:
```
{
  "edc:operandLeft": "'https://w3id.org/edc/v0.0.1/ns/nested'.'https://w3id.org/edc/v0.0.1/ns/key'",
  "edc:operator": "=",
  "edc:operandRight": "value"
}
```

to navigate through the property tree, the wrapping `'` are needed to clearly separate path items (because keys could contain the `.`).
Current behavior without wrapping `'` for a not nested property is still valid.

## Why it does that

fix bug

## Further notes

- refactored the asset related sql tables, now collapsed into a single table with JSON fields. **THIS IS A BREAKING CHANGE**. please refer to the `asset-index-sql` extension README to get migration instructions.
- removed the unused `updateDataAddress` method from `AssetService` 

## Linked Issue(s)

Closes #3423 

### DIFF
--- a/core/common/util/build.gradle.kts
+++ b/core/common/util/build.gradle.kts
@@ -18,6 +18,8 @@ plugins {
 }
 
 dependencies {
+    api(project(":spi:common:core-spi"))
+
     testImplementation(libs.junit.pioneer)
 }
 
--- a/core/common/util/src/main/java/org/eclipse/edc/util/reflection/ReflectionUtil.java
+++ b/core/common/util/src/main/java/org/eclipse/edc/util/reflection/ReflectionUtil.java
@@ -14,6 +14,8 @@
 
 package org.eclipse.edc.util.reflection;
 
+import org.eclipse.edc.spi.types.PathItem;
+
 import java.lang.reflect.Field;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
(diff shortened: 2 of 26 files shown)


jimmarino (approved):

wolf4ood (approved):

arnoweiss:
The segments of "operandLeft" must be in extended form, right? I observed that nothing in there resolves against a @context. This isn't necessarily a problem, but I'll document it that way if you confirm.

ndr-brt:
> The segments of `operandLeft` must be in extended form, right? I observed that nothing in there resolves against a `@context`. This isn't necessarily a problem, but I'll document it that way if you confirm.

yes, it must be extended, because it is a simple "value", not an "id"

PR #3571, opened 2023-10-29
feat: add private properties to policy definition entity

suh-rao:
### What this PR changes/adds
Add private properties to policy definition entity. The implementation supports in-memory and postgres as a store.

### Why it does that
The private properties are required to provide the owner with the possibility to add extra properties which should not be exposed outside. This requirement was initiated when we want to add administrative properties like (createdBy, createdAt, chnagedBy, chnagedAt), policyType(Access, Usage). We assume that the private properties can be used to store such a requirement after registering a listener on the policy definition.
More information is in this discussion:
https://github.com/eclipse-edc/Connector/discussions/2592#discussioncomment-7379257

Further notes
List other areas of code that have changed but are not necessarily linked to the main feature. This could be method
signature changes, package declarations, bugs that were encountered and were fixed inline, etc.

Linked Issue(s)
Closes https://github.com/eclipse-edc/Connector/issues/3490


paullatzelsperger (changes requested):
No pr description, lacking test coverage, no query by private property functionality and testing of it. This pr is insufficient all around.

PR #3597, opened 2023-11-10
refactor(api): counterPartyAddress

### LINKED ISSUE #3343: rename `providerUrl` and `connectorAddress` to `counterPartyAddress` on management api
# Feature Request

Currently there are multiple way of calling the same field: `counterPartyAddress` as it was called in the internal domain model (`RemoteMessage` included) is also called `providerUrl` and `connectorAddress`.

management api

consistency

## Solution Proposal

we should need to avoid breaking changes, printing a log in the case the old name is used, as done in #3342

saschaisele-zf:
## What this PR changes/adds

Deprecates the remaining uses of `connectorAddress` and `providerUrl` in the `ContractNegotiationApi` and `TransferProcessApi`.
Instead, `counterPartyAddress` is used going forward.

## Why it does that

Use of consistent naming across the management api.

## Further notes

It is important to note that this is a breaking change.
CAUTION ADVICED WHEN MERGING.

## Linked Issue(s)

Closes #3343 


extensions/control-plane/api/management-api/contract-negotiation-api/src/main/java/org/eclipse/edc/connector/api/management/contractnegotiation/validation/ContractRequestValidator.java
```diff
@@ -40,4 +43,22 @@ public static Validator<JsonObject> instance() {
                 )
                 .build();
     }
+
+    private record MandatoryCounterPartyAddressOrConnectorAddress(JsonLdPath path) implements Validator<JsonObject> {
+
+        @Override
+        public ValidationResult validate(JsonObject input) {
```
jimmarino, line 50: This can be simplified by checking if the first value is set and returning

extensions/control-plane/api/management-api/transfer-process-api/src/main/java/org/eclipse/edc/connector/api/management/transferprocess/validation/TransferRequestValidator.java
```diff
@@ -43,4 +46,23 @@ public static Validator<JsonObject> instance() {
...
                 .build();
     }
+
+    private record MandatoryCounterPartyAddressOrConnectorAddress(JsonLdPath path) implements Validator<JsonObject> {
+
+        @Override
+        public ValidationResult validate(JsonObject input) {
+            var counterPartyAddress = new MandatoryValue(path.append(TRANSFER_REQUEST_COUNTER_PARTY_ADDRESS));
+            var connectorAddress = new MandatoryValue(path.append(TRANSFER_REQUEST_CONNECTOR_ADDRESS));
+
```
jimmarino, line 56: Please optimize by checking if the first succeeded and return

jimmarino (changes requested):
Can we also note that this is a breaking change because it is not backwards compatible with prior connector versions before this commit?

Note, we should also discuss when to merge this.

jimmarino:
Added a note in the review: we need to discuss when to introduce this change due to DSP backward-compatibility issues.

extensions/control-plane/api/management-api/contract-negotiation-api/src/main/java/org/eclipse/edc/connector/api/management/contractnegotiation/transform/JsonObjectToContractRequestTransformer.java
```diff
@@ -69,8 +70,13 @@ private String getProviderId(@NotNull JsonObject jsonObject, @NotNull Transforme
             return transformString(providerId, context);
         }
 
-        return transformString(jsonObject.get(CONNECTOR_ADDRESS), context);
+        return counterPartyAddressOrConnectorAddress(jsonObject, context);
 
     }
 
+    private String counterPartyAddressOrConnectorAddress(@NotNull JsonObject jsonObject, @NotNull TransformerContext context) {
+        var counterPartyAddress = transformString(jsonObject.get(CONTRACT_REQUEST_COUNTER_PARTY_ADDRESS), context);
```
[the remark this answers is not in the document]
saschaisele-zf: Added a note to mark it as obsolete once the fields are removed.
Also added the note to the other validators where it applies.
Call for `CONNECTOR_ADDRESS` is now only made if `counterPartyAddress` is `null`.

extensions/control-plane/api/management-api/contract-negotiation-api/src/main/java/org/eclipse/edc/connector/api/management/contractnegotiation/validation/ContractRequestValidator.java
saschaisele-zf: Rearranged the order to return as soon as `coutnerPartyAddress` is found.
Did the some for the `TransferRequestValidator` and `CatalogRequestValidator`.

extensions/control-plane/api/management-api/transfer-process-api/src/main/java/org/eclipse/edc/connector/api/management/transferprocess/validation/TransferRequestValidator.java
saschaisele-zf: Done.

extensions/control-plane/api/management-api/contract-negotiation-api/src/main/java/org/eclipse/edc/connector/api/management/contractnegotiation/validation/ContractRequestValidator.java
```diff
@@ -40,4 +43,25 @@ public static Validator<JsonObject> instance() {
...
+            var counterPartyAddress = new MandatoryValue(path.append(CONTRACT_REQUEST_COUNTER_PARTY_ADDRESS));
+            var validateCounterParty = counterPartyAddress.validate(input);
+            if (validateCounterParty.succeeded()) {
+                return ValidationResult.success();
+            }
+            var connectorAddress = new MandatoryValue(path.append(CONNECTOR_ADDRESS));
+            var validateConnectorAddress = connectorAddress.validate(input);
+            if (validateConnectorAddress.succeeded()) {
+                return ValidationResult.success();
```
ndr-brt, line 62: Please log a warning message that suggest to switch to the new attribute

ndr-brt (changes requested):
just minor comments about the `warning` messages.
@jimmarino I think this PR is not a breaking change (everything will work in the same way), the break part of the change will happen when we'll remove the old attributes. I don't see change in the DSP protocol either

extensions/control-plane/api/management-api/transfer-process-api/src/main/java/org/eclipse/edc/connector/api/management/transferprocess/transform/JsonObjectToTransferRequestTransformer.java
```diff
@@ -54,7 +60,12 @@ public JsonObjectToTransferRequestTransformer() {
 
         builder.id(nodeId(input));
         visitProperties(input, k -> switch (k) {
-            case TRANSFER_REQUEST_CONNECTOR_ADDRESS -> v -> builder.connectorAddress(transformString(v, context));
+            case TRANSFER_REQUEST_CONNECTOR_ADDRESS -> v -> {
+                monitor.warning(format("The attribute %s has been deprecated in type %s, please use %s",
+                        TRANSFER_REQUEST_CONNECTOR_ADDRESS, TRANSFER_REQUEST_TYPE, TRANSFER_REQUEST_COUNTER_PARTY_ADDRESS));
```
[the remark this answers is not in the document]
saschaisele-zf: Since working a `Monitor` into the existing `Validator` interface would be a big change in multiple places not related to this PR, I reckon this should either be a seperate Issue or a new `ValidatorWithMonitor` Interface should be created specificly for this case. Any thoughts here?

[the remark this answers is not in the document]
ndr-brt: My suggestion would be to pass the monitor in the constructor, like we usually do with dependency injection.

extensions/control-plane/api/management-api/contract-negotiation-api/src/main/java/org/eclipse/edc/connector/api/management/contractnegotiation/validation/ContractRequestValidator.java
saschaisele-zf: Added the warning message.

extensions/control-plane/api/management-api/transfer-process-api/src/main/java/org/eclipse/edc/connector/api/management/transferprocess/transform/JsonObjectToTransferRequestTransformer.java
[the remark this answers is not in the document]
saschaisele-zf: Moved the warning messages from transformers to their respective validators.

ndr-brt (approved):
LGTM

PR #3621, opened 2023-11-16
feat: add prefix monitor

### LINKED ISSUE #3510: monitor type builder
# Feature Request

Currently the monitor instance is passed to all the components that needs it, but it doesn't get "specialized", so if from a component an info is logged:
```java
monitor.info("interesting information");
```

on the console will be printed without an indication of the component, this could make logs hard to read.
A current workaround is to manually add a prefix on the log message, but a more holistic solution could be to have a way to build a `Monitor` instance that will add a prefix, like
```java
var componentMonitor = monitor.forComponent("Component");
new Component(monitor);
```
then every `Component` log will have the `[Component]` prefix

_Are there any requirements?_

## Solution Proposal

hamidonos:
## What this PR changes/adds

This will add a new monitor implementation so that logs can be prefixed with a custom string.

## Why it does that

Give logs more meaning by providing a context.

## Linked Issue(s)

Closes #3510


spi/common/core-spi/src/main/java/org/eclipse/edc/spi/monitor/PrefixMonitor.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+package org.eclipse.edc.spi.monitor;
+
+import java.util.function.Supplier;
+
+/**
+ * Monitor implementation which will prefix a supplied text to any log message.
+ * Note that this monitor will delegate the actual log execution to the underlying monitor provided by the context.
+ */
+public class PrefixMonitor implements Monitor {
```
jimmarino, line 23: This can be made package private

extensions/common/monitor/monitor-jdk-logger/src/test/java/org/eclipse/edc/monitor/logger/PrefixMonitorTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+import java.util.function.Supplier;
+
+import static java.lang.String.format;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.verify;
+
+public class PrefixMonitorTest {
```
jimmarino, line 31: This test class is too complicated and opaque. It should be apparent what is being asserted by quickly looking at the individual test methods. Please:

1. Replace the use of captors with argument matchers and strict mocks. This will make the test easier to read and less verbose. 
2. Inline the assert methods in all test methods rather than having two separate private methods, one of which calls a supplier and delegates to the other

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/monitor/PrefixMonitor.java
hamidonos: Are you sure? Class won't be accessible anywhere (also in the test)
 
Did you actually mean the `PrefixMonitorTest` class?

ndr-brt (changes requested):

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/monitor/Monitor.java
```diff
@@ -64,4 +65,8 @@ default String sanitizeMessage(Supplier<String> supplier) {
                 .orElse(null);
     }
 
+    default Monitor withPrefix(String prefix) {
```
ndr-brt, line 68: add documentation

jimmarino:
Please resolve the outstanding comments and dependency review. Thx.

jimmarino (approved):

PR #3630, opened 2023-11-20
refactor: make DataPlanePublicApiController use PipelineService

### LINKED ISSUE #3629: refactor: make `DataPlanePublicApiController` use `PipelineService`
# Feature Request

Currently `DataPlanePublicApiController` uses `DataPlaneManager` to transfer data from source to `OutputStreamDataSink` but in fact it should do this directly with `PipelineService`, doing that the `transfer` method on `DataPlaneManager` can be removed, as that's not its job to trigger a pipeline transfer flow (for this we have the `PipelineService`

data-plane

refactor

## Solution Proposal

ndr-brt:
## What this PR changes/adds

Make `DataPlanePublicApiController` go directly through `PipelineService` instead of passing through `DataPlaneManager`

## Why it does that

avoid useless indirection

## Further notes

- refactored tests

## Linked Issue(s)

Closes #3629


extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/DataPlaneApiExtension.java
```diff
@@ -67,6 +65,9 @@ public class DataPlaneApiExtension implements ServiceExtension {
     @Inject
     private DataPlaneManager dataPlaneManager;
```
bscholtes1A, line 66: Can be removed?

bscholtes1A (approved):

extensions/data-plane/data-plane-api/src/main/java/org/eclipse/edc/connector/dataplane/api/DataPlaneApiExtension.java
ndr-brt: no, because it is used by `DataPlaneControlApiController`

jimmarino (approved):

wolf4ood (approved):

PR #3644, opened 2023-11-23
chore: fix Kafka flaky test

wolf4ood:
## What this PR changes/adds

This PR should fix the test `EndToEndKafkaTransferTest#kafkaToHttpTransfer` which currently fails sometime.

I've tracked down the reason and it seems due the fact that even though at consumer side the transfer is terminated
and on the provider side the data source is closed, there could be some stream records still waiting to be pushed via HTTP.

The fix is in the assert lambda clear the expectations on the `destinationServer` if the verify fails.

## Why it does that

fix flaky test


ndr-brt (approved):
looks good thanks!

PR #3651, opened 2023-11-23
feat: add LRU cache for DID Documents

### LINKED ISSUE #3645: Create a LRU cache for resolved DidDocuments
# Feature Request

In the context of IATP, many things are based on DID documents, such as resolving key material, resolving service URLs (e.g. for the `CredentialService`), and possibly others.

In order to decrease the amount of network calls to resolve the DID document, we will implement a LRU cache, that the `DidResolverRegistry` uses.

`identity-did-core`

Not all of the operations that require information extracted from a DID are done in the same place, and tunneling the did document through everywhere would introduce weird and incorrect dependencies in SPI modules.

## Solution Proposal

Create a LRU cache based on a `LinkedHashmap`, cache DID documents, as it is safe to assume, that a DID document will be used multiple times in close temporal vicinity.

paullatzelsperger:
## What this PR changes/adds

- Modifies the existing `LruCache` to become a `ConcurrentLruCache`.
- uses the `ConcurrentLruCache` in the `DidResolverRegistryImpl` to avoid excessive network traffic.

## Why it does that

Avoid excessive network traffic, when resolving the same did in close temporal vicinity.

## Further notes

## Linked Issue(s)

Closes #3645


jimmarino (approved):

bscholtes1A:
@paullatzelsperger  sorry for the late review, but shouldn't we fetch again the DID document after some time? (even if it is present in the cache), in order to avoid to continue using an outdated version of the DID document?

paullatzelsperger:
> @paullatzelsperger  sorry for the late review, but shouldn't we fetch again the DID document after some time? (even if it is present in the cache), in order to avoid to continue using an outdated version of the DID document?

Yep, good point. Sounds like a task for a follow-up issue :)
I didn't bother with it now, because for our intents and purposes, the dids are rather long-living

PR #3659, opened 2023-11-25
feat: relying-party creates SI token for VP query

paullatzelsperger:
## What this PR changes/adds

This PR extends the `IdentityAndTrustService` and generates another SI token on the relying-party side 
which is then used to execute the VP query against the counter-party's CredentialService (= presentation endpoint).

## Why it does that

Complete the circle of the presentation flow

## Further notes

- **Be aware that currently we need to parse the `access_token` to get the `scope` string. This is a workaround, until we can extract the scope from the verification context. This behaviour will change in subsequent developments and is only there for the immediate future. The related code has been marked accordingly.**
- the `access_token` constant was renamed to `PRESENTATION_ACCESS_TOKEN_CLAIM` and is now used everywhere
- this PR also adds the temporary `AudienceResolver` object, which maps a DSP url to an audience, which can be used to resolve key material

## Linked Issue(s)


wolf4ood (approved):
LGTM

PR #3660, opened 2023-11-26
feat: various minor improvements / patches for IATP

### LINKED ISSUE #3650: LDP VC/VP Verification: create URI-equality comparison that ignores the `fragment`
# Feature Request

the [VC/VP Data Integrity specification](https://w3c.github.io/vc-data-integrity/#proofs) requires in numerous places, that two URIs be the same, for example the `issuer` and the `verificationMethod` contained in the proof object. These items typically are URIs.

Normally, the issuer would contain the general resource identifier, e.g. `did:web:the-issuer`, while the verification method would contain a URI fragment, indicating the key material used to create the proof, e.g. `did:web:the-issuer#signing-key-1`. 

Using the JDKs `URI#equals` method will always fail in those cases, because it will include the fragment. 

Thus, we need our use-case specific URI comparison.

VC/VP verification

Compliance with the spec.

## Solution Proposal

convert the URIs to a string, chop off the fragment, perform string equals comparison.

paullatzelsperger:
## What this PR changes/adds

This PR adds several minor features and fixes for IATP:
- extracts a `ClaimToken` from a VC
- performs URI comparison without fragment when comparing Issuer and VerificationMethod
- fixes a wrong extracton of DID#VerificationMethod.controller and DID#VerificationMethod.type
- adds a customizable accumulator operation to the predicate-based query resolver

## Further notes

## Linked Issue(s)

Closes #3650

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -81,7 +81,7 @@ maven/mavencentral/com.jcraft/jzlib/1.1.3, BSD-2-Clause, approved, CQ6218
 maven/mavencentral/com.lmax/disruptor/3.4.4, Apache-2.0, approved, clearlydefined
 maven/mavencentral/com.networknt/json-schema-validator/1.0.76, Apache-2.0, approved, CQ22638
 maven/mavencentral/com.nimbusds/nimbus-jose-jwt/9.28, Apache-2.0, approved, clearlydefined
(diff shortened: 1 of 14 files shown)


wolf4ood (approved):
LGTM

PR #3665, opened 2023-11-27
docs: decision record about Protocol Services refactor

wolf4ood:
## What this PR changes/adds

Adds a Decision Record about moving the security checks (`IdentityService#verifyJwtToken`) from the DSP layer
to the protocol service layer.

## Why it does that

This refactor will allow us to attach contextual information of the current DSP request that can be used in `IdentityService` implementations.

## Linked Issue(s)

Relates #3662 

_Please be sure to take a look at the [contributing guidelines](https://github.com/eclipse-edc/.github/blob/main/CONTRIBUTING.md#submit-a-pull-request) and our [etiquette for pull requests](https://github.com/eclipse-edc/.github/blob/main/contributing/pr_etiquette.md)._

PR #3668, opened 2023-11-29
refactor: moved security checks in protocol service layer

### LINKED ISSUE #3667: Protocol Service refactoring
# Feature Request

Implements the refactoring outlined in the DR #3665  by pushing the security checks into the protocol service layer

dsp, core services

iatp, protocol agnostic security checks

wolf4ood:
## What this PR changes/adds

Implements the refactor outlined in DR #3665 

## Why it does that

Decoupling the net protocol from the security checksm which are now delegated to the protocol service layer.

This also will allow use to attach contextual information to the verification process.

## Further notes

In this PR since we are pushing here the invocation of the `IdentityService` we don't know the audience here/ The audience removal will be tackle in subsequent PRs. 

`IdentityService` that relies on this parameter would not work for the time being.

## Linked Issue(s)

Closes #3667 

### DIFF
--- a/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/protocol/BaseProtocolService.java
+++ b/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/protocol/BaseProtocolService.java
@@ -20,8 +20,6 @@ import org.eclipse.edc.spi.iam.TokenRepresentation;
 import org.eclipse.edc.spi.monitor.Monitor;
 import org.eclipse.edc.spi.result.ServiceResult;
 
-import java.util.function.Function;
-
 /**
  * Base class for all protocol service implementation. This will contain common logic such as validating the JWT token
  * and extracting the {@link ClaimToken}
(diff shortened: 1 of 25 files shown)


core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/service/protocol/BaseProtocolService.java
```diff
@@ -0,0 +1,59 @@
...
+        // TODO: since we are pushing here the invocation of the IdentityService we don't know the audience here
+        //  The audience removal will be tackle next. IdentityService that relies on this parameter would not work
+        //  for the time being.
+        var result = identityService.verifyJwtToken(tokenRepresentation, null);
+
+        if (result.failed()) {
+            monitor.debug(() -> "Unauthorized: %s".formatted(result.getFailureDetail()));
+            return ServiceResult.unauthorized("Unauthorized");
+        }
+        return callback.apply(result.getContent());
```
ndr-brt, line 57: I think this could just return the `ServiceResult<T>` without the callback parameter, letting the caller chaining the "callback" behavior using the `.map` method or ServiceResult, this will avoid the callback nested block and likely have the code more testable

ndr-brt (approved):
:rocket:

jimmarino (approved):

PR #3669, opened 2023-11-29
refactor: create distributions from `DataFlowManager` transfer types

### LINKED ISSUE #3631: Use `DataFlowManager` to create `Distribution` list
# Feature Request

Currently the `DistributionResolver` is using the `DataPlaneInstanceStore` to create the `Distribution` list that then is used in the `Catalog`, but, as we need to distinguish the transfer type (`push`/`pull`) from the `DataAddress` type (`Http`, `Kafka`, ...).

The component that's holding this knowledge is `DataPlaneManager`, so it should have the responsibility to give a list of distribution format.

transfer

dsp

ndr-brt:
## What this PR changes/adds

Makes `Catalog`'s `Distribution` list being built by the knowledge of the `DataFlowManager`, because that's where the supported transfer types knowledge is stored.

This is the `DataFlowController` details:
- the `ProviderPush` will get the distributions out of the `DataPlaneInstance`s registered on the connector and associate them with the `PUSH` flow type
- the `ConsumerPull` for the moment will only return a single `Http-PULL` type, in the future it will be able to provide other pull mechanism as well (ref. #3637)

## Why it does that

interoperability

## Further notes

-

## Linked Issue(s)

Closes #3631

### DIFF
--- a/core/control-plane/catalog-core/build.gradle.kts
+++ b/core/control-plane/catalog-core/build.gradle.kts
@@ -19,7 +19,7 @@ plugins {
 dependencies {
     api(project(":spi:common:catalog-spi"))
     api(project(":spi:control-plane:contract-spi"))
-    api(project(":spi:data-plane-selector:data-plane-selector-spi"))
+    api(project(":spi:control-plane:transfer-spi"))
 
     testImplementation(project(":core:common:junit"))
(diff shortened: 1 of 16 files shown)


wolf4ood (approved):
:rocket:

PR #3677, opened 2023-11-30
feat: adds verification context

### LINKED ISSUE #3675: Identity Service: introduce VerificationContext
# Feature Request

As part of the #3662 we want the `IdentityService` to receive extra/additional contextual information about the current verification call.

core

`IdentityService` implementors could provide additional checks/logic based on the verification context

## Solution Proposal

The `IdentityService#verifyJwtToken`  will be changed:

from:

```java
Result<ClaimToken> verifyJwtToken(TokenRepresentation tokenRepresentation, String audience);
``` 

to: 

```java
Result<ClaimToken> verifyJwtToken(TokenRepresentation tokenRepresentation, VerificationContext context); 
```

For now the `VerificationContext` will contain the `audience` for backward compatibility and a `Policy` that's associated
with the current verification process.

Policy extraction for the current context will be tacked in separated PRs accordingly to their area (catalog,transfer process, contract negotiation)

wolf4ood:
## What this PR changes/adds

Adds `VerificationContext` in the `IdentityService#verifyJwt` methods

## Why it does that

To be able to attach contextual information in the verification phase

## Further notes

For now the `Policy` it's always an empty one. Policy extraction will be tackle in subsequent PRs in each area (catalog/transfer process/ contract negotiation) 

## Linked Issue(s)

Closes #3675


spi/common/core-spi/src/main/java/org/eclipse/edc/spi/iam/VerificationContext.java
```diff
@@ -0,0 +1,96 @@
...
+package org.eclipse.edc.spi.iam;
+
+import org.eclipse.edc.policy.model.Policy;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Representation an additional context for {@link IdentityService} verifiers.
```
jimmarino, line 24:
```suggestion
 * Provides additional context for {@link IdentityService} verifiers.
```

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/iam/VerificationContext.java
```diff
@@ -0,0 +1,96 @@
...
+    private Policy policy;
+
+    private final Map<Class<?>, Object> additional;
+
+    private VerificationContext() {
+        additional = new HashMap<>();
+    }
+
+    /**
+     * Returns the audience if existent otherwise null.
```
jimmarino, line 40:
```suggestion
     * Returns the audience or null if not available.
```

jimmarino (approved):

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/iam/IdentityService.java
```diff
@@ -38,8 +38,8 @@ public interface IdentityService {
      * Verifies a JWT bearer token.
      *
      * @param tokenRepresentation A token representation including the token to verify.
-     * @param audience            Expected audience.
+     * @param context             The {@link  VerificationContext}.
      * @return Result of the validation.
      */
-    Result<ClaimToken> verifyJwtToken(TokenRepresentation tokenRepresentation, String audience);
+    Result<ClaimToken> verifyJwtToken(TokenRepresentation tokenRepresentation, VerificationContext context);
```
paullatzelsperger, line 44: I can already here voices whining about backwards compatibility and breaking changes... Could we leave the old signature in place in a default method, and have it delegate to this new method?
the `VerificationContext` contains the audience anyway

paullatzelsperger (approved):

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/iam/IdentityService.java
wolf4ood: We can, but that method will not be called by our code in the protocol service layer, as we would need to populate the `VerificationContext` . Also the `audience` there will go away soon :)

ndr-brt: @paullatzelsperger hey, that's me :laughing: 
In my opinion it really depends on how many extension of this could exists, seems to be a pretty used interface so I'd go for the deprecation, but I will be also ok in changing without deprecating
(interface changes are generally more tolerated because only devs are involved, instead rest apis... well, you know)

ndr-brt (approved):

wolf4ood:
I've restored as default method the one with `audience`, but anyway the audience is going away soon :)

PR #3714, opened 2023-12-15
feat: add token renewal mechanism for hashicorp vault

### LINKED ISSUE #3610: HashiCorp Vault: Add token rotation mechanism
# Feature Request

HashiCorp Vault (HCV) supports multiple types of tokens to authenticate. So far, we've only ever used the root token, which has an infinite TTL and cannot be renewed. Using the root token is not recommended for obvious reasons, and might not even be possible in managed situations.

All other tokens have a finite TTL, thus they must be renewed. Non-renewable tokens exist, but they are not relevant here.

There are several ways to renew a token through the REST API: 
- renew a token ([doc](https://developer.hashicorp.com/vault/api-docs/auth/token#renew-a-token))
- renew a token by its accessor ([doc](https://developer.hashicorp.com/vault/api-docs/auth/token#renew-a-token-accessor))
- renew a token "itself" ([doc](https://developer.hashicorp.com/vault/api-docs/auth/token#renew-a-token))

The most useful to us will likely be either the accessor or the "self" renewal.

HashiCorp Vault extension

Security: should not use root token
Convenience: could be impractical or even impossible to use root token (managed environments)
Necessity: all tokens other than the root token have a finite lifespan

## Solution Proposal

- the `edc.vault.hashicorp.token` property contains the non-root token
- a new `edc.vault.hashicorp.token.ttl` property is added containing the TTL in seconds, defaulting to 300 (5 min)
- upon boot-up, the implementation checks if the token is renewable (via [token lookup](https://developer.hashicorp.com/vault/api-docs/auth/token#lookup-a-token))
- if the token is not renewable, a warning is logged
- the token is immediately renewed, setting it to the configured TTL
- a scheduled executor is created to run at TTL - 10 seconds (to account for transient network errors) that performs the renewal

## Q&A
- how will this behave in a clustered environment? most likely, it will just work, as the token itself will not change, only it's TTL. At worst, the token will get renewed much more often than necessary.
- how does one create the "initial" token? This is out of scope of this issue
- what happens if my runtime crashes? Most likely nothing, as the token string will still be the same. No reconfiguration is needed
- Can I still use my root token? Yes, but you really shouldn't.

hamidonos:
## What this PR changes/adds

This PR adds an automatic token renewal mechanism to the Hashicorp Vault extension. 

## Why it does that

Currently the EDC uses the root token, which for obvious reasons, is not advised in production.

## Further notes

Note that Vault tokens don't rotate. They are renewed/refreshed.

A token has a TTL. The TTL defines the interval in which the token has to be renewed, otherwise it will be invalidated forever. If a token expires for any reason the EDC will go into a unhealthy state. In this case the EDC has to be restarted with a fresh token.

The Vault token is still supplied by the configurations. Furthermore these configurations have been added:

- VAULT_TOKEN_TTL: Defines the TTL of the token which is passed to the Vault upon renewal. The TTL might not be honored by the Vault (e.g. in the case of periodic tokens). Therefore the TTL value returned by the Vault should be respected for each renewal.
- VAULT_TOKEN_RENEW_BUFFER: Buffer before a renewal is triggered to account for network issues etc.

Upon start the EDC validates the token. If the token is not valid EDC will crash. If the token is renewable a scheduled one-shot task is submitted. The task executes after a delay defined as `TTL - VAULT_TOKEN_RENEW_BUFFER`. If the renewal was successful the next task is scheduled again at `TTL - VAULT_TOKEN_RENEW_BUFFER` and so on.  The cycle is broken if the renewal failed for **any** reason. It does not make sense to re-schedule the renewal if the renewal failed before, since renewable tokens have a strict time window in which they have to be renewed again.

For non-renewable but valid tokens the EDC will be healthy until the token expires.

Renewable tokens have a max TTL, if they're not periodic. After that time the token will be invalidated no matter what. Further renewals will fail.
The max TTL can either be a system default config on Vault side, or a explicit max TTL set by the admin at token creation.

Periodic tokens can be renewed forever as long they're renewed within the TTL.

## Linked Issue(s)

Closes #3610


core/common/junit/src/main/java/org/eclipse/edc/junit/extensions/DependencyInjectionExtension.java
```diff
@@ -41,20 +42,19 @@
...
  * The {@link ServiceExtensionContext} instance is wrapped by a Mockito Spy.
  */
-public class DependencyInjectionExtension extends BaseRuntime implements BeforeEachCallback, ParameterResolver {
+public class DependencyInjectionExtension extends BaseRuntime implements BeforeEachCallback, BeforeAllCallback, ParameterResolver {
 
     private ServiceExtensionContext context;
     private ObjectFactory factory;
 
     @Override
     public void beforeEach(ExtensionContext extensionContext) {
```
jimmarino, line 51: Why does this need to be done before all and before each?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -9,69 +9,195 @@
...
+                    .build();
+            return Result.success(healthCheckResponse);
+        } catch (IOException e) {
+            return Result.failure(e.getMessage());
+        }
+    }
+
+    void inspectToken(TokenLookUpResponsePayloadToken token) {
+        if (token.isRootToken()) {
+            monitor.warning("Root token detected. Don't use root tokens in production environment!");
```
jimmarino, line 105:
```suggestion
            monitor.warning("Root token detected. Do not use root tokens in production.");
```

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -9,69 +9,195 @@
...
+        } else {
+            if (token.isRenewable()) {
+                monitor.info("Token is renewable");
```
jimmarino, line 108: What purpose does logging this at INFO level serve? INFO level is for important information and I don't think this should be logged at all (not even at DEBUG).

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -9,69 +9,195 @@
...
+            } else {
+                monitor.warning("Token is not renewable");
```
jimmarino, line 110: Ditto see above

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -9,69 +9,195 @@
...
+            }
+
+            if (token.isPeriodicToken()) {
+                monitor.info("Token has a renewal period of %s".formatted(token.getPeriod()));
```
jimmarino, line 114: This does not make sense to log related information at two different levels. Please convert it to something that would be meaningful to an operator:

1. Is it a problem or something the operator must be made aware of? The fact that a token has a particular renewable period does not need to be reported. It pollutes the log stream.
2. If it is a problem, what is the specific remedy? Telling the operator to look at a property in a configuration file (for what?) does not indicate what they need to do. 
3. Do not report related information separately using different log levels.
4. Which token is it?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -9,69 +9,195 @@
...
+                monitor.warning("Configured time-to-live (ttl) will not be honored for periodic tokens");
+            } else {
+                monitor.warning("Non-periodic Token will expire at some point. Check 'max_ttl' property inside Vault server configuration file for more information");
+            }
+
+            if (token.hasExplicitMaxTimeToLive()) {
+                var explicitMaxTtl = token.getExplicitMaxTimeToLive();
+                var formattedTime = TIME_FORMAT.formatted(explicitMaxTtl / 3600, (explicitMaxTtl % 3600) / 60, explicitMaxTtl % 60);
+                monitor.warning("Token will ultimately expire in %s".formatted(formattedTime));
```
jimmarino, line 123: Again, the same issues as above. Why is this a warning? What will happen? What token is it?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -9,69 +9,195 @@
...
+                .newInstance()
+                .increment(hashicorpVaultConfig.timeToLive())
+                .build();
+        var request = httpPost(uri, requestPayload);
+
+        try (var response = httpClient.execute(request)) {
+            if (response.isSuccessful()) {
+                var responseBody = Objects.requireNonNull(response.body()).string();
+                var payload = objectMapper.readValue(responseBody, TokenRenewalResponsePayload.class);
+                payload.getWarnings().forEach(monitor::warning);
```
jimmarino, line 163: Please do not log this way. This needs to be refactored.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -9,69 +9,195 @@
...
+                return Result.success(payload.getToken());
+            } else {
+                return Result.failure("%d".formatted(response.code()));
+            }
+        } catch (IOException e) {
+            return Result.failure(e.getMessage());
```
jimmarino, line 169: Ditto above

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -9,69 +9,195 @@
...
 
-    public Result<String> getSecretValue(@NotNull String key) {
+    void scheduleNextTokenRenewal(long timeToLive) {
+        var delay = timeToLive - hashicorpVaultConfig.renewBuffer();
+
+        scheduledExecutorService.schedule(() -> {
+            var tokenRenewResult = renewToken();
+
+            if (tokenRenewResult.succeeded()) {
+                monitor.info("Token was renewed successfully");
```
jimmarino, line 180: Remove this line

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -9,69 +9,195 @@
...
+                var token = tokenRenewResult.getContent();
+                scheduleNextTokenRenewal(token.getTimeToLive());
+            } else {
+                monitor.warning("Failed to renew token: %s".formatted(tokenRenewResult.getFailureDetail()));
```
jimmarino, line 184: Which token?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -9,69 +9,195 @@
...
+            }
+        }, delay, TimeUnit.SECONDS);
+
+        var formattedTime = TIME_FORMAT.formatted(delay / 3600, (delay % 3600) / 60, delay % 60);
+        monitor.info("Next token renewal scheduled in %s".formatted(formattedTime));
```
jimmarino, line 189: Do not log this. It will spam the output

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -9,69 +9,195 @@
...
-        var request = new Request.Builder().url(requestUri).headers(headers).get().build();
+        var request = httpGet(requestUri);
 
         try (var response = httpClient.execute(request)) {
 
             if (response.isSuccessful()) {
                 if (response.code() == HTTP_CODE_404) {
-                    return Result.failure(
-                            String.format(CALL_UNSUCCESSFUL_ERROR_TEMPLATE, "Secret not found"));
+                    return Result.failure(String.format(CALL_UNSUCCESSFUL_ERROR_TEMPLATE, "Secret not found"));
```
jimmarino, line 200: Secret for which key?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultConfig.java
```diff
@@ -0,0 +1,227 @@
...
+import static org.eclipse.edc.vault.hashicorp.model.Constants.VAULT_TIMEOUT_SECONDS;
+import static org.eclipse.edc.vault.hashicorp.model.Constants.VAULT_TIMEOUT_SECONDS_DEFAULT;
+import static org.eclipse.edc.vault.hashicorp.model.Constants.VAULT_TOKEN;
+import static org.eclipse.edc.vault.hashicorp.model.Constants.VAULT_TOKEN_RENEW_BUFFER;
+import static org.eclipse.edc.vault.hashicorp.model.Constants.VAULT_TOKEN_RENEW_BUFFER_SECONDS_DEFAULT;
+import static org.eclipse.edc.vault.hashicorp.model.Constants.VAULT_TOKEN_TTL;
+import static org.eclipse.edc.vault.hashicorp.model.Constants.VAULT_TOKEN_TTL_SECONDS_DEFAULT;
+import static org.eclipse.edc.vault.hashicorp.model.Constants.VAULT_URL;
+
+class HashicorpVaultConfig {
```
jimmarino, line 42: Documentation? What does this class do?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultConfig.java
```diff
@@ -0,0 +1,227 @@
...
+    private boolean healthStandbyOk;
+    private Duration timeout;
+    private String token;
+    private int timeToLive;
+    private int renewBuffer;
+    private String secretPath;
+
+    private HashicorpVaultConfig() {}
+
+    public static HashicorpVaultConfig create(ServiceExtensionContext context) {
```
jimmarino, line 56: Please refactor this into a method that is used by the extension. Passing a context to a static that uses a builder like this is convoluted and pollutes it with extension concepts.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/model/TokenLookUpResponsePayload.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ */
+
+package org.eclipse.edc.vault.hashicorp.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
jimmarino, line 21: Documentation? What does this class do?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/model/TokenLookUpResponsePayloadToken.java
```diff
@@ -0,0 +1,111 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSetter;
+import com.fasterxml.jackson.annotation.Nulls;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
jimmarino, line 25: Documentation? What does this class do?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/model/TokenRenewalRequestPayload.java
```diff
@@ -0,0 +1,50 @@
...
+ *       Mercedes-Benz Tech Innovation GmbH - Add token rotation mechanism
+ *
+ */
+
+package org.eclipse.edc.vault.hashicorp.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+public class TokenRenewalRequestPayload {
```
jimmarino, line 20: Documentation? What does this class do?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/model/TokenRenewalResponsePayload.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSetter;
+import com.fasterxml.jackson.annotation.Nulls;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
jimmarino, line 25: Documentation? What does this class do?

jimmarino, line 25: Why ignore unknown?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/model/TokenRenewalRequestPayload.java
jimmarino: Serialization/deserialization must have unit tests

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/model/TokenRenewalResponsePayload.java
jimmarino: Serialization/deserialization must have unit tests

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/model/TokenRenewalResponsePayloadToken.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ */
+
+package org.eclipse.edc.vault.hashicorp.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+
+public class TokenRenewalResponsePayloadToken {
```
jimmarino, line 21: Documentation?

jimmarino: Serialization/deserialization must have unit tests

jimmarino (changes requested):
Submitting an initial round of change requests. Once these are cleaned up, I'll take a closer look.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
```diff
@@ -18,31 +19,40 @@
...
     private TypeManager typeManager;
 
+    @Inject
+    private ExecutorInstrumentation executorInstrumentation;
+
     private Vault vault;
     private PrivateKeyResolver privateKeyResolver;
+    private ScheduledExecutorService scheduledExecutorService;
+    private HashicorpVaultClient hashicorpVaultClient;
+    private boolean isTokenRenewable;
```
paullatzelsperger, line 55: this is information that the extension does not need. it should be handled inside the vault client.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
```diff
@@ -59,15 +69,54 @@ public PrivateKeyResolver privateKeyResolver() {
...
+            hashicorpVaultClient.scheduleNextTokenRenewal(token.getTimeToLive());
+        }
+    }
+
+    @Override
+    public void shutdown() {
+        scheduledExecutorService.shutdownNow();
     }
 
+    private void initHashicorpVaultClient(HashicorpVaultConfig config, Monitor monitor) {
```
paullatzelsperger, line 107: nit: i don't like void-returning init methods too much. why not return the vault client and rename the method `createHashicorpVaultClient`?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
```diff
@@ -59,15 +69,54 @@ public PrivateKeyResolver privateKeyResolver() {
...
+        hashicorpVaultClient = new HashicorpVaultClient(config, httpClient, typeManager.getMapper(), scheduledExecutorService, monitor);
+
+        var tokenLookUpResult = hashicorpVaultClient.lookUpToken();
+
+        if (tokenLookUpResult.failed()) {
+            throw new EdcException("[%s] Initial token look up failed: %s".formatted(NAME, tokenLookUpResult.getFailureDetail()));
```
paullatzelsperger, line 113: do **not** throw an exception here as it will cause a hard crash in the runtime and prevent the connector to come up. even if the initial token lookup failed, it could be a transient error (e.g. the vault pod is not (yet) reachable in K8S) that your implementation should be able to handle.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
```diff
@@ -59,15 +69,54 @@ public PrivateKeyResolver privateKeyResolver() {
...
+        }
+
+        monitor.info("Token is valid");
```
paullatzelsperger, line 116: spurious: what's the point of this?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthCheck.java
```diff
@@ -40,51 +38,62 @@ public HashicorpVaultHealthCheck(HashicorpVaultClient client, Monitor monitor) {
...
-        } catch (EdcException e) {  // can be thrown by the client, e.g. on JSON parsing error, etc.
-            var exceptionMsg = format(HEALTH_CHECK_ERROR_TEMPLATE, "EdcException: " + e.getMessage(), "");
-            monitor.severe(exceptionMsg, e);
-            return HealthCheckResult.failed(exceptionMsg).forComponent(COMPONENT_NAME);
+        var healthCheckResponseResult = client.doHealthCheck();
+        if (healthCheckResponseResult.failed()) {
+            monitor.severe(HEALTH_CHECK_ERROR);
+            return HealthCheckResult
+                    .failed(healthCheckResponseResult.getFailureMessages())
+                    .forComponent(HashicorpVaultExtension.NAME);
```
paullatzelsperger, line 46: not the same thing. The extension's name should be separate from the health component.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthCheck.java
```diff
@@ -40,51 +38,62 @@ public HashicorpVaultHealthCheck(HashicorpVaultClient client, Monitor monitor) {
...
         }
 
+        var errors = new ArrayList<String>(0);
```
paullatzelsperger, line 49: spurious: no point in initializing with an initial cap of 0.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthCheck.java
```diff
@@ -40,51 +38,62 @@ public HashicorpVaultHealthCheck(HashicorpVaultClient client, Monitor monitor) {
...
-                var sealedMsg = format(HEALTH_CHECK_ERROR_TEMPLATE, "Vault is sealed", response.getPayload());
+                var sealedMsg = HEALTH_CHECK_UNSUCCESSFUL_TEMPLATE.formatted("Vault is sealed", response.getPayload());
                 monitor.warning(sealedMsg);
-                result = HealthCheckResult.failed(sealedMsg);
+                errors.add(sealedMsg);
             }
             default -> {
-                var unspecifiedMsg = format(HEALTH_CHECK_ERROR_TEMPLATE, "Unspecified response from vault. Code: " + response.getCode(), response.getPayload());
+                var unspecifiedMsg = HEALTH_CHECK_UNSUCCESSFUL_TEMPLATE.formatted("Unspecified response from vault. Code: " + response.getCode(), response.getPayload());
                 monitor.warning(unspecifiedMsg);
```
paullatzelsperger, line 83: maybe all those monitor invocations could be handled in _one_ single line after the switch statement?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthCheck.java
```diff
@@ -40,51 +38,62 @@ public HashicorpVaultHealthCheck(HashicorpVaultClient client, Monitor monitor) {
...
+        var tokenLookUpResult = client.lookUpToken();
+
+        if (tokenLookUpResult.failed()) {
+            var tokenNotValidMsg = "Token look up failed: %s".formatted(tokenLookUpResult.getFailureDetail());
+            monitor.warning(tokenNotValidMsg);
+            errors.add(tokenNotValidMsg);
+        }
+
+        var result = errors.isEmpty() ? HealthCheckResult.success() : HealthCheckResult.failed(errors);
+        return result.forComponent(HashicorpVaultExtension.NAME);
```
paullatzelsperger, line 97: see above

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -20,217 +21,392 @@
...
-        // invoke
-        var result = vaultClient.getSecretValue(KEY);
-
-        // verify
-        assertNotNull(result);
-        verify(edcClientMock).execute(argThat(request -> request.method().equalsIgnoreCase("GET") &&
-                request.url().encodedPath().contains(CUSTOM_SECRET_PATH + "/data") &&
-                request.url().encodedPathSegments().contains(KEY)));
+    void beforeEach() {
+        reset(monitor);
```
paullatzelsperger, line 89: why do you need to reset the mock? it's not a static variable and gets initialized on every test

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -20,217 +21,392 @@
...
+            // prepare
+            var response = mock(Response.class);
+            var body = mock(ResponseBody.class);
+            when(httpClient.execute(any(Request.class))).thenReturn(response);
+            when(response.code()).thenReturn(200);
+            when(response.isSuccessful()).thenReturn(true);
+            when(response.body()).thenReturn(body);
+            when(body.string())
+                    .thenReturn(
+                            "{ " +
```
paullatzelsperger, line 105: nit: use text blocks (`"""`) to increase readability

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -20,217 +21,392 @@
...
+                                    "\"performance_standby\": false," +
+                                    "\"replication_performance_mode\": \"mode\"," +
+                                    "\"replication_dr_mode\": \"mode\"," +
+                                    "\"server_time_utc\": 100," +
+                                    "\"version\": \"1.0.0\"," +
+                                    "\"cluster_name\": \"name\"," +
+                                    "\"cluster_id\": \"id\" " +
+                                    " }");
+
+            var healthCheckResponseResult = vaultClient.doHealthCheck();
```
paullatzelsperger, line 118: should assert that the result was successful first.

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -20,217 +21,392 @@
...
+                    HealthCheckResponse.HashicorpVaultHealthResponseCode
+                            .INITIALIZED_UNSEALED_AND_ACTIVE,
+                    healthCheckResponse.getCodeAsEnum());
+
+            var healthCheckResponsePayload = healthCheckResponse.getPayload();
+
+            assertNotNull(healthCheckResponsePayload);
+            Assertions.assertTrue(healthCheckResponsePayload.isInitialized());
+            Assertions.assertFalse(healthCheckResponsePayload.isSealed());
+            Assertions.assertFalse(healthCheckResponsePayload.isStandby());
```
paullatzelsperger, line 141: static imports and AssertJ assertions please.

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -20,217 +21,392 @@
...
-
-        // verify
-        assertNotNull(result);
-        verify(edcClientMock).execute(argThat(request -> request.method().equalsIgnoreCase("POST") &&
-                request.url().encodedPath().contains(CUSTOM_SECRET_PATH + "/data") &&
-                request.url().encodedPathSegments().contains(KEY)));
+    @Nested
+    class HealthCheck {
+        @Test
+        void returnsSuccessfulResult_whenHealthCheckCouldBePerformed() throws IOException {
```
paullatzelsperger, line 95: please read our guidelines on how to name test methods: https://github.com/eclipse-edc/docs/blob/main/developer/contributing/coding-principles.md#v-testing

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -20,217 +21,392 @@
...
+                                            "{" +
+                                            "\"lease_duration\": 100" +
+                                            "}" +
+                                            "}");
+                    var captor = ArgumentCaptor.forClass(Runnable.class);
+                    // break second token renewal call
+                    doNothing().when(vaultClientSpy).scheduleNextTokenRenewal(100L);
+
+                    vaultClientSpy.scheduleNextTokenRenewal(500L);
+                    verify(scheduledExecutorService).schedule(captor.capture(), eq(500L), eq(TimeUnit.SECONDS));
```
paullatzelsperger, line 309: what's going on here? please use Awaitility and a small timeout for testing?

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -20,217 +21,392 @@
...
-        assertThat(result.succeeded()).isTrue();
-        verify(edcClientMock).execute(argThat(request -> request.method().equalsIgnoreCase("DELETE") &&
-                        request.url().encodedPath().contains(CUSTOM_SECRET_PATH + "/metadata")
-                /*request.url().encodedPathSegments().contains(KEY)*/));
+    @Nested
+    class Secret {
+        @Test
+        void getSecretValue() throws IOException {
+            // prepare
+            var response = mock(Response.class);
```
paullatzelsperger, line 347: please do not mock value objects. They can be constructed relatively easily.

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/system/health/HealthCheckResult.java
```diff
@@ -43,6 +44,11 @@ public static HealthCheckResult failed(String... errors) {
         return new HealthCheckResult(false, new Failure(errorList));
     }
 
+    public static HealthCheckResult failed(List<String> errors) {
```
paullatzelsperger, line 47: this seems fishy: why can this list ever contain `null` elements?

paullatzelsperger (changes requested):
I see some major design and implementation flaws with this PR. In addition to the inline comments in the code:
- please don't use package-private classes and methods unless there is a compelling reason to do so. Especially if methods are only package-private so they can be used from tests is big design smell and should be avoided.
- testing: method names. please read our [guidelines](https://github.com/eclipse-edc/docs/blob/main/developer/contributing/coding-principles.md#v-testing) re test method names
testing: from what I can tell, argument captors are used wrongly. async tests should be done using Awaitility
- testing: do not mock value objects, such as REsponses. there is an relatively easy way to [build them](https://stackoverflow.com/a/56044698)

There are a few other smelly areas that I didn't even get into yet, and if I'm honest I'm not sure this PR is salvageable at this point.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthCheck.java
hamidonos: The list is empty in 99.9% of cases (happy path) so why would you need a bigger list from the get go if java can just resize the list if needed ?

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
hamidonos: if I don't reset it the verifications will fail

core/common/junit/src/main/java/org/eclipse/edc/junit/extensions/DependencyInjectionExtension.java
```diff
@@ -41,20 +42,19 @@
...
-        factory = new ReflectiveObjectFactory(
-                new InjectorImpl(Mockito::mock),
-                new InjectionPointScanner(),
-                context
-        );
+        createContext();
+    }
+
+    @Override
+    public void beforeAll(ExtensionContext extensionContext) {
```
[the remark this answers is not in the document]
hamidonos: because initializing the context before each test is not necessary in my opinion

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/system/health/HealthCheckResult.java
hamidonos: You have no control of what downstream projects will call this method with so better check for null

core/common/junit/src/main/java/org/eclipse/edc/junit/extensions/DependencyInjectionExtension.java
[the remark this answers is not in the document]
jimmarino: Please do not change things in a PR that are not specifically related to the task at hand.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
hamidonos: What do you propose?
 
- Retry X times and crash ultimatelly?
- Don't crash at all and still schedule next renew?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthCheck.java
jimmarino: This will have a negligible impact on runtime resource consumption, particularly since it will be dereferenced quickly. Please keep with the coding style we currently have, it's easier to read.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/model/TokenRenewalResponsePayload.java
hamidonos: I didn't want to serialize unnecessary properties from the API Response to 
a) reduce the performance impact
b) make the implementation less dependent on the api

If you want I can serialize all properties from the response but then I will also have to extend the tests

hamidonos:
@jimmarino @paullatzelsperger thanks for the thorough review
I've fixed many of your findings and submitted a new version 

@paullatzelsperger why do you want to expose classes outside of the package if not needed?

Also the guidelines say this:
`10. Use package-private classes if they're not needed outside the package, e.g. implementation classes`

Can you give me an example where I've made something package private just for the sake of testing?

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
hamidonos: Shouldn't the extension take care of initalizing/starting the client properly since it also has to take care of shutting it down properly?

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
hamidonos: I'm not sure if Awaitility is the correct choice for this particular unit test

With Awaitility I would poll for a state. What is the state I can poll here? I want to verify invocations on my mocks.

Furthermore this will introduce real threads which in turn will make this an integration test since I have to test the scheduled executor service in combination with my `runnable`. With this I will also have to think about race conditions between tests etc.

While I agree with you that captors usually have different use cases I would still like to go with this solutions as it's the most elegant solution

Other options could be:
- Extract the `runnable` in a separate package-private method or nested class and test it separately. But this will go against the guidelines since it's a package-private code for testing only
- Extract the `runnable` in a separate public class. This doesn't feel right as well

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/model/TokenRenewalResponsePayload.java
jimmarino: There will be negligiable performance and memory impact, particularly given the context in which these operations are happening (Jackson, HTTP stack, remote call to a REST interface, non-critical code path, etc.). 

Instead of using Jackson to deserialize this to a POJO, please put it in a map and pull the values from the keys.

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/system/health/HealthCheckResult.java
ndr-brt: looks too defensive for me as well, a null error message is just a null error message, won't hurt anyone

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
paullatzelsperger: I would getting more familiar with Awaitility. It has the `untilAsserted()` method that takes a lambda, in which you can verify a mock.

To make it clear: ArgumentCaptors have a very different use case, and you **will not** abuse them in this way. 

One way you could do this is as follows:
- the `HashicorpVaultClient` takes the executor and the schedule period in its CTor
- in the test you pass a very small timeout, eg. 5 ms
- with awaitility you wait until the mock verifies, an assertion is true, etc.

paullatzelsperger:
Several comments from my initial review were ignored. I understand that the PR is still in draft, but the comments were resolved without implementation. This is very annoying and a waste of both our time. Judging by the current state of affairs, and by the amount of comments on this PR, my recommendation would be to close it, to take all the input and to start over fresh, possibly with a design proposal first. 
When doing that, the following aspects should be taken in to consideration:

- package private classes: we typically use them when implementing an SPI interface, which is not the case here. Implementors can be package-private, but should still declare their methods `private` and `public`, simply to declare intent. It is hard/impossible for a reader to discern whether a method was intended to be called from outside the class or not. All public methods **must** be documented and tested. 
- `monitor.warning()` and above: always pass the exception instance so the stack trace does not get swallowed
- when returning failures, the exception message should be part of the failure message
- Config values should be validated early on, e.g. when reading the values from the settings, or when assembling a config object, not upon their usage. this causes repeated, weird, unexpected and unnecessary runtime errors.
- Do **NOT** ignore comments of a previous review (vault config, `HashicorpVault#getHeaders()`,...)
- Use AssertJ assertions, no `assertEquals`, `assertNotEquals`
- All serializable classes must have tests to ensure correct serialization and deserialization
- Wherever builders are used, which perform some validation in their `build()` method, then that requires testing too.
- Regarding partially parsing API responses (i.e. ignoring some properties of the responses JSON), there are several ways to do it properly:
  - If you are using only a small number of properties: deserialize the response into a `Map` and parse "manually". Good enough for simple strings, but not practical for complex or nested structures.
  - Create a POJO for the _entire_ response and just don't use some properties. This will break when the Vault changes its API response, e.g. by adding optional fields etc. Personally, I still like this, as it forces a conscientious decision. It may be impractical though, if a large number of properties are to be ignored. Also, it results in unused methods on the POJO. 
  - For the sake of completeness: in your extension, you could create an `ObjectMapper`, that is based on the injected one, and configure it with the `FAIL_ON_UNKNOWN_PROPERTIES` flag:
  ```java
  var yourMapper = typeManager.getMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
  ```
    I would **not** do this however, as it makes testing more complicated and coupled more tightly. Further, the `ObjectMapper` itself and instantiating it is a resource-hog, so we wanna be careful there. Also, I am a proponent of strict verification, i.e. a "silently" changed Vault API should not go unnoticed.

Issue #3455, opened 2023-09-14
dsp: adapt to the new `processId` attribute names


ndr-brt:
# Feature Request

On the DSP specification there was a change about the `processId` field for the `ContractNegotiation` and `TransferProcess` messags: https://github.com/International-Data-Spaces-Association/ids-specification/issues/106

We should adapt to it.

## Which Areas Would Be Affected?

DSP

## Why Is the Feature Desired?

dsp completion

PR #3717, opened 2023-12-18
feat: support DSP `consumerPid` and `providerPid` for ContractNegotiation

ndr-brt:
## What this PR changes/adds

Supports new DSP fields `consumerId` and `providerId` as specified in the [dataspace protocol specs](https://docs.internationaldataspaces.org/ids-knowledgebase/v/dataspace-protocol/contract-negotiation/contract.negotiation.protocol)

## Why it does that

dsp protocol

## Further notes

- transfer process implementation will come in a subsequent PR
- every message will contain both `consumerPid`, `providerPid` and `processId`, the latter will be the same as the `consumerPid`
- the functionality has been implemented in a way that it will maintain backward compatibility with previous versions connectors. Interoperability cases:

#### NEW consumer -> OLD provider
on contract request:
- consumer:
  - generate negotiation with new id and null correlation id
  - create and dispatch request message with:
    - consumerPid valued with its own id.
    - processId is valued with the same id because correlation id is null
- provider:
  - receives the message:
    - it is on an old version so it get the processId and it stores it as correlationId,
    - sends a message back with the `processId` valued with the correlationId
- consumer receives, fetches the negotiation by the processId in the URL that's its own id (correct)
  - it saves the received processId as correlationId (see ContractNegotiationProtocolServiceImpl)
  - on the next message, it will sent it using that correlationId.
- provider will fetch negotiation correctly because it used the `findByCorrelationId`, and that's the id it received.

#### OLD consumer  -> NEW provider
on contract request:
- consumer:
  - generate negotiation with id and correlation id of the same value
  - create and dispatch request message with:
    - processId valued with the id/correlationId value
- provider:
  - receives the message:
    - it is on a new version so it values both consumerPid and providerPid with the processId value
    - create a negotiation with its own id and the correlationId equal to the consumerPid/processId
    - sends a message back putting the correlationId as processId so it will be used for every communication
- consumer:
  - receives the message:
    - it will use the processId to fetch the negotiation
    - the next message will contain the `processId` that will be used by the provider to successfully fetch the negotiation by correlationIs (as fallback)

## Linked Issue(s)

Part of #3455

### DIFF
--- a/core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/negotiation/AbstractContractNegotiationManager.java
+++ b/core/control-plane/contract-core/src/main/java/org/eclipse/edc/connector/contract/negotiation/AbstractContractNegotiationManager.java
@@ -72,15 +72,26 @@ public abstract class AbstractContractNegotiationManager extends AbstractStateEn
      */
     @WithSpan
     protected boolean processTerminating(ContractNegotiation negotiation) {
-        var termination = ContractNegotiationTerminationMessage.Builder.newInstance()
+        var terminationBuilder = ContractNegotiationTerminationMessage.Builder.newInstance()
                 .protocol(negotiation.getProtocol())
                 .counterPartyAddress(negotiation.getCounterPartyAddress())
-                .processId(negotiation.getCorrelationId())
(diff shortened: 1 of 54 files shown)


paullatzelsperger (approved):
LGTM

PR #3727, opened 2023-12-20
feat: Introduce transfer type

### LINKED ISSUE #3724: Introducing transfer type
# Feature Request

As outlined in the DR #3637, we want introduce the concept of transfer type to remove the coupling with the `DataAddress#type`  

transfer, mgmt api, dsp

Extensibility of the `DataFlowManager` and `DataFlowController,` and  prep for #3705

## Solution Proposal

The introduction of a `transferType` will impact the following area/components:

- Catalog: already covered in #3669
- Transfer
- DataFlowControllers

### Transfer

we will introduce the `transferType` property from the MGMT api flowing through the DSP protocol in the `TransferRequestMessage#format` field.

### DataFlowControllers

For now the will remain untouched waiting for the #3705 work. We will just do some modification related to backward compatibility for making the current `DataFlowControllers` work with or without the `transferType`

wolf4ood:
## What this PR changes/adds

Introduce the transfer type concept as outlined in the DR #3637.

Adds the transfer type field in: 

- `TransferRequest`
- `DataRequest`
- `TransferRequestMessage` 
- `DataFlowRequest`

and their transformers when necessary.

The `TransferRequest#dataTestination` (DataAddress) is still mandatory, even in the case where the `transferType` could be enough for the transfer, like in the `PULL`s scenarios  (Backward compatibiliy).

The EDC `DataFlowController`s :

- `ConsumerPullTransferDataFlowController`
- `ProviderPushTransferDataFlowController`

have been slightly modified to be backward compatible in case they receive the `transferType` field.

## Why it does that

Decouple the type of transfer from the `DataAddress#type` of the destination in a transfer request

## Linked Issue(s)

Closes #3724


ndr-brt (commented):
why is `transferType` added on `DataRequest`? This value object has no particular use in the current implementation so I would add `transferType` directly on the `TransferProcess`

wolf4ood:
@ndr-brt  no particilar reason, all the DSP/ protocol related informations are stored in the `DataRequest` that's why I put that there. If i remember correctly we wanted also the merge those two. For me it's fine to put that in `TransferProcess`

ndr-brt:
> @ndr-brt no particular reason, all the DSP/ protocol related informations are stored in the `DataRequest` that's why I put that there. If i remember correctly we wanted also the merge those two. For me it's fine to put that in `TransferProcess`

yes, the merging is proceeding really slowly :grin: I think we should not add new stuff on the `DataRequest` but that should be done directly in the `TransferProcess`.
To inline the `DataRequest` we'll need to change some apis first, like the `ResourceManifestGenerator`.
I created the issue so that we won't forget: https://github.com/eclipse-edc/Connector/issues/3732

wolf4ood:
Cool make sense :)

extensions/control-plane/store/sql/transfer-process-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/transferprocess/store/schema/TransferProcessStoreStatements.java
```diff
@@ -75,6 +75,10 @@ default String getConnectorAddressColumn() {
         return "connector_address";
     }
 
+    default String getTransferTypeColumn() {
+        return "transfer_type";
```
paullatzelsperger, line 79: this needs to be added to the `TransferProcessMapping` as well, otherwise querying won't work. While doing that, I would
also suggest to add a test to query for that property in the `TransferProcessStoreTestBase`:

```java

@Test
void queryByTransferType() {
    range(0, 10).forEach(i -> getTransferProcessStore().save(createTransferProcessBuilder("test-tp-" + i)
            .transferType("type" + i)
            .build()));
    var querySpec = QuerySpec.Builder.newInstance().filter(Criterion.criterion("transferType", "=", "type4")).build();

    var result = getTransferProcessStore().findAll(querySpec);

    assertThat(result).extracting(TransferProcess::getTransferType).containsOnly("type4");
}
```

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/types/domain/transfer/DataFlowRequest.java
```diff
@@ -158,6 +165,11 @@ public Builder destinationDataAddress(DataAddress destination) {
             return this;
         }
 
+        public Builder transferType(String transferType) {
+            request.transferType = transferType;
```
paullatzelsperger, line 169: I presume this is a required property, and if so there should be a validation in the `build()` method, plus a test asserting this

wolf4ood: It's not a mandatory property for backward compatibility

extensions/control-plane/store/sql/transfer-process-store-sql/src/main/java/org/eclipse/edc/connector/store/sql/transferprocess/store/schema/TransferProcessStoreStatements.java
wolf4ood: done

extensions/control-plane/transfer/transfer-data-plane/src/main/java/org/eclipse/edc/connector/transfer/dataplane/flow/ConsumerPullTransferDataFlowController.java
```diff
@@ -39,36 +40,52 @@ public class ConsumerPullTransferDataFlowController implements DataFlowControlle
     private final DataPlaneSelectorService selectorService;
     private final ConsumerPullDataPlaneProxyResolver resolver;
 
+    private final Set<String> transferTypes = Set.of("%s-%s".formatted("Http", PULL));
```
ndr-brt, line 43: currently the http type is `HttpData`, not `Http`. I know that I wrote that code (:smiling_face_with_tear: ), but the `Push` dataflowcontroller will create `HttpData-PUSH` transer types, so I think this should be consistent.

Furthermore, this can be a string instead of a single entry Set

wolf4ood: Ok so i will change to `HttpData-PULL` ok?

wolf4ood: This is a set so we can reuse it for `transferTypesFor` method

wolf4ood: @ndr-brt  renamed

paullatzelsperger (approved):

jimmarino (approved):

PR #3754, opened 2024-01-02
refactor: simplify private key resolver

### LINKED ISSUE #3740: Token Refactor Step 1: Simplify `PrivateKeyResolver`
# Feature Request

Step 1 of the token refactoring deals with simplifying the `KeyParser`, `PrivateKeyResolver` and `KeyPairFactory`.

- the `PrivateKeyResolver` simply obtains the private key by its key-id. Implementations for the following sources should exist:
  - vault: key is held in the vault (recommended)
  - config: key is resolved from config/prop/env (not recommended, just for testing)
  - file: key is resolved from a file, the path of which is resolved through a config

- the `KeyParser` can interpret the string representation, for instance JWK, and converts it into a runtime-type, for example `ECKey`.
- remove `KeyPairFactory` (both of them)

`connector-core`, `jwt-core`, `identity-did-service`, `oauth2-core`, `data-plane-http-oauth2-core`

Simplification and streamlining of how private keys are handled/resolved in EDC.

## Solution Proposal

Parsing a private key from its String representation gets decoupled from the resolver. Implementations for the resolver would deal with
the _location_, where the private key is held (e.g. a vault), whereas the `KeyParser` is responsible for getting the key from its textual
representation (JWK, PEM) into a JCA type (`PublicKey`), possibly going via an intermediary type, such as `ECKey` (from Nimbus).

```java
interface KeyParser {
   boolean canHandle(String encoded);
   PrivateKey parse(String encoded);
}

interface KeyParserRegistry {
   void register(KeyParser parser);
   PrivateKey parse(String value);
}

@FunctionalInterface
@ExtensionPoint
public interface PrivateKeyResolver {
    @Nullable PrivateKey resolvePrivateKey(String id);
}
```

[edit 1]: initially, `KeyParsers` for the following formats will be provided out-of-the-box: PEM, PKCS#12, and JWK. **The recommended serialization format is JWK**

paullatzelsperger:
## What this PR changes/adds

Most notably, this PR changes the following aspects:
- refactor the signature of `KeyParser`
- implement `PemParser` and `JwkParser` implementations
- introduce a `KeyParserRegistry`
- refactor the signature of `PrivateKeyResolver`
- adapt the capabilities of `VaultPrivateKeyResolver`, so that it first checks the Vault, and if no Pk is found, falls back to the `Config`
- `VaultPrivateKeyResolver` is provided by a new extension `SecurityDefaultServicesExtension`
- rename `FsPrivateKeyResolver` -> `JksPrivateKeyResolver`

## Why it does that

## Further notes

## Linked Issue(s)

Closes #3740 


core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/security/keyparsers/PemParser.java
```diff
@@ -0,0 +1,152 @@
...
+import java.security.PrivateKey;
+import java.security.interfaces.RSAPrivateCrtKey;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.RSAPublicKeySpec;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.regex.Pattern;
+
+public class PemParser implements KeyParser {
```
jimmarino, line 43: doc

jimmarino (approved):

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/security/keyparsers/PemParser.java
paullatzelsperger: it's there, I think your review and my latest commit had a race condition

core/common/connector-core/src/test/java/org/eclipse/edc/connector/core/security/keyparsers/JwkParserTest.java
```diff
@@ -0,0 +1,78 @@
...
+import static org.mockito.Mockito.mock;
+
+class JwkParserTest {
+
+    private final JwkParser parser = new JwkParser(new ObjectMapper(), mock());
+
+    @Test
+    void canHandle() {
+
+    }
```
wolf4ood, line 44: Remove or implement?

wolf4ood (approved):

Issue #3742, opened 2023-12-28
Token Refactor Step 2.2: introduce security context


paullatzelsperger:
# Feature Request

Step 2.2 of the token refactoring deals with the refactoring of the `TokenValidationService`. Different types of tokens require different verification and validation rules. Furthermore, resolving the public key with which to verify the token could depend on the calling code: some public keys are resolved from a DID, some are resolve from a PKI, others may be configured in the settings etc.

## Which Areas Would Be Affected?

`jwt-spi`, `oauth2-core`

## Why Is the Feature Desired?

the same type (as in: Java type) of token may need be validated and verified differently depending on the calling code. 

## Solution Proposal

```java
public interface TokenValidationService {
    Result<ClaimToken> validate(TokenRepresentation representation, PublicKeyResolver resolver, TokenValidationRule... rules);
    Result<ClaimToken> validate(@NotNull String token, PublicKeyResolver resolver, TokenValidationRule... rules);
}
```

PR #3758, opened 2024-01-04
Refactor/3757 improve publickeyresolver

### LINKED ISSUE #3757: Token Refactor Step 2.4: align the `PublicKeyResolver` with the `PrivateKeyResolver`
# Feature Request

Like the `PrivateKeyResolver`, we need to decouple _resolving_ a key from _parsing_ the key. Thus, the `PublicKeyResolver` will have an (abstract) implementation, that delegates the actual resolution down to subclasses, but handles the parsing (using the `KeyParserRegistry`).

`TokenValidationService`

symmetry in handling key material

## Solution Proposal

Out-of-the-box, we will provide key resolvers for:
- DID: public key is obtained through the DIDResolver
- URL: public key is obtained from a URL (cf. Oauth2, OIDC's `jwks_uri`)

paullatzelsperger:
## What this PR changes/adds

This PR makes the `PublicKeyResolver` symmetric with the `PrivateKeyResolver` by implementing an `AbstractPublicKeyResolver` that uses the same `KeyParser[Registry]` infrastructure. Only the actual resolution of the key material is delegated down to implementors.

In addition, the `DidPublicKeyResolver` now extends the `AbstractPublicKeyResolver` base class.

To achieve that, the `KeyParsers` were improved so that they can handle public _and_ private keys alike.

## Why it does that

Preparatory work necessary for issue #3742

## Further notes

## Linked Issue(s)

Closes #3757


wolf4ood (approved):
LGTM

Issue #3744, opened 2023-12-28
Token Refactor Step 2.3: registries for validation rules and decorators


paullatzelsperger:
# Feature Request

Step 2.3 of the token refactor deals with harmonizing the way how token decorators and validation rules are handled.
To that end, the notion of a _token context_ is introduced. This allows us to register decorators and validation rules specifically for
a particular token context, and extend them through extension points.

For example, some downstream projects may require additional validation checks when handling tokens of the `"oauth2"` context.

Note that while the technical type of some tokens may be the same, i.e. JWT, the "semantic" type, is different. OAuth2 tokens have a different
set of decorators and rules than Self-Issued ID Tokens or other access tokens. The term "token context" refers to that "semantic" type.

## Which Areas Would Be Affected?

dsp

## Why Is the Feature Desired?

Harmonized, symmetrical and extensible way to handle both token decoration (adding claims) and validation (asserting claims)

## Solution Proposal

```java
public interface TokenDecoratorRegistry {
    void register(String context, TokenDecorator decorator);
    void unregister(String context, TokenDecorator decorator);
    List<TokenDecorator> decoratorsFor(String context);
}
```

```java
public interface TokenValidationRulesRegistry {
    void addRule(String context, TokenValidationRule rule);
    List<TokenValidationRule> rulesFor(String context);
}
```

the `context` should be an easily identifiable string, such as `"oauth2"` or `"siop"`.

> Note that token contexts **MUST** be documented by the extension that introduces them!

PR #3760, opened 2024-01-08
refactor: add context to TokenValidationService

### LINKED ISSUE #3742: Token Refactor Step 2.2: introduce security context

### LINKED ISSUE #3757: Token Refactor Step 2.4: align the `PublicKeyResolver` with the `PrivateKeyResolver`

paullatzelsperger:
## What this PR changes/adds

This PR introduces the notion of a "context" to the `TokenValidationService` and the `TokenValidationRuleRegistry`. That way, there is now only _one central instance_ of each.
Various `TokenValidationRules` get registered for their respective context, e.g. `"oauth2"`.

Impls for both are registered by the `token-core` module (formerly called `jwt-core`).

## Why it does that

harmonization of token validation

## BREAKING CHANGES
- `jwt-spi` was split up in `jwt-spi` and `token-spi`. **THIS IS A BREAKING CHANGE!**
- the module `jwt-core` was renamed to `token-core`. **THIS IS A BREAKING CHANGE!**
- `JwtDecorator` was renamed `TokenDecorator`

## Further notes

- much of the changeset stems from that renaming and the accompanying package change
- the use of `JwtRegisteredClaimNames` was mostly replaced with the constants from the Nimbus library
- upgrades to Gradle 8.5 due to some weird class file version errors
 
## Linked Issue(s)

Closes #3742 
Contributes to #3744


extensions/common/iam/identity-trust/identity-trust-sts/identity-trust-sts-embedded/src/main/java/org/eclipse/edc/iam/identitytrust/sts/embedded/SelfIssuedTokenDecorator.java
```diff
@@ -45,7 +44,7 @@ class SelfIssuedTokenDecorator implements JwtDecorator {
     @Override
     public Map<String, Object> claims() {
         var claims = new HashMap<String, Object>(this.claims);
-        claims.put(ISSUED_AT, Date.from(clock.instant()));
+        claims.put("iat", Date.from(clock.instant()));
```
bscholtes1A, line 47: Shouldnt this remain a constant?

bscholtes1A (approved):

extensions/common/iam/identity-trust/identity-trust-sts/identity-trust-sts-embedded/src/main/java/org/eclipse/edc/iam/identitytrust/sts/embedded/SelfIssuedTokenDecorator.java
paullatzelsperger: absolutely. IDK why that changed -> using `JWTClaimNames` now.

Issue #3743, opened 2023-12-28
Story: Refactoring of token handling in EDC


paullatzelsperger:
# Feature Request

this is a story issue that tracks all necessary steps outlined in [decision-record 2023-12-23](https://github.com/eclipse-edc/Connector/blob/main/docs/developer/decision-records/2023-12-19-token-handling-refactor/README.md).

## Which Areas Would Be Affected?

wide areas of the EDC code base, plus downstream EDC projects

## Why Is the Feature Desired?

token handling is currently all over the place, and it needs to be streamlined and unified.

## Solution Proposal

Implement the following issues:
- [x] #3740 
- [x] #3741 
- [x] #3742 
- [x] #3744 
- [x] #3757 
- [x] #3745 
- [x] #3746 
- [ ] #3747 
- [x] #3748 

All PRs should be merged into a temporary branch `refactor/token-handling` (https://github.com/eclipse-edc/Connector/tree/refactor/token-handling) to avoid constant build breakages and then merged into `main` in one fell swoop.

xdxxxdx:
Hello @paullatzelsperger 
Thanks so much for your reply, what does IATP stand for?

PR #3770, opened 2024-01-11
feat: add KeyPair-to-JWK conversion

paullatzelsperger:
## What this PR changes/adds

This PR adds a conversion method from `KeyPair` (JDK) to `JWK` (Nimbus lib) for easy use downstream.

## Why it does that

In some cases, such as when signing VerifiablePresentations, we may need the actual JWK representation, and Nimbus doesn't have 
good support for converting.

## Further notes

- the JDK does not have good support for Elliptic Curve multiplication, which we need to regenerate the PublicKey from a PrivateKey. BouncyCastle does, so I added it 
as implementation dependency. 
- some dependencies of the `identity-trust-service` module got cleaned up
- renamed the `JwsSignerVerifierFactory` to `CryptoConverter`

## Open questions

- ~~Do we want to rename the `JwsSignerVerifierFactory`? The name seems kinda weird, but I can't think of a better one.~~ -> now named `CryptConverter`
  [edit]: ~~maybe `NimbusHelper`?~~
- ~~Is the `jwt-spi` package the best place for it? probably some crypto lib module would be more suitable.~~ moved to `crypto-core`

## Linked Issue(s)

Contributes to #3743


core/common/token-core/src/main/java/org/eclipse/edc/token/JwtGenerationService.java
```diff
@@ -37,11 +37,11 @@
 import java.util.function.Supplier;
 
 public class JwtGenerationService implements TokenGenerationService {
-    private final JwsSignerVerifierFactory factory;
+    private final CryptoConverter factory;
```
ndr-brt, line 40:
```suggestion
    private final CryptoConverter converter;
```

core/common/token-core/build.gradle.kts
```diff
@@ -18,8 +18,10 @@ plugins {
 }
 
 dependencies {
+    api(project(":spi:common:token-spi"))
     api(project(":spi:common:jwt-spi"))
 
+    implementation(project(":extensions:common:crypto:crypto-core")) // for the CryptoConverter
```
ndr-brt, line 24: given that `crypto-core` is not an extension but a library, it could stay in the `core` folder

spi/common/oauth2-spi/build.gradle.kts
```diff
@@ -18,6 +18,7 @@ plugins {
 
 dependencies {
     api(project(":spi:common:jwt-spi"))
+    implementation(libs.nimbus.jwt)
```
ndr-brt, line 21: it should be not necessary

extensions/common/crypto/crypto-core/src/main/java/org/eclipse/edc/security/token/jwt/CryptoConverter.java
```diff
@@ -131,16 +178,97 @@ public JWSAlgorithm getRecommendedAlgorithm(JWSSigner signer) {
...
+        for (var i = 0; i < array.length / 2; i++) {
             var temp = array[i];
             array[i] = array[array.length - 1 - i];
             array[array.length - 1 - i] = temp;
         }
         return array;
     }
 
-    private Ed25519Verifier createOctetKeyPair(PublicKey publicKey) throws JOSEException {
+    private Ed25519Verifier createEdDsaSigner(PublicKey publicKey) throws JOSEException {
```
ndr-brt, line 271:
```suggestion
    private Ed25519Verifier createEdDsaVerifier(PublicKey publicKey) throws JOSEException {
```

extensions/common/crypto/crypto-core/src/main/java/org/eclipse/edc/security/token/jwt/CryptoConverter.java
```diff
@@ -131,16 +178,97 @@ public JWSAlgorithm getRecommendedAlgorithm(JWSSigner signer) {
...
         var edKey = (EdECPublicKey) publicKey;
         var curveName = edKey.getParams().getName();
```
ndr-brt, line 273: not part of this PR, but I noticed that the process to get the `Curve` from the `EdECKey` (both public and private) is the same and it could be extracted in a common method

paullatzelsperger: easy fix. will do.

core/common/token-core/build.gradle.kts
ndr-brt: currently `core` contains also library modules (as `util`), anyway I'm ok for leaving it between the extensions as well

ndr-brt (approved):

PR #3771, opened 2024-01-11
refactor: get rid of DspHttpDispatcherDelegate

### LINKED ISSUE #3344: get rid of `DspHttpDispatcherDelegate`
# Feature Request

In the DSP dispatching components we have this "delegate" that takes care of both creating the request and handling the response for the messages through the DSP wire, other than the evident SRP violation, it is designed using abstractions, and this brought in issues in tests (that have an hierarchy themselves) but also it's not easy to change to handle "exceptional cases" (see dataset request in #3342).

A first refactoring work has already been done in #3342 extracting the "create request" behavior, but also the "handle response" could be modeled in a different way, using composition over abstraction.

dsp dispatch

refactoring

ndr-brt:
## What this PR changes/adds

Get rid of DspHttpDispatcherDelegate and subclasses, preferring composition over inheritance.

## Why it does that

cleanup

## Further notes

- deprecated two `EdcHttpClient.executeAsync` methods as they are not really needed, in fact have the mapping function passed  to the method makes really hard to test, plus having an overarching `response` closure is not the best approach because if the body is null that will end up with an exception. (is the `ResponseBody` that needs to be closed, please refer to the way it was handled in the `DspHttpRemoteMessageDispatcherImpl.handleResponse` method.

## Linked Issue(s)

Closes #3344


data-protocols/dsp/dsp-http-core/src/main/java/org/eclipse/edc/protocol/dsp/dispatcher/DspHttpRemoteMessageDispatcherImpl.java
```diff
@@ -108,12 +115,44 @@ public <M extends RemoteMessage> void registerPolicyScope(Class<M> messageClass,
...
+
+    private String asString(ResponseBody it) {
+        try {
+            return it.string();
+        } catch (IOException e) {
+            return "Cannot read response body: " + e.getMessage();
+        }
+    }
+
+    private record MessageHandler<M extends RemoteMessage, R>(DspHttpRequestFactory<M> requestFactory,
```
jimmarino, line 150: formatting

data-protocols/dsp/dsp-http-core/src/main/java/org/eclipse/edc/protocol/dsp/dispatcher/DspHttpRemoteMessageDispatcherImpl.java
```diff
@@ -108,12 +115,44 @@ public <M extends RemoteMessage> void registerPolicyScope(Class<M> messageClass,
...
+                                                              DspHttpResponseBodyExtractor<R> bodyExtractor) {
+    }
+
+    private record PolicyScope<M extends RemoteMessage>(Class<M> messageClass, String scope,
```
jimmarino, line 154: formatting

jimmarino (approved):
two minor formatting comments

PR #3777, opened 2024-01-15
feat: support end-to-end idempotency for process state replication

### LINKED ISSUE #2993: Support end-to-end idempotency for process state replication
The runtime must support end-to-end idempotency for process state replication. Specifically, when process state changes and a 'ProcessRemoteMessage' is dispatched, the receiving end must be able to perform message de-duplication. This means the message id must remain constant for the state change. 

The provider must guarantee that the same message id is used when dispatching messages representing a state change. This will likely entail the message id being persisted since message send retries must survive runtime restarts.

The client will need to perform de-duplication by checking the message id against the last one it received.

ndr-brt:
## What this PR changes/adds

Store and evaluate protocol messages ids into `ContractNegotiation` and `TransferProcess` to support end-to-end idempotency. (ref. https://github.com/eclipse-edc/Connector/issues/2993#issuecomment-1700967610):

> ### sender side
> - when it sends a message, stores the id
> - when it changes state, wipes the id stored
> - when it tries to send a message, if the id store is set, it gets set on the message, otherwise a new id will be created

> ### receiver side
> - stores a list of already received id
> - when a message is received, checks if the message has been already received, if so, it discards it
> - when the entity reaches a "final" state (like `COMPLETED` or `TERMINATED`), ignore all the messages received.

## Why it does that

.

## Further notes
- this PR provides a **breaking change** because there a new column to be added to the `ContractNegotiation` and `TransferProcess` tables called `protocol_messages` of type JSON

## Linked Issue(s)

Closes #2993 


paullatzelsperger (approved):

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/entity/ProtocolMessages.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ */
+
+package org.eclipse.edc.spi.entity;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Describe the history of the wire messages sent or received.
```
jimmarino, line 21:
```suggestion
 * Records the history of the wire messages sent or received.
```

jimmarino (approved):

wolf4ood (approved):

PR #3779, opened 2024-01-16
feat: add index for lease_id in edc_contract_negotiation, edc_edr_cache and edc_transfer_process

### LINKED ISSUE #985: Remove `contractOffer` with `contractOfferId` on `ContractOfferRequest`
# Feature Request

Currently, to start a negotiation, a complete `ContractOffer` entity is requested into the `ContractOfferRequest` object.
We should rely on the `catalog`, in particular on the `QueryEngine` to get the contract offer instance and pass the id only.

This will fix a problem existent in the `ContractNegotiationApi`, since at the moment (ref #957)
> The NegotiationInitiateRequestDto contains an offerId, while the ContractOfferRequest in which it is transformed contain a full ContractOffer object. For the moment I created it with a ContractOffer that has only the id, but's obvious that we need a better process.

## Solution Proposal
Replace `contractOffer` with `contractOfferId` in `ContractOfferRequest` and use `QueryEngine` to retrieve the full entity

## Type of Issue
improvement

aahrensmaterna:
Implement indices on lease_id foreign key colums on the tables
- edc_contract_negotiation
- edc_edr_cache
- edc_transfer_process

Closes #985


aahrensmaterna:
Pull request ended up in the wrong project, sorry for the noise.

Issue #3736, opened 2023-12-22
refactor: make `TranslationMapping` building whole SQL condition


ndr-brt:
# Feature Request

Currently the `TranslationMapping` class maps a `Criterion` into the left operand of a sql condition.
Making it build the full condition will permit to be more flexible on conditions that requires custom operators (as implementing the `contains` for the postgresql implementation. 

The idea is to follow the approach we already have in the `CriterionToPredicateConverter`

## Which Areas Would Be Affected?

sql store

## Why Is the Feature Desired?

support more operators in sql stores

## Solution Proposal

_If possible, provide a (brief!) solution proposal._

PR #3780, opened 2024-01-16
refactor: extract the translation logic from the TranslationMapping class

ndr-brt:
## What this PR changes/adds

second refactoring toward the solution of #3736 

in this PR the "translation" logic that was implemented in the `TranslationMapping` hierarchy has been extracted to a new interface `FieldTranslator` and its hierarchy, while `TranslationMapping` remains a simple tree data structure that can provide the `FieldTranslator` to the caller (we could find a better name for it, suggestions welcome)

## Why it does that

now that we have the `FieldTranslator` we can add a method there that can return a `WhereClause` directly, giving the possibility to translate the whole criteria following the specific SQL implementation

## Linked Issue(s)

Part of #3736 

### DIFF
--- a/extensions/common/sql/sql-core/src/main/java/org/eclipse/edc/sql/translation/FieldTranslator.java
+++ b/extensions/common/sql/sql-core/src/main/java/org/eclipse/edc/sql/translation/FieldTranslator.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *  Copyright (c) 2024 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
  *
  *  This program and the accompanying materials are made available under the
  *  terms of the Apache License, Version 2.0 which is available at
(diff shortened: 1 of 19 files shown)


extensions/common/sql/sql-core/src/main/java/org/eclipse/edc/sql/translation/FieldTranslator.java
```diff
@@ -0,0 +1,31 @@
+/*
+ *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
```
paullatzelsperger, line 2: seems an update to intelliJ's copyright profile is needed :)

extensions/common/sql/sql-core/src/main/java/org/eclipse/edc/sql/translation/SortFieldConverter.java
```diff
@@ -0,0 +1,29 @@
...
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.sql.translation;
+
+/**
+ * Converts a sort field in the canonical format to the sql representation.
+ */
+interface SortFieldConverter {
```
paullatzelsperger, line 20: @FunctionalInterface?

paullatzelsperger (approved):

extensions/common/sql/sql-core/src/main/java/org/eclipse/edc/sql/translation/FieldTranslator.java
ndr-brt: ouch, I thought I updated it but likely I did it in the wrong place

jimmarino (approved):

PR #3782, opened 2024-01-17
feat: implement automatic hashicorp token renewal

### LINKED ISSUE #3610: HashiCorp Vault: Add token rotation mechanism

hamidonos:
## What this PR changes/adds

This PR adds an automatic token renewal mechanism to the Hashicorp Vault extension. 

## Why it does that

Currently the EDC uses the root token which is not suitable for production stage.

## Further notes

Note that the Vault token does not rotate. It is renewed.

A token has a TTL. The TTL defines the interval in which the token has to be renewed, otherwise it will be invalidated forever. If a token expires for any reason the EDC will go into a unhealthy state. Currently there is no other way to recover from a invalid token other than restarting the EDC with a fresh token since the token is provided in the boot configurations.

Brief explanation of new configurations added:

- VAULT_TOKEN_TTL: Defines the TTL of the token which is passed to the Vault upon renewal. The TTL might not be honored by the Vault (e.g. in the case of periodic tokens). Therefore the TTL value returned by the Vault should be respected for each renewal. Vault requires a minimum TTL of 5 seconds.
- VAULT_TOKEN_RENEW_BUFFER: Buffer before a renewal is triggered to account for network issues etc. Has to be smaller than the TTL.
- VAULT_TOKEN_SCHEDULED_RENEWAL_ENABLED: Whether the automatic renewal feature added by this PR is active or not. Defaults to true. This should be only set to false for development or testing cases.

Upon start the EDC validates the token by calling the Vault API. For non valid tokens a error message will be logged and the Connector will enter a unhealthy state. If the token is renewable it will be renewed instantly. If the renewal was successful a scheduled one-shot task is submitted. The task executes after a delay defined as `TTL - VAULT_TOKEN_RENEW_BUFFER`. If the renewal was successful the next task is scheduled again at `TTL - VAULT_TOKEN_RENEW_BUFFER` and so on. Http requests are retried if they are retryable. [See Hashicorp status codes](https://developer.hashicorp.com/vault/api-docs)

For non-renewable but valid tokens the EDC will stay healthy until the token expires.

Tokens have a max TTL even if they're renewable (unless they're periodic). After the max ttl expires the token will be invalidated for ever. Further renewals will fail.
The max TTL can either be a system default config on Vault side, or a explicit max TTL set by the admin at token creation.

Periodic tokens can be renewed forever as long they're renewed within the TTL.

## Linked Issue(s)

Closes #3610

<hr/>
 
<sub>Hamid Ahmetovic<hamid.ahmetovic@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md) </sub>


extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -185,25 +347,94 @@ private HttpUrl getHealthUrl() {
...
+     *
+     * @param request        the request to execute
+     * @param timeoutSeconds timeout which will break the loop ultimately
+     * @return result of the http request
+     * @throws IOException          httpclient was not able to execute the request
+     * @throws ExecutionException   result of a retried http request could not be retrieved
+     * @throws InterruptedException execution was interrupted while waiting for the result of a retried http request
+     */
+    @NotNull
+    private Response executeWithRetry(Request request, long timeoutSeconds) throws IOException, ExecutionException, InterruptedException {
```
ndr-brt, line 410: retry is handled automatically by the `EdcHttpClient` component, you can pass a `FallbackFactory` to tell when the call needs to be retried (take a look at `FallbackFactories` for examples), please avoid this custom implementation.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
```diff
@@ -17,54 +18,117 @@
...
+        scheduledExecutorService = executorInstrumentation.instrument(Executors.newScheduledThreadPool(CORE_POOL_SIZE), NAME);
+        isScheduledTokenRenewalEnabled = context.getSetting(VAULT_TOKEN_SCHEDULED_RENEWAL_ENABLED, VAULT_TOKEN_SCHEDULED_RENEWAL_ENABLED_DEFAULT);
+    }
 
-            vault = new HashicorpVault(client, context.getMonitor());
-        }
-        return vault;
+    @Override
+    public void start() {
+        if (isScheduledTokenRenewalEnabled) hashicorpVaultClient.scheduleTokenRenewal();
```
ndr-brt, line 96: add missing brackets

ndr-brt (changes requested):
.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthExtension.java
```diff
@@ -39,31 +42,26 @@ public class HashicorpVaultHealthExtension implements ServiceExtension {
     @Inject
     private EdcHttpClient httpClient;
```
ndr-brt, line 43: not used anymore (also `TypeManager`

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
hamidonos: But, I cannot influence for how long it will be retried since those configurations are set globally for every request.

Imagine a scenario where the K8S Cluster & EDC boots with a fresh token of TTL=30 Minutes.
Now Vault takes 3 Minutes to boot for whatever reason. 
Will the default http retrypolicy be sufficient enough in this use case? The way I see it I have no control over that.

Also the Vault can also go into recovery mode while running so this does not affect start up phase only.

The EDC shouldn't enter unhealthy states because the retry setting wasn't sufficient enough.

In the descrption above I proposed to move this logic into the `EdcHttpClient` since I (also) think that it's misplaced here but that would require a different issue and PR.

ndr-brt: I cannot see a case where EDC starts and fundamental services (like Vault or database) are not already online.
I don't think we should manage hypothetical cluster issues here. 

The logic should be: is the service down/not staisfying the connector requests for some reasons? the connector will try again for some time, if after some time the external service is not able to execute what is requested, then the issue is at the network side.

Having the retry policy being configurable for every call is a different feature that needs to be discussed and tackled in a different discussion/issue/pr

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -185,25 +347,94 @@ private HttpUrl getHealthUrl() {
...
+        return new Request.Builder()
+                .url(requestUri)
+                .headers(headers)
+                .post(createRequestBody(requestBody))
+                .build();
+    }
+
+    @NotNull
+    private Headers getHeaders() {
+        var headersBuilder = new Headers.Builder().add(VAULT_REQUEST_HEADER, Boolean.toString(true));
```
denisneuling, line 383: Would rather do ```Boolean.TRUE.toString()``` or simply ```"true"```

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -9,228 +9,619 @@
...
+                    assertThat(data.getPeriod()).isEqualTo("1h");
+                    assertThat(data.getPolicies()).isEqualTo(List.of("default", "root"));
+                    assertThat(data.isRenewable()).isTrue();
+                    assertThat(data.getTtl()).isEqualTo(2589147L);
+                    assertThat(data.getType()).isEqualTo("service");
+                }
+            }
+
+            @ParameterizedTest
+            @ValueSource(ints = {400, 403, 404, 405, 412, 429, 472, 473, 500, 501, 502, 503})
```
ndr-brt, line 296: I think we're putting too much detail in here, with the `FallbackFactory` abstraction we have the possibility to split the concerns. Personally I would create a `HashicorpVaultClientFallbackFactory` that implements the retry logic when specific status are returned, it can be tested separately (please look at `EdcHttpClientImplTest.execute_fallback_shouldRetryIfStatusIsNot2xxOr4xx` as an example).

Then here that FallbackFactory will be injected and mocked: we can then simply verify its usage.

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -9,228 +9,619 @@
...
+                    verify(httpClient).execute(requestCaptor.capture(), anyList());
+                    var request = requestCaptor.getValue();
+                    var copy = Objects.requireNonNull(request.newBuilder().build());
+                    var buffer = new Buffer();
+                    Objects.requireNonNull(copy.body()).writeTo(buffer);
+                    var tokenRenewRequest = OBJECT_MAPPER.readValue(buffer.readUtf8(), TokenRenewRequest.class);
+                    // given a configured ttl of 5 this should equal "5s"
+                    assertThat(tokenRenewRequest.getIncrement()).isEqualTo("%ds".formatted(HASHICORP_VAULT_CLIENT_CONFIG_VALUES.ttl()));
+                    assertThat(tokenRenewResult.succeeded()).isTrue();
+                    assertThat(tokenRenewResult.succeeded()).isTrue();
```
ndr-brt, line 385: duplicate line.
(PS: `AbstractResultAssert` can be used here)

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthCheck.java
```diff
@@ -40,51 +45,27 @@ public HashicorpVaultHealthCheck(HashicorpVaultClient client, Monitor monitor) {
...
 
-        HealthResponse response;
-        HealthCheckResult result;
-        try {
-            response = client.getHealth();
-        } catch (EdcException e) {  // can be thrown by the client, e.g. on JSON parsing error, etc.
-            var exceptionMsg = format(HEALTH_CHECK_ERROR_TEMPLATE, "EdcException: " + e.getMessage(), "");
-            monitor.severe(exceptionMsg, e);
-            return HealthCheckResult.failed(exceptionMsg).forComponent(COMPONENT_NAME);
+        var errors = new ArrayList<String>();
```
ndr-brt, line 48: to gather errors from multiple result the `merge` method can be used. actually this method could be implemented as:

```java
        return client.doHealthCheck().merge(client.lookUpToken())
                .flatMap(result -> {
                    if (result.succeeded()) {
                        return HealthCheckResult.success();
                    } else {
                        monitor.warning("Healthcheck failed with reason(s): " + result.getFailureDetail());
                        return HealthCheckResult.failed(result.getFailureMessages());
                    }
                }).forComponent(HashicorpVaultHealthExtension.NAME);
```

ndr-brt (changes requested):

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/model/TokenLookUpData.java
```diff
@@ -0,0 +1,269 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Token data returned by the Vault token lookup operation.
+ */
+@JsonDeserialize(builder = TokenLookUpData.Builder.class)
+public class TokenLookUpData {
```
jimmarino, line 31: We keep going over the same ground as the previous PR:

1. There are no unit tests that verifiy serialization-deserialziation. Integration tests should not be relied on for this.
2. Why even have these classes (including TokenLookUpResponse) when only a few data items are accessed. It introduces a lot of boilerplate code that is brittle because it will fail if additional properties are added.

Per my comments on the previous PR, please remove these classes.

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientIntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+    void beforeEach(EdcExtension extension) throws IOException, InterruptedException {
+        assertThat(CREATION_TTL).isGreaterThan(TTL);
+        extension.setConfiguration(getConfig());
+    }
+
+    @Test
+    void lookUpToken_whenTokenNotExpired_shouldSucceed(HashicorpVaultClient hashicorpVaultClient) {
+        var tokenLookUpResult = hashicorpVaultClient.lookUpToken();
+
+        assertThat(tokenLookUpResult.succeeded()).isTrue();
```
jimmarino, line 70: The number of assertions here is a code smell. Please see my comments about using POJOs to deserialize a small number of properties. Related to this, integration tests shoould not be used to cover what can be done in unit tests.

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientIntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+                    assertThat(tokenLookUpResult.failed()).isTrue();
+                    assertThat(tokenLookUpResult.getFailureDetail()).isEqualTo("Token look up failed with status 403");
+                });
+    }
+
+    @Test
+    void renewToken_whenTokenNotExpired_shouldSucceed(HashicorpVaultClient hashicorpVaultClient) {
+        var tokenRenewResult = hashicorpVaultClient.renewToken();
+
+        assertThat(tokenRenewResult.succeeded()).isTrue();
```
jimmarino, line 115: Ditto about the number of assertions

jimmarino (changes requested):

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +24,248 @@
...
+            if (response.isSuccessful()) {
+                var responseBody = response.body();
+                if (responseBody == null) {
+                    return Result.failure("Token look up returned empty body");
+                }
+                var payload = objectMapper.readValue(responseBody.string(), TokenLookUpResponse.class);
+                return Result.success(payload);
+            } else {
+                return Result.failure("Token look up failed with status %d".formatted(response.code()));
+            }
```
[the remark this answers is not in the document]
hamidonos: I would keep it separated since the log messages differ unless you want a common method where you pass those log messages as parametersm but this feels weird. 
I'm verifying the log messages for some tests.

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -9,228 +9,400 @@
...
+                    .body(ResponseBody.create(body, MediaType.get("application/json")))
+                    .protocol(Protocol.HTTP_1_1)
+                    .request(new Request.Builder().url("http://any").build())
+                    .build();
+            when(httpClient.execute(any(Request.class), anyList())).thenReturn(response);
+
+            var tokenLookUpResult = vaultClient.lookUpToken();
+
+            verify(httpClient).execute(any(Request.class), argThat((List<FallbackFactory> factories) -> factories.get(0) instanceof HashicorpVaultClientFallbackFactory));
+            AbstractResultAssert.assertThat(tokenLookUpResult).isSucceeded().satisfies(tokenRenewResponse -> {
```
ndr-brt, line 223: nit: static import

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +25,192 @@
...
+    }
+
+    /**
+     * Attempts to look up the Vault token defined in the configurations.
+     * <p>
+     * Will retry in some error cases.
+     *
+     * @return the result of the token lookup operation
+     */
+    public Result<Map<String, Object>> lookUpToken() {
```
ndr-brt, line 119: I'm not an huge fan of these "raw" return objects, because then the caller need to know how to get the data out of it. As the only usage of this response body extracts the `isRenevable` boolean, I would make this method return a `Result<Boolean>` so the caller won't need to deserialize further

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +25,192 @@
...
+    /**
+     * Attempts to renew the Vault token with the configured ttl. Note that Vault will not honor the passed
+     * ttl (or increment) for periodic tokens. Therefore, the ttl returned by this operation should always be used
+     * for further calculations.
+     * <p>
+     * Will retry in some error cases.
+     *
+     * @return the result of the token renewal operation
+     */
+    public Result<Map<String, Object>> renewToken() {
```
ndr-brt, line 153: Same as previous, but the content here will be `Long` (TTL)

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultTokenRenewTask.java
```diff
@@ -0,0 +1,188 @@
...
+
+                var tokenRenewResponse = tokenRenewResult.getContent();
+                var parseTtlResult = parseTtl(tokenRenewResponse);
+
+                if (parseTtlResult.failed()) {
+                    monitor.warning(INITIAL_TOKEN_RENEW_ERR_MSG_FORMAT.formatted(parseTtlResult.getFailureDetail()));
+                    return;
+                }
+
+                scheduleNextTokenRenewal(parseTtlResult.getContent());
```
ndr-brt, line 122: this code is the same as L135:L153, please extract a common method

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultTokenRenewTask.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    private Object getValueFromMap(Map<String, Object> map, String key) {
+        var value = map.get(key);
+        if (value == null) {
+            throw new IllegalArgumentException("Key '%s' does not exist".formatted(key));
+        }
+        return value;
+    }
```
ndr-brt, line 187: this logic should stay wrapped in the client

ndr-brt (changes requested):

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientFallbackFactoryTest.java
```diff
@@ -0,0 +1,38 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.mockito.Mockito.mockStatic;
+
+class HashicorpVaultClientFallbackFactoryTest {
+
+    private static final int[] NON_RETRYABLE_STATUS_CODES = {200, 204, 400, 403, 404, 405};
+
+    @Test
+    void create_shouldInitializeWithCorrectStatusCodes() {
```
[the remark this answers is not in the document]
hamidonos: I'm curious why you'd want to test the implementation of the httpClient.execute method inside the Hashicorp Vault Extension. As a caller, would you not expect it to work when you pass the Fallback Factory?

This ensures the correct set of http status codes. The usage of this fallback factory is verified inside the HashicorpVaultClient. I wouldn't go beyond that.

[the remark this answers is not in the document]
ndr-brt: This test is actually not verifying a behavior (that's already tested in the http client), but some magic values (the status code list).
So you won't be able to separate the values from the behavior (what's done in the httpclient test).
The alternative is to test this in an integration level, but it would be too effort for little value.

As I said I would delete this, but I could also get along, it's just philosophy of unit-testing :)

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultTokenRenewTask.java
```diff
@@ -0,0 +1,131 @@
...
+
+    private void initialize() {
+        var tokenLookUpResult = client.lookUpToken();
+
+        if (tokenLookUpResult.succeeded()) {
+            var isRenewable = tokenLookUpResult.getContent();
+
+            if (isRenewable) {
+                renewToken(INITIAL_TOKEN_RENEW_ERR_MSG_FORMAT);
```
ndr-brt, line 94: shouldn't this schedule the renewal instead of calling it immediately?

hamidonos: I believe the EDC should instantly look up and renew the token

1) you don't know the actual ttl of the token (the token could be really old already) -> I wouldn't want to schedule a renewal with a fictional ttl which might not fit the actual ttl. You could go around this by scheduling with 0s but then the log message will spew out that something went wrong with the scheduling instead of the "initialization" logic.
2) The token might be invalid from the start and you'd want to know that instantly
3) It was also requested this way & I think it makes sense to do so (or at the very least it doesn't hurt)

ndr-brt: sounds ok to me but that's not what's stated in the PR description:
>  If the token is renewable a scheduled one-shot task is submitted. The task executes after a delay defined as TTL - VAULT_TOKEN_RENEW_BUFFER

hamidonos: updated the text

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +25,206 @@
...
+                .build();
+        var request = httpGet(uri);
+
+        try (var response = httpClient.execute(request, FALLBACK_FACTORIES)) {
+            if (response.isSuccessful()) {
+                var responseBody = response.body();
+                if (responseBody == null) {
+                    return Result.failure("Token look up returned empty body");
+                }
+                var payload = objectMapper.readValue(responseBody.string(), new TypeReference<Map<String, Object>>() {});
```
jimmarino, line 136: TypeReference can be made a static constant

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +25,206 @@
...
+                if (parseRenewableResult.failed()) {
+                    return Result.failure("Token look up response could not be parsed: %s".formatted(parseRenewableResult.getFailureDetail()));
+                }
+                var isRenewable = parseRenewableResult.getContent();
+                return Result.success(isRenewable);
+            } else {
+                return Result.failure("Token look up failed with status %d".formatted(response.code()));
+            }
+        } catch (IOException e) {
+            var errMsg = "Failed to look up token with reason: %s".formatted(e.getMessage());
```
jimmarino, line 147: If thee error is reported back, there is no need to send it to the monitor here as that can be done by the client

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +25,206 @@
...
+                if (parseTtlResult.failed()) {
+                    return Result.failure("Token renew response could not be parsed: %s".formatted(parseTtlResult.getFailureDetail()));
+                }
+                var ttl = parseTtlResult.getContent();
+                return Result.success(ttl);
+            } else {
+                return Result.failure("Token renew failed with status: %d".formatted(response.code()));
+            }
+        } catch (IOException e) {
+            var errMsg = "Failed to renew token with reason: %s".formatted(e.getMessage());
```
jimmarino, line 190: Ditto above

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +25,206 @@
...
                 if (response.code() == HTTP_CODE_404) {
-                    return Result.failure(
-                            String.format(CALL_UNSUCCESSFUL_ERROR_TEMPLATE, "Secret not found"));
+                    return Result.failure("Secret %s not found".formatted(key));
                 }
 
                 var responseBody = response.body();
                 if (responseBody == null) {
-                    return Result.failure(String.format(CALL_UNSUCCESSFUL_ERROR_TEMPLATE, "Response body empty"));
+                    return Result.failure("Secret %s response body is empty".formatted(key));
```
jimmarino, line 209: The secret key should not be recorded. The client knows which key it asked for and can handle it appropriately.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +25,206 @@
...
-        var request = new Request.Builder().url(requestUri).headers(headers).get().build();
+        var request = httpGet(requestUri);
 
         try (var response = httpClient.execute(request)) {
 
             if (response.isSuccessful()) {
                 if (response.code() == HTTP_CODE_404) {
-                    return Result.failure(
-                            String.format(CALL_UNSUCCESSFUL_ERROR_TEMPLATE, "Secret not found"));
+                    return Result.failure("Secret %s not found".formatted(key));
```
jimmarino, line 204: The secret key should not be recorded. The client knows which key it asked for and can handle it appropriately.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +25,206 @@
...
-                    return Result.failure(String.format(CALL_UNSUCCESSFUL_ERROR_TEMPLATE, "Response body empty"));
+                    return Result.failure("Secret %s response body is empty".formatted(key));
                 }
                 var payload = objectMapper.readValue(responseBody.string(), GetEntryResponsePayload.class);
                 var value = payload.getData().getData().get(VAULT_DATA_ENTRY_NAME);
 
                 return Result.success(value);
             } else {
-                return Result.failure(String.format(CALL_UNSUCCESSFUL_ERROR_TEMPLATE, response.code()));
+                return Result.failure("Failed to get secret %s with status %d".formatted(key, response.code()));
```
jimmarino, line 216: The secret key should not be recorded. The client knows which key it asked for and can handle it appropriately

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +25,206 @@
...
             }
 
         } catch (IOException e) {
-            return Result.failure(e.getMessage());
+            var errMsg = "Failed to get secret %s with reason: %s".formatted(key, e.getMessage());
```
jimmarino, line 220: The secret key should not be recorded. The client knows which key it asked for and can handle it appropriately

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +25,206 @@
...
+            monitor.warning(errMsg, e);
```
jimmarino, line 221: Ditto about sending the message to the monitor

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -105,76 +236,48 @@ public Result<CreateEntryResponsePayload> setSecret(@NotNull String key, @NotNul
...
+                if (response.body() == null) {
+                    return Result.failure("Setting secret %s returned empty body".formatted(key));
+                }
+                var responseBody = response.body().string();
                 var responsePayload =
                         objectMapper.readValue(responseBody, CreateEntryResponsePayload.class);
                 return Result.success(responsePayload);
             } else {
-                return Result.failure(String.format(CALL_UNSUCCESSFUL_ERROR_TEMPLATE, response.code()));
+                return Result.failure("Failed to set secret %s with status %d".formatted(key, response.code()));
```
jimmarino, line 247: The secret key should not be recorded. The client knows which key it asked for and can handle it appropriately

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -105,76 +236,48 @@ public Result<CreateEntryResponsePayload> setSecret(@NotNull String key, @NotNul
 
         try (var response = httpClient.execute(request)) {
             if (response.isSuccessful()) {
-                var responseBody = Objects.requireNonNull(response.body()).string();
+                if (response.body() == null) {
+                    return Result.failure("Setting secret %s returned empty body".formatted(key));
```
jimmarino, line 240: The secret key should not be recorded. The client knows which key it asked for and can handle it appropriately

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -105,76 +236,48 @@ public Result<CreateEntryResponsePayload> setSecret(@NotNull String key, @NotNul
...
     public Result<Void> destroySecret(@NotNull String key) {
         var requestUri = getSecretUrl(key, VAULT_SECRET_METADATA_PATH);
-        var headers = getHeaders();
         var request = new Request.Builder().url(requestUri).headers(headers).delete().build();
 
         try (var response = httpClient.execute(request)) {
             return response.isSuccessful() || response.code() == HTTP_CODE_404
                     ? Result.success()
-                    : Result.failure(String.format(CALL_UNSUCCESSFUL_ERROR_TEMPLATE, response.code()));
+                    : Result.failure("Failed to destroy secret %s with status %d".formatted(key, response.code()));
```
jimmarino, line 263: The secret key should not be recorded. The client knows which key it asked for and can handle it appropriately

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -185,26 +288,85 @@ private HttpUrl getHealthUrl() {
...
+
+    private Result<Boolean> parseRenewable(Map<String, Object> map) {
+        try {
+            var data = objectMapper.convertValue(getValueFromMap(map, DATA_KEY), new TypeReference<Map<String, Object>>() {
+            });
+            var isRenewable = objectMapper.convertValue(getValueFromMap(data, RENEWABLE_KEY), Boolean.class);
+            return Result.success(isRenewable);
+        } catch (IllegalArgumentException e) {
+            var errMsg = "Failed to parse renewable flag from %s with reason: %s".formatted(map, e.getMessage());
+            monitor.warning(errMsg);
```
jimmarino, line 347: Ditto about monitoring

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -185,26 +288,85 @@ private HttpUrl getHealthUrl() {
...
+
+    private Result<Long> parseTtl(Map<String, Object> map) {
+        try {
+            var auth = objectMapper.convertValue(getValueFromMap(map, AUTH_KEY), new TypeReference<Map<String, Object>>() {
+            });
+            var ttl = objectMapper.convertValue(getValueFromMap(auth, LEASE_DURATION_KEY), Long.class);
+            return Result.success(ttl);
+        } catch (IllegalArgumentException e) {
+            var errMsg = "Failed to parse ttl from %s with reason: %s".formatted(map, e.getMessage());
+            monitor.warning(errMsg);
```
jimmarino, line 360: In this case, log the exception with a more descriptive error message (token renewal TTL) including the entire error so that the stack trace is preserved and return a simple failure. The caller should then simply abort the process. As it stands, the error is essentially logged twice, which does not add value to the output.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -185,26 +288,85 @@ private HttpUrl getHealthUrl() {
...
+            return Result.failure(errMsg);
+        }
+    }
+
+    private Object getValueFromMap(Map<String, Object> map, String key) {
+        var value = map.get(key);
+        if (value == null) {
+            throw new IllegalArgumentException("Key '%s' does not exist".formatted(key));
```
jimmarino, line 368: Ditto about key

jimmarino (changes requested):
Looking much better but a couple of things to clean up:

1. Secret keys should not be outputted to the monitor. 
2. Take a close look and come up with a more streamlined approach to how messages are logged and returned. Either log the message in the processing method, especially when an error is thrown (e.g., parsing the TTL) and return a simple error response, or return a message and let the client decide how to log it with additional information. 

Point 2 will also help with point 1 since the client can provide contextual information for a failure without revealing a secret key.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/model/EntryMetadata.java
```diff
@@ -33,8 +33,7 @@ public class EntryMetadata {
     @JsonProperty()
     private Integer version;
 
-    EntryMetadata() {
-    }
+    private EntryMetadata() {}
```
paullatzelsperger, line 36: the empty `@JsonProperty()` annotations are not needed and can be removed.

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -9,228 +9,447 @@
...
+            var response = new Response.Builder()
+                    .code(200)
+                    .message("any")
+                    .body(ResponseBody.create(body, MediaType.get("application/json")))
+                    .protocol(Protocol.HTTP_1_1)
+                    .request(new Request.Builder().url("http://any").build())
+                    .build();
+            when(httpClient.execute(any(Request.class))).thenReturn(response);
+
+            // invoke
```
paullatzelsperger, line 119: nit: those spurious comments can be removed. goes for all tests

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -9,228 +9,447 @@
...
+            var healthCheckResponseResult = vaultClient.doHealthCheck();
+
+            // verify
+            assertThat(healthCheckResponseResult.succeeded()).isTrue();
```
paullatzelsperger, line 123: please consider using the `AbstractResultAssert`, which allows to directly assert on a result:
```java
assertThat(healthCheckResponseResult).isSucceeded();
// or:
assertThat(someResult).isFailed().detail().isEqualTo("foobar");
```

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientTest.java
```diff
@@ -9,228 +9,447 @@
...
-        // verify
-        assertNotNull(result);
-        verify(edcClientMock).execute(argThat(request -> request.method().equalsIgnoreCase("GET") &&
-                request.url().encodedPath().contains(CUSTOM_SECRET_PATH + "/data") &&
-                request.url().encodedPathSegments().contains(KEY)));
+    @Nested
+    class Token {
+        @Test
+        void lookUpToken_whenApiReturns200_shouldSucceed() throws IOException {
+            var body = """
```
paullatzelsperger, line 204: inner monk says: spacing, emtpy lines

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultTokenRenewTask.java
```diff
@@ -0,0 +1,131 @@
...
+    @NotNull
+    private final HashicorpVaultClient client;
+    @NotNull
+    private final Monitor monitor;
+    private final long renewBuffer;
+    private ScheduledExecutorService scheduledExecutorService;
+    private Future<?> tokenRenewTask;
+    private final AtomicBoolean isRunning = new AtomicBoolean(false);
+
+    public HashicorpVaultTokenRenewTask(@NotNull ExecutorInstrumentation executorInstrumentation,
```
paullatzelsperger, line 48: at least to me it is not obvious what the intended use of `renewBuffer` is or what values I should pass. Please add javadoc to the CTor

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthCheck.java
```diff
@@ -39,52 +40,16 @@ public HashicorpVaultHealthCheck(HashicorpVaultClient client, Monitor monitor) {
...
-                result = HealthCheckResult.failed(sealedMsg);
-            }
-            default -> {
-                var unspecifiedMsg = format(HEALTH_CHECK_ERROR_TEMPLATE, "Unspecified response from vault. Code: " + response.getCode(), response.getPayload());
-                monitor.warning(unspecifiedMsg);
-                result = HealthCheckResult.failed(unspecifiedMsg);
-            }
-        }
-        return result.forComponent(COMPONENT_NAME);
+        return client
```
paullatzelsperger, line 43: much cleaner :thumbsup:

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClientIntegrationTest.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Test
+    void lookUpToken_whenTokenExpired_shouldFail() {
+        await()
+                .pollDelay(CREATION_TTL, TimeUnit.SECONDS)
+                .atMost(CREATION_TTL + 1, TimeUnit.SECONDS)
+                .untilAsserted(() -> {
+                    var tokenLookUpResult = client.lookUpToken();
+                    assertThat(tokenLookUpResult.failed()).isTrue();
```
paullatzelsperger, line 82: nit: use `AbstractResultAssert`. goes for the entire class

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthCheckTest.java
```diff
@@ -9,67 +9,109 @@
...
+        when(client.lookUpToken()).thenReturn(Result.failure(tokenLookUpErr));
 
         var result = healthCheck.get();
 
         assertThat(result).isFailed();
+        assertFailureMessagesSize(result, 2);
+        verify(monitor).warning("Healthcheck failed with reason(s): %s, %s".formatted(healthCheckErr, tokenLookUpErr));
+    }
+
+    private void assertFailureMessagesSize(HealthCheckResult result, int i) {
```
paullatzelsperger, line 114: personally, i really dislike these custom "assert-something" methods, especially if the assertion is as trivial as it is here. They make the code less readable. pls inline.

paullatzelsperger (changes requested):

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultClient.java
```diff
@@ -24,76 +25,192 @@
...
+    }
+
+    /**
+     * Attempts to look up the Vault token defined in the configurations.
+     * <p>
+     * Will retry in some error cases.
+     *
+     * @return boolean indicating if the token is renewable
+     */
+    public Result<Boolean> lookUpToken() {
```
paullatzelsperger, line 119: minor nit: if this method's purpose is to look up, whether the token is renewable, then it should be called `isTokenRenewable` or `checkTokenRenewable`. A name `lookUpToken` may leave people puzzled as to why this returns a boolean and not a token.

paullatzelsperger (approved):
just a minor naming nit, but once that is resolved I think we can merge this PR.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthExtension.java
```diff
@@ -9,61 +9,51 @@
...
-        healthCheckService.addReadinessProvider(healthCheck);
-        healthCheckService.addStartupStatusProvider(healthCheck);
+        var monitor = context.getMonitor().withPrefix(NAME);
+        var healthCheckEnabled = context.getSetting(VAULT_HEALTH_CHECK_ENABLED, VAULT_HEALTH_CHECK_ENABLED_DEFAULT);
+        if (healthCheckEnabled) {
+            var healthCheck = new HashicorpVaultHealthCheck(client, monitor);
+            healthCheckService.addLivenessProvider(healthCheck);
+            healthCheckService.addReadinessProvider(healthCheck);
+            healthCheckService.addStartupStatusProvider(healthCheck);
+            monitor.info("Healthcheck initialization complete");
```
jimmarino, line 54: Please change this to debug level and be more descriptive, e.g. "Vault health check." Even though the message is prefixed with the extension name. Also, I think health check is two words.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthExtension.java
```diff
@@ -9,61 +9,51 @@
...
+        } else {
+            monitor.info("Healthcheck disabled");
```
jimmarino, line 56: See comment above

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultHealthCheck.java
```diff
@@ -39,52 +40,16 @@ public HashicorpVaultHealthCheck(HashicorpVaultClient client, Monitor monitor) {
...
+                .doHealthCheck()
+                .merge(client.lookUpToken())
+                .flatMap(result -> {
+                    if (result.succeeded()) {
+                        return HealthCheckResult.success();
+                    } else {
+                        monitor.warning("Healthcheck failed with reason(s): " + result.getFailureDetail());
```
jimmarino, line 50: See comment about being more descriptive below.

jimmarino (changes requested):

paullatzelsperger:
@hamidonos is any work being done on this? Seeing as the latest comments from @jimmarino are 2 weeks old.

paullatzelsperger:
Notice: should this PR not receive any updates by Friday March 15 2024 5pm CET, I will close, adopt and re-submit it under my authorship. 
The Copyright will remain with the original author.

DominikPinsel:
> Notice: should this PR not receive any updates by Friday March 15 2024 5pm CET, I will close, adopt and re-submit it under my authorship. The Copyright will remain with the original author.

We kindly request reconsideration of the one-day response time, as your team member, Hamid, will be out of the office until Friday. Adjusting the response timeframe would greatly aid in maintaining workflow efficiency.

paullatzelsperger:
@DominikPinsel this PR has been open forever, and no activity was visible for two weeks after the latest comments. We have a release scheduled for [March 27](https://github.com/eclipse-edc/Connector/milestone/16) which should have this feature in it.

If it is a matter of a few days, then I'm sure we can accomodate, but we shouldn't wait too long. Shall we say +1week, i.e. Friday March 22nd then?

DominikPinsel:
> @DominikPinsel this PR has been open forever, and no activity was visible for two weeks after the latest comments. We have a release scheduled for [March 27](https://github.com/eclipse-edc/Connector/milestone/16) which should have this feature in it.
> 
> If it is a matter of a few days, then I'm sure we can accomodate, but we shouldn't wait too long. Shall we say +1week, i.e. Friday March 22nd then.

This would be great! :)

jimmarino (approved):

ndr-brt:
hey @hamidonos we had some changes that conflicted with this PR, please solve conflicts and then we'll be ready to merge

PR #3797, opened 2024-01-24
fix: stop `method` property from being mandatory for push http transfers

### LINKED ISSUE #3768: Assets Containing the `proxyMethod=true` Property can not be used with the ProviderPushTransfer
# Bug Report

## Describe the Bug

Assets containing the `proxyMethod=true` property can not be used with the ProviderPushTransfer.

### Expected Behavior

When starting a new `TransferProcess` with type `HttpData` over the POST `/transferprocesses` endpoint of the consumer the provider data-plane should use the `properties` of the `TransferRequest` to define the http-method for accessing the data-source.

### Observed Behavior

The validation in the `BaseSourceHttpParamsDecorator` fails with "DataFlowRequest %s: 'method' property is missing".

## Context Information

- Used version 0.2.1
- But should also hold for newer versions

## Detailed Description

This should origin from the `ProviderPushTransferDataFlowController`, which does not create an `DataFlowRequest` with the properties of the `TransferRequest`:
https://github.com/eclipse-edc/Connector/blob/cf3b2450b9bdcff749566e95b087b211b5d70fd9/extensions/control-plane/transfer/transfer-data-plane/src/main/java/org/eclipse/edc/connector/transfer/dataplane/flow/ProviderPushTransferDataFlowController.java#L64-L72

ndr-brt:
## What this PR changes/adds

The `method` property was mandatory when the `proxyMethod` was true, but, in fact, also if a source `DataAddress` defines that property, it could be that the same asset is used in a `PUSH` type transfer, this fix permits to do that by adding an additional check.

## Why it does that

fix bug

## Further notes

- in the future (see dataplane signaling DR) this information will be taken out from the `transferType` field that will be mandatory, while the `dataDestination` won't be supplied for `pull` transfers.

## Linked Issue(s)

Closes #3768 

### DIFF
--- a/extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/params/decorators/BaseSourceHttpParamsDecorator.java
+++ b/extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/params/decorators/BaseSourceHttpParamsDecorator.java
@@ -53,7 +53,7 @@ public class BaseSourceHttpParamsDecorator implements HttpParamsDecorator {
     }
 
     private @NotNull String extractMethod(HttpDataAddress address, DataFlowRequest request) {
-        if (Boolean.parseBoolean(address.getProxyMethod())) {
+        if (Boolean.parseBoolean(address.getProxyMethod()) && "HttpProxy".equals(request.getDestinationDataAddress().getType())) {
             return Optional.ofNullable(request.getProperties().get(METHOD))
                     .orElseThrow(() -> new EdcException(format("DataFlowRequest %s: 'method' property is missing", request.getId())));
         }
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -180,10 +180,8 @@ maven/mavencentral/javax.ws.rs/javax.ws.rs-api/2.1, (CDDL-1.1 OR GPL-2.0 WITH Cl
 maven/mavencentral/joda-time/joda-time/2.10.5, Apache-2.0, approved, clearlydefined
 maven/mavencentral/junit/junit/4.13.2, EPL-2.0, approved, CQ23636
 maven/mavencentral/net.bytebuddy/byte-buddy-agent/1.14.1, Apache-2.0, approved, #7164
-maven/mavencentral/net.bytebuddy/byte-buddy-agent/1.14.10, Apache-2.0, approved, #7164
-maven/mavencentral/net.bytebuddy/byte-buddy/1.12.21, Apache-2.0 AND BSD-3-Clause, approved, #1811
+maven/mavencentral/net.bytebuddy/byte-buddy-agent/1.14.11, Apache-2.0, approved, #7164
 maven/mavencentral/net.bytebuddy/byte-buddy/1.14.1, Apache-2.0 AND BSD-3-Clause, approved, #7163
-maven/mavencentral/net.bytebuddy/byte-buddy/1.14.10, Apache-2.0 AND BSD-3-Clause, approved, #7163
 maven/mavencentral/net.bytebuddy/byte-buddy/1.14.11, Apache-2.0 AND BSD-3-Clause, approved, #7163
 maven/mavencentral/net.java.dev.jna/jna/5.13.0, Apache-2.0 AND LGPL-2.1-or-later, approved, #6709
 maven/mavencentral/net.javacrumbs.json-unit/json-unit-core/2.36.0, Apache-2.0, approved, clearlydefined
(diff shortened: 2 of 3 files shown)


jimmarino (approved):

wolf4ood (approved):

bscholtes1A (approved):

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/params/decorators/BaseSourceHttpParamsDecorator.java
```diff
@@ -53,7 +53,7 @@ public HttpRequestParams.Builder decorate(DataFlowRequest request, HttpDataAddre
     }
 
     private @NotNull String extractMethod(HttpDataAddress address, DataFlowRequest request) {
-        if (Boolean.parseBoolean(address.getProxyMethod())) {
+        if (Boolean.parseBoolean(address.getProxyMethod()) && "HttpProxy".equals(request.getDestinationDataAddress().getType())) {
```
efiege, line 56: I don't expect that this will solve the issue regarding the `HttpPush`. This will disable the feature of proxying http methods for the `HttpPush`, as now just "GET" requests will be sent to the providing backend.

Our expectation would be that http-methods also could be proxied for the `HttpPush`

efiege (changes requested):

extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/params/decorators/BaseSourceHttpParamsDecorator.java
ndr-brt: simply, for http push this condition won't ever met, and it will pass through line 60, so the method will be taken from the `DataAddress` (as it already does) if it's defined, otherwise will use `GET`.

To verify that, just try to remove the `proxyMethod` property and set `method` instead on the source data address.

efiege: > simply, for http push this condition won't ever met, and it will pass through line 60, so the method will be taken from the DataAddress (as it already does) if it's defined, otherwise it will use the default GET.

Ohh yes, I have overlooked this: The `HttpPush` will use the `method` from the `DataAddress`.

> The proxy was made to work with pull, not for push

This would implicate that certain `HttpData` assets can just be used for `HttpProxy` transfer-types and not for `HttpData` transfer-types. Would't it make sense than to use `HttpProxy` type also for the providing asset?

ndr-brt: theoretically yes, I'm not sure which would be the applications for that: the consumer should be the one who chose how the call to its service will be made.

efiege: The applications would be the same as for the `http-pull` I think. 

So for the `http-pull` the `consumer` is able to select the preferred http-method, which is used to access the `provider` system. This is enabled by using the `prxoyMethod=true` property in the asset of the provider.

For the `http-push` the transfer is also started by the `consumer`. Like for the `http-pull` the `consumer` should be able to select the prefered http-method, which is used to access the `provider` system.

ndr-brt: in any case, also without the `proxyMethod` flag to true, the consumer can choose the method, that by default is `GET`. Really, all the `proxyXXX` stuff is there for the `proxy` use case ("nomen omen").

efiege: Hey @ndr-brt,

I'm sry for coming back on this so lately, but this change is really important to us and I may not have understood this correctly.

After giving this a second look, I am also convinced, that this change won't affect passing methods for the `HttpPush` from the consumer.

My problem was, that I though the `HttpDataAddress address` parameter is always populated by the `SourceDataAddress`:
https://github.com/eclipse-edc/Connector/blob/cf629c963c51f74157eef4d03f10dbf476c6e625/extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/params/decorators/BaseSourceHttpParamsDecorator.java#L55-L61

but it turns out it also gets called for the `DestinationDataAddress`:
https://github.com/eclipse-edc/Connector/blob/cf629c963c51f74157eef4d03f10dbf476c6e625/extensions/data-plane/data-plane-http/src/main/java/org/eclipse/edc/connector/dataplane/http/params/HttpRequestParamsProviderImpl.java#L54-L68

So indeed the `extractMethod` returns the method specified by the `DestinationDataAddress`.

Did I understood this correctly?

Thanks for fixing this!

PR #3799, opened 2024-01-24
refactor: improve Public Key Resolver

### LINKED ISSUE #3796: Improve PublicKeyResolver
# Feature Request

`PublicKeyResolver` as of today it's an injectable component, meaning only once instance can be the runtime. During the last token refactor we also require the `PublicKeyResolver` for validating tokens. But the PublicKeyResolver can be different depending on the context e.g. iatp, did, dataplane, oauth. Also the `PublicKeyResolver` 

We should remove the `PublicKeyResolver` as injectable and for local public key resolver (vault,filesystem, etc) we should have a separated service that can be injected when a public key needs to be resolved by id.
 
`PublicKeyResolver` and usage of it

refactoring, clean up after https://github.com/eclipse-edc/Connector/issues/3743

wolf4ood:
## What this PR changes/adds

- `PublicKeyResolver` it's not injectable anymore and needs to be instantiated depending on the use case.
- Specialized version of `PublicKeyResolver` for resolving public keys from DIDs  has been introduced (`DidPublicKeyResolver`).
- `LocalPublicKeyService` has been introduced for resolving public keys needed for the runtime (vault, config. etc)
- A default implementation of `LocalPublicKeyService` has been introduced in `connector-core` as default which resolve the public keys from the cached public keys (config) or vault as fallback. It replace the `VaultPublicKeyResolver` effectively 

## Linked Issue(s)

Closes #3796


core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/LocalPublicKeyDefaultExtension.java
```diff
@@ -0,0 +1,89 @@
...
+    public KeyParserRegistry keyParserRegistry;
+
+    private Config keysConfiguration;
+
+    private LocalPublicKeyServiceImpl localPublicKeyService;
+    @Inject
+    private Vault vault;
+
+    @Provider(isDefault = true)
+    public LocalPublicKeyService localPublicKeyService() {
```
paullatzelsperger, line 62: could this be instantiated in the `SecurityDefaultServicesExtension`?

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/LocalPublicKeyDefaultExtension.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        keysConfiguration = context.getConfig(EDC_PUBLIC_KEYS_PREFIX);
+    }
+
+    @Override
+    public void prepare() {
+        if (keysConfiguration != null) {
+            var result = keysConfiguration.partition().map(this::readPublicKey)
+                    .map(entry -> localPublicKeyService.addRawKey(entry.getKey(), entry.getValue()))
```
paullatzelsperger, line 76: there is a chance that `localPublicKeyService` will be `null` here, i.e. when the provider method has not been (yet) initialized. At the very least I would propose the following: 

```java
    @Provider(isDefault = true)
    public LocalPublicKeyService localPublicKeyService() {
        if(localPublicKeyService == null) {
            localPublicKeyService = new LocalPublicKeyServiceImpl(vault, keyParserRegistry);
        }
        return localPublicKeyService;
    }
```

and in `prepare()`: 
```java
   .map(entry -> localPublicKeyService().addRawKey(.....))
```

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/security/LocalPublicKeyServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+import java.security.PublicKey;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Implementation of {@link LocalPublicKeyService} which looks-up for the key by id
+ * first in the locally cached keys and as fallback in the vault.
+ */
+public class LocalPublicKeyServiceImpl implements LocalPublicKeyService {
```
paullatzelsperger, line 31: Contrary to the PR Description, the local service does not resolve keys from the config (via path or raw string). I think we will need that though.

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/iam/LocalPublicKeyService.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.security.PublicKey;
+
+/**
+ * Resolve a public key from an id. This is generic injectable component for resolving public
+ * keys configured in the runtime (vault, config, etc.)
+ */
+@ExtensionPoint
+@FunctionalInterface
+public interface LocalPublicKeyService {
```
paullatzelsperger, line 28: this could just be a marker interface and extend `PublicKeyResolver`, as it has the exact same signature and similar semantics.

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/LocalPublicKeyDefaultExtension.java
wolf4ood: I didn't want to 'pollute' `SecurityDefaultServicesExtension` with all the logic of  `LocalPublicKeyServiceImpl`

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/security/LocalPublicKeyServiceImpl.java
wolf4ood: It resolve from the config actually, just it sees the public keys from the config and cached locally

wolf4ood: The path is not supported yet, but do we need it?

paullatzelsperger: putting a raw PEM string as config value may be cumbersome, and may cause issues with multiline etc., so being able reference a key file (`.pem`) may come in handy

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/LocalPublicKeyDefaultExtension.java
paullatzelsperger: ok, fair enough. but is there a reason to read the keys in `prepare()` rather than `initialize()`?

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/LocalPublicKeyDefaultExtension.java
```diff
@@ -0,0 +1,89 @@
...
+
+import static org.eclipse.edc.connector.core.SecurityDefaultServicesExtension.NAME;
+
+@Extension(value = NAME)
+@Provides(LocalPublicKeyService.class)
+public class LocalPublicKeyDefaultExtension implements ServiceExtension {
+    
+    public static final String NAME = "Local Public Key Default Extension";
+
+    public static final String EDC_PUBLIC_KEYS_PREFIX = "edc.iam.publickeys";
```
paullatzelsperger, line 42: @Setting

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/security/LocalPublicKeyServiceImpl.java
wolf4ood: I can add another field to the config which is `path` instead of `value` and read it from a path in case

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/LocalPublicKeyDefaultExtension.java
wolf4ood: It's not needed this is only the prefix the `@Setting` is on `id` and `value`

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/security/LocalPublicKeyServiceImpl.java
paullatzelsperger: i would do both: check if a `*.path` config entry was specified, and if not, check if a `.value` config entry is there. That would give: vault -> path -> raw value

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/LocalPublicKeyDefaultExtension.java
wolf4ood: Yes because additional Key parsers could be registered in the `initialize` phase and if we create the cache from the config in that phase it might not work with custom key parsers

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/security/LocalPublicKeyServiceImpl.java
wolf4ood: done

paullatzelsperger (approved):

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/security/LocalPublicKeyServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private Optional<PublicKey> resolveFromCache(String id) {
+        return Optional.ofNullable(cachedKeys.get(id));
+    }
+
+    private Result<PublicKey> parseKey(String encodedKey) {
+        return registry.parse(encodedKey).compose(pk -> {
+            if (pk instanceof PublicKey publicKey) {
+                return Result.success(publicKey);
+            } else return Result.failure("The specified resource did not contain public key material.");
```
ndr-brt, line 63: brackets

ndr-brt (approved):

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/LocalPublicKeyDefaultExtension.java
```diff
@@ -0,0 +1,118 @@
...
+    private LocalPublicKeyServiceImpl localPublicKeyServiceImpl() {
+        if (localPublicKeyService == null) {
+            localPublicKeyService = new LocalPublicKeyServiceImpl(vault, keyParserRegistry);
+        }
+        return localPublicKeyService;
+    }
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        keysConfiguration = context.getConfig(EDC_PUBLIC_KEYS_PREFIX);
```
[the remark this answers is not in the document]
wolf4ood: I'd have to cache the context anyway in the initialize right?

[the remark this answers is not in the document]
ndr-brt: woa, you're right, I was pretty sure that also the prepare phase had the context injected... at this point it could also remain there, but in my opinion I don't see why context shouldn't be passed to all the phases.

PR #3809, opened 2024-01-29
feat: add operators validation

### LINKED ISSUE #3713: api: add validation on the `Criterion` object to avoid unsupported operators
# Feature Request

Currently is possible to create a `Criterion` that uses unsupported operators, as reported here:
https://github.com/eclipse-tractusx/tractusx-edc/issues/915

Let's add a validator for avoiding that upfront.

api validators

_Are there any requirements?_

## Solution Proposal

ndr-brt:
## What this PR changes/adds

Add operator validation.

## Why it does that

avoid problems with unsupported operators upfront

## Further notes

- Had to refactor the way we translate from a criterion operator to a `Predicate`, this way it will be possible for adopters to implement custom operators with custom validation. the `CriterionOperatorRegistry` keeps track of the supported operators and the way to translate a `Criterion` into a `Predicate`.

## Linked Issue(s)

Closes #3713


core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/store/ContainsOperatorPredicate.java
```diff
@@ -0,0 +1,31 @@
...
+import org.eclipse.edc.spi.query.OperatorPredicate;
+
+import java.util.Collection;
+
+public class ContainsOperatorPredicate implements OperatorPredicate {
+
+    @Override
+    public boolean test(Object property, Object operandRight) {
+        if (property instanceof Collection<?> collection) {
+            return collection.contains(operandRight);
```
paullatzelsperger, line 26: theoretically `operandRight` could be a `List` itself, IDK what would happen then...

paullatzelsperger (approved):

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/store/ContainsOperatorPredicate.java
ndr-brt: in the (postgre)SQL context it can be only a single object, because eventually I guess another operator should be used.
I think if we want to being able to pass a list as `operandRight` maybe we'll need another operator (`containsAll`?)

Issue #3819, opened 2024-02-01
 Policy scopes for contract negotiation ingress calls


wolf4ood:
# Feature Request

As we do for contract negotiation `egress` call, we should be able to attach custom information to the current contract negotiation `ingress` call, like running policy scope extractors.

Part of #3662

wolf4ood:
Closed by https://github.com/eclipse-edc/Connector/pull/3834

PR #3822, opened 2024-02-02
feat: policy scope extractors for transfer process ingress calls

### LINKED ISSUE #3818: Policy scopes for transfer process ingress calls
# Feature Request

As we do for transfer process `egress` call, we should be able to attach custom information to the current transfer process `ingress` call, like running policy scope extractors.

Part of #3662

wolf4ood:
## What this PR changes/adds

Introduces the policy scopes extractor in transfer process messages at protocol layers:

- Injected in all `*ProtocolService` the `policy` engine for scope extraction
- When evaluating the `policy` for the request scope the in context object has been changed from `TokenParameters.Builder`  to `RequestScope.Builder` for ingress/egress symmetry
- Now we need to fetch some additional data before verifying the actual token. To prevent possible leakage of resource presence to unauthenticated calls, all the transfer process apis returns `404` not found, in case of resource (transferprocess/contract negotiation) not present in the database and in the case token verification failure as outlined in the [spec](https://docs.internationaldataspaces.org/ids-knowledgebase/v/dataspace-protocol/transfer-process/transfer.process.binding.https#id-2.4.1-get).
- In order to prevent unwanted writes on transfer process lease in the database before the token verification, on all `TransferRemoteMessage` with the exception of `TransferRequestMessage`, the transfer process will be fetched two times, the first time in read-only mode for getting the associated policy and the second one after the token has been verified with the lease attached.

## Linked Issue(s)

Closes #3818

### DIFF
--- a/core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/negotiation/ContractNegotiationIntegrationTest.java
+++ b/core/control-plane/contract-core/src/test/java/org/eclipse/edc/connector/contract/negotiation/ContractNegotiationIntegrationTest.java
@@ -148,8 +148,8 @@ class ContractNegotiationIntegrationTest {
                 .build();
 
         when(identityService.verifyJwtToken(eq(tokenRepresentation), isA(VerificationContext.class))).thenReturn(Result.success(token));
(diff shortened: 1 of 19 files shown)


paullatzelsperger (approved):

wolf4ood:
@ndr-brt  noted i will do in the next PR since i will work on contract negotiation messages  #3819

PR #3827, opened 2024-02-05
feat: avoid fetching body in memory during data plane proxy call

### LINKED ISSUE #3826: `OutputStreamDataSink` is fetching response body in memory
# Feature Request

Looks like `OutputStreamDataSink` is not streaming directly the proxied call to the caller but it fetches the data in memory using an additional stream:
https://github.com/eclipse-edc/Connector/blob/9e9eeab26fece75bb1b2d6df86f9b3ceb879af47/core/data-plane/data-plane-util/src/main/java/org/eclipse/edc/connector/dataplane/util/sink/OutputStreamDataSink.java#L67

An improvement would be to stream directly from the call made on the source to the caller to save memory and permit long-running calls.

data plane

performances

## Solution Proposal

ndr-brt:
## What this PR changes/adds

Use the `AsyncStreamingDataSink` in the `DataPlanePublicApiController`, to avoid buffering response body in memory.

## Why it does that

optimization, permit long-running calls

## Linked Issue(s)

Closes #3826


jimmarino (approved):

wolf4ood (approved):
:rocket:

bscholtes1A (approved):

PR #3834, opened 2024-02-07
feat: policy scope extractors for contract negotiation ingress calls

### LINKED ISSUE #2819: Callback implementation : propagate the EDR

wolf4ood:
## What this PR changes/adds

Introduces the policy scopes extractor in contract negotiation messages at protocol layers:

- in all contract negotiation messages, `ContractRequestMessage` excluded, we first load the contract negotiation in read mode based on the message, then we extract the policy and verify the token with the scope attached running the policy engine. Same as in #3822,  if the contract negotiation is not found or the token verification fails the returned error is `404`
- When dealing with `ContractRequestMessage` we changed a bit how the sequence of fetching stuff from the database for validation is done. We first fetch all the information what we need in order to validate the `ContractRequestMessage`  through the new core service `ConsumerOfferResolver`. Then we verify the token with the extracted scopes, if successful we run the validation of the `ValidatableConsumerOffer` with the `ContractValidationService` 
- The default implementation of `ContractValidationService` has been modified in order to validate `ValidatableConsumerOffer`. Additional data like `Policy` or `ContractDefinition` from the database are not fetched anymore here since they are already contained  in the input `ValidatableConsumerOffer`

## Linked Issue(s)

Closes #2819

### DIFF
--- a/spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/offer/ConsumerOfferResolver.java
+++ b/spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/offer/ConsumerOfferResolver.java
@@ -1,0 +1,35 @@
+/*
+ *  Copyright (c) 2024 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.connector.contract.spi.offer;
+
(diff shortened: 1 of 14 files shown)


spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/contract/spi/offer/ConsumerOfferResolver.java
```diff
@@ -0,0 +1,35 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Resolve the consumer offer into a {@link ValidatableConsumerOffer} which can be used
+ * to validate incoming offer through {@link ContractValidationService#validateInitialOffer}
+ */
+public interface ConsumerOfferResolver {
+
+    @NotNull
+    ServiceResult<ValidatableConsumerOffer> resolveOffer(ContractOffer offer);
```
ndr-brt, line 30: I'd leave only the second method, and let the caller extract the id from the offer to call the `resolveOffer` method

jimmarino (approved):

PR #3859, opened 2024-02-14
chore: move observability api docs to default api docs

### LINKED ISSUE #3759: observaility api should not be listed in the management api documentation
# Feature Request

Currently the observability endpoints are listed under the [`management-api` swaggerhub documentation](https://app.swaggerhub.com/apis/eclipse-edc-bot/management-api/0.4.1#/Application%20Observability/checkHealth), but in fact they are exposed in the default api context.

The proposal would be to remove it from that one and publish a `default-api` openapi spec.

api, documentation

_Are there any requirements?_

## Solution Proposal

git-masoud:
## What this PR changes/adds

Moves move observability api docs to default api documentations.

## Why it does that

Currently the observability endpoints are listed under the [management-api swaggerhub documentation](https://app.swaggerhub.com/apis/eclipse-edc-bot/management-api/0.5.1#/Application%20Observability/checkHealth), but in fact they are exposed in the default api context.

## Linked Issue(s)

Closes #3759

### DIFF
--- a/.github/workflows/apidoc.yaml
+++ b/.github/workflows/apidoc.yaml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        apiGroup: [ 'default-api', 'public-api', 'management-api', 'control-api' ]
+        apiGroup: [ 'observability-api', 'public-api', 'management-api', 'control-api' ]
     env:
       rootDir: resources/openapi/yaml/${{ matrix.apiGroup }}
       SWAGGERHUB_API_KEY: ${{ secrets.SWAGGERHUB_TOKEN }}
--- a/.github/workflows/apidoc.yaml
+++ b/.github/workflows/apidoc.yaml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        apiGroup: [ 'public-api', 'management-api', 'control-api' ]
+        apiGroup: [ 'default-api', 'public-api', 'management-api', 'control-api' ]
     env:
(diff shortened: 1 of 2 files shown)


.github/workflows/apidoc.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        apiGroup: [ 'public-api', 'management-api', 'control-api' ]
+        apiGroup: [ 'default-api', 'public-api', 'management-api', 'control-api' ]
```
bscholtes1A, line 26: @paullatzelsperger  do we really want to publish these in Swagger hub?

paullatzelsperger, line 26: I don't like the name "default-api", IMO "observability-api" would be better. The bigger problem is, that in practice it will not (yet) work, because SwaggerHub has a limit of 3 APIs on the free tier which we are using. 

We have contacted SmartBear, but so far we have yet to receive a response.

paullatzelsperger: well, we need to document it _somewhere_, so why not Swaggerhub.

git-masoud: Ok, then let's leave this PR open/stale until there is an answer from them.

bscholtes1A (approved):

paullatzelsperger:
@git-masoud update here: unfortunately SmartBear is unable to increase the number of free APIs, so we decided to move our api documentation to Github Pages. For example, the management API will be available at https://eclipse-edc.github.io/Connector/openapi/management-api/0.5.2-SNAPSHOT/.

For this PR, I think nothing changes, we can just merge it.

PR #3875, opened 2024-02-18
fix: multiple fix in `Iatp` modules

bscholtes1A:
## What this PR changes/adds

 - Avoid `NullPointerException` in `TokenValidationRulesRegistryImpl` if there is no rules registered for a given context
 - In `JwtPresentationVerifier`:
 - do not validate VC `aud` claim, as they don't have one
 - do not validate VP `sub` claim, as they don't have one
 - **MAKE VERIFIER CONTEXT IMMUTABLE**: this was causing very dubious behavior as `JwtPresentationVerifier.verify` was modifying the `audience` in the `VerificationContext` in the 
recursive call but it was never restored to the initial and expected value for the subsequent presentation verifications! So basically, first presentation verification was fine, but all consecutives were failing because the context audience was changed at the end of the first execution!
 - In `DidPublicKeyResolverImpl`: improve error logging when the provided did is null (happens when the validated token does not have the `kid` claim)
 - Make `JwtToVerifiableCredentialTransformer` and `JwtToVerifiablePresentationTransformer` more robust by also handling the cases where `credentialSubject` and `credentialStatus` are list!
 - Fix missing `credentialSubject.id` mapping in `JwtToVerifiableCredentialTransformer`
 - Fix the serialization of `VerifiableCredential` `type` field: current implem was giving `types` instead of `type` (see spec: https://www.w3.org/TR/vc-data-model-2.0/#example-a-simple-example-of-a-verifiable-presentation)
 
## Why it does that

Fix.

## Further notes

## Linked Issue(s)

### DIFF
--- a/extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IdentityAndTrustExtension.java
+++ b/extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IdentityAndTrustExtension.java
@@ -14,7 +14,6 @@
 
 package org.eclipse.edc.iam.identitytrust.core;
 
-import com.nimbusds.jwt.JWTClaimNames;
 import jakarta.json.Json;
 import org.eclipse.edc.iam.did.spi.resolution.DidPublicKeyResolver;
 import org.eclipse.edc.iam.did.spi.resolution.DidResolverRegistry;
@@ -37,7 +36,6 @@ import org.eclipse.edc.runtime.metamodel.annotation.Setting;
 import org.eclipse.edc.security.signature.jws2020.JwsSignature2020Suite;
 import org.eclipse.edc.spi.http.EdcHttpClient;
 import org.eclipse.edc.spi.iam.IdentityService;
-import org.eclipse.edc.spi.result.Result;
 import org.eclipse.edc.spi.system.ServiceExtension;
 import org.eclipse.edc.spi.system.ServiceExtensionContext;
 import org.eclipse.edc.spi.types.TypeManager;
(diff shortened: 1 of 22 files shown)


extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IdentityAndTrustExtension.java
```diff
@@ -125,12 +124,7 @@ public void initialize(ServiceExtensionContext context) {
         rulesRegistry.addRule(IATP_SELF_ISSUED_TOKEN_CONTEXT, new AccessTokenNotNullRule());
 
         // add all rules for validating VerifiableCredential JWTs
-        rulesRegistry.addRule("iatp-vc", (toVerify, additional) -> Optional.ofNullable(toVerify.getStringClaim(JWTClaimNames.SUBJECT)).map(s ->
```
bscholtes1A, line 128: There is not `sub` in VC

spi/common/identity-trust-spi/src/main/java/org/eclipse/edc/identitytrust/model/VerifiableCredential.java
```diff
@@ -41,6 +42,8 @@ public class VerifiableCredential {
     public static final String VERIFIABLE_CREDENTIAL_PROOF_PROPERTY = "https://w3id.org/security#proof";
     private List<CredentialSubject> credentialSubject = new ArrayList<>();
     private String id; // must be URI, but URI is less efficient at runtime
+
+    @JsonProperty("type")
```
paullatzelsperger, line 46: nit: should we do the same for `credentialSubject`? i.e. call the field `credentialSubjects` (plural), but annotating with a JsonProperty to conform to the spec

paullatzelsperger (approved):
:rocket:

spi/common/identity-trust-spi/src/main/java/org/eclipse/edc/identitytrust/model/VerifiableCredential.java
bscholtes1A: done

PR #3912, opened 2024-02-22
feat(dsp): expose version endpoint

### LINKED ISSUE #3842: DSP 1.0: Support `/version` endpoint
# Feature Request

Ref. 
- https://docs.internationaldataspaces.org/ids-knowledgebase/v/dataspace-protocol/common-functionalities/common.protocol

This means that we'll need a version mechanism on the DSP endpoints exposed, not part of this issue, for the moment we will just expose `1.0` endpoints and there will be no support for older versions.

dsp

dsp compliance

## Solution Proposal

ndr-brt:
## What this PR changes/adds

Add modules to expose the `.well-known/dspace-version` endpoint.
the design follows the standard way, so:
controller -> service -> registry

then the versions are registered on the registry by the other dsp endpoint extensions

## Why it does that

DSP compliance

## Further notes

- the current "not versioned" endpoints are still exposed for backward compatibility
- the versioned `2024/1` are the same exact endpoints as the "not versioned" ones
- no additional work has been done to support multiple different versions, it will be done when it'll be needed.

## Linked Issue(s)

Closes #3842


wolf4ood (approved):
LGTM :rocket:

PR #3915, opened 2024-02-23
feat: implement DataPlaneAccessTokenService

### LINKED ISSUE #3895: DPS: implement `DataPlaneAccessTokenService`
# Feature Request

The `DataPlaneAccessTokenService` has to basic tasks:
- generating a `TokenRepresentation` based on `TokenParameters`, using the `TokenGenerationService`
- resolving an `AccessTokenData` object based on an incoming token string, using the `AccessTokenDataStore` (#3893)

```java
public interface DataPlaneAccessTokenService {
   Result<TokenRepresentation> obtainToken(TokenParameters parameters, DataAddress address);
   Result<AccessTokenData> resolve(String token);
}
```

Note that when the `TokenRepresentation` is created, a corresponding entry in the `AccessTokenDataStore` must be created, marking the token as "valid".

The `TokenRepresentation` is a minimal JWT, that contains the following claims:
- `jti`
- `iss` + `sub`: the `participantId` of the organization to which the issuing data plane belongs
- `iat`

in addition, the JWT _must_ have contain the `kid` and `alg` headers.

All other headers should be ignored.

Dataplane modules

paullatzelsperger:
## What this PR changes/adds

This PR adds an implementation for the `DataPlaneAccessTokenService`, that is based on JsonWebTokens (JWTs).

When new tokens are issued the context data they are associated with (`DataAddress`) is persisted in the `AccessTokenDataStore` using the token ID ("jti").

Then, when tokens are restored (via `resolve()`) they are verified using the `TokenValidationService`:
- verify signature
- assert `sub == iss`
- assert `jti != null`

Then, the `jti` claim is used as token ID to resolve the context data (=`DataAddress`) from storage. If no entry for that token ID is found, the token is rejected.

## Why it does that

DataPlane Signaling

## Further notes

## Linked Issue(s)

Closes #3895


core/data-plane/data-plane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/DataPlaneDefaultIamServicesExtension.java
```diff
@@ -0,0 +1,91 @@
+/*
+ *  Copyright (c) 2022 Microsoft Corporation
```
bscholtes1A, line 2: SHould be 2024 no?

bscholtes1A (approved):

core/data-plane/data-plane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/DataPlaneDefaultIamServicesExtension.java
paullatzelsperger: and it should not be Microsoft :) fixed

wolf4ood (approved):

PR #3939, opened 2024-02-29
feat: introduce DPS-API JSON-LD transformer registry

paullatzelsperger:
## What this PR changes/adds

This PR introduces the `data-plane-signaling-transform` module, which contains JSON-LD transformers specific to the DP-Signaling API as well as
an explicitly typed `SignalingApiTransformerRegistry`.

## Why it does that

The Signaling API will employ different serialization formats than e.g. the DSP APIs or the management API. Thus, we need to contain
the transformers to that particular HTTP context.

## Further notes

## Linked Issue(s)

### DIFF
--- a/extensions/data-plane/data-plane-signaling/data-plane-signaling-api-configuration/src/main/java/org/eclipse/edc/connector/api/signaling/configuration/SignalingApiConfigurationExtension.java
+++ b/extensions/data-plane/data-plane-signaling/data-plane-signaling-api-configuration/src/main/java/org/eclipse/edc/connector/api/signaling/configuration/SignalingApiConfigurationExtension.java
@@ -14,19 +14,30 @@
 
 package org.eclipse.edc.connector.api.signaling.configuration;
 
+import jakarta.json.Json;
+import org.eclipse.edc.connector.api.signaling.transform.SignalingApiTransformerRegistry;
+import org.eclipse.edc.connector.api.signaling.transform.SignalingApiTransformerRegistryImpl;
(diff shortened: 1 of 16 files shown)


extensions/data-plane/data-plane-signaling/data-plane-signaling-api-configuration/src/main/java/org/eclipse/edc/connector/api/signaling/configuration/SignalingApiConfigurationExtension.java
```diff
@@ -78,4 +91,16 @@ public void initialize(ServiceExtensionContext context) {
         webService.registerResource(webServiceConfiguration.getContextAlias(), new ObjectMapperProvider(jsonLdMapper));
         webService.registerResource(webServiceConfiguration.getContextAlias(), new JerseyJsonLdInterceptor(jsonLd, jsonLdMapper, SIGNALING_SCOPE));
     }
+
+    @Provider
+    public SignalingApiTransformerRegistry managementApiTypeTransformerRegistry() {
+        var factory = Json.createBuilderFactory(Map.of());
+
```
wolf4ood, line 98: Shouldn't the provider be in the `data-plane-signaling-transform` ? We would need  `SignalingApiTransformerRegistry` injected for also for the `signaling` client

paullatzelsperger: then it should the a provider in an extension on the client side IMO

wolf4ood (approved):

PR #3951, opened 2024-03-04
feat: implement DataPlaneSignalingAPI

### LINKED ISSUE #3911: DPS: implement the Data plane Signaling API
# Feature Request

Implements a new controller for the Data plane signaling API

The definition of the Signaling API and the OpenAPI spec has already been addressed in [this](https://github.com/eclipse-edc/Connector/pull/3906) PR. 

dpf

data plane signaling

_depends on https://github.com/eclipse-edc/Connector/issues/3934_

paullatzelsperger:
## What this PR changes/adds

This PR implements the actual API controller of the Signaling API that delegates down to collaborator services

## Why it does that

Full functionality of the signaling API
## Further notes

- terminating `DataFlows` for which no source was found, is now possible. Previously that would have resulted in a failure.
- terminating `DataFlows` now takes an optional `reason` argument that gets persisted in the data flow entity.

## Linked Issue(s)

Closes #3911


extensions/data-plane/data-plane-signaling/data-plane-signaling-api/build.gradle.kts
```diff
@@ -24,8 +24,11 @@ dependencies {
     api(project(":spi:data-plane:data-plane-spi"))
 
     implementation(project(":extensions:common:api:control-api-configuration"))
+    implementation(project(":extensions:data-plane:data-plane-signaling:data-plane-signaling-transform"))
     implementation(project(":extensions:data-plane:data-plane-signaling:data-plane-signaling-api-configuration"))
     implementation(libs.jakarta.rsApi)
+    implementation(libs.restAssured)
```
ndr-brt, line 30:
```suggestion
    testImplementation(libs.restAssured)
```

spi/data-plane/data-plane-spi/src/main/java/org/eclipse/edc/connector/dataplane/spi/DataFlow.java
```diff
@@ -103,8 +107,12 @@ public void transitToNotified() {
         transitionTo(NOTIFIED.code());
     }
 
-    public void transitToTerminated() {
+    public void transitToTerminated(@Nullable String reason) {
         transitionTo(TERMINATED.code());
+        if (reason != null) {
+            properties = new HashMap<>(properties); //properties could be immutable
```
ndr-brt, line 113: this shoulnd't be the case, I would change the instantiation at line 47 and the way the `properties` builder behaves, otherwise we'll need to instantiate a new map every time we need to add a property (in this object this is actually the only place where we do that, but generally speaking an entity should be mutable by default.)

ndr-brt (approved):

wolf4ood (approved):

Issue #3901, opened 2024-02-21
DPS: Implement a new `DataFlowController` compliant with the signaling protocol


wolf4ood:
Currently we have two DataFlowController.

- `ConsumerPullTransferDataFlowController`
- `ProviderPushTransferDataFlowController`

For backward compatibility for now we will keep them but we need to implenent a new `DataFlowController` that speaks and handle the DPS protocol. 

Ref to the data plane signaling section [here](https://github.com/eclipse-edc/Connector/blob/main/docs/developer/data-plane-signaling/data-plane-signaling.md#3-dataflowcontroller)

PR #3952, opened 2024-03-04
feat: introduce allowedTransferTypes in DataPlaneInstance

### LINKED ISSUE #3943: DPS: DataPlane selection refactoring
# Feature Request

The data plane selection process should be refactor according the data plane signaling format mapping https://github.com/eclipse-edc/Connector/pull/3924

wolf4ood:
## What this PR changes/adds

Introduces `allowedTransferTypes` in DataPlaneInstance.

A new `select` method has been added to `DataPlaneSelectorService` to support selection based also on the `transferType`. If it's  not present the behaviour will be as today only inspecting source/destination.

The current `DataFlowController`s don't use the `transferType` for selection. For the data plane signaling a new `DataFlowController` will be introduced #3901 that will take into account the `transferType` for dispatching the request to the right data plane.  

## Why it does that

enabling data plane signaling protocol for selecting the correct data plane instance

## Linked Issue(s)

Closes #3943

### DIFF
--- a/extensions/data-plane-selector/data-plane-selector-api/src/main/java/org/eclipse/edc/connector/dataplane/selector/transformer/JsonObjectToDataPlaneInstanceTransformer.java
+++ b/extensions/data-plane-selector/data-plane-selector-api/src/main/java/org/eclipse/edc/connector/dataplane/selector/transformer/JsonObjectToDataPlaneInstanceTransformer.java
@@ -32,6 +32,7 @@ import java.util.stream.Collectors;
 
 import static org.eclipse.edc.connector.dataplane.selector.spi.instance.DataPlaneInstance.ALLOWED_DEST_TYPES;
 import static org.eclipse.edc.connector.dataplane.selector.spi.instance.DataPlaneInstance.ALLOWED_SOURCE_TYPES;
+import static org.eclipse.edc.connector.dataplane.selector.spi.instance.DataPlaneInstance.ALLOWED_TRANSFER_TYPES;
 import static org.eclipse.edc.connector.dataplane.selector.spi.instance.DataPlaneInstance.Builder;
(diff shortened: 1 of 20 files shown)


extensions/data-plane-selector/data-plane-selector-api/src/main/java/org/eclipse/edc/connector/dataplane/selector/transformer/JsonObjectToDataPlaneInstanceTransformer.java
```diff
@@ -71,6 +72,10 @@ private void transformProperties(String key, JsonValue jsonValue, DataPlaneInsta
                 var set = jsonValue.asJsonArray().stream().map(jv -> transformString(jv, context)).collect(Collectors.toSet());
                 builder.allowedSourceTypes(set);
             }
+            case ALLOWED_TRANSFER_TYPES -> {
+                var set = jsonValue.asJsonArray().stream().map(jv -> transformString(jv, context)).collect(Collectors.toSet());
```
paullatzelsperger, line 76: wouldn't it be more robust to use `transformArray` or `transformArrayOrObject` instead? IIRC they should be able to handle string arrays

paullatzelsperger (approved):

extensions/data-plane-selector/data-plane-selector-api/src/main/java/org/eclipse/edc/connector/dataplane/selector/transformer/JsonObjectToDataPlaneInstanceTransformer.java
wolf4ood: Not sure we handle correctly a `transforArray` or `transformArrayOrObject` with `String.class`

ndr-brt (approved):

PR #4005, opened 2024-03-14
refactor: make the annotation name for PostgreSQL tests consistent with the tag

### LINKED ISSUE #4001: The tag for PostgreSQL integration test should be consistent with the annotation name
# Bug Report

## Describe the Bug

[The tag is PostgresqlIntegrationTest while the annotation is PostgresqlDbIntegrationTest](https://github.com/eclipse-edc/Connector/blob/v0.5.1/core/common/junit/src/main/java/org/eclipse/edc/junit/annotations/PostgresqlDbIntegrationTest.java#L33-L34).

### Expected Behavior

It would be better to have consistent name as [done for IntegrationTest annotation](https://github.com/eclipse-edc/Connector/blob/v0.5.1/core/common/junit/src/main/java/org/eclipse/edc/junit/annotations/IntegrationTest.java#L31-L32) in order to keep users from setting invalid option like `-DincludeTags=PostgresqlDbIntegrationTest`.

## Possible Implementation

renaming the PostgresqlDbIntegrationTest to PostgresqlIntegrationTest.

iwasakims:
## What this PR changes/adds

renaming PostgresqlDbIntegrationTest to PostgresqlIntegrationTest as refactoring.

## Why it does that

using different name for the annotation and the tag is confusing for developers.

## Linked Issue(s)

Closes #4001


ndr-brt (approved):
thanks

PR #4009, opened 2024-03-15
refactor: inline `DataRequest` value object into `TransferProcess`

### LINKED ISSUE #3732: refactor: inline `DataRequest` in `TransferProcess`
# Feature Request

the `DataRequest` value object is a leftover from the old IDS protocol spec, but in fact it does not represent anything in our domain model, so we should "inline" it (moving all its fields in the `TransferProcess`)

transfer

cleanup

## Solution Proposal

first we'll need to change the signatures types of the `ResourceManifestGenerator` and `*ResourceDefinitionGenerator`  from `DataRequest` to `TransferProcess`

ndr-brt:
## What this PR changes/adds

Inline `DataRequest` properties into `TransferProcess`
**BREAKING CHANGE** the sql schema related to the transfer process has been changed, please refer to the `transfer-process-store-sql` module readme for details

## Why it does that

`DataRequest` does not represent any specific concept in the domain model.

## Linked Issue(s)

Closes #3732 


wolf4ood (approved):
:rocket:

PR #4019, opened 2024-03-18
feat: implements revocation in DataPlaneAuthorizationService

### LINKED ISSUE #3997: DPS: add support for terminate message
When a `DataFlowTerminateMessage` reaches the data plane supporting the DPS protocol, currently we only 
stop the transfer if it's `PUSH` and mark the `DataFlow` terminated. In the `PULL` scenario we should interact
`DataPlaneAuthorizationService` in order to be able to revoke eventual tokens associated with the `DataFlow`

wolf4ood:
## What this PR changes/adds

implements revocation in DataPlaneAuthorizationService

## Why it does that

Notify the `DataPlaneAuthorizationService` that the transfer has been terminated or suspended
## Further notes

## Linked Issue(s)

Closes #3997

### DIFF
--- a/spi/data-plane/data-plane-spi/src/main/java/org/eclipse/edc/connector/dataplane/spi/iam/DataPlaneAccessTokenService.java
+++ b/spi/data-plane/data-plane-spi/src/main/java/org/eclipse/edc/connector/dataplane/spi/iam/DataPlaneAccessTokenService.java
@@ -51,4 +51,14 @@ public interface DataPlaneAccessTokenService {
      * @return A {@link AccessTokenData} that contains the original claims and the data resource ({@link DataAddress}. If the token could not be restored, a failure is returned.
      */
     Result<AccessTokenData> resolve(String token);
+
+    /**
+     * Revokes the {@link AccessTokenData} associated with the id
+     *
+     * @param transferProcessId The id of the {@link AccessTokenData}
+     * @param reason            The reason for the revocation
+     * @return Success if revoked, failure otherwise
+     */
+    Result<Void> revoke(String transferProcessId, String reason);
+
 }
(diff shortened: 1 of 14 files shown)


jimmarino (approved):

spi/data-plane/data-plane-spi/src/main/java/org/eclipse/edc/connector/dataplane/spi/iam/DataPlaneAccessTokenService.java
```diff
@@ -51,4 +51,14 @@ public interface DataPlaneAccessTokenService {
...
     Result<AccessTokenData> resolve(String token);
+
+    /**
+     * Revokes the {@link AccessTokenData} associated with the id
+     *
+     * @param transferProcessId The id of the {@link AccessTokenData}
+     * @param reason            The reason for the revocation
+     * @return Success if revoked, failure otherwise
+     */
+    Result<Void> revoke(String transferProcessId, String reason);
```
ndr-brt, line 62: is the reason needed here? It makes sense in the `TransferProcess` context because the counter part then knows why, but I'm not sure the data plane needs to know. In fact in the implementation is not used

ndr-brt (approved):

spi/data-plane/data-plane-spi/src/main/java/org/eclipse/edc/connector/dataplane/spi/iam/DataPlaneAccessTokenService.java
wolf4ood: For our internal impl we don't use it, but that might be useful for other impl? that's why i left it

Issue #4028, opened 2024-03-21
Improve library modularity and runtime separation


jimmarino:
# Overview

This issue is an umbrella issue for refactoring the codebase to improve library modularity and better separate the runtime subsystem from core connector services.

The `connector-spi` and `connector-core` modules have grown too large and have introduced a web of dependencies and tangles into the codebase. These dependencies will be cleaned up by introducing specific `spi` modules and library modules (suffixed with `-lib`) containing shared implementations.     

The goals of this PR are:

### Improve visibility boundaries
- SPI Modules define extension points, generally do not contain implementations
- Lib Modules contain implementations and other code that may be reused by multiple extensions
- Extensions contain implementations that should not be shared with other modules. Features these implementations provide may be accessed by other extensions via the SPI 

### Separate the runtime subsystem from connector services
- Make it possible to boot an EDC runtime without connector services. This will streamline downstream repositories such as Identity Hub and other projects that use the EDC as a lightweight runtime.

## Which Areas Would Be Affected?

The entire code base

## Why Is the Feature Desired?

Modularity

PR #4030, opened 2024-03-21
feat: Introduce boot-lib and http-lib modules  

### LINKED ISSUE #4029: Introduce `boot-lib` and `http-lib` modules
As part of #4028 the `boot-lib` and `http-lib` modules will be introduced. This will break the circular dependency between the `connector-core` and `junit` modules and allow the latter not to depend on the former.

jimmarino:
As described in #4028 and #4029 this PR introduces the `boot-lib` and `http-lib` modules. Specifically:

- It breaks the cyclic dependency between `connector-core` and `Junit` by moving `EdcHttpClientImpl` into the `http-lib` modules and `getFreePorts()` to a utility method in the core `util` module.
- Moves the `InMemoryVault` implementation to the `boot-lib` module so that it can be referenced without exposing visibility to core services
- Starts to clean up dependencies on `connector-core`. These will be refined in subsequent PRs.

## Linked Issue(s)

Closes #4029


wolf4ood (approved):
:broom:

PR #4044, opened 2024-03-25
feat: introduces properties provider for DataFlowStartMessage

### LINKED ISSUE #4043: DPS: extensions points to enrich DataFlowStartMessage
# Feature Request

we should introduce extension points to enrich `DataFlowStartMessage` and make it possible to vehicle additional custom properties into the DPS protocol

dpf, control plane

extensibility

wolf4ood:
## What this PR changes/adds

Allows adding additional/custom properties when sending the `DataFlowStartMessage` with the DPS protocol

A default provider is created, which returns an empty map

## Why it does that

extensibility

## Further notes

## Linked Issue(s)

Closes #4043


paullatzelsperger (approved):

extensions/control-plane/transfer/transfer-data-plane-signaling/src/test/java/org/eclipse/edc/connector/transfer/dataplane/flow/DataPlaneSignalingFlowControllerTest.java
```diff
@@ -350,4 +315,63 @@ private TransferProcess.Builder transferProcessBuilder() {
                 .counterPartyAddress("test.connector.address")
                 .dataDestination(DataAddress.Builder.newInstance().type("test").build());
     }
+
+    @Nested
+    class Suspend {
```
ndr-brt, line 320: I was late for review, but why put tests under private methods? Doesn't look correct

wolf4ood: which tests?

PR #4045, opened 2024-03-26
fix: add missing SQL mapping in `EndpointDataReferenceEntry`

bscholtes1A:
## What this PR changes/adds

Add missing mapping of `createdAt` and `id` for SQL `EnpointDataReferenceEntry`.

## Why it does that

Fix

## Further notes

## Linked Issue(s)


extensions/common/store/sql/edr-index-sql/src/main/java/org/eclipse/edc/connector/store/sql/edr/schema/postgres/EndpointDataReferenceEntryMapping.java
```diff
@@ -24,6 +24,8 @@
  */
 public class EndpointDataReferenceEntryMapping extends TranslationMapping {
     public EndpointDataReferenceEntryMapping(EndpointDataReferenceEntryStatements statements) {
+        add("id", statements.getAssetIdColumn());
```
wolf4ood, line 27: I think we don't need the `id` in the database the "id" is the `transfer_process_id`

wolf4ood: if we want to make available the search with `id` i would map it to `transfer_process_id` column

ndr-brt (changes requested):
It would be nice to see a test that demonstrates the need for this addition

PR #4052, opened 2024-03-27
feat: adds request context in the policy evaluation context

### LINKED ISSUE #4050: feat: policy scope extractor additional info
# Feature Request

When writing a scope extractor based on policies, the only information received in the `PolicyContext` is `RequestScope.Builder`

for adding scopes extracted from the policies. For some use case this is not enough for deciding if the scopes should be added or not. We should add more information about the request itself like the type of message of the request and the direction if its
egress or ingress call.

wolf4ood:
## What this PR changes/adds

Adds additional context information for policy scope extractors:

- Added `RequestContext` which carries the `RemoteMessage` and the direction `Egress`/`Ingress`
- The `RequestContext` is injected in:
  - `DspHttpRemoteMessageDispatcherImpl` (Egress)
  - `ProtocolTokenValidatorImpl` (Ingress)

## Why it does that

extensibility of scope extractors

## Further notes

For those message which can be sent from both side consumer/provider we should also carry in the future this information in the `RequestContext` for maximum flexibility

The `message` of `RequestContext` currently can be null in the `findById` scenario where there is no actual `RemoteMessage` involved

## Linked Issue(s)

Closes #4050


core/control-plane/control-plane-aggregate-services/src/test/java/org/eclipse/edc/connector/service/protocol/ProtocolTokenValidatorImplTest.java
```diff
@@ -51,19 +53,22 @@ void shouldVerifyToken() {
         when(identityService.verifyJwtToken(any(), any())).thenReturn(Result.success(claimToken));
         when(agentService.createFor(any())).thenReturn(participantAgent);
 
-        var result = validator.verify(tokenRepresentation, "scope", policy);
+        var result = validator.verify(tokenRepresentation, "scope", policy, mock());
```
paullatzelsperger, line 56: nit: not the biggest fan of mocking value objects. Maybe a `TestRemoteMessage` could be used?

jimmarino (approved):

core/control-plane/control-plane-aggregate-services/src/test/java/org/eclipse/edc/connector/service/protocol/ProtocolTokenValidatorImplTest.java
wolf4ood: Added a `TestMessage`

PR #4058, opened 2024-03-28
feat: console colors

paullatzelsperger:
## What this PR changes/adds

This PR adds colored console output to the `ConsoleMonitor`. This is achieved by inserting special unicode characters before log lines to access ANSI coloring.
Note that the _actual_ color depends on the color configuration in your terminal.

This behaviour is on by default, and it can be deactivated by passing the `--no-color` command line argument. This might be useful in certain environments, for example running in clustered environments, where log tools like [stern](https://github.com/stern/stern) are used, that do coloring on their own.

Also, some docker base images may not print colored output correctly, depending on the shell they use. There are ways to mitigate it, e.g. using the `tty: true` property in docker-compose, or passing the `--it` parameter to `docker run`.

## Why it does that

Better readability.

## Further notes

I had considered several ways to toggle the coloring: 
1. using a config property: does not work properly, because the default `ConsoleMonitor` is instantiated before the `ServiceExtensionContext`, so we'd have a chicken-and-egg-problem
2. using a dedicated `monitor-coloring` module that contains a custom monitor. Decided against it, because I wanted to be able to toggle within the same runtime.
3. passing a command line arg `--no-color`. Needs to be implemented by the runtime base class, but has the least impact on the code base. Also, central configuration of the monitor in test runtimes was possible. 
## Linked Issue(s)

### DIFF
--- a/spi/common/core-spi/src/main/java/org/eclipse/edc/spi/monitor/ConsoleMonitor.java
+++ b/spi/common/core-spi/src/main/java/org/eclipse/edc/spi/monitor/ConsoleMonitor.java
@@ -20,8 +20,6 @@ import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.function.Supplier;
 
-import static java.lang.String.format;
-
 /**
  * Default monitor implementation. Outputs messages to the console.
  */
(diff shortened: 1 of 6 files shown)


spi/common/core-spi/src/main/java/org/eclipse/edc/spi/monitor/ConsoleMonitor.java
```diff
@@ -76,16 +84,35 @@ public void debug(Supplier<String> supplier, Throwable... errors) {
 
     private void output(String level, Supplier<String> supplier, Throwable... errors) {
         var time = ZonedDateTime.now().format(DateTimeFormatter.ISO_LOCAL_DATE_TIME);
-        System.out.println(prefix + level + " " + time + " " + sanitizeMessage(supplier));
+        var colorCode = getColorCode(level);
```
ndr-brt, line 87: the `useColor` evaluation could be done here instead of in the `getColorCode` method:
```
        var colorCode = useColor ? getColorCode(level) : "";
        var resetCode = useColor ? ConsoleColor.RESET : "";
```

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/monitor/ConsoleMonitor.java
```diff
@@ -76,16 +84,35 @@ public void debug(Supplier<String> supplier, Throwable... errors) {
...
+        var resetCode = useColor ? ConsoleColor.RESET : "";
+
+        System.out.println(colorCode + prefix + level + " " + time + " " + sanitizeMessage(supplier) + resetCode);
         if (errors != null) {
             for (Throwable error : errors) {
```
ndr-brt, line 92: not on your changelog, but since you're here... `var`

ndr-brt (approved):
just nits, all good otherwise

PR #4065, opened 2024-03-29
chore: separate boot and connector modules and normalize module and package structure

### LINKED ISSUE #4028: Improve library modularity and runtime separation

jimmarino:
## What this PR changes/adds

This PR separates the boot runtime modules from connector modules. It also normalizes the module and package naming scheme to make it consistent across the codebase. The following rules are applied:

# Naming Conventions

Modules names adopt the following conventions:

- SPI modules must have an `-spi` suffix
- Library modules must have a `-lib` suffix
- Control plane modules MAY have a `control-plane-` prefix  (decided on a case-by-case basis)
- Data plane modules MAY have a `data-pane-` prefix (decided on a case-by-case basis)

All package names begin with `org.eclipse.edc` and follow the module name:
- Common modules append the module name to the package: `auth-basic` becomes `org.eclipse.edc.auth.basic`
- Control plane modules append `connector.controlplane` to the package: `control-plane-catalog` becomes `org.eclipse.edc.connector.controlplane.catalog`
- Data plane modules append `connector.dataplane` to the package:  `data-plane-http` becomes `org.eclipse.edc.connector.dataplane.http`
- Core modules may choose to include `core` in the package name or not
- `boot-spi` does not include `boot` in the package name (this is for historical reasons) 

# Layout Conventions
## /spi

Extension points and shared interfaces

### /spi/common

Extension points used by various runtimes.

## /core

Required extensions and libraries for various runtimes. These extensions implement basic functionality and are not designed to be substituted.
- Core modules may opt to include `core` in package name
#### /core/common 

Required extensions and libraries used across multiple runtimes. All common modules begin with `org.edc`.

### /core/control-plane 

Required extensions and libraries for control plane runtimes.
`org.edc.connector.controlplane`

### /core/data-plane 

Required extensions and libraries for data plane runtimes. 
`org.edc.connector.dataplane`
## /extensions

Optional extensions and libraries for various runtimes

### /common/extensions

Optional extensions and libraries used across multiple runtimes

### /extensions/control-plane

Optional extensions and libraries used in control plane runtime
### /extensions/data-plane/

Optional extensions and libraries used in data plane runtimes

## Linked Issue(s)

Relates to  #4028


paullatzelsperger (approved):
I went through it line-by-line, and all seems good.

PR #4091, opened 2024-04-05
refactor(test): Refactor QueryValidator usage in tests

### LINKED ISSUE #3731: refactor(test): improve `QueryValidator` testing strategy
# Feature Request

Currently the `QueryValidator` is tested integrated within the services, that's not a great approach.
It would be better to separate the responsibility (test the behavior in the validator test and the interaction in the service test)

tests

avoid unnecessary integration tests

## Solution Proposal

enchobelezirev:
## What this PR changes/adds

* Fix usages of `QueryValidator` in tests
* Extracted the parameterized tests for validating query parameters in a @Nested classes inside each test
* Created a Provider class for providing the subtypes for each Service class when creating the QueryValidator

## Why it does that

* Enhances the tests under `control-plane-aggregate-services` project

## Further notes

*Packages org.eclipse.edc.connector.controlplane.services.contractagreement and their test packages

## Linked Issue(s)

Closes #3731 


paullatzelsperger:
@enchobelezirev can you please sign the ECA?

enchobelezirev:
Hello @paullatzelsperger, I have fixed it :)

ndr-brt (changes requested):
the point in the issue was to avoid this "integration test" thing and test validators and services separately. please provide unit tests for the validators separated from the services ones

enchobelezirev:
Hello,

I have seen that there is already defined class for testing the `QueryValidator` called `QueryValidatorTest`. I thought that the idea behind the issue was to have main logic of the services tested(including the "search" methods) by working with a mock object of `QueryValidator`. That is the reason why I have added mock object inside the tests of the services which is just returning the result without executing the actual logic of the `QueryValidator`. 
As for the "integration tests", I thought that testing the logic which is using the QueryValidator is part of the tests of the services and is not needed as a separate tests classes. This is the reason why I have left the "integrationtest" inside each test of the service, as a Nested tests.

Please, correct me if I am wrong or I do not understand something. I am new to the project and there is high chance that I misunderstood the issue.

Thanks!

ndr-brt (commented):
this is not about the project but to keep a good testing strategy, here we have 2 components tested together in the same integration test for no reason. Dependency injection is used to make classes being testable by their own apart from specific cases, otherwise we'll end up bloated by tons of (harder to maintain) integration tests.

The issue is pretty clear about the desired outcome: "avoid unnecessary integration tests". The validator should be unit tested, since it's dependent from the object type it's validating, there should be a "generic behavior" test (that as you said it already exist, and then some tests for every class that's used (behaviors that currently are verified in the service tests and just need to be moved out somewhere else)

enchobelezirev:
@ndr-brt, I will see how to modify the changes so that it complies with the issue.

core/control-plane/control-plane-aggregate-services/src/test/java/org/eclipse/edc/connector/controlplane/services/asset/AssetServiceImplIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.eclipse.edc.spi.query.Criterion.criterion;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+class AssetServiceImplIntegrationTest {
```
ndr-brt, line 42: As I already said in the past 2 reviews, please avoid integration tests and transform them in unit tests on the validator.

ndr-brt (changes requested):

PR #4108, opened 2024-04-11
chore: add support for fetching properties w/wo namespace + bitstring

wolf4ood:
## What this PR changes/adds

- Adds the API for fetching properties from CredentialSubject and CredentialStatus with the tuple (namespace, property)
- Removes the usage of Bitset and implemented custom Bitstring for parsing the bitstring encoded list and checking if an index is `1`

### DIFF
--- a/spi/common/boot-spi/src/main/java/org/eclipse/edc/spi/result/Result.java
+++ b/spi/common/boot-spi/src/main/java/org/eclipse/edc/spi/result/Result.java
@@ -22,6 +22,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 /**
  * A generic result type.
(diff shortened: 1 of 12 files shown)


spi/common/boot-spi/src/main/java/org/eclipse/edc/spi/result/Result.java
```diff
@@ -48,6 +49,21 @@ public static <T> Result<T> failure(List<String> failures) {
...
     }
 
+
+    /**
+     * Runs the block provided by the {@link Supplier} and wrap the return into a Result
+     *
+     * @param supplier The block to execute
+     * @return The Result of the supplier call. Success if no {@link Exception} were thrown. Failure otherwise
+     */
+    public static <T> Result<T> ofFallible(Supplier<T> supplier) {
```
paullatzelsperger, line 59: maybe `ofThrowable` would convey better what is expected here.

paullatzelsperger (approved):

PR #4138, opened 2024-04-26
feat: add a new management api for secrets

### LINKED ISSUE #4090: Management API for Secrets
# Feature Request

As discussed during committer group session, there is currently no management API enabling a participant to add/delete/update secrets from the Vault. In some cases, participant might have a direct access to the Vault itself to do the operations. But if an operator manages the connector for another entity (connector-as-a-service), then an API access is required for the participant to manage its secrets in self-service mode.

Management API of the Control Plane.

In cases where a participant connector is managed by another entity (connector-as-a-service), then a management API for the secrets is required for the participant to be able to manage its secrets in self-service mode.

## Solution Proposal

A `SecretApi` will be introduced which will be very similar to the other management APIs (assets, policy definitions...), with the exception that the secret API will not be included in the management API BOM.

sbeuzit06:
## What this PR changes/adds

This PR adds a new management API that enables to manage secrets from the vault.

## Why it does that

In case of the connector is hosted on behalf of another entity, this entity might not have access to the underlying vault. In that case, this API can be used to add or remove secrets from the vault.

## Linked Issue(s)

Closes #4090


spi/control-plane/secret-spi/src/main/java/org/eclipse/edc/connector/secret/spi/event/SecretEvent.java
```diff
@@ -0,0 +1,63 @@
...
+
+        public abstract B self();
+
+        public B secretId(String secretId) {
+            event.secretId = secretId;
+            return self();
+        }
+
+        public T build() {
+            Objects.requireNonNull(event.secretId);
```
paullatzelsperger, line 59: nit: should probably add some description as second argument to `requireNonNull`, something like "Mandatory value secretId missing" or similar

core/common/lib/transform-lib/src/test/java/org/eclipse/edc/transform/transformer/edc/from/JsonObjectFromSecretTransformerTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+class JsonObjectFromSecretTransformerTest {
+
+    private final JsonBuilderFactory jsonFactory = Json.createBuilderFactory(Map.of());
+    private final TransformerContext context = mock();
+
+    private final JsonObjectFromSecretTransformer transformer = new JsonObjectFromSecretTransformer(jsonFactory);
+
+    @Test
+    void transform_returnJsonObject() {
```
paullatzelsperger, line 41: please add some negative tests, like when the secret id is missing, etc. Generally, transformers should not have any validation, as that is handled elsewhere, so we expect null values then (as opposed to: exceptions)

core/common/lib/transform-lib/src/test/java/org/eclipse/edc/transform/transformer/edc/to/JsonObjectToSecretTransformerTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @BeforeEach
+    void setUp() {
+        transformer = new JsonObjectToSecretTransformer();
+    }
+
+    @Test
+    void transform_returnSecret() {
+
+        var secret = jsonFactory.createObjectBuilder()
```
paullatzelsperger, line 53: please add some negative tests here, like missing values should raise exceptions in the Secret builder

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/types/domain/secret/Secret.java
```diff
@@ -0,0 +1,92 @@
...
+
+import static org.eclipse.edc.spi.constants.CoreConstants.EDC_NAMESPACE;
+
+
+/**
+ * The {@link Secret} contains the metadata and describes the data itself or a collection of data.
+ */
+@JsonDeserialize(builder = Secret.Builder.class)
+public class Secret extends Entity {
+    public static final String PROPERTY_ID = EDC_NAMESPACE + "id";
```
paullatzelsperger, line 33: generally, we tend to call the ID of a secret the "alias", at least in consumer-facing APIs like configuration. Maybe calling it `alias` here may be worth considering. 
Just a suggestion though.

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/types/domain/secret/Secret.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public Builder toBuilder() {
+        // No key, Id is used as vault key
+        return Builder.newInstance()
+                .value(value);
+    }
+
+    @JsonPOJOBuilder(withPrefix = "")
+    public static class Builder extends Entity.Builder<Secret, Builder> {
```
paullatzelsperger, line 56: do we expect this to be extensible? if not, i would consider making the constructor private, and possibly making the Builder class `final`

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/types/domain/secret/Secret.java
```diff
@@ -0,0 +1,92 @@
...
+
+        @Override
+        public Secret build() {
+            super.build();
+
+            if (entity.getId() == null) {
+                id(UUID.randomUUID().toString());
+            }
+
+            Objects.requireNonNull(entity.value, "`value` is missing");
```
paullatzelsperger, line 86: careful with backticks! single quotes "'" are almost always better, and we use them everywhere else.

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApi.java
```diff
@@ -0,0 +1,148 @@
...
+                    @ApiResponse(responseCode = "204", description = "Secret was deleted successfully"),
+                    @ApiResponse(responseCode = "400", description = "Request was malformed, e.g. id was null",
+                            content = @Content(array = @ArraySchema(schema = @Schema(implementation = ApiCoreSchema.ApiErrorDetailSchema.class)))),
+                    @ApiResponse(responseCode = "404", description = "A secret with the given ID does not exist",
+                            content = @Content(array = @ArraySchema(schema = @Schema(implementation = ApiCoreSchema.ApiErrorDetailSchema.class)))),
+                    // TODO: check if it makes sense to check id secret is referenced from any asset
+                    @ApiResponse(responseCode = "409", description = "The secret cannot be deleted, because it is referenced by an asset",
+                            content = @Content(array = @ArraySchema(schema = @Schema(implementation = ApiCoreSchema.ApiErrorDetailSchema.class))))
+            })
+    void removeSecret(String id);
```
paullatzelsperger, line 91: We do not have any domain relation between `Assets`, `ContractAgreements` and secrets. If one should be introduced, this requires discussion in - and approval of - the technical committers and must be preceded by a decision-record. 

In addition, if we wanted to do that (which I do not), we would need to make sure that secrets are cleaned up when the parent entity, e.g. Assets, are deleted, etc.

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApi.java
```diff
@@ -0,0 +1,148 @@
...
+            requestBody = @RequestBody(
+                    content = @Content(schema = @Schema(implementation = ApiCoreSchema.QuerySpecSchema.class))
+            ),
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "The secrets matching the query",
+                            content = @Content(array = @ArraySchema(schema = @Schema(implementation = SecretOutputSchema.class)))),
+                    @ApiResponse(responseCode = "400", description = "Request body was malformed",
+                            content = @Content(array = @ArraySchema(schema = @Schema(implementation = ApiCoreSchema.ApiErrorDetailSchema.class))))
+            })
+    JsonArray requestSecrets(JsonObject querySpecJson);
```
paullatzelsperger, line 64: I do not think we should offer a way to do broad secret searches, first because our `Vault` SPI does not provide that, and second because I don't see the use case for that.

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApiController.java
```diff
@@ -0,0 +1,136 @@
...
+                .peek(r -> r.onFailure(f -> monitor.warning(f.getFailureDetail())))
+                .filter(Result::succeeded)
+                .map(Result::getContent)
+                .collect(toJsonArray());
+    }
+
+    @GET
+    @Path("{id}")
+    @Override
+    public JsonObject getSecret(@PathParam("id") String id) {
```
paullatzelsperger, line 108: putting the ID in the path will cause it to be transmitted unencrypted. I would transmit the secret ID in the body. same for the `delete` api.

spi/control-plane/control-plane-spi/src/main/java/org/eclipse/edc/connector/spi/service/SecretService.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    Secret findById(String secretId);
+
+    /**
+     * Search Secrets
+     *
+     * @param query the query
+     * @return the collection of secrets that matches the query
+     */
+    ServiceResult<List<Secret>> search(QuerySpec query);
```
paullatzelsperger, line 39: see my comment in `SecretApi`. I don't think this makes sense.

data-protocols/dsp/dsp-http-api-configuration/src/main/java/org/eclipse/edc/protocol/dsp/http/api/configuration/DspApiConfigurationExtension.java
```diff
@@ -141,6 +143,7 @@ private void registerTransformers() {
         dspApiTransformerRegistry.register(new JsonObjectFromDataAddressDspaceTransformer(jsonBuilderFactory, mapper));
         dspApiTransformerRegistry.register(new JsonObjectFromQuerySpecTransformer(jsonBuilderFactory));
         dspApiTransformerRegistry.register(new JsonObjectFromCriterionTransformer(jsonBuilderFactory, mapper));
+        dspApiTransformerRegistry.register(new JsonObjectFromSecretTransformer(jsonBuilderFactory));
```
paullatzelsperger, line 146: The registration of the secrets transformers has nothing to do with the DSP protocol, so its registration should **not** be done here, as that would introduce a dependency from DSP -> Secrets. Better to do this in the `ManagementApiConfigurationExtension`, or the `SecretsApiExtension`

paullatzelsperger (changes requested):
A few general remarks:
- I think it should be called `SecretsApi` etc., i.e. using the plural, otherwise people could be mislead into thinking it's some sort of secret API
- I would generally avoid transmitting stuff in the request URL. secret aliases are not secrets per se, but I wouldn't take any chances. 
-  Please add an End-To-End Test. Feel free to take a look at the `AssetApiEndToEndTest` or the `CatalogApiEndToEndTest` etc. for an example. 

Other comments inline.

data-protocols/dsp/dsp-http-api-configuration/src/main/java/org/eclipse/edc/protocol/dsp/http/api/configuration/DspApiConfigurationExtension.java
ndr-brt: Agree and I would put it in the `SecretsApiExtension`

extensions/common/api/management-api-configuration/src/main/java/org/eclipse/edc/connector/api/management/configuration/ManagementApiConfigurationExtension.java
```diff
@@ -115,12 +117,15 @@ public void initialize(ServiceExtensionContext context) {
...
         managementApiTransformerRegistry.register(new JsonObjectFromCriterionTransformer(factory, jsonLdMapper));
+        managementApiTransformerRegistry.register(new JsonObjectFromSecretTransformer(factory));
 
         OdrlTransformersFactory.jsonObjectToOdrlTransformers(participantIdMapper).forEach(managementApiTransformerRegistry::register);
         managementApiTransformerRegistry.register(new JsonObjectToDataAddressTransformer());
         managementApiTransformerRegistry.register(new JsonObjectToQuerySpecTransformer());
         managementApiTransformerRegistry.register(new JsonObjectToCriterionTransformer());
         managementApiTransformerRegistry.register(new JsonObjectToAssetTransformer());
         managementApiTransformerRegistry.register(new JsonValueToGenericTypeTransformer(jsonLdMapper));
+        managementApiTransformerRegistry.register(new JsonObjectToSecretTransformer());
```
ndr-brt, line 128: as above, likely this would fit better in the `SecretsApiExtension`

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/secret/SecretServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+                .map(secretValue -> Secret.Builder.newInstance()
+                        .value(secretValue)
+                        .id(secretId)
+                        .build())
+                .orElse(null);
+    }
+
+    @Override
+    public ServiceResult<List<Secret>> search(QuerySpec query) {
+        throw new UnsupportedOperationException("Query operation is not supported for secrets");
```
ndr-brt, line 53: please just remove this method, it does not make any sense for secrets

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApi.java
```diff
@@ -0,0 +1,148 @@
...
+
+import static io.swagger.v3.oas.annotations.media.Schema.RequiredMode.REQUIRED;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.CONTEXT;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.ID;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.TYPE;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.EDC_SECRET_TYPE;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.EDC_SECRET_VALUE;
+
+@OpenAPIDefinition(
+        info = @Info(description = "This contains the secret management API, which allows a participant to manage the API keys associated to the API they expose.", title = "Secret API"))
```
ndr-brt, line 38: please rephrase this in a meaningful way

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApi.java
```diff
@@ -0,0 +1,148 @@
...
+@Tag(name = "Secret")
+public interface SecretApi {
+
+    @Operation(description = "Creates a new secret, as a key,value pair",
```
ndr-brt, line 42: nit: spacing

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApi.java
```diff
@@ -0,0 +1,148 @@
...
+    record SecretOutputSchema(
+            @Schema(name = ID)
+            String id,
+            @Schema(name = TYPE, example = EDC_SECRET_TYPE)
+            String type,
+            @Schema(name = EDC_SECRET_VALUE, requiredMode = REQUIRED)
+            String value,
+            long createdAt
+    ) {
+        // TODO: check key and value names
```
ndr-brt, line 135: please solve this TODO

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApi.java
```diff
@@ -0,0 +1,148 @@
...
+        public static final String SECRET_OUTPUT_EXAMPLE = """
+                {
+                    "@context": { "@vocab": "https://w3id.org/edc/v0.0.1/ns/" },
+                    "@id": "secret-id",
+                    "@type": "https://w3id.org/edc/v0.0.1/ns/Secret",
+                    "key": "secret-key",
+                    "value": "secret-value",
+                    "createdAt": 1688465655
```
ndr-brt, line 143: I'm not sure this field is actually returned

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApiController.java
```diff
@@ -0,0 +1,136 @@
...
+import static jakarta.json.stream.JsonCollectors.toJsonArray;
+import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON;
+import static java.util.Optional.of;
+import static org.eclipse.edc.spi.query.QuerySpec.EDC_QUERY_SPEC_TYPE;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.EDC_SECRET_TYPE;
+import static org.eclipse.edc.web.spi.exception.ServiceResultHandler.exceptionMapper;
+
+@Consumes(APPLICATION_JSON)
+@Produces(APPLICATION_JSON)
+@Path("/v3/secrets")
```
ndr-brt, line 49: why `v3`? please just use `v1`

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApiController.java
```diff
@@ -0,0 +1,136 @@
...
+                .orElseThrow(exceptionMapper(Secret.class, secret.getId()));
+
+        return transformerRegistry.transform(idResponse, JsonObject.class)
+                .orElseThrow(f -> new EdcException(f.getFailureDetail()));
+    }
+
+    @POST
+    @Path("/request")
+    @Override
+    public JsonArray requestSecrets(JsonObject querySpecJson) {
```
ndr-brt, line 86: not needed, please delete this method

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApiController.java
ndr-brt: as long as TLS is used, the URL will be encrypted, to me it's ok to keep it in the url.
also because the secret key is not really a "sensitive" value

extensions/control-plane/api/management-api/secret-api/src/test/java/org/eclipse/edc/connector/api/management/secret/validation/SecretValidatorTest.java
```diff
@@ -0,0 +1,84 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.ID;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.VALUE;
+import static org.eclipse.edc.junit.assertions.AbstractResultAssert.assertThat;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.EDC_SECRET_VALUE;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.PROPERTY_ID;
+
+class SecretValidatorTest {
+
+    private final Validator<JsonObject> validator = org.eclipse.edc.connector.api.management.secret.validation.SecretValidator.instance();
```
ndr-brt, line 33: nit: add import

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/types/domain/secret/Secret.java
ndr-brt: as I stated before, in the json-ld representation I'd use the `@id` term.

spi/common/core-spi/src/main/java/org/eclipse/edc/spi/types/domain/secret/Secret.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Objects;
+import java.util.UUID;
+
+import static org.eclipse.edc.spi.constants.CoreConstants.EDC_NAMESPACE;
+
+
+/**
+ * The {@link Secret} contains the metadata and describes the data itself or a collection of data.
+ */
+@JsonDeserialize(builder = Secret.Builder.class)
```
ndr-brt, line 31: Is jackson json serdes needed? on the api side we use dedicated transformers and in the vault they are stored in a key-value manner, I would remove these unused annotations

spi/control-plane/secret-spi/src/main/java/org/eclipse/edc/connector/secret/spi/observe/SecretObservableImpl.java
```diff
@@ -0,0 +1,23 @@
...
+ *       Amadeus - Initial API and Implementation
+ *
+ */
+
+package org.eclipse.edc.connector.secret.spi.observe;
+
+import org.eclipse.edc.spi.observe.ObservableImpl;
+
+/**
+ * Observes Secret livecycle state changes.
```
ndr-brt, line 20:
```suggestion
 * Observes Secret lifecycle state changes.
```

ndr-brt (changes requested):
please add some end-to-end tests (as done in `system-tests/management-api` for the other endpoints

rafaelmag110:
I have one concern regarding a possible exploit this API opens.
As is, an ill intentioned actor can use this to mess with secrets that have been registered by the operator of the connector, and are necessary for its proper functioning (public/private certificates, idp releated secrets, ...).

Two proposals that come to mind are: 
- Adding a config that enables an operator of the connector to config the API so a prefix can be added to the secrets created by the it.
- Adding the possibility to change the underlying vault directory where the secrets created by the API will be stored

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApi.java
bscholtes1A: Hello @paullatzelsperger 
we discussed that item a few weeks ago during the committer meeting already. I explained the use-case and we concluded it was ok to have it in the codebase as long as this secret API was not embedded by default in the management API BOM which is the case here. 
If needed, I can write a decision-record tough.

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApiController.java
paullatzelsperger: yeah, true, it's only the host part that goes unencrypted. I'm OK with leaving it in the path since that makes things a bit easier.

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApi.java
paullatzelsperger: having the API is fine, and we discussed that. Introducing a domain relation between Assets, Contracts and Secrets is something different entirely, which - to my recollection - was not discussed. 
In other words, I object to the 409 here, not the Secrets API per se.

paullatzelsperger:
> I have one concern regarding a possible exploit this API opens. As is, an ill intentioned actor can use this to mess with secrets that have been registered by the operator of the connector, and are necessary for its proper functioning (public/private certificates, idp releated secrets, ...).

That is true for any other Management API as well. Malicious actors could easily nuke their connector deployment if they delete stuff they shouldn't. 


> * Adding a config that enables an operator of the connector to config the API so a prefix can be added to the secrets created by the it.

not sure how that would help the supposed exploit

> * Adding the possibility to change the underlying vault directory where the secrets created by the API will be stored

Some powerful vault impls (Hashicorp Vault) support that, but some vaults, such as Azure KeyVault are more geared towards role-based-access-control, so there is no way that I know of to realize user-based access control on "subfolders".

rafaelmag110:
> That is true for any other Management API as well. Malicious actors could easily nuke their connector deployment if they delete stuff they shouldn't. Which is all the more reason to properly secure the Management API.

Isn't it a bit different? If someone who has access to the DMA mistakenly deletes a resource (i.e. asset or policy) its up to them to restore it. But if they are able to, say, delete the private certificate the connector uses to encrypt EDRs is a concern for the operator of said connector. At least from my point of view...

> not sure how that would help the supposed exploit

If the SecretsApi always prefixes its keys, it is unlikely that a malicious actor can access non prefixed keys, only if the operator made the mistake to prefix them in the same way.

paullatzelsperger:
> Isn't it a bit different? If someone who has access to the DMA mistakenly deletes a resource (i.e. asset or policy) its up to them to restore it. But if they are able to, say, delete the private certificate the connector uses to encrypt EDRs is a concern for the operator of said connector. At least from my point of view...

I don't think so, because if you give users access to a certain class of information, be they assets, policies or secrets, you're handing over full control. If an operational scenario requires that users have only _read_ access to secrets, but _write_ access should be blocked, that could be solved in the vault directly. Most vaults have some sort of RBAC.


> If the SecretsApi always prefixes its keys, it is unlikely that a malicious actor can access non prefixed keys, only if the operator made the mistake to prefix them in the same way.

I may not be understanding this fully, but it sounds like a very brittle and insecure solution, that could easily be circumvented by guessing the secret alias, or looking at EDC code.

bscholtes1A:
> A few general remarks:
> 
> * I think it should be called `SecretsApi` etc., i.e. using the plural, otherwise people could be mislead into thinking it's some sort of secret API
> * Please do not introduce domain relations (Secrets -> Assets) just out-of-hand. This is a more complicated topic, and one that must be discussed and designed first. If this is a requirement, then I suggest opening a scoped discussion first.
> * I would generally avoid transmitting stuff in the request URL. secret aliases are not secrets per se, but I wouldn't take any chances.
> * Please add an End-To-End Test. Feel free to take a look at the `AssetApiEndToEndTest` or the `CatalogApiEndToEndTest` etc. for an example.
> 
> Other comments inline.

Agree with all remarks. 
They have been all processed as per last commit.

core/common/lib/transform-lib/src/test/java/org/eclipse/edc/transform/transformer/edc/from/JsonObjectFromSecretTransformerTest.java
bscholtes1A: Not sure we add a lot here.
If we forget the `id` then we wont even be able to create a `Secret` and call the function. 
If we forget the `value` then nothing special will happen.

bscholtes1A:
> > Isn't it a bit different? If someone who has access to the DMA mistakenly deletes a resource (i.e. asset or policy) its up to them to restore it. But if they are able to, say, delete the private certificate the connector uses to encrypt EDRs is a concern for the operator of said connector. At least from my point of view...
> 
> I don't think so, because if you give users access to a certain class of information, be they assets, policies or secrets, you're handing over full control. If an operational scenario requires that users have only _read_ access to secrets, but _write_ access should be blocked, that could be solved in the vault directly. Most vaults have some sort of RBAC.
> 
> > If the SecretsApi always prefixes its keys, it is unlikely that a malicious actor can access non prefixed keys, only if the operator made the mistake to prefix them in the same way.
> 
> I may not be understanding this fully, but it sounds like a very brittle and insecure solution, that could easily be circumvented by guessing the secret alias, or looking at EDC code.

On top of what @paullatzelsperger mentioned, it is also worth emphasizing that such API will typically be exposed over the internet through an API gateway (such as Azure APIM) which gives you full freedom to create inbound policies for the incoming traffic so that only secrets with alias matching a given pattern can be updated, removed...

extensions/control-plane/api/management-api/secrets-api/src/main/java/org/eclipse/edc/connector/api/management/secrets/SecretsApi.java
```diff
@@ -0,0 +1,125 @@
...
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.TYPE;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.EDC_SECRET_TYPE;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.EDC_SECRET_VALUE;
+
+@OpenAPIDefinition(
+        info = @Info(description = "This contains the secret management API, which allows to add, remove and update secrets in the Vault.", title = "Secret API"))
+@Tag(name = "Secret")
+public interface SecretsApi {
+
+    @Operation(description = "Creates a new secret.",
```
paullatzelsperger, line 41: nit: pls add `operationId` properties, to avoid name clashes and help with OpenAPI generation

extensions/control-plane/api/management-api/secrets-api/src/main/java/org/eclipse/edc/connector/api/management/secrets/SecretsApi.java
```diff
@@ -0,0 +1,125 @@
...
+            @Schema(name = TYPE, example = EDC_SECRET_TYPE)
+            String type,
+            @Schema(name = EDC_SECRET_VALUE, requiredMode = REQUIRED)
+            String value
+    ) {
+        public static final String SECRET_INPUT_EXAMPLE = """
+                {
+                    "@context": { "@vocab": "https://w3id.org/edc/v0.0.1/ns/" },
+                    "@id": "secret-id",
+                    "value" : "secret-value"
```
paullatzelsperger, line 100: `@value` is a reserved JSON-LD keyword. Not sure if this could cause any troubles. Maybe @ndr_brt knows?

extensions/control-plane/api/management-api/secrets-api/src/test/java/org/eclipse/edc/connector/api/management/secrets/SecretsApiControllerTest.java
```diff
@@ -0,0 +1,324 @@
...
+        baseRequest()
+                .body(createSecretJson().build())
+                .contentType(JSON)
+                .post("/secrets")
+                .then()
+                .statusCode(409);
+    }
+
+    @Test
+    void createSecret_emptyAttributes() {
```
paullatzelsperger, line 186: isn't this the same as `..._whenValidationFails`?

extensions/control-plane/api/management-api/secrets-api/src/test/java/org/eclipse/edc/connector/api/management/secrets/SecretsApiControllerTest.java
```diff
@@ -0,0 +1,324 @@
...
+
+        baseRequest()
+                .contentType(JSON)
+                .delete("/secrets/not-existent-id")
+                .then()
+                .statusCode(404);
+    }
+
+    @Test
+    void deleteSecret_conflicts() {
```
paullatzelsperger, line 223: obsolete

extensions/control-plane/api/management-api/secrets-api/src/test/java/org/eclipse/edc/connector/api/management/secrets/SecretsApiTest.java
```diff
@@ -0,0 +1,84 @@
...
+import static org.eclipse.edc.connector.api.management.secrets.SecretsApi.SecretOutputSchema.SECRET_OUTPUT_EXAMPLE;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.ID;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.TYPE;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.VALUE;
+import static org.eclipse.edc.junit.assertions.AbstractResultAssert.assertThat;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.EDC_SECRET_TYPE;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.EDC_SECRET_VALUE;
+import static org.mockito.Mockito.mock;
+
+class SecretsApiTest {
```
paullatzelsperger, line 43: not exactly sure what this is supposed to test. If it tests the transformers, then it should be named accordingly and be annotated with `@ComponentTest`

extensions/control-plane/api/management-api/secrets-api/src/test/java/org/eclipse/edc/connector/api/management/secrets/validation/SecretsValidatorTest.java
```diff
@@ -0,0 +1,78 @@
...
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.VALUE;
+import static org.eclipse.edc.junit.assertions.AbstractResultAssert.assertThat;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.EDC_SECRET_VALUE;
+
+class SecretsValidatorTest {
+
+    private final Validator<JsonObject> validator = SecretsValidator.instance();
+
+    @Test
+    void shouldSucceed_whenValidInput() {
```
paullatzelsperger, line 35: naming pattern should be `<methodName>_when<Condition>[_then<ExpectedEvent>]`

core/common/lib/transform-lib/src/main/java/org/eclipse/edc/transform/transformer/edc/from/JsonObjectFromSecretTransformer.java
```diff
@@ -0,0 +1,47 @@
...
+import org.eclipse.edc.transform.spi.TransformerContext;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.ID;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.TYPE;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.EDC_SECRET_TYPE;
+import static org.eclipse.edc.spi.types.domain.secret.Secret.EDC_SECRET_VALUE;
+
+public class JsonObjectFromSecretTransformer extends AbstractJsonLdTransformer<Secret, JsonObject> {
```
paullatzelsperger, line 30: The transformers should be placed in the extension for the Secrets API, not in the core transform lib, as they are only used by the Secrets API, analogous to the `JsonObjectFromContractDefinitionTransformer`

.gitignore
```diff
@@ -42,6 +42,7 @@ distributions/**/github.properties
 
 .DS_Store
 **/secrets
+!**/secrets-api/**/secrets
```
paullatzelsperger, line 45: I think this is not needed.

paullatzelsperger (commented):
a few nits an niggles here and there

bscholtes1A:
> please add some end-to-end tests (as done in `system-tests/management-api` for the other endpoints

done

ndr-brt (changes requested):
there are still 6 unresolved comments on my previous review (and 1 from Paul's one).
please solve them

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApi.java
bscholtes1A: I don't get this one

extensions/control-plane/api/management-api/secrets-api/src/main/java/org/eclipse/edc/connector/api/management/secrets/SecretsApi.java
bscholtes1A: it does not seem to be a problem, at least from what we have experimented

extensions/control-plane/api/management-api/secret-api/src/main/java/org/eclipse/edc/connector/api/management/secret/SecretApi.java
ndr-brt: space missing after the second comma :)

extensions/control-plane/api/management-api/secrets-api/src/main/java/org/eclipse/edc/connector/api/management/secrets/SecretsApi.java
ndr-brt: I don't know XD maybe @wolf4ood 
it should be good anyway

paullatzelsperger: ok, then lets put it in, we can always fix it later should it be a problem.

extensions/control-plane/api/management-api/secrets-api/src/test/java/org/eclipse/edc/connector/api/management/secrets/SecretsApiTest.java
paullatzelsperger: mmh, thats true. lets put it in then, to have it consistent, but we should probably make all of these a bit more descriptive.

extensions/control-plane/api/management-api/secrets-api/src/main/java/org/eclipse/edc/connector/api/management/secret/validation/JsonObjectToSecretTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Amadeus - Initial API and Implementation
+ *
+ */
+
+package org.eclipse.edc.connector.api.management.secret.validation;
```
paullatzelsperger, line 15: wrong package

paullatzelsperger (approved):
one wrong package, but good for me otherwise

bscholtes1A:
> one wrong package, but good for me otherwise

fixed

ndr-brt (approved):
LGTM

PR #4152, opened 2024-05-02
chore: removes OAuth2 capabilities from the data sink

wolf4ood:
## What this PR changes/adds

removes OAuth2 capabilities from the data sink


extensions/data-plane/data-plane-http-oauth2-core/src/test/java/org/eclipse/edc/connector/dataplane/http/oauth2/DataPlaneHttpOauth2ExtensionTest.java
```diff
@@ -0,0 +1,47 @@
+/*
+ *  Copyright (c) 2023 Amadeus
```
paullatzelsperger, line 2: seems incorrect

paullatzelsperger (approved):

jimmarino (approved):

florianrusch-zf:
It would be great to know what the reason for the removal is. I have no argument against the removal, I'm just curious and the PR description is more or less non-existent...

jimmarino:
The reasons for this removal will be explained by the committers later.

florianrusch-zf:
@jimmarino where can I find the explanation?

jimmarino:
This was a security vulnerability that should be described in the release notes. You can look for it there.

PR #4159, opened 2024-05-03
fix: issuanceDate fallback in vc transformer

### LINKED ISSUE #4113: JwtToVerifiableCredentialTransformer considers wrong fallback claim for issuance date
In the current main branch we have identified a small bug.
[JwtToVerifiableCredentialTransformer ](https://github.com/eclipse-edc/Connector/blob/e4e71b9f087b06d670fa323d6b8cd9d0438e7078/extensions/common/iam/identity-trust/identity-trust-transform/src/main/java/org/eclipse/edc/iam/identitytrust/transform/to/JwtToVerifiableCredentialTransformer.java#L77) considers the wrong fallback claim for the issuance date and hence the expiry date is used instead. 

Currently it is:
```java
extractDate(vc.get(ISSUANCE_DATE_PROPERTY), claims.getExpirationTime()).ifPresent(builder::issuanceDate);
```

But is should be:
```java
extractDate(vc.get(ISSUANCE_DATE_PROPERTY), claims.getIssueTime()).ifPresent(builder::issuanceDate);
```

lholthof:
## What this PR changes/adds

In case a JWT's VC does not contain a claim `issuanceDate` the fallback is to take the date from the JWT itself. Currently the fallback picks the `exp` claim for this, which is applicable for expiration date but not for issuance date. Therefore the claim `iat`, the issued at time of the JWT, should be used instead.

## Why it does that

Misleading issuanceDate would be used, by setting it to the JWT's expiration date.

## Further notes

This is only a minor issue as it is only the fallback value.
Minor cleanup in the TestData.

## Linked Issue(s)

Closes #4113

### DIFF
--- a/extensions/common/iam/identity-trust/identity-trust-transform/src/main/java/org/eclipse/edc/iam/identitytrust/transform/to/JwtToVerifiableCredentialTransformer.java
+++ b/extensions/common/iam/identity-trust/identity-trust-transform/src/main/java/org/eclipse/edc/iam/identitytrust/transform/to/JwtToVerifiableCredentialTransformer.java
@@ -74,7 +74,7 @@ public class JwtToVerifiableCredentialTransformer extends AbstractJwtTransformer
                 extractDate(vc.get(EXPIRATION_DATE_PROPERTY), claims.getExpirationTime()).ifPresent(builder::expirationDate);
 
                 // issuance date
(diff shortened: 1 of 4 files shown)


jimmarino (approved):

jimmarino:
Thanks for the PR!

paullatzelsperger (approved):
LGTM - thanks!

PR #4166, opened 2024-05-08
feat: dataplane self-registration

### LINKED ISSUE #2534: Self Registration of Data Planes
# Feature Request

As discussed [here](https://github.com/eclipse-edc/Connector/discussions/2198) it would be cool to provide a way for a data plane to register itself into the available data plane instances. 

Data Plane Selector, DPF

In use cases where elasticity of data planes matters this could help spinning new data planes without the efforts manually maintaining all the data plane instances available.  

## Solution Proposal

We could write a new extensions that self register itself using the `selector` api based on a configuration.

If we support self registration we should also support de-registration if the instance
is not available. Probably we could have some sort of heartbeat mechanism that ensure that the registered instances are up and running.

ndr-brt:
## What this PR changes/adds

Give the data-plane "self registration" capabilities, through:
- `data-plane-self-registration` extension, that will obtain the supported src/dest/transfer types and call the data plane selector api
- `data-plane-selector-control-api` extension, that offers endpoints to permit the data-planes to interact with the control-plane, particularly, they offer "registration" and "unregistration" (that will provided in a subsequent issue/PR)

some diff is caused by moving stuff around and replacing the swagger plugin string with the reference to the version catalog.

## Why it does that

simplify data-plane management

## Further notes

- now the destination type is not used anymore to select dataplanes, because `transferType` already represent the handled types and it's already mandatory. in fact, the destination types field could be removed from `DataPlaneInstance` (upcoming issue)
- `addDataplane` endpoint on the management API has been deprecated (already removed usage from E2E tests)
- what was `turnCount` used for? now it isn't used by anyone, I deprecated it.
- `signaling` api context looks to be not necessary, their api should be exposed on the `control` one (upcoming issue).

## Linked Issue(s)

Closes #2534 


extensions/data-plane-selector/data-plane-selector-client/src/main/java/org/eclipse/edc/connector/dataplane/selector/RemoteDataPlaneSelectorService.java
```diff
@@ -123,38 +124,52 @@ public DataPlaneInstance select(DataAddress source, DataAddress destination, Str
 
     @Override
     public ServiceResult<Void> addInstance(DataPlaneInstance instance) {
-        throw new UnsupportedOperationException("not implemented");
-    }
+        var jsonObject = typeTransformerRegistry.transform(instance, JsonObject.class).getContent();
```
paullatzelsperger, line 127: here we should probably evaluate the result

extensions/data-plane-selector/data-plane-selector-client/src/main/java/org/eclipse/edc/connector/dataplane/selector/RemoteDataPlaneSelectorService.java
```diff
@@ -123,38 +124,52 @@ public DataPlaneInstance select(DataAddress source, DataAddress destination, Str
...
+                handleResponse(response, null, null);
+                return ServiceResult.success();
             }
+        } catch (IOException e) {
+            return ServiceResult.badRequest(e.getMessage());
+        }
+    }
 
-            return mapper.readValue(body, tr);
+    private <R> R handleResponse(Response response, TypeReference<? extends R> tr, R defaultValue) {
```
paullatzelsperger, line 147: how about returning a `Result<R>` here and evaluating that in client code, e.g. the `addInstance` method?

extensions/data-plane-selector/data-plane-selector-client/src/test/java/org/eclipse/edc/connector/dataplane/selector/RemoteDataPlaneSelectorServiceTest.java
```diff
@@ -109,7 +113,17 @@ void find_withTransferType() {
         var result = service.select(DataAddress.Builder.newInstance().type("test1").build(), DataAddress.Builder.newInstance().type("test2").build(), "random", "transferType");
 
         assertThat(result).usingRecursiveComparison().isEqualTo(expected);
+    }
+
+    @Test
+    @Disabled // this will be tested when there will be a clear separation between the selector management and control api
```
paullatzelsperger, line 119: active code should be tested. The `addInstance` method has an impl, so why not test it?

paullatzelsperger (commented):
the `turnCount` was originally intended to track how often each data plane was selected to serve a transfer, to implement a round-robin-like selector strategy.
Together with the `turnCount` I would put the entire selection strategy idea up for discussion - it seems dataplanes are selected by capability much more than by some high-level load-balancing strategy

extensions/data-plane-selector/data-plane-selector-client/src/main/java/org/eclipse/edc/connector/dataplane/selector/RemoteDataPlaneSelectorService.java
ndr-brt: totally agree, given the size of the PR I added that to my TODO list, but let's see if doing that will help me solve the other 2 comments

jimmarino:
> the `turnCount` was originally intended to track how often each data plane was selected to serve a transfer, to implement a round-robin-like selector strategy. Together with the `turnCount` I would put the entire selection strategy idea up for discussion - it seems dataplanes are selected by capability much more than by some high-level load-balancing strategy

Yes, from a control plane standpoint, that type of load balancing should be performed by the data plane "instance" itself. In other words, the control plane assumes a particular data plane instance is capable of handling those conditions.

extensions/data-plane-selector/data-plane-selector-client/src/test/java/org/eclipse/edc/connector/dataplane/selector/RemoteDataPlaneSelectorServiceTest.java
ndr-brt: the problem is that this class is tested in a "bizzarre" way, by loading the "server" controller, that, currently, is a different one for this method (because of the separation between management/control api that will be solved in subsequent PRs). Anyway this logic is tested by the E2E tests currently.
Let's see if can I do something anyway

ndr-brt:
@paullatzelsperger solved two comments, but those tests need to be changed, because the integration scope is definitely too broad (they depend to the actual controller class!), I added that to my TODO list will do it in the next PRs.

@jimmarino @paullatzelsperger so definitely the `turnCount` can be removed here, also because it is a value that can only be set at creation (that currently is made on the management-api), so it's definitely not useful.

paullatzelsperger (approved):

paullatzelsperger:
> @jimmarino @paullatzelsperger so definitely the `turnCount` can be removed here, also because it is a value that can only be set at creation (that currently is made on the management-api), so it's definitely not useful.

yep, makes sense

wolf4ood (approved):

mhellmeier:
To be honest, I wonder about this pull request. We volunteered to create such a DP self-registration extension. We followed your [CONTRIBUTING guidelines](https://github.com/eclipse-edc/docs/blob/main/CONTRIBUTING.md) and started different discussions on GitHub (see [here ](https://github.com/eclipse-edc/Connector/discussions/2188) or [here](https://github.com/eclipse-edc/Connector/discussions/2198)) and on Discord (see [here](https://discord.com/channels/973118216152973323/973118216152973326/1139172147185582180)).
The result of all these discussions with @paullatzelsperger , @ndr-brt and @jimmarino was that such an extension should not be implemented in this core repository due to specific deployments:

> TL;DR: either of those two solutions could be viable solutions, but due to their specific nature are not fit for general use, thus I don't think they should go in upstream EDC.

Source: @paullatzelsperger -- https://github.com/eclipse-edc/Connector/discussions/2198#discussioncomment-6720253

Since this project is an Eclipse open source project which should follow the open source best practices and the [Eclipse Foundation Project Handbook](https://www.eclipse.org/projects/handbook/): Where is this change in decision discussed and documented? Why isn't it discussed in the specific discussions linked above?

paullatzelsperger:
First of all, the discussions you mentioned are (as of today) a year and a half old, and in the world of Software Development this can mean a drastic change of landscape which is exactly what happened here. Furthermore, you can't seriously expect us to remember and relate all current goings-on to age-old discussions.

Second, could you please provide any reference, which forces the maintainers of a project to get in contact with and align with the original authors of a previously discarded idea, if said idea became viable at any time?

Third, this is quite a delicate are of code, so I'm sure you understand that we want this to be implemented by be someone, who is intimately familiar with the code base and has a proven track record of stellar contributions, for example, a committer. So the chances of getting something like this accepted for someone _without_ a track record or any sort are slim, to say the least.

Lastly as I'm sure you know, since you mentioned the EF Handbook and our Contributing guidelines, it generally remains at the discretion of the committers to accept or reject any contribution based on a variety of reasons. 

So from my point of view, all this is perfectly aligned with OSS practices, even if you seem offended by this PR. I'm also going to lock this discussion now, because this PR is merged, so we're really beating a dead horse at this point.

jimmarino:
Sorry, I need to chime in here.

First, you mischaracterized what three committers (@jimmarino, @ndr-brt, and @paullatzelsperger) stated in the referenced discussions. The consensus among the committers was that any solution would need to be generalized and reconciled with work underway in the DSP specification (a.k.a. IDS). There were also requirements for idempotent behavior, which DPF was not equipped to support at the time.  

In the roughly year and a half since DSP has been updated to include more explicit state transitions and data formats, and the Data Plane Signaling API was introduced. Moreover, we are close to drafting a Data Plane API standard based on the latter. All of these were prerequisites to this PR, all were extensively discussed, and all have been vetted by implementation experience. This is in line with our positions expressed in the above-referenced discussions. 

I also don't understand how this project fails to follow Eclipse procedures. This PR was authored by a committer and was approved by two other committers. I would also have approved it, but I did not have time to finalize my review (I had no issues with it). Furthermore, the PR was the product of extensive discussions and collaboration among the committers. That's precisely how Eclipse projects should be run.

Given the volume of requests and work associated with this project, it is not reasonable to expect the committers to reach out or explain their actions to every potential contributor, interested party, or person who may have raised a question in the past. In order to stay informed and up-to-date, the best approach is to attend EDC office hours, monitor discussions, and read decision records.

ndr-brt:
I think @jimmarino and @paullatzelsperger already covered the point extensively, I just wanted to add that the related issue (https://github.com/eclipse-edc/Connector/issues/2534) it remained open for like more than a year (read: IT ages :) ), so the idea to have this feature implemented was never discarded in any way, at the end it made it, so I guess it could be considered an happy ending for everyone (a committer contribution is typically less expensive in terms of time both on the "code" side and on the "review" side).

PR #4172, opened 2024-05-13
feat: add getStatusPurpose method

paullatzelsperger:
## What this PR changes/adds

This PR adds a method `getStatusPurpose` to the `RevocationListService` that returns either the active status (`"suspension"`, `"revocation"`) or `null` if no status is active.

## Why it does that

In IdentityHub we need to update the internal status of a `VerifiableCredentialResource` depending on this status.
## Linked Issue(s)

Relates https://github.com/eclipse-edc/IdentityHub/issues/340


spi/common/boot-spi/src/main/java/org/eclipse/edc/spi/result/Result.java
```diff
@@ -41,6 +41,10 @@ public static <T> Result<T> success(@NotNull T content) {
         return new Result<>(content, null);
     }
 
+    public static <T> Result<T> successNullable(@Nullable T content) {
```
ndr-brt, line 44: why not remove the `@NotNull` annotation from the `success` method? I think it's perfectly fine to have a null content (e.g. for `Void` type)

extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/StatusList2021RevocationService.java
```diff
@@ -55,7 +60,45 @@ public Result<Void> checkValidity(VerifiableCredential credential) {
                 .orElse(Result.failure("Could not check the validity of the credential with ID '%s'".formatted(credential.getId())));
     }
 
+    @Override
+    public Result<String> getStatusPurpose(VerifiableCredential credential) {
+        if (credential.getCredentialStatus().isEmpty()) {
+            return successNullable(null);
+        }
```
ndr-brt, line 67: this guard is not necessary, when the status is empty the method will return a success(null) anyway

extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/StatusList2021RevocationService.java
```diff
@@ -55,7 +60,45 @@ public Result<Void> checkValidity(VerifiableCredential credential) {
...
+                .collect(Collectors.groupingBy(AbstractResult::succeeded)); //partition by succeeded/failed
+
+        //if any of the credentialStatus checks failed, return a failure
+        if (res.containsKey(false)) {
+            return Result.failure(res.get(false).stream().map(AbstractResult::getFailureDetail).toList());
+        }
+
+        // otherwise map all
+        var list = res.get(true).stream()
+                .filter(r -> r.getContent() != null) //
```
ndr-brt, line 80:
```suggestion
                .filter(r -> r.getContent() != null)
```

extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/StatusList2021RevocationService.java
```diff
@@ -55,7 +60,45 @@ public Result<Void> checkValidity(VerifiableCredential credential) {
...
+    public Result<String> getStatusPurpose(VerifiableCredential credential) {
+        if (credential.getCredentialStatus().isEmpty()) {
+            return successNullable(null);
+        }
+        var res = credential.getCredentialStatus().stream()
+                .map(StatusListStatus::parse)
+                .map(this::getStatusInternal)
+                .collect(Collectors.groupingBy(AbstractResult::succeeded)); //partition by succeeded/failed
+
+        //if any of the credentialStatus checks failed, return a failure
```
ndr-brt, line 73: I think the comment is not needed as the code is expressive enough

extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/StatusList2021RevocationService.java
```diff
@@ -55,7 +60,45 @@ public Result<Void> checkValidity(VerifiableCredential credential) {
...
+        if (res.containsKey(false)) {
+            return Result.failure(res.get(false).stream().map(AbstractResult::getFailureDetail).toList());
+        }
+
+        // otherwise map all
```
ndr-brt, line 78: same

extensions/common/iam/verifiable-credentials/src/test/java/org/eclipse/edc/iam/verifiablecredentials/StatusList2021RevocationServiceTest.java
```diff
@@ -110,4 +109,58 @@ void checkRevocation_whenCached_valid() {
...
+
+    @Test
+    void getStatusPurpose_whenNoCredentialStatus() {
+        clientAndServer.reset();
+        clientAndServer.when(request().withMethod("GET").withPath("/credentials/status/3"))
+                .respond(HttpResponse.response().withStatusCode(200).withBody(TestData.STATUS_LIST_CREDENTIAL_SINGLE_SUBJECT));
+        var credential = TestFunctions.createCredentialBuilder().build();
+        assertThat(revocationService.getStatusPurpose(credential))
+                .isSucceeded()
+                .isNull(); //content is null, not the result!
```
ndr-brt, line 163: adding a `.isNotNull()` before `.isSucceeded()` could remove the need for the comment

spi/common/boot-spi/src/main/java/org/eclipse/edc/spi/result/Result.java
paullatzelsperger: that was the alternative approach. My thinking was to make it more explicit, since returning `null` should be more of an edge case.
I was kinda in-between anyway, so i'll update the code.

extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/StatusList2021RevocationService.java
paullatzelsperger: it's either the guard here, or I refactor or add a similar check in Line 76:
```java
 var list = res.get(true).stream()
                .filter(r -> r.getContent() != null)
                .map(AbstractResult::getContent).toList();
```

ndr-brt: nope, because the stream will be empty so the filter function will never be evaluated

paullatzelsperger: yes, but the `.get(true)` will, which is what'll cause the NPE

spi/common/boot-spi/src/main/java/org/eclipse/edc/spi/result/Result.java
ndr-brt: In any case on the "client" side there will no way to know if this edge case was put in place other than checking if the content is null or not. IIRC at first it was put "not nullable" because the logic was: "if content is null result is failed", but it ended with "if failure is not null result is failed"

paullatzelsperger: yep, makes sense. users of a method that returns a `Result` would be well-advised to read javadoc anyway.

wolf4ood (approved):

extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/StatusList2021RevocationService.java
ndr-brt: ouch!

ndr-brt (approved):

PR #4197, opened 2024-05-20
feat(api): enable authentication for control-api

### LINKED ISSUE #839: API Security: enable authentication for (internal) REST APIs
~~All (internal) REST APIs, i.e. the DataMgmt API, the DataPlane API, ControlPlane API(?), should implement a consistent (possibly shared) use of tokens with regards to creation, refreshing, revoking.~~

~~Each API should have its own Jetty Context, but they should use the same `AuthenticationService`. Currently there is only the `TokenBasedAuthenticationService`, which may need to get adapted.~~

Currently, only the `control-API` is unsecured, so we need to make sure that users have a way to supply an `AuthenticationService` for it.

ndr-brt:
## What this PR changes/adds

Permit control-api authentication via extensions.

Server side:
- a `ApiAuthenticationRegistry` service has been introduced that can be used both for `management-api` and `control-api`, it is used to register an `AuthenticationService` for the specified context.

Client side:
- the `ControlClientAuthenticationProvider` can be registered to provide the authentication headers. It has already been injected on all the "control-api client" services to decorate the request with the headers (by default it returns an empty map)

## Why it does that

authentication

## Further notes

- currently the contexts name are defined as `control-api`, `management-api`, ... as done for the transformers, but, the context alias as it is handled by the configuration is in the form of `control`, `management`. It would be better to uniform to a single convention, and the latter would be preferred. Subsequent PR will come

## Linked Issue(s)

Closes #839


spi/common/auth-spi/src/main/java/org/eclipse/edc/api/auth/spi/ControlClientAuthenticationProvider.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.eclipse.edc.api.auth.spi;
+
+import java.util.Map;
+
+/**
+ * Provide headers to be used in intra-components (control-api) communication.
+ */
+public interface ControlClientAuthenticationProvider {
```
paullatzelsperger, line 22:
```suggestion
@FunctionalInterface
public interface ControlClientAuthenticationProvider {
```

extensions/common/auth/auth-basic/src/main/java/org/eclipse/edc/api/auth/basic/ConfigCredentials.java
```diff
@@ -0,0 +1,18 @@
...
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.api.auth.basic;
+
+public record ConfigCredentials(String username, String vaultKey) {
```
paullatzelsperger, line 17: I'd either make this package-private, or put bit of documentation on it

paullatzelsperger (approved):

extensions/common/auth/auth-basic/src/main/java/org/eclipse/edc/api/auth/basic/ConfigCredentials.java
ndr-brt: Done, I had to make package-private also the `BasicAuthenticationProvider` to avoid a warning about visibility, anyway the module has been deprecated as a whole.

PR #4235, opened 2024-06-04
fix: adds checks on prefixes post JSON-LD expansion process

### LINKED ISSUE #4160: JSON-LD compaction error on catalog-request
# Bug Report

## Describe the Bug

Under very particular circumstances, a consumer's catalog-request fails if the provider has configured his assets in such a way that the JSON-LD compaction breaks. 

### Expected Behavior

`HTTP 200` - serve the Provider's catalog.

### Observed Behavior

`HTTP 502` 
```json
[
    {
        "message": "{\"@type\":\"dspace:CatalogError\",\"dspace:code\":\"401\",\"dspace:reason\":\"Unauthorized\",\"@context\":{\"@vocab\":\"[https://w3id.org/edc/v0.0.1/ns/\",\"edc\":\"https://w3id.org/edc/v0.0.1/ns/\",\"tx\":\"https://w3id.org/tractusx/v0.0.1/ns/\",\"tx-auth\":\"https://w3id.org/tractusx/auth/\",\"cx-policy\":\"https://w3id.org/catenax/policy/\",\"dcat\":\"http://www.w3.org/ns/dcat#\",\"dct\":\"http://purl.org/dc/terms/\",\"odrl\":\"http://www.w3.org/ns/odrl/2/\",\"dspace\":\"https://w3id.org/dspace/v0.8/\"}}"](https://w3id.org/edc/v0.0.1/ns//%22,/%22edc/%22:/%22https://w3id.org/edc/v0.0.1/ns//%22,/%22tx/%22:/%22https://w3id.org/tractusx/v0.0.1/ns//%22,/%22tx-auth/%22:/%22https://w3id.org/tractusx/auth//%22,/%22cx-policy/%22:/%22https://w3id.org/catenax/policy//%22,/%22dcat/%22:/%22http://www.w3.org/ns/dcat#\%22,\%22dct\%22:\%22http://purl.org/dc/terms/\%22,\%22odrl\%22:\%22http://www.w3.org/ns/odrl/2/\%22,\%22dspace\%22:\%22https://w3id.org/dspace/v0.8/\%22%7D%7D%22),
        "type": "BadGateway",
        "path": null,
        "invalidValue": null
    }
]
```

## Steps to Reproduce

1. Provider-side: Create Asset via POST `/v3/assets` with payload. Note the missing binding of the `dct` prefix in the context.

```json
{
  "@context": {
    "edc": "https://w3id.org/edc/v0.0.1/ns/"
  },
  "@id": "breakingAsset",
  "properties": {
    "dct:type": {"@id": "https://w3id.org/my-taxonomy#MyAssetType"}
  },
  "privateProperties": {
  },
  "dataAddress": {
    "@type": "DataAddress",
    "type": "HttpData",
    "baseUrl": "https://my-backend.url",
    "oauth2:tokenUrl": "https://my-auth.server/token",
    "oauth2:clientId": "myclient",
    "oauth2:clientSecretKey":"mykey",
    "type": "HttpData",
    "proxyMethod": "true",
    "proxyQueryParams": "true",
    "proxyPath": "true"
  }
}
```

2. Provider-side: create contract-definition with arbitrary policy, allowing the consumer to satisfy the access policy.
3. Consumer-side: query catalog

## Context Information

Provider logs:

```text
jakarta.ws.rs.InternalServerErrorException: Failed to compact JsonObject: When compacting an IRI would result in an IRI which could be confused with a compact IRI [code=IRI_CONFUSED_WITH_PREFIX].
```

The error is triggered [here in the titanium json-ld lib](https://github.com/filip26/titanium-json-ld/blob/500f065e9466a2cff4f89f97af7352146004ae21/src/main/java/com/apicatalog/jsonld/compaction/UriCompaction.java#L531). An issue with the same exception (though on a different endpoint) was fixed before, see #3172.

Setup: eclipse-edc 0.6.2 (as part of tractusx-edc 0.7.0).

## Possible Root Cause

The error is triggered by the property `"dct:type": {"@id": "https://w3id.org/my-taxonomy#MyAssetType"}` ONLY IF the `@context` in the `POST /v3/assets` request does not bind the `dct` namespace. `dct:type` is interpreted as an expanded, absolute URI and persisted properly. However, upon catalog-request the property is loaded from the database and titanium processes it according to the standard context of the EDC that includes `dct` as set here:
https://github.com/eclipse-edc/Connector/blob/d77b676c58a8a8aa1e8881250e5b1b389a364e74/data-protocols/dsp/dsp-http-api-configuration/src/main/java/org/eclipse/edc/protocol/dsp/http/api/configuration/DspApiConfigurationExtension.java#L121

All prerequisites of the [IRI compaction algorithm, step 9](https://www.w3.org/TR/json-ld11-api/#iri-compaction) are then true and an exception is thrown. The same issue would likely occur with all other EDC-registered prefixes (`dcat`, `dspace`, ...).

wolf4ood:
## What this PR changes/adds

A new JSON-LD validator has been introduced in order to walk the input JsonObject after the expansion process
and check if some `property` name or `@id`/`@type` values still contains prefixed notation and checks those prefixes against the configured ones for the runtime. 
This can happen on custom properties (that we have in almost all entities) when the input JSON-LD object does not contains the prefix alias in the `@context`.

For example like in the #4160  issue if a user post a property `dct:type` without explicitly  mapping `dct` in the context :

```json
{
    "@context": {
        "dct": "http://purl.org/dc/terms/"
    },
    "dct:type": "mytype"
}
``` 

It will cause the property to be stored as `dct:type` instead of the expanded form. This is not always an issue but it can lead to JsonLD error  in compaction as `Absolute IRI  confused with prefix`. 

Since there are multiple places where we support custom properties in our entities, almost in every entities
the prefix check is added directly after the expansion process in the default JSON-LD service implementation.

The prefixes checker can be disabled with the setting `edc.jsonld.prefixes.check`, but it's activated by default

## Why it does that

Fixes eventual compaction issue when returning JSON-LD objects

## Linked Issue(s)

Closes #4160 

### DIFF
--- a/core/common/lib/json-ld-lib/src/main/java/org/eclipse/edc/jsonld/JsonLdConfiguration.java
+++ b/core/common/lib/json-ld-lib/src/main/java/org/eclipse/edc/jsonld/JsonLdConfiguration.java
@@ -32,7 +32,7 @@ public class JsonLdConfiguration {
         return httpsEnabled;
     }
 
-    public boolean isCheckPrefixes() {
+    public boolean shouldCheckPrefixes() {
         return checkPrefixes;
     }
 
--- a/core/common/lib/json-ld-lib/src/main/java/org/eclipse/edc/jsonld/JsonLdConfiguration.java
+++ b/core/common/lib/json-ld-lib/src/main/java/org/eclipse/edc/jsonld/JsonLdConfiguration.java
@@ -18,6 +18,7 @@ public class JsonLdConfiguration {
 
     private boolean httpEnabled = false;
     private boolean httpsEnabled = false;
+    private boolean checkPrefixes = true;
 
     private JsonLdConfiguration() {
 
@@ -29,6 +30,10 @@ public class JsonLdConfiguration {
 
     public boolean isHttpsEnabled() {
         return httpsEnabled;
+    }
+
+    public boolean isCheckPrefixes() {
+        return checkPrefixes;
     }
 
     public static class Builder {
@@ -46,6 +51,11 @@ public class JsonLdConfiguration {
 
         public Builder httpsEnabled(boolean httpsEnabled) {
             configuration.httpsEnabled = httpsEnabled;
+            return this;
+        }
+
+        public Builder checkPrefixes(boolean checkPrefixes) {
+            configuration.checkPrefixes = checkPrefixes;
             return this;
         }
 
--- a/core/common/lib/json-ld-lib/src/test/java/org/eclipse/edc/jsonld/TitaniumJsonLdTest.java
+++ b/core/common/lib/json-ld-lib/src/test/java/org/eclipse/edc/jsonld/TitaniumJsonLdTest.java
@@ -77,6 +77,36 @@ class TitaniumJsonLdTest {
     }
 
     @Test
+    void expand_shouldFail_whenMissingRegisteredPrefix() {
+        var jsonObject = createObjectBuilder()
+                .add(JsonLdKeywords.CONTEXT, createObjectBuilder().build())
+                .add("custom:item", "foo")
+                .build();
+        var jsonLd = defaultService(JsonLdConfiguration.Builder.newInstance().checkPrefixes(false).build());
+
+        jsonLd.registerNamespace("custom", "https://custom.namespace.org/schema/");
+
+        var expanded = jsonLd.expand(jsonObject);
+
+        AbstractResultAssert.assertThat(expanded).isSucceeded();
+    }
+
+    @Test
+    void expand_shouldSucceed_whenMissingRegisteredPrefix() {
+        var jsonObject = createObjectBuilder()
+                .add(JsonLdKeywords.CONTEXT, createObjectBuilder().build())
+                .add("custom:item", "foo")
+                .build();
+        var jsonLd = defaultService();
+
+        jsonLd.registerNamespace("custom", "https://custom.namespace.org/schema/");
+
+        var expanded = jsonLd.expand(jsonObject);
+
+        AbstractResultAssert.assertThat(expanded).isFailed();
+    }
+
+    @Test
     void expand_withCustomContext() {
         var jsonObject = createObjectBuilder()
                 .add(JsonLdKeywords.CONTEXT, createObjectBuilder().add("custom", "https://custom.namespace.org/schema/").build())
@@ -96,6 +126,7 @@ class TitaniumJsonLdTest {
                 .contains("https://custom.namespace.org/schema/key2")
                 .contains("@value\":\"value2\"");
     }
+
 
     @Test
     void compact() {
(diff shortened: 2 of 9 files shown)


jimmarino (approved):

core/common/lib/json-ld-lib/src/main/java/org/eclipse/edc/jsonld/JsonLdConfiguration.java
```diff
@@ -31,6 +32,10 @@ public boolean isHttpsEnabled() {
         return httpsEnabled;
     }
 
+    public boolean isCheckPrefixes() {
```
paullatzelsperger, line 35: minor nit: `shouldCheckPrefixes` sounds more idiomatic to me...

core/common/lib/json-ld-lib/src/test/java/org/eclipse/edc/jsonld/TitaniumJsonLdTest.java
```diff
@@ -75,6 +75,36 @@ void expand_shouldFail_whenPropertiesWithoutNamespaceAndContextIsMissing() {
...
+                .add(JsonLdKeywords.CONTEXT, createObjectBuilder().build())
+                .add("custom:item", "foo")
+                .build();
+        var jsonLd = defaultService(JsonLdConfiguration.Builder.newInstance().checkPrefixes(false).build());
+
+        jsonLd.registerNamespace("custom", "https://custom.namespace.org/schema/");
+
+        var expanded = jsonLd.expand(jsonObject);
+
+        AbstractResultAssert.assertThat(expanded).isSucceeded();
```
ndr-brt, line 90: nit: static import

core/common/lib/json-ld-lib/src/test/java/org/eclipse/edc/jsonld/TitaniumJsonLdTest.java
```diff
@@ -75,6 +75,36 @@ void expand_shouldFail_whenPropertiesWithoutNamespaceAndContextIsMissing() {
         AbstractResultAssert.assertThat(expanded).isFailed();
     }
 
+    @Test
+    void expand_shouldFail_whenMissingRegisteredPrefix() {
```
ndr-brt, line 79: the test name is not coherent with the assertion

core/common/lib/json-ld-lib/src/test/java/org/eclipse/edc/jsonld/TitaniumJsonLdTest.java
```diff
@@ -75,6 +75,36 @@ void expand_shouldFail_whenPropertiesWithoutNamespaceAndContextIsMissing() {
...
+
+        jsonLd.registerNamespace("custom", "https://custom.namespace.org/schema/");
+
+        var expanded = jsonLd.expand(jsonObject);
+
+        AbstractResultAssert.assertThat(expanded).isSucceeded();
+    }
+
+    @Test
+    void expand_shouldSucceed_whenMissingRegisteredPrefix() {
```
ndr-brt, line 94: ditto

ndr-brt (approved):

PR #4251, opened 2024-06-07
chore: correct null check message to corresponding field

### LINKED ISSUE #4250: EndpointDataReferenceEntry: null check for providerId refers to the wrong message
Hi there,

the null check for providerId will give the wrong validation message. Actually it is pointing to `transferProcessId` but should be `providerId`.

`Objects.requireNonNull(((EndpointDataReferenceEntry)this.entity).providerId, "transferProcessId");`

https://github.com/eclipse-edc/Connector/blob/893296628a643154cb8811038c06e14c3483f739/spi/common/edr-store-spi/src/main/java/org/eclipse/edc/edr/spi/types/EndpointDataReferenceEntry.java#L121C13-L121C68

domreuter:
## What this PR changes/adds

This PR corrects the null check error message for the providerId within the EDREntry. 

## Why it does that

Wrong validation message was given for the providerId field.

## Linked Issue(s)

Closes #4250

### DIFF
--- a/spi/common/edr-store-spi/src/main/java/org/eclipse/edc/edr/spi/types/EndpointDataReferenceEntry.java
+++ b/spi/common/edr-store-spi/src/main/java/org/eclipse/edc/edr/spi/types/EndpointDataReferenceEntry.java
@@ -23,8 +23,7 @@ import static org.eclipse.edc.spi.constants.CoreConstants.EDC_NAMESPACE;
 /**
  * Represents metadata associated with an EDR
  */
-public class
-EndpointDataReferenceEntry extends Entity {
+public class EndpointDataReferenceEntry extends Entity {
 
     public static final String SIMPLE_TYPE = "EndpointDataReferenceEntry";
     public static final String EDR_ENTRY_TYPE = EDC_NAMESPACE + SIMPLE_TYPE;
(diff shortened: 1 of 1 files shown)


spi/common/edr-store-spi/src/main/java/org/eclipse/edc/edr/spi/types/EndpointDataReferenceEntry.java
```diff
@@ -23,7 +23,8 @@
 /**
  * Represents metadata associated with an EDR
  */
-public class EndpointDataReferenceEntry extends Entity {
+public class
+EndpointDataReferenceEntry extends Entity {
```
wolf4ood, line 27: This formatting seems wrong

wolf4ood (changes requested):
Just the formatting and it should be good

Thanks

spi/common/edr-store-spi/src/main/java/org/eclipse/edc/edr/spi/types/EndpointDataReferenceEntry.java
domreuter: Sure, don't know how that came in 🤦🏻

wolf4ood (approved):
Thanks!

PR #4256, opened 2024-06-10
feat: first JSON-LD context for management api

### LINKED ISSUE #4255: Draft JSON-LD context for management API.
# Feature Request

For improving the usability of management API we should draft a JSON-LD context for the management API that will be used in ingress/egress for all the requests/responses (catalog response excluded).

Currently we did improved the "usability" of JSON-LD (removing the need of prefixed properties) by using the `@vocab` in ingress and egress bound to the `edc` namespace (`https://w3id.org/edc/v0.0.1/ns/`). That works for removing the all edc relevant properties (even custom one) but it has some limitations. 

For example if we want to enforce at JSON-LD level that a property should be an `@id` or that `@container` of the property is always a `@set`, that is not possible only with `@vocab`.

Also at the current state when dealing with the `odrl` policies we can use the `odrl` context in ingress but when reading out the policy definition the `odrl` policy contains always `odrl` prefixes. 

 
management api

dev experience

_Are there any requirements?_

## Solution Proposal

Draft an experimental JSON-LD context of management api in a separated experimental extension that will register a new context url and cached file for the edc management api.

wolf4ood:
## What this PR changes/adds

- Draft a JSON-LD context for the management api
- Introduces an experimental extension that register the context in the management api scope
- E2E serialization/deserialization of management api entities using the context.

## Why it does that

dev experience

## Further notes

Currently the context is cached and associated with the URL `https://w3id.org/edc/connector/management/v0.0.1` which is yet to be decided in the final form.

Since using the `odrl` context at root level could lead some issue when compacting on egress, it's automatically defined as local context when defining the `policy` term.

There is still an issue when compacting the `@type` of the `odrl` policy where currently the local context is not applied  there

For performance reason all `ser/de` tests are in a single class with a class level e2e runtime.

## Linked Issue(s)

Closes #4255 

### DIFF
--- a/extensions/common/api/management-api-json-ld-context/src/main/resources/META-INF/services/org.eclipse.edc.spi.system.ServiceExtension
+++ b/extensions/common/api/management-api-json-ld-context/src/main/resources/META-INF/services/org.eclipse.edc.spi.system.ServiceExtension
@@ -1,5 +1,5 @@
 #
-#  Copyright (c) 2020 - 2022 Microsoft Corporation
+#  Copyright (c) 2024 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
 #
 #  This program and the accompanying materials are made available under the
 #  terms of the Apache License, Version 2.0 which is available at
@@ -8,7 +8,7 @@
 #  SPDX-License-Identifier: Apache-2.0
 #
 #  Contributors:
-#       Microsoft Corporation - initial API and implementation
+#       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
 #
 #
 
--- a/extensions/common/api/management-api-json-ld-context/src/main/resources/META-INF/services/org.eclipse.edc.spi.system.ServiceExtension
+++ b/extensions/common/api/management-api-json-ld-context/src/main/resources/META-INF/services/org.eclipse.edc.spi.system.ServiceExtension
@@ -1,0 +1,15 @@
+#
+#  Copyright (c) 2020 - 2022 Microsoft Corporation
+#
+#  This program and the accompanying materials are made available under the
+#  terms of the Apache License, Version 2.0 which is available at
+#  https://www.apache.org/licenses/LICENSE-2.0
+#
+#  SPDX-License-Identifier: Apache-2.0
+#
+#  Contributors:
+#       Microsoft Corporation - initial API and implementation
+#
+#
+
+org.eclipse.edc.connector.api.management.jsonld.ManagementApiJsonLdContextExtension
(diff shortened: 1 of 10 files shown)


extensions/common/api/management-api-json-ld-context/src/main/resources/META-INF/services/org.eclipse.edc.spi.system.ServiceExtension
```diff
@@ -0,0 +1,15 @@
+#
+#  Copyright (c) 2020 - 2022 Microsoft Corporation
```
paullatzelsperger, line 2: copy-paste :)?

extensions/common/api/management-api-json-ld-context/src/test/java/org/eclipse/edc/connector/api/management/jsonld/serde/ManagementApiJsonLdContextExtensionTest.java
```diff
@@ -0,0 +1,46 @@
+/*
+ *  Copyright (c) 2020 - 2022 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
```
paullatzelsperger, line 2:
```suggestion
 *  Copyright (c) 2024 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
```

paullatzelsperger (approved):

extensions/common/api/management-api-json-ld-context/src/main/resources/META-INF/services/org.eclipse.edc.spi.system.ServiceExtension
wolf4ood: as always :D

extensions/common/api/management-api-json-ld-context/src/test/java/org/eclipse/edc/connector/api/management/jsonld/serde/TestFunctions.java
```diff
@@ -0,0 +1,316 @@
...
+                .add(CONTEXT, createContextBuilder().build())
+                .add(ID, "id")
+                .add("counterPartyAddress", "address")
+                .add("contractId", "contractId")
+                .add("dataDestination", dataDestination)
+                .add("properties", propertiesJson)
+                .add("privateProperties", privatePropertiesJson)
+                .add("protocol", "protocol")
+                .add("callbackAddresses", createCallbackAddress())
+                .add("assetId", "assetId")
```
ndr-brt, line 166: this is deprecated so it could be avoided

ndr-brt (approved):

PR #4260, opened 2024-06-11
fix: enforce contract negotiation request and transfer request consistency

### LINKED ISSUE #4239: Control Plane does not asset `target` value when opening Contract Negotiation
# Bug Report

## Describe the Bug

When consumer initiates a contract negotiation, the control plane does not check if the content of the `target` field (which should contain the asset id) is equal to the asset id defined in the asset id portion of the policy `@id`.

In fact, the content of this `target` field is not even used, and 

### Expected Behavior

Control Plane should enforce that the content of the `target` field contains the same asset id as the one defined in the asset id portion of the policy `@id`, and return a validation error if it's not the case.

## Steps to Reproduce

Initiate a contract negotiation through the control plane management API, and use a random put a random string in the `target` field of the request.

### LINKED ISSUE #4240: Control Plane does not check asset id when initiating transfer process
# Bug Report

## Describe the Bug

The request for initiating a transfer process contains an `assetId` and a `contractId` field. The `contractId` points to a contract agreement, which itself contains an asset id. 

No check is performed in order to asset that the asset id passed in the request matches the one defined in the contract agreement. 

If there is a mismatch, this results in an inconsistency in the Endpoint Data Reference that is generated. 

### Expected Behavior

Control Plane should return a validation error if the `assetId` in the request does not match the one defined in the contract agreement. 

### Observed Behavior

No error is returned, and an inconsistent Endpoint Data Reference is generated.

## Steps to Reproduce

Initiate a transfer process through the control plane management API, and use a random `assetId` in the request instead of the one stored in the contract.

bscholtes1A:
## What this PR changes/adds

This PR fixes two issues of the current implementation:
- in contract negotiation request, the asset id part of the offer id can be different from the asset id provided in the `target` field of the `Policy`.
- in transfer request, the asset id provided in the request can be different to the one referred by the contract id also provided in this same request. This leads to inconsistent `EndpointDataReferenceEntry` being generated.

## Why it does that

Fix.

## Linked Issue(s)

Closes #4239 
Closes #4240 


spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/controlplane/contract/spi/types/offer/ContractOffer.java
```diff
@@ -103,8 +104,17 @@ public Builder policy(Policy policy) {
 
         public ContractOffer build() {
             Objects.requireNonNull(contractOffer.id, "Id must not be null");
+            var offerId = ContractOfferId.parseId(contractOffer.id)
+                    .orElseThrow(failure -> new IllegalArgumentException(failure.getFailureDetail()));
             Objects.requireNonNull(contractOffer.assetId, "Asset id must not be null");
             Objects.requireNonNull(contractOffer.policy, "Policy must not be null");
+            if (!offerId.assetIdPart().equals(contractOffer.assetId)) {
+                throw new IllegalArgumentException("Asset id %s must be equal to asset id part of offer id %s".formatted(contractOffer.assetId, offerId.assetIdPart()));
+            }
```
ndr-brt, line 113: this check will cause issues in the communication with connectors that are not EDC (the contract offer id structure is not standardized, it is EDC specific)

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/controlplane/contract/spi/types/offer/ContractOffer.java
```diff
@@ -103,8 +104,17 @@ public Builder policy(Policy policy) {
 
         public ContractOffer build() {
             Objects.requireNonNull(contractOffer.id, "Id must not be null");
+            var offerId = ContractOfferId.parseId(contractOffer.id)
+                    .orElseThrow(failure -> new IllegalArgumentException(failure.getFailureDetail()));
```
ndr-brt, line 108: same here, this validation would break for every contract offer created by a connector that's not an EDC.

spi/control-plane/contract-spi/src/main/java/org/eclipse/edc/connector/controlplane/contract/spi/types/offer/ContractOffer.java
```diff
@@ -103,8 +104,17 @@ public Builder policy(Policy policy) {
...
             Objects.requireNonNull(contractOffer.assetId, "Asset id must not be null");
             Objects.requireNonNull(contractOffer.policy, "Policy must not be null");
+            if (!offerId.assetIdPart().equals(contractOffer.assetId)) {
+                throw new IllegalArgumentException("Asset id %s must be equal to asset id part of offer id %s".formatted(contractOffer.assetId, offerId.assetIdPart()));
+            }
+
+            if (!contractOffer.assetId.equals(contractOffer.policy.getTarget())) {
+                throw new IllegalArgumentException("Policy target %s must be equal to asset id %s".formatted(contractOffer.policy.getTarget(), contractOffer.assetId));
+            }
```
ndr-brt, line 117: generally speaking these builder validations are somewhat dangerous e.g. if in the database we have entities that won't pass this validation there will be un-manageable extensions thrown, plus it makes testing really annoying (as demonstrated by the changelog of this PR)

We have 4 `ContractOffer` new instance creations in the codebase, and 3 of them already take the policy asset id to set the `assetId`, it's just a matter to add a validation before the remaining call (that, looking at the code, it's not really needed).

ndr-brt (changes requested):
I think that the point 1 reported in the PR does not need any work, because the contract id "structure" is EDC dependant, adding a validation on that one will raise interoperability issues.
About the point 2, a validation in the `TransferRequestValidator` would do the job.

An improval that could be made is to remove the `assetId` from that `TransferRequest` as it could be obtained from the `ContractAgreement` instead of having the client to specify it

Issue #4262, opened 2024-06-11
Management Domains


jimmarino:
# Feature Request

This is an umbrella issue for Management Domains.

## Which Areas Would Be Affected?

FCC and the Control Plane

## Why Is the Feature Desired?

Support for distributed deployment topologies.

PR #4263, opened 2024-06-11
docs: Add architecture document for Management Domains

jimmarino:
## What this PR changes/adds

Adds an architecture document for Management Domains.

## Why it does that

## Further notes

One or more DRs will be added later.

## Linked Issue(s)

Related to #4262


ndr-brt (approved):

mspiekermann (approved):
While I have nothing against the content (quite the opposite), I would move section 3 out of this specific document. I think it would be better to split the explanation of the management-domain concept and how it is implemented (which is better placed in a story, issues and/or a decision record).

jimmarino:
> While I have nothing against the content (quite the opposite), I would move section 3 out of this specific document. I think it would be better to split the explanation of the management-domain concept and how it is implemented (which is better placed in a story, issues and/or a decision record).

Yes, more implementation specifics will be added in subsequent DRs. This was just intended as an outline of what needs to be done. I think I mentioned that in the associated issue.

wolf4ood (approved):

PR #4264, opened 2024-06-11
fix: enforce consistency of contract negotiation request and transfer request

### LINKED ISSUE #4239: Control Plane does not asset `target` value when opening Contract Negotiation

### LINKED ISSUE #4240: Control Plane does not check asset id when initiating transfer process

bscholtes1A:
## What this PR changes/adds

This PR fixes two issues of the current implementation:
- in contract negotiation request, the asset id part of the offer id can be different from the asset id provided in the `target` field of the `Policy`.
- in transfer request, the asset id provided in the request can be different to the one referred by the contract id also provided in this same request. This leads to inconsistent `EndpointDataReferenceEntry` being generated.

## Why it does that

Fix.

## Linked Issue(s)

Closes #4239 
Closes #4240 

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -80,7 +80,7 @@ maven/mavencentral/com.jcraft/jzlib/1.1.3, BSD-2-Clause, approved, CQ6218
 maven/mavencentral/com.lmax/disruptor/3.4.4, Apache-2.0, approved, clearlydefined
 maven/mavencentral/com.networknt/json-schema-validator/1.0.76, Apache-2.0, approved, CQ22638
 maven/mavencentral/com.nimbusds/nimbus-jose-jwt/9.28, Apache-2.0, approved, clearlydefined
-maven/mavencentral/com.nimbusds/nimbus-jose-jwt/9.40, , restricted, clearlydefined
+maven/mavencentral/com.nimbusds/nimbus-jose-jwt/9.40, Apache-2.0, approved, #15156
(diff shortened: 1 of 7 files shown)


ndr-brt (approved):
LGTM

PR #4270, opened 2024-06-13
feat: implement Delegated Authentication Service

### LINKED ISSUE #4261: Management API: add delegating authentication service
# Feature Request

Currently, we have two implementation of the `AuthenticationService`: a `BasicAuthenticationService` which was deprecated in version 0.6.5, and the `TokenBasedAuthenticationService`, which lacks some key features, such as token rotation, token invalidation, etc.

Thus, another implementation will be added that delegates the authentication decision to a third-party Identity Provider (IdP).

Management API clients must obtain a JSON Web token from that third-party IdP system and attach it to the request's `Authorization` header as `Bearer` token. The connector then resolves the IdP's public key and verifies the token.

NB: The abbreviation `dac` stands for "delegated access control" and is used only to better group config values.

## Limitations

- We will only implement verification of OAuth2 tokens (JWT format). 
- No evaluation of claims, i.e. no authorization.
- Only JWS are supported, i.e. no encrypted tokens (JWEs)
- Public keys must be provided as JWK or in PEM format. If JWK, then the expected format is [JWKS](https://datatracker.ietf.org/doc/html/rfc7517#section-5)

Initially, this system will be deployed for the Management API context, but in the future it could be used for other API contexts as well.

Management API

to have a more production-grade security solution for the Management API

## Solution Proposal

- Both the `TokenBasedAuthenticationService` extension and the `DelegatingAuthenticationService` extension will be loaded at runtime. The `DelegatingAuthenticationService` will be registered in the `ServiceExtensionContext` if - and only if - all config parameters it requires are provided. Otherwise, we fall back to the `TokenBasedAuthenticationService`, which will be registered as default provider.

- By default, the `DelegatingAuthenticationService` only verifies the JWT, plus some basic plausibility verification:
  - `nbf` if present, `nbf` must be before the current time (allowing for some epsilon)
  - `exp` if present, `exp` must be after the current time (allowing for some epsilon)

- additional token validation can be contributed to the `TokenValidationRuleRegistry` using extensions using the `management-api` context string.

- Required configuration values 
  - `edc.api.auth.dac.key.url`: URL of the IdP where the connector can download the public key(s) (JWK or PEM format).

- Optional configuration values
  - `edc.api.auth.dac.cache.validty`: time (in milliseconds) that the public key(s) of the IdP should be cached. Set to <= 0 to disable the cache.

- to allow for some backwards compatibility, the `DelegatingAuthenticationService` will also evaluate the `x-api-key` header and log a warning. After the usual deprecation period of two milestones this will be removed.

[edit 1]: removed optional header configuration
[edit 2]: added JWKS requirement
[edit 3]: renamed config property to be more generic

paullatzelsperger:
## What this PR changes/adds

Implements the `DelegatedAuthenticationService` that takes a bearer token and validates it against a public key resolved from a JWKS url.

## Why it does that

Delegating auth and related tasks (rotation, revocation,...) away from the connector

## Further notes

- Currently, only JWKS are supported (no PEM endpoints). let's see how far we get. We can always add that functionality to the `JwksPublicKeyResolver`
- `NotBeforeValidationRule` and `ExpirationIssuedAtValidationRule` have received an optional `allowNull` argument, because `nbf` and `exp` are not mandatory claims in general (they might be in certain OAuth2 flows)
- the `TokenBasedAuthenticationService` is only registered, if no other service is registered for the `"management-api"` context. That way, it yields the way to other implementations.

## Linked Issue(s)

Closes #4261


paullatzelsperger:
FYI @lgblaumeiser

extensions/common/auth/auth-delegated/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
+/*
+ *  Copyright (c) 2020 - 2022 Microsoft Corporation
+ *
```
wolf4ood, line 3: copy and paste? :)

wolf4ood (approved):

core/common/token-core/src/test/java/org/eclipse/edc/jwt/rules/ExpirationIssuedAtValidationRuleTest.java
```diff
@@ -60,15 +62,19 @@ void validationKoBecauseExpirationTimeNotRespected() {
...
         var token = ClaimToken.Builder.newInstance().build();
 
-        var result = rule.checkRule(token, emptyMap());
+        var result = r.checkRule(token, emptyMap());
 
-        assertThat(result.succeeded()).isFalse();
-        assertThat(result.getFailureMessages()).hasSize(1)
-                .contains("Required expiration time (exp) claim is missing in token");
+        assertThat(result.succeeded()).isEqualTo(allowNull);
+        if (!allowNull) {
```
ndr-brt, line 74: conditionals in test are generally not good, given that this parameterized test has only two inputs, I'd just split in two tests

core/common/token-core/src/test/java/org/eclipse/edc/jwt/rules/NotBeforeValidationRuleTest.java
```diff
@@ -60,15 +62,19 @@ void validationKoBecauseNotBeforeTimeNotRespected() {
                 .contains("Current date/time with leeway before the not before (nbf) claim in token");
     }
 
-    @Test
-    void validationKoBecauseNotBeforeTimeNotProvided() {
+    @ParameterizedTest(name = "Allow nbf null: {0}")
+    @ValueSource(booleans = { true, false })
+    void validationKoBecauseNotBeforeTimeNotProvided(boolean allowNull) {
```
ndr-brt, line 67: ditto

extensions/common/auth/auth-delegated/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+plugins {
+    `java-library`
+}
+
+dependencies {
+    api(project(":spi:common:auth-spi"))
+    api(project(":spi:common:token-spi"))
+    api(project(":core:common:token-core")) // for the validation rules
```
ndr-brt, line 22: implementation, as only `spi`s should be registered as `api`

extensions/common/auth/auth-delegated/src/main/java/org/eclipse/edc/api/auth/delegated/DelegatedAuthenticationService.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static jakarta.ws.rs.core.HttpHeaders.AUTHORIZATION;
+
+public class DelegatedAuthenticationService implements AuthenticationService {
+
+    public static final String MANAGEMENT_API_CONTEXT = "management-api";
+    @Deprecated
```
ndr-brt, line 33: pls add `since` attribute

extensions/common/auth/auth-delegated/src/main/java/org/eclipse/edc/api/auth/delegated/DelegatedAuthenticationService.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    public boolean isAuthenticated(Map<String, List<String>> headers) {
+
+        if (headers == null) {
+            var msg = "Headers were null";
+            monitor.warning(msg);
+            throw new AuthenticationFailedException(msg);
+        }
```
ndr-brt, line 59: can this happen? looking at the callers (only `AuthenticationRequestFilter` I'd say no

extensions/common/auth/auth-delegated/src/main/java/org/eclipse/edc/api/auth/delegated/JwksPublicKeyResolver.java
```diff
@@ -0,0 +1,116 @@
...
+                if (cacheValidityMs < DEFAULT_REFRESH_AHEAD_TIME + DEFAULT_CACHE_REFRESH_TIMEOUT) {
+                    builder.refreshAheadCache(false);
+                }
+
+            }
+            jwkSource = builder.build();
+        } catch (MalformedURLException e) {
+            monitor.warning("Malformed JWK URL: " + jwksUrl, e);
+            throw new EdcException(e);
+        }
```
ndr-brt, line 82: looks like this would fit better in a factory method and have the `JWKSource` passed already built to the constructor

paullatzelsperger: that would leak implementation details (`JWKSource`) outside of the class, and it would put the need to know the intrinsics of the `JWKSource` on the caller.

extensions/common/auth/auth-delegated/src/main/java/org/eclipse/edc/api/auth/delegated/DelegatedAuthenticationService.java
paullatzelsperger: this could (theoretically) be called from anywhere, not just request filters. We have the same check in the `TokenBased` one, and it's just a bit more defensive IMO.

extensions/common/auth/auth-delegated/src/main/java/org/eclipse/edc/api/auth/delegated/JwksPublicKeyResolver.java
ndr-brt: the factory method can actually stay in the class. I'm always a bit reclutant to add logic into a constructor, plus, as a rule of thumb, it's always good to know that a constructor cannot fail, leaving the dirty work to factories/builders

paullatzelsperger: ah, got you. yeah, that sounds good

extensions/common/auth/auth-delegated/src/main/java/org/eclipse/edc/api/auth/delegated/DelegatedAuthenticationService.java
ndr-brt: To me it is too defensive because it could happen only because a developer error, not a "flow" error (e.g. a request that brings null headers cannot happen in fact) so an error like this it would be spotted at the CI level.

ndr-brt (approved):

PR #4301, opened 2024-06-24
feat: implement Asset/Dataset specialization pt. 2

### LINKED ISSUE #4296: Management Domains: `Asset` and `Dataset` specialization pt. 2
# Feature Request

Part 2 of the implementation work for [Management Domains](https://github.com/eclipse-edc/Connector/blob/main/docs/developer/management-domains/management-domains.md) involves the following items:
- let `Catalog` extend `Dataset`
- `DatasetResolverImpl`: when it encounters an `Asset` that is a catalog (`isCatalog()` method), it creates a `Catalog` object (instead of a `Dataset`) containing the URL of the linked asset in its `service` property
- if the `JsonObjectFromDatasetTransformer` encounters a `Dataset`, that is a `Catalog`, it delegates its serialization down to the `JsonObjectFromCatalogTransformer`

`Dataset`, `Catalog`, `DatasetResolverImpl` and their transformers

to enable the implementation Management Domains, in particular linked catalogs

## Solution Proposal

- as part of this issue, the `Dataset#distributions` list should be initialized with an empty `Hashmap` to avoid NPEs.

paullatzelsperger:
## What this PR changes/adds

implements the second part of the preparatory work for Management Domains, specifically:

- makes `Catalog` extend `Dataset`
- improves the `DatasetResolver` and `DistributionResolver` so that they can handle `CatalogAsset`s
- updates the JSON-LD transformers, so they can handle `CatalogAsset`s

## Why it does that

Management Domain feature, specifically handling distributed catalogs

## Further notes

## Linked Issue(s)

Closes #4296

### DIFF
--- a/core/control-plane/control-plane-catalog/build.gradle.kts
+++ b/core/control-plane/control-plane-catalog/build.gradle.kts
@@ -22,7 +22,8 @@ dependencies {
     api(project(":spi:control-plane:transfer-spi"))
     api(project(":spi:control-plane:asset-spi"))
 
-    testImplementation(project(":tests:junit-base"));
+    implementation(project(":spi:common:data-address:data-address-http-data-spi"))
+    testImplementation(project(":tests:junit-base"))
 
     testImplementation(project(":core:common:connector-core"))
     testImplementation(project(":core:control-plane:control-plane-core"))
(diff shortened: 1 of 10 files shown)


wolf4ood (approved):
:rocket:

core/control-plane/control-plane-catalog/build.gradle.kts
```diff
@@ -22,7 +22,8 @@ dependencies {
     api(project(":spi:control-plane:transfer-spi"))
     api(project(":spi:control-plane:asset-spi"))
 
-    testImplementation(project(":tests:junit-base"));
+    implementation(project(":spi:common:data-address:data-address-http-data-spi"))
```
jimmarino, line 25: Why is an implementation dependency (or even test implementation) dependency needed? Can't the test use a fake type?

jimmarino (commented):
Question inline related to dependencies

ndr-brt (approved):

core/control-plane/control-plane-catalog/build.gradle.kts
paullatzelsperger: it requires the `HttpDataAddressSchema.HTTP_DATA_TYPE` and `HttpDataAddressSchema.BASE_URL` constants

jimmarino: Ah that's right, OK

PR #4302, opened 2024-06-25
refactor(test): speed up e2e tests

ndr-brt:
## What this PR changes/adds

Export classpath loading from `EmbeddedRuntime`

## Why it does that

Classpath loading is a time consuming operation (approx 14 seconds first time it runs, 4-5 seconds other times) as it needs to invoke gradle wrapper and the `printClasspath` task. By doing it just once for every runtime in the `e2e-transfer-test` makes tests run faster (approx 1 minute less than before)

## Further notes

## Linked Issue(s)

### DIFF
--- a/core/common/junit/src/main/java/org/eclipse/edc/junit/extensions/ClasspathReader.java
+++ b/core/common/junit/src/main/java/org/eclipse/edc/junit/extensions/ClasspathReader.java
@@ -1,0 +1,97 @@
+/*
+ *  Copyright (c) 2024 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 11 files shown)


paullatzelsperger (approved):

paullatzelsperger:
i'm wondering: the improvement consists of not doing the CP scanning in the `boot()` method, but in the constructor, right? so it would only come into play when the runtime is initialized on every method, but not on every class, right?

ndr-brt:
> i'm wondering: the improvement consists of not doing the CP scanning in the `boot()` method, but in the constructor, right? so it would only come into play when the runtime is initialized on every method, but not on every class, right?

the advantage kicks in also in the "runtime per class" case, e.g. in the "e2e transfer tests" where a single runtime is used multiple time across different tests (see `backend-service` runtime, used in 7 different test classes). Loading its classpath once will make things faster.

paullatzelsperger:
> the advantage kicks in also in the "runtime per class" case, e.g. in the "e2e transfer tests" where a single runtime is used multiple time across different tests (see `backend-service` runtime, used in 7 different test classes). 

true. seriously cool shit!!

PR #4304, opened 2024-06-26
feat: make pub/priv aliases mandatory at startup

### LINKED ISSUE #3994: DPS: public/private key aliases should be mandatory at boot time
maybe the key could be extracted and verified in the `@Provider` method, so it will avoid lazy runtime errors

_Originally posted by @ndr-brt in https://github.com/eclipse-edc/Connector/pull/3988#discussion_r1520971243_

ndr-brt:
## What this PR changes/adds

Read priv/pub aliases from configuration at startup, ensuring that the runtime won't start if they're not provided.

## Why it does that

Avoid lazy failures.

## Further notes

- Fixed some test tags (that will distribute tests in different ci jobs in a better way)
- in the transfer e2e created the `noPolicyConstraints` just once.
- EDIT: I took the opportunity to get rid of the infamous `backend-service`, relying only on instances of `mockserver` instead

## Linked Issue(s)

Closes #3994


jimmarino:
> * EDIT: I took the opportunity to get rid of the infamous `backend-service`, relying only on instances of `mockserver` instead

😄 🪓

jimmarino (approved):

PR #4316, opened 2024-07-01
fix: adds order by timestamp 

### LINKED ISSUE #4308: Missing order by on `nextNotLeased`  in stateful entity store (SQL)
# Bug Report

## Describe the Bug

Implementations of SQL variant `StateEntityStore` for the `nextNotLeased` method miss the order by
when looking for the next items to process. This could lead to starvation for some state and it could 
prevent some optimization (index creation) when fetching the items to process.

### Observed Behavior

For example putting an index like this 

```sql
CREATE INDEX IF NOT EXISTS transfer_process_state
    ON edc_transfer_process (state);
```

Makes the `NextNotLeased.avoidsStarvation`  on `TransferProcessStoreTestBase` fails

## Possible Implementation

add order by `state_time_stamp` on `QuerySpec` in all `nextNotLeased` SQL implementation

wolf4ood:
## What this PR changes/adds

Adds a deterministic order by `stateTimestamp` field in `nextNotLeased` method of state machines:

- `PolicyMonitor`
- `DataPlane`
- `ContractNegotiation`
- `TransferProcess`

This will allow a consistent and fair processing order for entities based on their state.
By using an index on `state` + timestamp it will also speedup processing states that need to transition and it will also lower down the usage of the database which we constantly poll for next states to process. State machine should not poll final state or states that needs to be monitored for a long period of time.

While profiling state machine the `PolicyMonitor` was on of those which spikes the CPU usage of both connector and
database since basically it's constantly polling the database for the entries in `STARTED` state, which grows over time without transitioning. Also the `edc.policy.monitor.state-machine.iteration-wait-millis` didn't help since the state machine does not wait if at least  one entry has been process in a tick. To lower down the load to the database usage of the `PolicyMonitor` state machine the `processMonitoring` processor now return `false` if no state transition is made. In this way the `wait` configuration can be applied.

## Why it does that

## Further notes

We might think of reworking `PolicyMonitor` state machine, since it's a bit different from the other state machine and it needs constantly checking on transfer process states and policy evaluation. In the current form can be a CPU hog 

## Linked Issue(s)

Closes #4308 

### DIFF
--- a/extensions/control-plane/store/sql/contract-negotiation-store-sql/docs/schema.sql
+++ b/extensions/control-plane/store/sql/contract-negotiation-store-sql/docs/schema.sql
@@ -80,3 +80,7 @@ CREATE UNIQUE INDEX IF NOT EXISTS contract_negotiation_id_uindex
 
 CREATE UNIQUE INDEX IF NOT EXISTS contract_agreement_id_uindex
     ON edc_contract_agreement (agr_id);
+
+
+-- This will help to identify states that needs to be transition without table scan when the entries start to grow
+CREATE INDEX IF NOT EXISTS contract_negotiation_state ON edc_contract_negotiation (state,state_timestamp);
--- a/extensions/control-plane/store/sql/contract-negotiation-store-sql/docs/schema.sql
+++ b/extensions/control-plane/store/sql/contract-negotiation-store-sql/docs/schema.sql
@@ -82,5 +82,5 @@ CREATE UNIQUE INDEX IF NOT EXISTS contract_agreement_id_uindex
     ON edc_contract_agreement (agr_id);
 
 
--- This will help to identify states that needs to be transition without table scan when the entries start to grow
+-- This will help to identify states that need to be transitioned without a table scan when the entries grow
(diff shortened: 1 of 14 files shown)


extensions/control-plane/store/sql/contract-negotiation-store-sql/docs/schema.sql
```diff
@@ -80,3 +80,7 @@ CREATE UNIQUE INDEX IF NOT EXISTS contract_negotiation_id_uindex
 
 CREATE UNIQUE INDEX IF NOT EXISTS contract_agreement_id_uindex
     ON edc_contract_agreement (agr_id);
+
+
+-- This will help to identify states that needs to be transition without table scan when the entries start to grow
```
jimmarino, line 85:
```suggestion
-- This will help to identify states that need to be transitioned without a table scan when the entries grow
```

jimmarino (approved):

ndr-brt (approved):
I also agree that the PolicyMonitor could be refactored into a watchdog, when it was implemented the state machine looked like the more natural thing to do, but in fact there's only one non final state (STARTED) so the SM is definitely overkill

PR #4321, opened 2024-07-02
feat: introduce auth configuration

### LINKED ISSUE #4294: Refactor auth extensions
# Feature Request

Apart from the `BasicAuthenticationExtension` which is deprecated and marked for removal, we have two auth extension

- `DelegatedAuthenticationExtension`
- `TokenBasedAuthenticationExtension`

Currently those two extension, after the refactor of the `ApiAuthenticationRegistry`, are bound to the `management-api` context. 
So reusing token based auth or delegated one in other contexts cannot be easily done.

Introduce

auth, api

_Are there any requirements?_

## Solution Proposal

Introduce another extension for configuring the association between web context and `AuthenticationService` as described [here](https://github.com/eclipse-edc/Connector/pull/4300)

wolf4ood:
## What this PR changes/adds

Introduces a configuration extension for associating an auth implementation to a web context with configuration.

Introduces an `ApiAuthenticationProvider` which creates a `AuthenticationService` based on the input config.

The configuration plugs into the `web.http.<context>` mechanism adding a nested key for configuring the authentication for the specific `<context>`. For example for configuring the tokenbased auth for a custom context `custom`:

```
web.http.custom.path=/custom
web.http.custom.port=8081
web.http.custom.auth.type=tokenbased
web.http.custom.auth.key=apiKey
```

## Why it does that

flexibility, cleanup, decoupling

## Further notes

Some API configuration module explicitly register the `AuthenticationRequestFilter` and uses custom `context` name to be used in the `ApiAuthenticationRegistry`. For backward compatibility that has not been changed. For overriding the AuthenticationService for that specific api context the additional config value `context` is available

For example for the management context:

```
web.http.management.auth.context=management-api
```

As future improvement we should remove `ApiAuthenticationProvider` and apply directly the `AuthenticationRequestFilter` with the configured `AuthenticationService`. We also might want to incorporate
this extension as default behavior in the web extension.

## Linked Issue(s)

Closes #4294 


ndr-brt (commented):
Great stuff!
I think it would be beneficial to have a piece of `.md` in the documentation that explains how to secure the APIs, it would be also good to mark the explicit registration of the filter as deprecated so we could get rid of them at the right time.

wolf4ood:
@ndr-brt 

added some docs on the md file

@paullatzelsperger 
I've added in the further notes the next step of refactoring that we could to to remove the `AuthRegistry`

ndr-brt (approved):

paullatzelsperger (approved):

PR #4325, opened 2024-07-03
docs: add decision record about optional params in CatalogRequest

### LINKED ISSUE #4322: Management API: add optional "scopes" parameter
# Feature Request

The `/catalog` endpoint of the Management API should take an optional parameter with which clients can inject additional DCP scope values.

This could either be a named field in the JSON structure, or an entry in extensible map, similar to various `properties` fields we have.

Catalog API

In typical DCP interactions, the consumer derives from the current request (e.g. policies, or existing agreements) which scopes need to be put in the access token. This does not work for a Catalog request, because there is no request context.

Further, in many dataspaces there are "default scopes", i.e. scopes that must be present on _every_ DSP interaction, but these are static - they can't be changed at runtime.

In some cases, it is necessary to add scopes to the default scopes. For example, a consumer may _know_ that a provider has certain assets available, but they have an access policy constraing on them, which makes them "invisible" unless a certain credential is presented. 

## Solution Proposal

- write a decision-record 
- add a new optional parameter to the `CatalogRequest` plus its transformers
- in IdentityHub, treat `PresentationQueryMessage#scope` as nullable

paullatzelsperger:
## What this PR changes/adds

adds a decision record about adding a new, optional `parameters` field to the `CatalogRequest` body

## Why it does that

documentation

## Further notes

## Linked Issue(s)

Closes #4322

### DIFF
--- a/docs/developer/decision-records/2024-07-03-additional-catalogrequest-param/README.md
+++ b/docs/developer/decision-records/2024-07-03-additional-catalogrequest-param/README.md
@@ -16,15 +16,14 @@ these are static - they can't be changed at runtime.
 So if a provider offers assets, that are only available is a special (non-default) credential is presented, the consumer
 must be able to attach the respective scope strings to the access token.
 
-_NB: the information \_which_ scopes must be added, has to be conveyed out-of-band.\_
+_NB: the information which scopes must be added, has to be conveyed out-of-band._
 
 In other words, a consumer may know that a provider has certain assets available, but they have an access policy
 constraint on them, which makes them "invisible" unless a certain credential is presented.
 
 ## Approach
 
-The `CatalogRequest` will receive a new field `parameters`, which is an extensible map. This map may contain an
-entry `additionalScopes` which is a list of scope strings (as
+The `CatalogRequest` will receive a new nullable field `additionalScopes` which is a list of scope strings (as
 per [DCP Specification, Section 3.1](https://github.com/eclipse-tractusx/identity-trust/blob/main/specifications/verifiable.presentation.protocol.md#31-access-scopes)):
 
 ```json
@@ -36,18 +35,16 @@ per [DCP Specification, Section 3.1](https://github.com/eclipse-tractusx/identit
   "querySpec": {
     //...
   },
-  "parameters": {
-    "additionalScopes": [
-      "org.eclipse.edc.vc.type.AdditionalCredentialType1:read",
-      "org.eclipse.edc.vc.type.AdditionalCredentialType2:*"
-    ]
-  }
+  "additionalScopes": [
+    "org.eclipse.edc.vc.type.AdditionalCredentialType1:read",
+    "org.eclipse.edc.vc.type.AdditionalCredentialType2:*"
+  ]
 }
 ```
 
 Scope strings provided in that fashion will get added to the default scopes.
 
-NB: both the `parameters` map and the `additionalScopes` entry are _OPTIONAL_.
+NB: the `additionalScopes` entry is _OPTIONAL_.
 
 This feature will be added in an Alpha version of the Management API first, specifically of `3.1.0-alpha`. The
 respective URL path will be `/v3.1alpha/api/management/catalog/request` as per
--- a/docs/developer/decision-records/2024-07-03-additional-catalogrequest-param/README.md
+++ b/docs/developer/decision-records/2024-07-03-additional-catalogrequest-param/README.md
@@ -1,0 +1,44 @@
+# Adding an additional parameter to the `CatalogRequest` (Management API)
+
+## Decision
+
+The `POST /catalog/request` endpoint to request a catalog will receive a new optional field in the request body with which the requesting participant (=consumer) can insert additional scopes into the DCP interaction.
+
+## Rationale
+
+In typical DCP interactions, the consumer derives the required scopes from the current request (e.g. policies, or existing agreements). This does not work for a Catalog request, because there is no request context.
+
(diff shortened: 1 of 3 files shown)


docs/developer/decision-records/2024-07-03-additional-catalogrequest-param/README.md
```diff
@@ -0,0 +1,44 @@
...
+```json
+{
+  "@type": "CatalogRequest",
+  "counterPartyAddress": "http://provider-address.com",
+  "counterPartyId": "providerId",
+  "protocol": "dataspace-protocol-http",
+  "querySpec": {
+    //...
+  },
+  "parameters": {
```
ndr-brt, line 32: is this `parameters` really needed? `additionalScopes` could stay at the root level, as `CatalogRequest` itself can already be considered "extensible" (`querySpec` is not mandatory so in fact it represent an object extension)

docs/developer/decision-records/2024-07-03-additional-catalogrequest-param/README.md
```diff
@@ -0,0 +1,44 @@
...
+        "org.eclipse.edc.vc.type.AdditionalCredentialType2:*",
+    ]
+  }
+}
+```
+Scope strings provided in that fashion will get added to the default scopes.
+
+NB: both the `parameters` map and the `additionalScopes` entry are _OPTIONAL_.
+
+This feature will be added in an Alpha version of the Management API first, specifically of `3.1.0-alpha`. The respective URL path will be `/v31alpha/api/management/catalog/request` as per our [deprecation policy](https://github.com/eclipse-edc/docs/tree/main/developer/decision-records/2024-05-27-maturity-levels-deprecation-policy).
```
ndr-brt, line 44: `v31alpha` is giving me mixed feelings, why don't use `v3.1-alpha` directly? dots and dashes are permitted in the path part of an http url

paullatzelsperger: `/v31alpha` is what the [deprecation policy](https://github.com/eclipse-edc/docs/tree/main/developer/decision-records/2024-05-27-maturity-levels-deprecation-policy) defines, so I went with it.

docs/developer/decision-records/2024-07-03-additional-catalogrequest-param/README.md
paullatzelsperger: it's not strictly needed, but it would fit the paradigm we often use. Also, if we have this container object, we could easily add other features in the future without changing the shape of the body (i.e. the `CatalogRequest` object)

docs/developer/decision-records/2024-07-03-additional-catalogrequest-param/README.md
ndr-brt: at least `v3.1alpha`, that is consistent with the deprecation policy (in which there's no reference to minor versions

paullatzelsperger: dots in urls... yeah they are allowed, but I slightly recoil everytime I see one :)

ndr-brt: well, `v31` looks like version 31.x.x :)

paullatzelsperger: that is true as well... so it's a beauty pageant between `/v31alpha`, `/v3-1-alpha` and `/v3.1alpha`

docs/developer/decision-records/2024-07-03-additional-catalogrequest-param/README.md
paullatzelsperger: it would be an extensible map, so we wouldn't need to change it. only whoever consumes the `CatalogRequest` objects needs to know to call `get(newParamterKey)`

ndr-brt: i think we are getting too technical for this scope, but nothing prevents extensible properties to be at the root level of an object (e.g. `Policy.extensibleProperties`, I know, they are not really used, but they are deserialized from the root level of a policy)

paullatzelsperger: if we use a `parameters` object, we only need to change the transformers and the API once, and the problem is solved for all future parameters. 
If we decide to remove the `additionalScopes` later, we could simply leave the outer wrapper object in place and not change the API layer, so no change. My objective was future-proofing this as much as possible.

ndr-brt: this should not be the case, if in the future we'll remove the `additionalScopes` in fact we'll need to change the API documentation, and that's should be transparent and described by a new version, because that's a breaking change, otherwise would be a sort of swept under the carpet.

"open" parameters are good when there's no knowledge in the domain about structure and content (see asset properties), but since we know how these `additionalScopes` are shaped and we know how to deserialize them, I would describe that clearly in our domain object (`CatalogRequest`).

paullatzelsperger: ok, let's agree to disagree, and not drag this on for too long. We're ultimately debating the dots on the "i"s. 
I'm just gonna put it directly into the root object.

jimmarino (approved):

wolf4ood (approved):

ndr-brt (approved):

PR #4326, opened 2024-07-03
feat: prevent using SQL modules with the NoopTransactionContext

paullatzelsperger:
## What this PR changes/adds

throws an error when the SQL modules, in particular `sql-core` is on the classpath, but no `TransactionContext` implementation (other than `NoopTransactionContext`) was provided.

## Why it does that

avoid pitfalls, where downstream projects have to explicitly add the `transaction-local` module, but there is no warning if they don't do that.
In those cases, there could be sporadic exceptions thrown by the connection pool

## Further notes

## Linked Issue(s)


ndr-brt (approved):
LGTM

wolf4ood (approved):

PR #4376, opened 2024-07-26
feat(sql): add SQL bootstrapper

### LINKED ISSUE #4356: Provide SQL Migrations by default
# Feature Request

Every PostgreSQL-based store implementation already has a `schema.sql` file, that is currently only used for testing against a real database.

This story deals with providing automatic database migration for each store using Flyway. I can see two ways to implement this on a high level:

1. as a separate module: for each store there would be an accompanying migration module, for example `asset-index-sql` and `asset-index-sql-migration`. Putting the migration module on the classpath executes the migration. To make things a bit easier for the users we should provide BOMs:

- `control-plane-stores-sql`: contains all the SQL stores for the control plane
- `control-plane-stores-sql-migration`: contains `control-plane-stores-sql` plus all migration modules
- `data-plane-stores-sql`: contains all SQL stores for the data plane
- `data-plane-stores-sql-migration`: contains `data-plane-stores-sql` plus all migration modules

2. embedded in the impl: each store module directly contains the migration, and there is a config switch that determines whether the migration is executed or not. This should be `true` by default. NB: this is the way how Tractus-X EDC does things at the moment.

_Personally, I would go for option 2., as it seems simpler and more compact. Also keeps the module count down, which keeps build times lower._

It must be stated in the clearest terms that this would be a **basic default solution** that is intended for simple test/demo deployments and **not** for live DB upgrades in production environments! In those scenarios a more elaborate migration strategy, eg. with specialized tools is needed.

All SQL stores

Currently, every downstream project using EDC has to come up with their own migration strategy, which is no easy thing and its repetitive work.

## Solution Proposal

- ~~a Decision-Record will outline the scope and limits~~
- implementation as bootstrap feature, which can be enabled globally (default=off)

paullatzelsperger:
## What this PR changes/adds

This PR adds the `SqlSchemaBootstrapper` class, which allows sql-related extensions to enqueue DDL statements. In the `prepare()` phase, the bootstrapper then executes these commands one after the other.

To do that, the schema files (formerly known as `schema.sql`) were renamed and moved to the `resources/` folder. The were given individual names to avoid name clashes.

In addition, all SQL-store tests were updated to read the schema from the resource rather than `docs/schema.sql`. 

## Why it does that

easy way to create database structures.

## Further notes

- `"edc.sql.schema.autocreate"` is `false` by default

## Linked Issue(s)

Closes #4356


extensions/common/sql/sql-bootstrapper/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+dependencies {
+    api(project(":spi:common:core-spi"))
+    api(project(":spi:common:transaction-spi"))
+    implementation(project(":spi:common:transaction-datasource-spi"))
+    implementation(project(":extensions:common:sql:sql-core")) // SqlQueryExecutor
+//
+    testImplementation(project(":core:common:junit"))
+//    testImplementation(project(":extensions:common:transaction:transaction-local"))
+//    testImplementation(testFixtures(project(":extensions:common:sql:sql-core")))
+//    testImplementation(libs.postgres)
```
ndr-brt, line 30: commented out code can be removed

extensions/common/sql/sql-bootstrapper/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
+/*
+ *  Copyright (c) 2021 Daimler TSS GmbH
```
ndr-brt, line 2: copyright

extensions/common/sql/sql-bootstrapper/src/main/java/org/eclipse/edc/sql/bootstrapper/SqlSchemaBootstrapperImpl.java
```diff
@@ -0,0 +1,88 @@
...
+        this.queryExecutor = queryExecutor;
+        this.dataSourceRegistry = dataSourceRegistry;
+        this.monitor = monitor;
+    }
+
+
+    @Override
+    public void addStatementFromResource(String datasourceName, String resourceName, ClassLoader classLoader) {
+        try (var sqlStream = classLoader.getResourceAsStream(resourceName)) {
+            var sql = new Scanner(Objects.requireNonNull(sqlStream)).useDelimiter("\\A").next();
```
ndr-brt, line 55: scanner needs to be closed, it could be included in the try-with-resources block

extensions/common/sql/sql-bootstrapper/src/main/java/org/eclipse/edc/sql/bootstrapper/SqlSchemaBootstrapperImpl.java
```diff
@@ -0,0 +1,88 @@
...
+            statements.add(new QueuedStatementRecord(resourceName, datasourceName, sql));
+        } catch (IOException e) {
+            throw new EdcException(e);
+        }
+    }
+
+    public Result<Void> executeSql() {
```
ndr-brt, line 62: this method should be put in the interface (that will save the explicit cast in the `SqlSchemaBootstrapperExtension`)

extensions/common/sql/sql-bootstrapper/src/main/java/org/eclipse/edc/sql/bootstrapper/SqlSchemaBootstrapperImpl.java
```diff
@@ -0,0 +1,88 @@
...
+        monitor.debug("Running DML statements: [%s]".formatted(statements.stream().map(QueuedStatementRecord::name).collect(Collectors.joining(", "))));
+        return transactionContext.execute(() -> {
+            Stream<Result<Void>> objectStream = statements.stream().map(statement -> {
```
ndr-brt, line 65: `var`, but maybe it's better to inline directly

extensions/common/sql/sql-core/src/testFixtures/java/org/eclipse/edc/sql/testfixtures/PostgresqlEndToEndInstance.java
```diff
@@ -42,18 +42,16 @@ static void createDatabase(String participantName) {
 
         var extensionsFolder = TestUtils.findBuildRoot().toPath().resolve("extensions");
```
ndr-brt, line 43: at this point the `createDatabase` could just create the database and leave the bootstrapper to create tables and indexes

extensions/control-plane/store/sql/asset-index-sql/README.md
```diff
@@ -4,7 +4,7 @@ Provides SQL persistence for assets.
 
 ## Prerequisites
 
-Please apply this [schema](docs/schema.sql) to your SQL database.
+Please apply this [schema](src/main/resources/schema.sql) to your SQL database.
```
ndr-brt, line 7:
```suggestion
Please apply this [schema](src/main/resources/asset-index-schema.sql) to your SQL database.
```

ndr-brt: same in the other readmes

extensions/common/sql/sql-bootstrapper/src/main/java/org/eclipse/edc/sql/bootstrapper/SqlSchemaBootstrapperImpl.java
paullatzelsperger: I don't want this method in the interface, for it could entice contributing modules to invoke it, which could potentially cause problems. The cast is ugly, I'll find another way.

extensions/common/sql/sql-bootstrapper/src/main/java/org/eclipse/edc/sql/bootstrapper/SqlSchemaBootstrapperImpl.java
paullatzelsperger: interesting. When I do that, IntelliJ complains about not "being able to reference a non-static method from a static context" (which threw me off initially), but compiling with Gradle seems to work 🤷

extensions/common/sql/sql-bootstrapper/src/main/java/org/eclipse/edc/sql/bootstrapper/SqlSchemaBootstrapperImpl.java
ndr-brt: ok, I understand it, but this necessity is not explicit (and being misinterpreted, e.g. from a future-me :) ), that could be achieved by keeping only the `addStatement` method in the bootstrapper and moving the execution logic to a module-private service, that won't be registered on the runtime, this way the `executeSql` will effectively become not callable from other modules.

paullatzelsperger: thats a lot of ceremony for one single method. if `executeSql()` is package-private and documented, it should be clear enough what is done. 
In addition, I didn't put the interface in an SPI package on purpose, because it is not intended as an extension point anyway.

paullatzelsperger: I tried it, and it looks OK, but I'm not really sure splitting such a trivial task in two classes (and two test classes) is really worth it TBH.

wolf4ood (approved):

extensions/common/sql/sql-bootstrapper/src/main/java/org/eclipse/edc/sql/bootstrapper/SqlSchemaBootstrapperImpl.java
```diff
@@ -15,74 +15,33 @@
...
-            return failure(e.getMessage());
-        }
-    }
-
-    private record QueuedStatementRecord(String name, String datasourceName, String sql) {
+    @Override
+    public Map<String, List<String>> getStatements() {
+        var map = new HashMap<String, List<String>>();
+        statements.forEach(qsr -> map.computeIfAbsent(qsr.datasourceName(), s -> new ArrayList<>()).add(qsr.sql()));
+        return map;
```
ndr-brt, line 45: nit: this could be replaced by
```java
return statements.stream().collect(groupingBy(QueuedStatementRecord::datasourceName, mapping(QueuedStatementRecord::sql, toList())));
```

ndr-brt (approved):

PR #4385, opened 2024-07-31
feat(vault): implements custom secret folder config

### LINKED ISSUE #4384: Hashicorp Vault folder configuration
### Discussed in https://github.com/eclipse-edc/Connector/discussions/4375

# Hashicorp Vault folder configuration

This proposal aims to introduce the possibility to configure dedicated folders inside Hashicorp Vault as secret storage.

### Administration

Having the option to sort the relevant secrets for multiple EDC instances in folders, enables an orderly structure inside Hashicorp Vault.

In this way, administration is made easier and less frustrating.

### Security

If you are using multiple instances of the EDC and/or other applications/services/components together with a single Hashicorp Vault instance, it is undesirable to have all the secrets accessible to everyone.

With the possibility to configure a folder for each EDC, every instance of the EDC can be separated in what secrets it can access.

This increases security.

## Affected Areas

Hashicorp Vault extension

## Solution Proposal

Introduce the optional configuration value `edc.vault.hashicorp.folder`.

If this value is set, the method `getSecretUrl` inside `HashicorpVaultClient.java` adds the folder to the URL that is generated.

With this, all operations will happen in the files of the folder instead.</div>

saschaisele-zf:
## What this PR changes/adds

Adds the possibility to configure a folder for the storage of secrets inside Hashicorp Vault.

## Why it does that

Additional feature, as described in the discussion and issue.

## Linked Issue(s)

Closes #4384 

### DIFF
--- a/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
+++ b/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
@@ -71,7 +71,7 @@ public class HashicorpVaultExtension implements ServiceExtension {
     @Setting(value = "The URL path of the vault's /secret endpoint", defaultValue = VAULT_API_SECRET_PATH_DEFAULT)
     public static final String VAULT_API_SECRET_PATH = "edc.vault.hashicorp.api.secret.path";
 
-    @Setting(value = "The path of the folder that the secret is stored in", required = false, defaultValue = "")
+    @Setting(value = "The path of the folder that the secret is stored in, relative to VAULT_FOLDER_PATH")
     public static final String VAULT_FOLDER_PATH = "edc.vault.hashicorp.folder";
 
     @Inject
@@ -147,7 +147,7 @@ public class HashicorpVaultExtension implements ServiceExtension {
         var ttl = context.getSetting(VAULT_TOKEN_TTL, VAULT_TOKEN_TTL_DEFAULT);
         var renewBuffer = context.getSetting(VAULT_TOKEN_RENEW_BUFFER, VAULT_TOKEN_RENEW_BUFFER_DEFAULT);
         var secretPath = context.getSetting(VAULT_API_SECRET_PATH, VAULT_API_SECRET_PATH_DEFAULT);
-        var folderPath = context.getSetting(VAULT_FOLDER_PATH, "");
+        var folderPath = context.getSetting(VAULT_FOLDER_PATH, null);
 
         return HashicorpVaultSettings.Builder.newInstance()
                 .url(url)
--- a/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
+++ b/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
@@ -71,6 +71,9 @@ public class HashicorpVaultExtension implements ServiceExtension {
     @Setting(value = "The URL path of the vault's /secret endpoint", defaultValue = VAULT_API_SECRET_PATH_DEFAULT)
     public static final String VAULT_API_SECRET_PATH = "edc.vault.hashicorp.api.secret.path";
 
+    @Setting(value = "The path of the folder that the secret is stored in", required = false, defaultValue = "")
+    public static final String VAULT_FOLDER_PATH = "edc.vault.hashicorp.folder";
+
     @Inject
     private EdcHttpClient httpClient;
 
@@ -144,6 +147,7 @@ public class HashicorpVaultExtension implements ServiceExtension {
         var ttl = context.getSetting(VAULT_TOKEN_TTL, VAULT_TOKEN_TTL_DEFAULT);
         var renewBuffer = context.getSetting(VAULT_TOKEN_RENEW_BUFFER, VAULT_TOKEN_RENEW_BUFFER_DEFAULT);
         var secretPath = context.getSetting(VAULT_API_SECRET_PATH, VAULT_API_SECRET_PATH_DEFAULT);
+        var folderPath = context.getSetting(VAULT_FOLDER_PATH, "");
 
         return HashicorpVaultSettings.Builder.newInstance()
                 .url(url)
@@ -155,6 +159,7 @@ public class HashicorpVaultExtension implements ServiceExtension {
                 .ttl(ttl)
                 .renewBuffer(renewBuffer)
                 .secretPath(secretPath)
+                .folderPath(folderPath)
                 .build();
     }
 }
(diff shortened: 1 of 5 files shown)


jimmarino:
We're putting the proverbial cart before the horse here. Please provide a Decision Record, including a testing proposal.

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
```diff
@@ -71,6 +71,9 @@ public class HashicorpVaultExtension implements ServiceExtension {
     @Setting(value = "The URL path of the vault's /secret endpoint", defaultValue = VAULT_API_SECRET_PATH_DEFAULT)
     public static final String VAULT_API_SECRET_PATH = "edc.vault.hashicorp.api.secret.path";
 
+    @Setting(value = "The path of the folder that the secret is stored in", required = false, defaultValue = "")
```
florianrusch, line 74:
```suggestion
    @Setting(value = "The path to the folder in which all the secrets of/for the EDC should be stored in", required = false, defaultValue = "")
```

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
```diff
@@ -71,6 +71,9 @@ public class HashicorpVaultExtension implements ServiceExtension {
...
+    public static final String VAULT_FOLDER_PATH = "edc.vault.hashicorp.folder";
```
florianrusch, line 75: Should we maybe name it subfolder?

florianrusch (changes requested):

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
jimmarino: FWIW, the original is correct and more concise

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
jimmarino: The question is whether the path is relative or absolute, not if it is a subfolder

florianrusch: I guess it depends on how you define relative/absolute in this case, right? 😄

For example the "full" URL would be: `https://the-vault.com/v1/secrets/data/some/deep/deep/deep/folder/my-secret`
This url splits up in the following parts:
- `https://the-vault.com/v1/secrets` (already covered by `VAULT_URL` and `VAULT_API_SECRET_PATH`)
- `data` (or `metadata`)
- `some/deep/deep/deep/folder` (will be covered by `VAULT_FOLDER_PATH`
- `my-secret`

jimmarino: An absolute path always starts with a scheme, so I don't think there is any ambiguity here. From the example, the path is always relative so it should explicitly state that, "The path of the folder relative to VAULT_FOLDER_PATH..."

florianrusch: @saschaisele can you please align the changes

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultClient.java
```diff
@@ -276,11 +276,17 @@ private HttpUrl getSecretUrl(String key, String entryType) {
         key = key.replace("%2F", "/");
 
         var vaultApiPath = settings.secretPath();
+        var folderPath = settings.getFolderPath();
+
+        if (folderPath == null) {
+            folderPath = "";
+        }
```
ndr-brt, line 283: this can never be null as per settings object instantiation, unneeded check

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
```diff
@@ -144,6 +147,7 @@ private HashicorpVaultSettings getSettings(ServiceExtensionContext context) {
         var ttl = context.getSetting(VAULT_TOKEN_TTL, VAULT_TOKEN_TTL_DEFAULT);
         var renewBuffer = context.getSetting(VAULT_TOKEN_RENEW_BUFFER, VAULT_TOKEN_RENEW_BUFFER_DEFAULT);
         var secretPath = context.getSetting(VAULT_API_SECRET_PATH, VAULT_API_SECRET_PATH_DEFAULT);
+        var folderPath = context.getSetting(VAULT_FOLDER_PATH, "");
```
ndr-brt, line 150: I'm not a fan of the empty string because it can lead to misunderstandings. 
I would set `null` as its default value, then in `HashicorpVaultClient` in the `getSecretUrl` method call the `addPathSegments` method with `folderPath` only if that's not null.

ndr-brt (changes requested):

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
florianrusch: :D But setting a proper default value instead of just `null` allows us to avoid any kind of nullpointer exceptions if someone forgets the null check.

SaschaIsele: changed the default to null

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/HashicorpVaultExtension.java
SaschaIsele: changed the description to reflect the relative nature of the path

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultClient.java
ndr-brt, line 283: in case of null folder path it would be better to avoid to add a path segments at all because:
- it will be more explicit
- it will avoid unnecessary calls to the `trimLeadingOrEndingSlash` method 
- it will not rely on an internal detail of okhttp (the fact that an empty string is ignored.)

SaschaIsele: in case of null, the `folderPath` will not be added now

PR #4402, opened 2024-08-08
feat: add component ID

### LINKED ISSUE #4399: Add `componentId` property
# Feature Request

In order to identify individual deployment components, we'll introduce a new property named `edc.component.id` that should be
a unique, stable identifier for a unit of deployment.

The default value is a randomized string, but a warning will be logged.

In addition, the `edc.runtime.id` value should always be a randomized string.

EDC core/boot

For data plane registration we need a stable identifier for the selection strategy.
The `runtimeId` is needed for exlusive locks on the DB level.

## Solution Proposal

- Introduce a new property `edc.component.id` and issue a `WARNING` message if it is not configured.
- Refactor `edc.runtime.id`: if it is configured, issue a `WARNING` and replace the value with a random string. If
  the `edc.component.id` is _not_ configured, use the configured `edc.runtime.id` as `edc.component.id`

paullatzelsperger:
## What this PR changes/adds

This PR adds the `edc.component.id` property with the following logic:

- if `edc.runtime.id` is configured, a warning is issued
- if `edc.component.id` is not configured, a warning is issued. 
  - if `edc.runtime.id` is configured, its value is used for `edc.component.id`, and `edc.runtime.id` receives a random UUID
  - if `edc.runtime.id` is not configured, a random UUID is used for `edc.component.id`

This way, backwards compatibility can be provided for deployments that previously had configured a static `edc.runtime.id`.

In addition, Dataplane Self Registration was modified to use the `edc.component.id` property.

## Why it does that

component IDs are used for idempotent processes, such as data plane registration. runtime IDs are used for lease acquisition.

## Further notes

## Linked Issue(s)

Closes #4399

### DIFF
--- a/core/common/boot/src/main/java/org/eclipse/edc/boot/system/DefaultServiceExtensionContext.java
+++ b/core/common/boot/src/main/java/org/eclipse/edc/boot/system/DefaultServiceExtensionContext.java
@@ -25,6 +25,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
 
+import static org.eclipse.edc.boot.BootServicesExtension.COMPONENT_ID;
 import static org.eclipse.edc.boot.BootServicesExtension.RUNTIME_ID;
 
 /**
@@ -33,14 +34,13 @@ import static org.eclipse.edc.boot.BootServicesExtension.RUNTIME_ID;
  */
 public class DefaultServiceExtensionContext implements ServiceExtensionContext {
 
-    @Deprecated(since = "0.6.2")
-    private static final String EDC_CONNECTOR_NAME = "edc.connector.name";
 
     private final Map<Class<?>, Object> services = new HashMap<>();
     private final Config config;
     private boolean isReadOnly = false;
     private String participantId;
     private String runtimeId;
+    private String componentId;
 
     public DefaultServiceExtensionContext(Monitor monitor, Config config) {
         this.config = config;
@@ -66,6 +66,11 @@ public class DefaultServiceExtensionContext implements ServiceExtensionContext {
     @Override
     public String getRuntimeId() {
         return runtimeId;
+    }
+
+    @Override
+    public String getComponentId() {
+        return componentId;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


core/common/boot/src/main/java/org/eclipse/edc/boot/system/DefaultServiceExtensionContext.java
```diff
@@ -108,20 +113,21 @@ public void initialize() {
...
+        componentId = getSetting(COMPONENT_ID, null);
+        if (componentId == null) {
+            if (runtimeId != null) {
+                getMonitor().warning("'%s' is not configured, but '%s' is. The value used for '%s' will be used for '%s', and a random string will be used for '%s'".formatted(COMPONENT_ID, RUNTIME_ID, RUNTIME_ID, COMPONENT_ID, RUNTIME_ID));
+                componentId = runtimeId;
             } else {
-                getMonitor().warning("%s is not configured and it will fallback to the deprecated %s value".formatted(RUNTIME_ID, EDC_CONNECTOR_NAME));
-                runtimeId = connectorName;
+                getMonitor().warning("%s is not configured so a random UUID is used. It is recommended to provide a static one.".formatted(COMPONENT_ID));
+                componentId = UUID.randomUUID().toString();
```
ndr-brt, line 128: this logic could be simplified, what about:
`runtime.id`:
- if `runtime.id` is not null print warning
- else set random `runtime.id`

`component.id`
- if `component.id` is null print warning and set it equal to `runtime.id` (that will be randomic if `runtime.id` was null otherwise will be the same value as `runtime.id`)

paullatzelsperger: that is not quite the same though, because I wanted `runtime.id` to be _always_ randomized.

ndr-brt: ok, gotcha, I guess in a future we could just avoid reading the `edc.runtime.id` setting at all.

paullatzelsperger: yes, this is for backwards compatibility mostly

PR #4408, opened 2024-08-12
refactor: extract data-plane-iam module from data-plane-core

### LINKED ISSUE #4407: data-plane requires public and private key settings also if not really needed
# Bug Report

## Describe the Bug

Currently to start a data-plane is always mandatory to provide private and public key configuration through settings:
`edc.transfer.proxy.token.signer.privatekey.alias`
`edc.transfer.proxy.token.verifier.publickey.alias`

but they are only needed if the data-plane provides `PULL` transfer mechanism

### Expected Behavior

to not require them in case of no `PULL` transfer mechanism is registered

### Observed Behavior

exception at startup

## Steps to Reproduce

build a barebone data-plane runtime and run it

## Context Information

EDC 0.8.1

## Detailed Description

## Possible Implementation

ndr-brt:
## What this PR changes/adds

Extract `data-plane-iam` module from `data-plane-core`

## Why it does that

modularity: it will permit data-plane runtime to run "only PUSH transfers": without requiring mandatory PULL-related settings.

## Further notes

## Linked Issue(s)

Closes #4407 


extensions/data-plane/data-plane-iam/src/main/java/org/eclipse/edc/connector/dataplane/iam/service/DataPlaneAuthorizationServiceImpl.java
```diff
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2024 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *  Copyright (c) 2024 Contributors to the Eclipse Foundation
```
paullatzelsperger, line 2: we don't have this copyright anywhere else, maybe it's better to leave the old one in place?

paullatzelsperger (approved):

extensions/data-plane/data-plane-iam/src/main/java/org/eclipse/edc/connector/dataplane/iam/service/DataPlaneAuthorizationServiceImpl.java
ndr-brt: yes, that was an intellij inititative :D

PR #4421, opened 2024-08-21
Improve QueryValidator testing strategy

### LINKED ISSUE #3731: refactor(test): improve `QueryValidator` testing strategy

harnvo:
## What this PR changes/adds

Currently I have only modified `TransferProcessServiceImpl` to see if I am on the right track.

What I have done so far is:
- Added a `QueryValidator` builder
- I used `QueryValidator.Builder` for dependency injection for instead of `QueryValidator`. This is because it is still `TransferProcessServiceImpl`'s liability to determine what should be validated.
- In `QueryValidatorTest` I created a nested class `QueryValidatorFromTransferProcessServiceImpl` to check if everything works fine with that specific example.

## Why it does that

## Further notes

There is a lot not done yet. In particular, I do not quite understand what you meant by `add some tests for every class that's used`
 in https://github.com/eclipse-edc/Connector/pull/4091#pullrequestreview-1990909492. 

I have moved some test cases from `TransferProcessServiceImplTest` into `QueryValidatorTest`, but IMO it is quite unreasonable to make the unit-test of `QueryValidator` dependent on any code other than `QueryValidator` (e.g. `TransferProcess`, `ProvisionedResource`, etc.)

## Linked Issue(s)

Closes #3731


harnvo:
@ndr-brt This is a draft PR, but I would like to know if I have understood everything correctly. Thanks!

harnvo:
@jimmarino Thanks for your suggestion.

There are two ways of introducing the `QueryValidatorFactory`.

The first and the most straight-forward one is to keep a HashMap that maps a `service` to`canonicalType` and `subtypeMap`. This is easy, but we would then need to modify `QueryValidatorFactory` everytime we modify/add a new Service.

Another option would be having two public static method `getCanonicalType` and `getSubtypeMap` for each service requiring query validation. The Factory would then look something like:
```
public class QueryValidatorFactory {
    public static QueryValidator createQueryValidator(Class<? extends QueryValidatable> clazz) throws RuntimeException {
        try {
            var canonicalType = clazz.getMethod("getCanonicalType").invoke(null);
            var typeHierarchyMap = clazz.getMethod("getSubtypeMap").invoke(null);
            return new QueryValidator((Class<?>) canonicalType, (Map<Class<?>, List<Class<?>>>) typeHierarchyMap);
        } catch (NoSuchMethodException | IllegalAccessException | java.lang.reflect.InvocationTargetException e) {
            throw new RuntimeException(e);
        }
    }
}
```
By doing this the service would still be responsible for determinating the actual `QueryValidator`. But I feel the code is too hacky for a proper PR.

My question is: who should actually be resposible for determinating the actual type to validate against? (Or, is this question even important?) If it should still be the service, then are there any general idea I could implement this in a more elegant way?

harnvo:
Closing in favor of #4430

Issue #3688, opened 2023-12-05
Support filter expressions for numbers


ndr-brt:
# Feature Request

Currently if we try to filter an asset by a numeric field with postgresql database, e.g.:
```json
{
    "@context": {
        "odrl": "http://www.w3.org/ns/odrl/2/"
    },
    "@type": "QuerySpec",
    "offset": 0,
    "limit": 10,
    "filterExpression": [
        {
            "operandLeft": "https://w3id.org/edc/v0.0.1/ns/month",
            "operator": "=",
            "operandRight": 12
        }
    ]
}
```

we'll receive this error:
```
SEVERE 2023-12-05T10:59:28.1771974 JerseyExtension: Unexpected exception caught
org.eclipse.edc.spi.persistence.EdcPersistenceException: org.postgresql.util.PSQLException: ERROR: operator does not exist: text = double precision
  Hint: No operator matches the given name and argument types. You might need to add explicit type casts.
```

It would be nice to have numbers supported as [it was done for boolean](https://github.com/eclipse-edc/Connector/blob/bdd6a2a425301676ebd72827f5ec6412935270e4/extensions/common/sql/sql-core/src/main/java/org/eclipse/edc/sql/translation/JsonFieldMapping.java#L43-L45).

## Which Areas Would Be Affected?

postgres statement building

## Why Is the Feature Desired?

_Are there any requirements?_

## Solution Proposal

_If possible, provide a (brief!) solution proposal._

bmg13:
Is this issue still applicable? I believe it was fixed in here: https://github.com/eclipse-edc/Connector/commit/703502f7be7d8bd8bc1951bb81df4397bfa4a5c0#diff-7a903fa35f24513d25fd5919768a14e2aafcb26b3e0e2d84487b189a13c022f8R51

ndr-brt:
Works for `int` but not for `float/double`

bmg13:
I see, can you please assign to me?
the idea would be to have a if() for each type (adding one for double here).

PR #4435, opened 2024-08-26
feat: Support filter expressions for additional numberic types

bmg13:
## What this PR changes/adds

Allow different types of numeric field with postgresql database.

## Why it does that

AS done for other types (Boolean, Integer), add the same validation for other numeric types to avoid additional logging errors that should not be reflected with expected user behaviour.

## Further notes

UT's were added to reflect new scenarios.
Took the opportunity of extending with primitive java types to include more potential query usages.

## Linked Issue(s)

Closes https://github.com/eclipse-edc/Connector/issues/3688

### DIFF
--- a/extensions/common/sql/sql-core/src/main/java/org/eclipse/edc/sql/translation/JsonFieldTranslator.java
+++ b/extensions/common/sql/sql-core/src/main/java/org/eclipse/edc/sql/translation/JsonFieldTranslator.java
@@ -47,6 +47,22 @@ public class JsonFieldTranslator implements FieldTranslator {
         return checkStatementByType(type, statement);
     }
 
+    @Override
+    public WhereClause toWhereClause(List<PathItem> path, Criterion criterion, SqlOperator operator) {
+        var leftOperand = getLeftOperand(path, criterion.getOperandRight().getClass());
+
+        var amendedLeftOperand = Optional.of(leftOperand)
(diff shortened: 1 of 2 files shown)


extensions/common/sql/sql-core/src/main/java/org/eclipse/edc/sql/translation/JsonFieldTranslator.java
```diff
@@ -71,4 +63,31 @@ public WhereClause toWhereClause(List<PathItem> path, Criterion criterion, SqlOp
         );
     }
 
+    private String checkStatementByType(Class<?> type, String statement) {
```
paullatzelsperger, line 66: nit: method name is not correct. It doesn't "check" anything, it creates a typecast expression based on a type. I would suggest a name like `createStatementForType` or similar.

bmg13: good point, thank you, changed in https://github.com/eclipse-edc/Connector/pull/4435/commits/8196e469c2dd48874989ac12188bbcb6757909cc

bmg13:
@ndr-brt @paullatzelsperger sorry for the tag, but I cannot update the label of this PR

paullatzelsperger (approved):

PR #4448, opened 2024-08-30
feat: adds REST API for policy validation

### LINKED ISSUE #4425: Add REST APIs for policy validation
# Feature Request

A new API in alpha should be exposes for validating `Policy` already stored.

The validation  should leverage the work done in #4423 and expose the results in the response

wolf4ood:
## What this PR changes/adds

adds REST API for policy validation

## Why it does that

improves policies usability and debuggability 

## Further notes

## Linked Issue(s)

Closes #4425

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -82,7 +82,7 @@ maven/mavencentral/com.lmax/disruptor/3.4.4, Apache-2.0, approved, clearlydefine
 maven/mavencentral/com.networknt/json-schema-validator/1.0.76, Apache-2.0, approved, CQ22638
 maven/mavencentral/com.nimbusds/nimbus-jose-jwt/9.28, Apache-2.0, approved, clearlydefined
 maven/mavencentral/com.nimbusds/nimbus-jose-jwt/9.40, Apache-2.0, approved, #15156
-maven/mavencentral/com.puppycrawl.tools/checkstyle/10.18.0, LGPL-2.1-or-later, restricted, clearlydefined
+maven/mavencentral/com.puppycrawl.tools/checkstyle/10.18.1, LGPL-2.1-or-later, restricted, clearlydefined
 maven/mavencentral/com.samskivert/jmustache/1.15, BSD-2-Clause AND BSD-3-Clause, approved, clearlydefined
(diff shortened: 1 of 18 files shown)


spi/control-plane/control-plane-spi/src/main/java/org/eclipse/edc/connector/controlplane/services/spi/policydefinition/PolicyService.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.eclipse.edc.policy.model.Policy;
+import org.eclipse.edc.runtime.metamodel.annotation.ExtensionPoint;
+import org.eclipse.edc.spi.result.Result;
+
+/**
+ * Service that permits actions on Policy entity, e.g. validation
+ */
+@ExtensionPoint
+public interface PolicyService {
```
jimmarino, line 25: Do we need this service? It doesn't seem to really do very much.

jimmarino (commented):
A question inline

spi/control-plane/control-plane-spi/src/main/java/org/eclipse/edc/connector/controlplane/services/spi/policydefinition/PolicyService.java
wolf4ood: It will host also the evaluation plan method in the next PR. The main purpose is to not expose at API level the `PolicyEngine`.  If it's not an issue i can chop the service and use directly the `PolicyEngine`

jimmarino: If it will contain other functionality, then I would keep it. If it is just a wrapper, I would 🪓 it.

jimmarino (approved):

spi/control-plane/control-plane-spi/src/main/java/org/eclipse/edc/connector/controlplane/services/spi/policydefinition/PolicyService.java
ndr-brt: I would put it into the `PolicyDefinitionService`

ndr-brt (commented):
.

PR #4454, opened 2024-09-05
feat: enable defining supported VC types for trusted issuers

### LINKED ISSUE #4453: feat: enable defining supported VC types for trusted issuers
# Feature Request

As of today the `TrustedIssuerRegistry` defines the list of trusted issuers ids. When validating the counter-party VCs, it is just checked that the issuer of each of these VC is contained in that list, which is quite limiting. 

A typical use-case is that we want to be able to define the types of VCs that are supported for each trusted issuer.

VC validation module.

Typical use-case for any dataspace.

bscholtes1A:
## What this PR changes/adds

As of today the TrustedIssuerRegistry defines the list of trusted issuers ids. When validating the counter-party VCs, it is just checked that the issuer of each of these VC is contained in that list, which is quite limiting.

A typical use-case is that we want to be able to define the types of VCs that are supported for each trusted issuer.

## Why it does that

VC validation module.

## Linked Issue(s)

Closes #4453 

### DIFF
--- a/extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/defaults/DefaultTrustedIssuerRegistry.java
+++ b/extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/defaults/DefaultTrustedIssuerRegistry.java
@@ -17,28 +17,26 @@ package org.eclipse.edc.iam.identitytrust.core.defaults;
 import org.eclipse.edc.iam.verifiablecredentials.spi.model.Issuer;
 import org.eclipse.edc.iam.verifiablecredentials.spi.validation.TrustedIssuerRegistry;
 
-import java.util.Collection;
-import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
- * Simple, memory-based implementation of a {@link TrustedIssuerRegistry}
+ * Simple, memory-based implementation of a {@link TrustedIssuerRegistry}.
  */
 public class DefaultTrustedIssuerRegistry implements TrustedIssuerRegistry {
-    private final Map<String, Issuer> store = new HashMap<>();
+
+    private final Map<String, Set<String>> store = new ConcurrentHashMap<>();
 
     @Override
-    public void addIssuer(Issuer issuer) {
-        store.put(issuer.id(), issuer);
+    public void register(Issuer issuer, String credentialType) {
+        store.computeIfAbsent(issuer.id(), k -> new HashSet<>()).add(credentialType);
     }
 
     @Override
-    public Issuer getById(String id) {
-        return store.get(id);
+    public Set<String> getSupportedTypes(Issuer issuer) {
+        return store.getOrDefault(issuer.id(), Set.of());
     }
 
-    @Override
-    public Collection<Issuer> getTrustedIssuers() {
-        return store.values();
-    }
 }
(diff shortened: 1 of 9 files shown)


extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/defaults/DefaultTrustedIssuerRegistry.java
```diff
@@ -17,28 +17,31 @@
...
+import java.util.Set;
 
 /**
- * Simple, memory-based implementation of a {@link TrustedIssuerRegistry}
+ * Simple, memory-based implementation of a {@link TrustedIssuerRegistry}.
  */
 public class DefaultTrustedIssuerRegistry implements TrustedIssuerRegistry {
-    private final Map<String, Issuer> store = new HashMap<>();
+
+    private final Map<String, Set<String>> store = new HashMap<>();
```
jimmarino, line 30: This should be made thread-safe using a ConcurrentHashMap

jimmarino (changes requested):

extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/defaults/DefaultTrustedIssuerRegistry.java
bscholtes1A: fixed

jimmarino (approved):

extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/rules/HasValidIssuer.java
```diff
@@ -42,6 +41,15 @@ public Result<Void> apply(VerifiableCredential credential) {
         if (issuer.id() == null) {
             return failure("Issuer did not contain an 'id' field.");
         }
-        return trustedIssuers.contains(issuer.id()) ? success() : failure("Issuer '%s' is not in the list of trusted issuers".formatted(issuer.id()));
+
+        if (!trustedIssuerRegistry.isTrusted(issuer)) {
```
paullatzelsperger, line 45: is this check needed? Simply checking whether a particular credential type is registered for the issuer seems enough. If the issuer is not registered, the registry would return an empty `Set`.

spi/common/verifiable-credentials-spi/src/main/java/org/eclipse/edc/iam/verifiablecredentials/spi/validation/TrustedIssuerRegistry.java
```diff
@@ -16,17 +16,38 @@
...
-    void addIssuer(Issuer issuer);
+    String WILDCARD = "*";
+
+    /**
+     * Register a supported type for a trusted issuer.
+     *
+     * @param issuer the issuer
+     * @param type   the type
+     */
+    void register(Issuer issuer, String type);
```
paullatzelsperger, line 34: nit: In this class, I would call it `credentialType`, and improve the Javadoc for it. The term `type` is not explicit enough IMO.
nit: 
```suggestion
void register(Issuer issuer, String credentialType);
```

spi/common/verifiable-credentials-spi/src/main/java/org/eclipse/edc/iam/verifiablecredentials/spi/validation/TrustedIssuerRegistry.java
```diff
@@ -16,17 +16,38 @@
...
+
+    /**
+     * Check if an issuer is trusted.
+     *
+     * @param issuer the issuer
+     * @return true if the issuer is trusted, false otherwise.
+     */
+    boolean isTrusted(Issuer issuer);
```
paullatzelsperger, line 42: is this really necessary? What is the use case of this method? couldn't someone just check if `getSupportedTypes(...).isEmpty()`?

jimmarino: I prefer having that method. It makes it explicit, and technically there could be a trusted issuer whose types are not necessarily supported by the EDC deployment, although that would be a stretch.

wolf4ood (approved):

PR #4476, opened 2024-09-13
feat(console monitor): Log level can be set via program arg

### LINKED ISSUE #4392: Permit `ConsoleMonitor` level configuration
# Feature Request

Currently it's not possible to configure the output level of the default `ConsoleMonitor` instance. The constructor is called without passing the `level`, that by default is `DEBUG`:
https://github.com/eclipse-edc/Connector/blob/8d259e911f2edef8fb036f515588b9b9e9030d16/core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java#L92

this should be configurable through command line arguments (preferred, as they are already used to configure it) or system properties.

_Are there any requirements?_

## Solution Proposal

rafaelmag110:
## What this PR changes/adds

This allows to set the log level of the `ConsoleMonitor` via program args (i.e. --log-level=INFO).

## Why it does that

It was not possible to set the level.

## Linked Issue(s)

Closes #4392

### DIFF
--- a/core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
+++ b/core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
@@ -23,8 +23,10 @@ import org.eclipse.edc.boot.system.injection.InjectorImpl;
 import org.eclipse.edc.boot.system.injection.ProviderMethod;
 import org.eclipse.edc.boot.system.injection.ProviderMethodScanner;
 import org.eclipse.edc.boot.system.injection.lifecycle.ExtensionLifecycleManager;
+import org.eclipse.edc.spi.EdcException;
 import org.eclipse.edc.spi.monitor.ConsoleMonitor;
 import org.eclipse.edc.spi.monitor.Monitor;
+import org.eclipse.edc.spi.result.Result;
 import org.eclipse.edc.spi.system.MonitorExtension;
 import org.eclipse.edc.spi.system.ServiceExtension;
 import org.eclipse.edc.spi.system.ServiceExtensionContext;
@@ -90,7 +92,11 @@ public class ExtensionLoader {
 
     static @NotNull Monitor loadMonitor(List<MonitorExtension> availableMonitors, String... programArgs) {
         if (availableMonitors.isEmpty()) {
-            return new ConsoleMonitor(parseLogLevel(programArgs), !Set.of(programArgs).contains(ConsoleMonitor.COLOR_PROG_ARG));
+            var parseResult = parseLogLevel(programArgs);
+            if (parseResult.failed()) {
+                throw new EdcException(parseResult.getFailureDetail());
+            }
+            return new ConsoleMonitor((ConsoleMonitor.Level) parseResult.getContent(), !Set.of(programArgs).contains(ConsoleMonitor.COLOR_PROG_ARG));
         }
 
         if (availableMonitors.size() > 1) {
@@ -131,19 +137,24 @@ public class ExtensionLoader {
     /**
      * Parses the ConsoleMonitor log level from the program args. If no log level is provided, defaults to Level default.
      */
-    private static ConsoleMonitor.Level parseLogLevel(String[] programArgs) {
+    private static Result<?> parseLogLevel(String[] programArgs) {
         return Stream.of(programArgs)
                 .filter(arg -> arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG + "="))
-                .map(arg -> arg.split("=").length == 2 ? arg.split("=")[1] : "null" )
-                .map(lvl -> {
+                .map(arg -> arg.split("="))
+                .map(arg -> {
+                    var validValueMessage = String.format("Valid values for the console level are %s", Stream.of(ConsoleMonitor.Level.values()).toList());
+
+                    if (arg.length != 2) {
+                        return Result.failure(String.format("Value missing for the --log-level argument. %s", validValueMessage));
+                    }
                     try {
-                        return ConsoleMonitor.Level.valueOf(lvl);
+                        return Result.success(ConsoleMonitor.Level.valueOf(arg[1]));
                     } catch (IllegalArgumentException e) {
-                        throw new IllegalArgumentException(String.format("Illegal Console Monitor log level value: %s. Possible values are %s", lvl, Stream.of(ConsoleMonitor.Level.values()).toList()));
+                        return Result.failure(String.format("Invalid value %s for the --log-level argument. %s", arg[1], validValueMessage));
                     }
                 })
                 .findFirst()
-                .orElse(ConsoleMonitor.Level.getDefaultLevel());
+                .orElse(Result.success(ConsoleMonitor.Level.getDefaultLevel()));
     }
 
 }
--- a/core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
+++ b/core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
@@ -87,7 +87,7 @@ public class ExtensionLoader {
         return loadMonitor(loader.stream().map(ServiceLoader.Provider::get).collect(Collectors.toList()), programArgs);
     }
 
-    public static @NotNull Monitor loadMonitor(List<MonitorExtension> availableMonitors, String... programArgs) {
+    static @NotNull Monitor loadMonitor(List<MonitorExtension> availableMonitors, String... programArgs) {
         if (availableMonitors.isEmpty()) {
             return new ConsoleMonitor(parseLogLevel(programArgs), !Set.of(programArgs).contains(ConsoleMonitor.COLOR_PROG_ARG));
         }
--- a/core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
+++ b/core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
@@ -31,6 +31,7 @@ import org.eclipse.edc.spi.system.ServiceExtensionContext;
 import org.eclipse.edc.spi.telemetry.Telemetry;
 import org.jetbrains.annotations.NotNull;
 
+import java.lang.String;
 import java.util.HashMap;
 import java.util.List;
 import java.util.ServiceLoader;
@@ -89,7 +90,7 @@ public class ExtensionLoader {
 
     static @NotNull Monitor loadMonitor(List<MonitorExtension> availableMonitors, String... programArgs) {
         if (availableMonitors.isEmpty()) {
-            return new ConsoleMonitor(!Set.of(programArgs).contains("--no-color"));
+            return new ConsoleMonitor(parseLogLevel(programArgs), !Set.of(programArgs).contains("--no-color"));
         }
 
         if (availableMonitors.size() > 1) {
@@ -110,6 +111,25 @@ public class ExtensionLoader {
             throw new IllegalStateException(String.format("Found %s OpenTelemetry implementations. Please provide only one OpenTelemetry service provider.", openTelemetries.size()));
         }
         return openTelemetries.isEmpty() ? GlobalOpenTelemetry.get() : openTelemetries.get(0);
+    }
+
+    /**
+     * Parses the ConsoleMonitor log level from the program args. If no log level is provided, defaults to Level default.
+     */
+    private static ConsoleMonitor.Level parseLogLevel(String[] programArgs) {
+        return Set.of(programArgs).stream()
+                .filter(arg -> arg.startsWith("--log-level"))
+                .map(arg -> getLogLevel(arg.split("=")[1]))
+                .findFirst()
+                .orElse(ConsoleMonitor.Level.getDefaultLevel());
+    }
+
+    private static ConsoleMonitor.Level getLogLevel(String level){
+        try{
+            return ConsoleMonitor.Level.valueOf(level);
+        } catch (IllegalArgumentException e){
+            return ConsoleMonitor.Level.getDefaultLevel();
+        }
     }
 
     /**
--- a/core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
+++ b/core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
@@ -113,21 +113,6 @@ public class ExtensionLoader {
     }
 
     /**
-     * Parses the ConsoleMonitor log level from the program args. If no log level is provided, defaults to Level default.
-     */
-    private static ConsoleMonitor.Level parseLogLevel(String[] programArgs) {
-        return Set.of(programArgs).stream().filter(arg -> arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG)).map(arg -> getLogLevel(arg.split("=")[1])).findFirst().orElse(ConsoleMonitor.Level.getDefaultLevel());
-    }
-
-    private static ConsoleMonitor.Level getLogLevel(String level) {
-        try {
-            return ConsoleMonitor.Level.valueOf(level);
-        } catch (IllegalArgumentException e) {
-            return ConsoleMonitor.Level.getDefaultLevel();
-        }
-    }
-
-    /**
      * Loads and orders the service extensions.
      */
     public List<InjectionContainer<ServiceExtension>> loadServiceExtensions(ServiceExtensionContext context) {
@@ -140,6 +125,23 @@ public class ExtensionLoader {
      */
     public <T> List<T> loadExtensions(Class<T> type, boolean required) {
         return serviceLocator.loadImplementors(type, required);
+    }
+
+    /**
+     * Parses the ConsoleMonitor log level from the program args. If no log level is provided, defaults to Level default.
+     */
+    private static ConsoleMonitor.Level parseLogLevel(String[] programArgs) {
+        return Set.of(programArgs).stream()
+                .filter(arg -> arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG))
+                .map(arg -> {
+                    try {
+                        return ConsoleMonitor.Level.valueOf(arg.split("=")[1]);
+                    } catch (IllegalArgumentException e) {
+                        throw new IllegalArgumentException(String.format("Illegal Console Monitor log level value: %s. Possible values are DEBUG, INFO, WARNING, SEVERE", arg.split("=")[1]));
+                    }
+                })
+                .findFirst()
+                .orElse(ConsoleMonitor.Level.getDefaultLevel());
     }
 
 }
--- a/core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
+++ b/core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
@@ -139,18 +139,17 @@ public class ExtensionLoader {
      */
     private static Result<?> parseLogLevel(String[] programArgs) {
         return Stream.of(programArgs)
-                .filter(arg -> arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG + "="))
-                .map(arg -> arg.split("="))
+                .filter(arg -> arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG))
                 .map(arg -> {
                     var validValueMessage = String.format("Valid values for the console level are %s", Stream.of(ConsoleMonitor.Level.values()).toList());
-
-                    if (arg.length != 2) {
+                    var splitArgs = arg.split("=");
+                    if (splitArgs.length != 2) {
                         return Result.failure(String.format("Value missing for the --log-level argument. %s", validValueMessage));
                     }
                     try {
-                        return Result.success(ConsoleMonitor.Level.valueOf(arg[1]));
+                        return Result.success(ConsoleMonitor.Level.valueOf(splitArgs[1].toUpperCase()));
                     } catch (IllegalArgumentException e) {
-                        return Result.failure(String.format("Invalid value %s for the --log-level argument. %s", arg[1], validValueMessage));
+                        return Result.failure(String.format("Invalid value \"%s\" for the --log-level argument. %s", splitArgs[1], validValueMessage));
                     }
                 })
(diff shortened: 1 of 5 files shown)


jimmarino:
Setting the log level should not be done like this, particularly having `BaseRuntime` downcast to `ConsoleMonitor. ' The log level can be set via a program argument. This PR can be simplified only to change `ExtensionLoader.loadMonitor(..)` to handle this case.

jimmarino (changes requested):
See my comment about redoing this PR.

rafaelmag110:
So you suggest removing the config option?

jimmarino:
It should be set the same way the color arg is done. Only one class needs to be modified for this.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/runtime/BaseRuntime.java
```diff
@@ -135,6 +142,27 @@ protected ServiceExtensionContext createServiceExtensionContext(Config config) {
         return context;
     }
 
+    /**
+     * Sets the ConsoleMonitor level from config, if specified. The config has precedence over the program argument.
+     */
+    @NotNull
+    protected String[] setLogLevelProgArgFromConfig(Config config, Monitor monitor, String[] programArgs) {
+        if (monitor instanceof ConsoleMonitor) {
```
paullatzelsperger, line 150: the fact aside that the `BaseRuntime` should not know anything about a specific monitor, why is the log level only relevant for a `ConsoleMonitor`?

core/common/boot/src/main/java/org/eclipse/edc/boot/system/runtime/BaseRuntime.java
```diff
@@ -135,6 +142,27 @@ protected ServiceExtensionContext createServiceExtensionContext(Config config) {
...
+            var levelConfig = config.getString(ConsoleMonitor.LOG_LEVEL_CONFIG, null);
+            if (levelConfig != null) {
+                var logLevelArgs = String.format("%s=%s", ConsoleMonitor.LEVEL_PROG_ARG, levelConfig);
+                var newProgArgs = new ArrayList<>(Arrays.asList(programArgs));
+                newProgArgs = (ArrayList<String>) newProgArgs.stream()
+                                .filter(arg -> !arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG))
+                                .collect(Collectors.toList());
```
paullatzelsperger, line 157: this could be simplified by `Stream.of(programArgs)`.
Putting the `LEVEL_PROG_ARG` into the `ConsoleMonitor` class is incorrect, b/c loglevels aren't restricted to the ConsoleMonitor.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/runtime/BaseRuntime.java
```diff
@@ -85,6 +88,10 @@ public static void main(String[] args) {
     public void boot(boolean addShutdownHook) {
         monitor = createMonitor();
         var config = configurationLoader.loadConfiguration(monitor);
+        if (monitor instanceof ConsoleMonitor) {
```
paullatzelsperger, line 91: this is not a very clean solution, because the monitor is already used during configuration loading, which could potentially result in log lines at a level that was "configured away". 
BTW this is the reason we put the color option as command line arg, I think the loglevel should be treated the same.

paullatzelsperger (changes requested):

core/common/boot/src/main/java/org/eclipse/edc/boot/system/runtime/BaseRuntime.java
rafaelmag110: It's not. But the `Monitor` interface doesn't declare a level attribute which in my opinion leaves this specificity to the implementations.
The reason why this is here is because I couldn't think of another option to load the log level from the config.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/runtime/BaseRuntime.java
rafaelmag110: Other Monitor implementations might have a specific way to set the log level.
The LEVEL_PROG_ARG is here because it's specific to the connector default monitor implementation.

rafaelmag110:
Thanks for your reviews. I do understand your points.
I'd still like the option to set the default monitor log level via a system / env property since changing the run command might not be ideal in certain situations.

I have two suggestions:
1 - Simply read from the system or environment properties using `propOrEnv()` on the constructor.
2 - Have a constructor that reads default monitor configs from a dedicated config file (similar to what the jdk-logger permits).

With option 2 we can easily add new configs later if we need them (i.e. like a formatter option).

We can also change the color option to be configurable that way.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
```diff
@@ -112,6 +112,21 @@ private static Supplier<Object> getDefaultProviderInvoker(ServiceExtensionContex
         return openTelemetries.isEmpty() ? GlobalOpenTelemetry.get() : openTelemetries.get(0);
     }
 
+    /**
+     * Parses the ConsoleMonitor log level from the program args. If no log level is provided, defaults to Level default.
+     */
+    private static ConsoleMonitor.Level parseLogLevel(String[] programArgs) {
+        return Set.of(programArgs).stream().filter(arg -> arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG)).map(arg -> getLogLevel(arg.split("=")[1])).findFirst().orElse(ConsoleMonitor.Level.getDefaultLevel());
```
ndr-brt, line 119: nit: it could be formatted to improve readability (and make the comment not necessary):
```java
return Set.of(programArgs).stream()
    .filter(arg -> arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG))
    .map(arg -> getLogLevel(arg.split("=")[1]))
    .findFirst()
    .orElse(ConsoleMonitor.Level.getDefaultLevel());
```

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
```diff
@@ -112,6 +112,21 @@ private static Supplier<Object> getDefaultProviderInvoker(ServiceExtensionContex
         return openTelemetries.isEmpty() ? GlobalOpenTelemetry.get() : openTelemetries.get(0);
     }
 
+    /**
+     * Parses the ConsoleMonitor log level from the program args. If no log level is provided, defaults to Level default.
+     */
+    private static ConsoleMonitor.Level parseLogLevel(String[] programArgs) {
```
ndr-brt, line 118: private methods should be placed under the public ones

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
```diff
@@ -112,6 +112,21 @@ private static Supplier<Object> getDefaultProviderInvoker(ServiceExtensionContex
...
+        return Set.of(programArgs).stream().filter(arg -> arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG)).map(arg -> getLogLevel(arg.split("=")[1])).findFirst().orElse(ConsoleMonitor.Level.getDefaultLevel());
+    }
+
+    private static ConsoleMonitor.Level getLogLevel(String level) {
+        try {
+            return ConsoleMonitor.Level.valueOf(level);
+        } catch (IllegalArgumentException e) {
+            return ConsoleMonitor.Level.getDefaultLevel();
```
ndr-brt, line 126: I think the runtime should not be started when the log level specified is not valid and the available log levels should be listed.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/runtime/BaseRuntime.java
ndr-brt: To be honest I don't think log level should be set with configuration, there's a clear reason why it gets set with program arguments. I would remove the if branch at lines 91:94 and this protected method as well

core/common/boot/src/test/java/org/eclipse/edc/boot/system/ExtensionLoaderTest.java
```diff
@@ -98,6 +98,36 @@ void loadMonitor_whenNoMonitorExtension() {
...
     }
 
+    @Test
+    void loadMonitor_programArgsSetConsoleMonitorLogLevel() {
+
+        var monitor = ExtensionLoader.loadMonitor(new ArrayList<>(), ConsoleMonitor.LEVEL_PROG_ARG + "=INFO");
+
+        var expectedMonitor = new ConsoleMonitor(ConsoleMonitor.Level.INFO, true);
+
+        assertEquals(monitor, expectedMonitor);
```
ndr-brt, line 108: please use assertjs `assertThat` kind of assertions

core/common/boot/src/test/java/org/eclipse/edc/boot/system/ExtensionLoaderTest.java
```diff
@@ -98,6 +98,36 @@ void loadMonitor_whenNoMonitorExtension() {
...
+
+        var monitor = ExtensionLoader.loadMonitor(new ArrayList<>());
+
+        var expectedMonitor = new ConsoleMonitor(ConsoleMonitor.Level.getDefaultLevel(), true);
+
+        assertEquals(monitor, expectedMonitor);
+    }
+
+    @Test
+    void loadMonitor_consoleMonitorDefaultLogLevelWhenWrongArgs() {
```
ndr-brt, line 122: as I already said, this should result in an exception thrown.

spi/common/boot-spi/src/main/java/org/eclipse/edc/spi/monitor/ConsoleMonitor.java
```diff
@@ -82,6 +88,27 @@ public void debug(Supplier<String> supplier, Throwable... errors) {
         output(DEBUG, supplier, errors);
     }
 
+    @Override
+    public boolean equals(Object o) {
```
ndr-brt, line 92: I don't think we should add equal/hashCode implementation for "behavior" classes.
The reason here looks to be testing, a better reason to not do that.
With assertJ is actually possible to check equality on private fields via `extracting` method

ndr-brt (changes requested):

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
rafaelmag110: couldn't checkstyle prevent this?

core/common/boot/src/main/java/org/eclipse/edc/boot/system/runtime/BaseRuntime.java
jimmarino: Agreed, the design of the Monitor is that the host (runtime bootstrap) decides how to source and configure it. In the default case, this is done by the bootstrapper and should never be sourced from configuration.

As I mentioned, this PR can be simplified to only modify one class.

jimmarino:
> Thanks for your reviews. I do understand your points. I'd still like the option to set the default monitor log level via a system / env property since changing the run command might not be ideal in certain situations.
> 
> I have two suggestions: 1 - Simply read from the system or environment properties using `propOrEnv()` on the constructor. ([function link](https://github.com/eclipse-edc/Connector/blob/c0e67c6bc46ea41cbf2ac1971304a155af12d2f4/core/common/lib/util-lib/src/main/java/org/eclipse/edc/util/configuration/ConfigurationFunctions.java#L28)) 2 - Have a constructor that reads default monitor configs from a dedicated config file (similar to what the jdk-logger permits).
> 
> With option 2 we can easily add new configs later if we need them (i.e. like a formatter option).
> 
> We can also change the color option to be configurable that way.

As I mentioned, this PR is too complex for what it needs to do. Program arguments can be sourced from environment variables for people who want to create their Docker images that way. There's no need to introduce this behavior in the codebase, and we should never introduce a dedicated configuration source such as a file. Let's keep this as simple as possible and try to only modify the single class.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
ndr-brt: definitely. this is material for a separated issue.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
rafaelmag110: [638be6a](https://github.com/eclipse-edc/Connector/pull/4476/commits/638be6acf4ced6b1909be5e3c50988cbedc7f2a0)

core/common/boot/src/test/java/org/eclipse/edc/boot/system/ExtensionLoaderTest.java
rafaelmag110: [638be6a](https://github.com/eclipse-edc/Connector/pull/4476/commits/638be6acf4ced6b1909be5e3c50988cbedc7f2a0)

spi/common/boot-spi/src/main/java/org/eclipse/edc/spi/monitor/ConsoleMonitor.java
rafaelmag110: [73f5085](https://github.com/eclipse-edc/Connector/pull/4476/commits/73f5085ddf516f1d8b29ec1c45e9c2c737ac0c90)

core/common/boot/src/test/java/org/eclipse/edc/boot/system/ExtensionLoaderTest.java
rafaelmag110: [73f5085](https://github.com/eclipse-edc/Connector/pull/4476/commits/73f5085ddf516f1d8b29ec1c45e9c2c737ac0c90)

rafaelmag110:
@jimmarino @paullatzelsperger @ndr-brt 
Adjusted the PR as per discussed. Please review.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
```diff
@@ -127,4 +127,21 @@ public <T> List<T> loadExtensions(Class<T> type, boolean required) {
...
+     * Parses the ConsoleMonitor log level from the program args. If no log level is provided, defaults to Level default.
+     */
+    private static ConsoleMonitor.Level parseLogLevel(String[] programArgs) {
+        return Set.of(programArgs).stream()
+                .filter(arg -> arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG))
+                .map(arg -> {
+                    try {
+                        return ConsoleMonitor.Level.valueOf(arg.split("=")[1]);
+                    } catch (IllegalArgumentException e) {
+                        throw new IllegalArgumentException(String.format("Illegal Console Monitor log level value: %s. Possible values are DEBUG, INFO, WARNING, SEVERE", arg.split("=")[1]));
```
ndr-brt, line 140: the "possible values" can be obtained from the `Level` enum, not that they will easily change or anything, but looks like the more natural thing to do instead of put them statically in the string.

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
```diff
@@ -127,4 +127,21 @@ public <T> List<T> loadExtensions(Class<T> type, boolean required) {
...
 
+    /**
+     * Parses the ConsoleMonitor log level from the program args. If no log level is provided, defaults to Level default.
+     */
+    private static ConsoleMonitor.Level parseLogLevel(String[] programArgs) {
+        return Set.of(programArgs).stream()
+                .filter(arg -> arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG))
+                .map(arg -> {
+                    try {
+                        return ConsoleMonitor.Level.valueOf(arg.split("=")[1]);
```
ndr-brt, line 138: the `arg.split("=")[1]` is duplicated at line 140, it could be extracted (also in a separated `map` function)

ndr-brt (commented):
a pair of nits, ready to go otherwise

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
rafaelmag110: [d8d26ca](https://github.com/eclipse-edc/Connector/pull/4476/commits/d8d26ca30fb53af3ea566fbb4551a16de443d9f8)

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
rafaelmag110: [d8d26ca](https://github.com/eclipse-edc/Connector/pull/4476/commits/d8d26ca30fb53af3ea566fbb4551a16de443d9f8)

ndr-brt (approved):

jimmarino (approved):

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
```diff
@@ -127,4 +128,22 @@ public <T> List<T> loadExtensions(Class<T> type, boolean required) {
         return serviceLocator.loadImplementors(type, required);
     }
 
+    /**
+     * Parses the ConsoleMonitor log level from the program args. If no log level is provided, defaults to Level default.
+     */
+    private static ConsoleMonitor.Level parseLogLevel(String[] programArgs) {
+        return Set.of(programArgs).stream()
+                .filter(arg -> arg.startsWith(ConsoleMonitor.LEVEL_PROG_ARG + "="))
+                .map(arg -> arg.split("=").length == 2 ? arg.split("=")[1] : "null" )
```
paullatzelsperger, line 137: nit: adding another `map(arg -> arg.split("=")` before this avoids having to do it twice. 

Also, providing `"null"` seems spurious. Better to provide `null` as default

core/common/boot/src/test/java/org/eclipse/edc/boot/system/ExtensionLoaderTest.java
```diff
@@ -98,6 +98,32 @@ void loadMonitor_whenNoMonitorExtension() {
         assertTrue(monitor instanceof ConsoleMonitor);
     }
 
+    @Test
+    void loadMonitor_programArgsSetConsoleMonitorLogLevel() {
+
+        var monitor = ExtensionLoader.loadMonitor(new ArrayList<>(), ConsoleMonitor.LEVEL_PROG_ARG + "=INFO");
```
paullatzelsperger, line 104: nit: this could be a `ParameterizedTest` with an `@EnumSource`

paullatzelsperger (commented):
small nits, OK for me otherwise

core/common/boot/src/main/java/org/eclipse/edc/boot/system/ExtensionLoader.java
rafaelmag110: [3b1253a](https://github.com/eclipse-edc/Connector/pull/4476/commits/3b1253a2a5d21ca8c584f8abf296ed69005e8727)

core/common/boot/src/test/java/org/eclipse/edc/boot/system/ExtensionLoaderTest.java
rafaelmag110: [3b1253a](https://github.com/eclipse-edc/Connector/pull/4476/commits/3b1253a2a5d21ca8c584f8abf296ed69005e8727)

core/common/boot/src/test/java/org/eclipse/edc/boot/system/ExtensionLoaderTest.java
```diff
@@ -98,6 +100,33 @@ void loadMonitor_whenNoMonitorExtension() {
         assertTrue(monitor instanceof ConsoleMonitor);
     }
 
+    @ParameterizedTest
+    @EnumSource
+    void loadMonitor_programArgsSetConsoleMonitorLogLevel(ConsoleMonitor.Level level) {
+
+        var monitor = ExtensionLoader.loadMonitor(new ArrayList<>(), ConsoleMonitor.LEVEL_PROG_ARG + "=" + level.toString());
```
paullatzelsperger, line 107: I would also test various combinations of spaces, upper, lower case etc.:

`"--log-level = INFO"`, `"--log-level= INFO"`, `"--log-level =INFO"`, `"--log-level=info"`,...

Not saying all of those should get accepted, IMO it's fine to only accept `"--log-level=XYZ"`, but to have a definitive test and to guard against regression.

core/common/boot/src/test/java/org/eclipse/edc/boot/system/ExtensionLoaderTest.java
```diff
@@ -98,6 +102,24 @@ void loadMonitor_whenNoMonitorExtension() {
...
+    @MethodSource("provideLogLevelVariantArgs")
+    void loadMonitor_programArgsSetConsoleMonitorLogLevel(String programArgs, ConsoleMonitor.Level level) {
+
+        var monitor = ExtensionLoader.loadMonitor(new ArrayList<>(), programArgs);
+
+        assertThat(monitor).extracting("level").isEqualTo(level);
+    }
+
+    @ParameterizedTest
+    @MethodSource("provideLogLevelWrongArgs")
```
paullatzelsperger, line 115: pls use the `ArgumentsProvider` b/c it is strongly typed and more resilient w.r.t. to refactoring

rafaelmag110: Is commit [aa0c624](https://github.com/eclipse-edc/Connector/pull/4476/commits/aa0c62454af93b2729374aa2dd0923e3c53fa1b9) ok? I can remove the last commit.

paullatzelsperger (approved):

PR #4490, opened 2024-09-25
fix: only return dataset with valid offers

ronjaquensel:
## What this PR changes/adds

Adds a check to the `DatasetResolverImpl` to only return a dataset if it has an offer attached.

## Why it does that

To avoid exposing sensitive information (asset metadata) if the requesting participant is not authorized to see it.

## Further notes

Also adds a check for `query` and `getById` to immediately return if no valid contract definitions exist for the requesting participant, as then no offers would be returned anyway, so it doesn't make sense to query and evaluate the assets in that case.

### DIFF
--- a/core/control-plane/control-plane-catalog/src/main/java/org/eclipse/edc/connector/controlplane/catalog/DatasetResolverImpl.java
+++ b/core/control-plane/control-plane-catalog/src/main/java/org/eclipse/edc/connector/controlplane/catalog/DatasetResolverImpl.java
@@ -62,6 +62,10 @@ public class DatasetResolverImpl implements DatasetResolver {
     @NotNull
     public Stream<Dataset> query(ParticipantAgent agent, QuerySpec querySpec) {
         var contractDefinitions = contractDefinitionResolver.definitionsFor(agent).toList();
+        if (contractDefinitions.isEmpty()) {
+            return Stream.empty();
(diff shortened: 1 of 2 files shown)


core/control-plane/control-plane-catalog/src/main/java/org/eclipse/edc/connector/controlplane/catalog/DatasetResolverImpl.java
```diff
@@ -62,6 +62,10 @@ public DatasetResolverImpl(ContractDefinitionResolver contractDefinitionResolver
     @NotNull
     public Stream<Dataset> query(ParticipantAgent agent, QuerySpec querySpec) {
         var contractDefinitions = contractDefinitionResolver.definitionsFor(agent).toList();
+        if (contractDefinitions.isEmpty()) {
+            return Stream.empty();
+        }
+
```
ndr-brt, line 68: add test (that ensures e.g. that the returned stream is empty and no calls to the assetIndex are made).

ronjaquensel: good point, done

ndr-brt (approved):

PR #4495, opened 2024-09-25
docs: decision record about multiple protocol versions

wolf4ood:
## What this PR changes/adds

Decision record about multiple protocol versions

## Why it does that

## Further notes

## Linked Issue(s)

### DIFF
--- a/docs/developer/decision-records/2024-09-25-multiple-protocol-versions/README.md
+++ b/docs/developer/decision-records/2024-09-25-multiple-protocol-versions/README.md
@@ -1,0 +1,73 @@
+# Multiple protocol versions
+
+## Decision
+
+We will introduce a `protocolVersion` concept that will be carried in the current `protocol` property, that will drive
+the selection the right protocol version for connector to connector communication.
+
+The protocol convention will be `<protocolName>:<protocolVersion>`
+
+## Rationale
+
+Currently, EDC supports only [DSP](https://docs.internationaldataspaces.org/ids-knowledgebase/v/dataspace-protocol)
+in the current stable form. As the spec is evolving, we should prepare the EDC for supporting multiple versions in the
+same EDC runtime. This will enable the communication between two EDC connectors as long as they agree on a common
+supported version.
+
+## Approach
+
+Since we are enriching the `protocol` property with a version, no additional changes is required in `RemoteMessage`s,
+`TransferProcess` and `ContractNegotiation`.
+
+We will refactor `RemoteMessageDispatcherRegistry` to:
+
+```java
+public interface RemoteMessageDispatcherRegistry {
+    /**
+     * Registers a dispatcher.
+     */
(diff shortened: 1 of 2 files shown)


ndr-brt (commented):
just an idea... why not use the `protocol` field that already exist? 
currently the value is always `dataspace-protocol-http`, it could become `dataspace-protocol-http:version`, that will likely simplify the work:
- no need to change CN and TP entities
- no need to change `RemoteMessage` interface and relative implementations
- no need to change `ContractRequest` and `TransferRequest`

wdyt?

wolf4ood:
I also though about this, it will probably reduce the work for supporting an additional fields everywhere. 
It make sense but it would require to register multiple `DspHttpRemoteMessageDispatcher` for different versions.
which has a downside that today we inject `DspHttpRemoteMessageDispatcher` and register all the message handling there

wolf4ood:
Probably by refactoring the `RemoteMessageDispatcherRegistry`

```java
@ExtensionPoint
public interface RemoteMessageDispatcherRegistry {

    /**
     * Registers a dispatcher.
     */
    void register(String protocol, RemoteMessageDispatcher dispatcher);
}
```

Could allow us to register the same `DspHttpRemoteMessageDispatcher` for different versions

ndr-brt:
> I also though about this, it will probably reduce the work for supporting an additional fields everywhere. It make sense but it would require to register multiple `DspHttpRemoteMessageDispatcher` for different versions. which has a downside that today we inject `DspHttpRemoteMessageDispatcher` and register all the message handling there

we could refactor the dispatcher registry to being able to support the version (e.g. by split the `protocol` field by `:` and detect the "name" part) so we can keep a single dispatcher implementation that will be able to discern between different versions and act properly.

otherwise also refactor the `DspHttpRemoteMessageDispatcher` to keep the common logic separated and make only the "delta" logic injectable so we can easily register multiple dispatcher for multiple versions with different injections/configurations.

Maybe we are getting too technical here, but it's just a matter of refactoring

wolf4ood:
i would try to stick to a single `DspHttpRemoteMessageDispatcher` since basically the diff is only the transformers and the json-ld context which can be configured in different context/scope per version. the logic then is the same.

Having automatically split in the registry will work but it will always dispatch even of unsupported versions

What about the refactoring I proposed above? in case i can amend the DR :) and go with the version in the `protocol` field

ndr-brt:
> i would try to stick to a single `DspHttpRemoteMessageDispatcher` since basically the diff is only the transformers and the json-ld context which can be configured in different context/scope per version. the logic then is the same.
> 
> Having automatically split in the registry will work but it will always dispatch even of unsupported versions. In case we could put a check in the implementation underlying
> 
> What about the refactoring I proposed above? in case i can amend the DR :) and go with the version in the `protocol` field

looks good to me :+1:

PR #4512, opened 2024-09-30
fix:  (hashicorp-vault) token self-lookup path as segments

### LINKED ISSUE #4508: Hashicorp Vault Token Renewal Initialization Failure
# Bug Report

## Describe the Bug

Starting TX EDC 0.8.0-RC3  in a local environment (based on EDC 0.10.0-20240918-SNAPSHOT, hence main)  outputs a hashicorp vault warning both in control-plane and data-plane images when using a static EDC_VAULT_HASHICORP_TOKEN from the environment

### Expected Behavior

No Warning, token refresh should work.

### Observed Behavior

WARNING 2024-09-27T14:06:12.663937303 [Hashicorp Vault] Initial token look up failed with reason: Failed to look up token with reason: Unexpected character ('<' (code 60)): expected a valid value (JSON String, Number, Array, Object or token 'null', 'true' or 'false')  at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 1]

Since the initial token lookup failed, the guess is that also the refresh does not work.

## Steps to Reproduce

1. Run a local vault with a preconfigured token
2. configure a tractus-x image (also tested with TX EDC 0.7.4 --> EDC 0.7.2 - same problem)  against the vault using that token in EDC_VAULT_HASHICORP_TOKEN
3. start the image and log at the log

## Context Information

- EDC main/0.10.0-20240918-SNAPSHOT inside of TX EDC 0.8.0-RC3
- hashicorp/vault:>1.17.3 as the vault implementation (docker releae "vault" stuck with <=1.13.3 7 months ago - see https://hub.docker.com/_/vault)
- OS: MacOS, Orbstack with linux/amd64 rosetta emulation

## Detailed Description

The culprit seems to lie in the HashiCorpVaultClient#isTokenRenewable method 

https://github.com/eclipse-edc/Connector/blob/067ded3635b33a81c645cfe061dec5bbf73e00c0/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultClient.java#L124

especially when compared to the renewToken implementation

https://github.com/eclipse-edc/Connector/blob/067ded3635b33a81c645cfe061dec5bbf73e00c0/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultClient.java#L161

The TOKEN_LOOK_UP_SELF_PATH is encoded as "v1%2Fauth%2Ftoken%2Flookup-self" and appended to the vault url.
A call is then made to (local cluster sample) "http://vault:8201/v1%2Fauth%2Ftoken%2Flookup-self". The default hashicorp/vault:latest image answers with "200 OK" from a redirection to "http://vault:8201/ui" and the content will be html, not the expected json response (which shows the seen parsing error/warning).

## Possible Implementation

Instead of 

```
        var uri = settings.url()
                .newBuilder()
                .addPathSegment(TOKEN_LOOK_UP_SELF_PATH)
                .build();
```

it rather should be

```
        var uri = settings.url()
                .newBuilder()
                .addPathSegments(TOKEN_LOOK_UP_SELF_PATH)
                .build();
```

to split the TOKEN_LOOK_UP_SELF_PATH before encoding only the remaining individual segments and not encoding the TOKEN_LOOK_UP_SELF_PATH as a whole.

drcgjung:
## What this PR changes/adds

HashciCorpClient should interpret token self-lookup path as several segments.

## Why it does that

Currently, the path is interpreted as  a single segment (which encodes the path slashes into %2F). The resulting endpoint will hit the hashicorp routing logic and will be redirected to "ui" and return an html page instead of the expected json response.

## Linked Issue(s)

Closes #4508  <-- _insert Issue number if one exists_


extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultClient.java
```diff
@@ -121,7 +121,7 @@ public Result<Void> doHealthCheck() {
     public Result<Boolean> isTokenRenewable() {
         var uri = settings.url()
                 .newBuilder()
-                .addPathSegment(TOKEN_LOOK_UP_SELF_PATH)
+                .addPathSegments(TOKEN_LOOK_UP_SELF_PATH)
```
ndr-brt, line 124: isn't there a way to justify this change through an automated test?

drcgjung: I switched [HashicorpVaultClientIntegrationTest](https://github.com/eclipse-edc/Connector/blob/8f5bff24a80c950a6c24cb09af1cc5695eefb56e/extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultClientIntegrationTest.java#L42) from "vault/1.9.6" to "hashicorp/vault:1.17.3" (and higher) -> lookUpToken_whenTokenNotExpired_shouldSucceed will fail without this patch.

docker stopped to update "vault" from 1.14 on - the newer "vault" implementations all come from hashicorp as a verified publisher.

I know the license problem why many people froze the reference implementation, but its a kind of bug anway if the reference implementation interprets its own spec now more literally.

ndr-brt: @drcgjung ok, but in the changelog there's no version update, please change the docker image version accordingly, there shouldn't be any licensing issue.

paullatzelsperger: For a bit of context: HashiCorp changed their licensing in August 2023, mainly to prevent companies from offering competing services that are based on Vault (and/or other HashiCorp offerings). 

Since the EDC project neither sells nor embeds HashiCorp community software, nor does it offer a product competitive to a HashiCorp product, I think we are in the clear. 

However, if there is a breaking API change between the freely available version (`1.9.6`) and the "commercial" version (`1.17.3`), that cannot be solved with a separate code path in the vault client, then this becomes a topic for the EDC Technical Committee how to proceed. 

Are you saying that there is a breaking API change between `1.9.6` and `1.17.3`?

ndr-brt: @paullatzelsperger we already have libraries that deal with proprietary software (see all the technology repos), we are not distributing hashicorp code.
License wise I see no issue here.
 
If a new version brought a breaking change I think we should provide either two versions of the module or a config to being able to make the EDC being able to interact with both "pre" and "post" breaking change, until the "pre" will get out of support.

drcgjung: > @drcgjung Are you saying that there is a breaking API change between `1.9.6` and `1.17.3`?

No, I am saying that AFAICS from version 1.17.3 on, the vault API does not tolerate/interpret url-encoded slashes in path prefixes as normal slashes anymore (which is IMHO quite a difference from the W3C spec perspective). So the older versions seems to have been more "tolerant" wrt the interpretation of the same API.

My patch works with ALL versions, only the integration test currently needs to select a single reference - maybe it could be subclassed/doubled to test integration with both versions?

paullatzelsperger: @ndr-brt my point exactly. _We (=EDC)_ are in the clear, legally. My point is that we can't really recommend publicly to use vault `1.17.3`, because that is not strictly OSS anymore and that would be a problem from an EF perspective I think. Yes, we have implementations for other non-free software, but there always has to be an OSS alternative.

I guess a good middle ground would be to test against `1.17.3` AND the last-known-free version `1.9.6`, as @drcgjung suggested. That way we can guarantee that our code works with "an open-source variant".

ndr-brt: > Just to be sure and to guard against regression, a good middle ground would be to test against `1.17.3` AND the last-known-free version `1.9.6`, as @drcgjung suggested. That way we can guarantee that our code works with "an open-source variant".

definitely agree

drcgjung: > > Just to be sure and to guard against regression, a good middle ground would be to test against `1.17.3` AND the last-known-free version `1.9.6`, as @drcgjung suggested. That way we can guarantee that our code works with "an open-source variant".
> 
> definitely agree

second integration test implemented.

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultClientIntegrationTest.java
```diff
@@ -37,22 +39,31 @@
 
 @ComponentTest
 @Testcontainers
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class HashicorpVaultClientIntegrationTest {
     @Container
-    static final VaultContainer<?> VAULT_CONTAINER = new VaultContainer<>("vault:1.9.6")
+    final VaultContainer<?> VAULT_CONTAINER = new VaultContainer<>(getVaultImage())
```
paullatzelsperger, line 45: please do not change that. it will spin up a separate container _for each test method_, which needlessly blows up test runtimes.

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultClientIntegrationTest.java
```diff
@@ -37,22 +39,31 @@
 
 @ComponentTest
 @Testcontainers
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
paullatzelsperger, line 42: not needed

paullatzelsperger (changes requested):
I get why you changed it to non-static, but there should be a better way to do this without blowing up test run times.

extensions/common/vault/vault-hashicorp/src/test/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultClientIntegrationTest.java
drcgjung: thanks for pointing this out ... I took over your gratious blueprint.

ndr-brt:
@drcgjung if you need another review, please re-request it.
looks like dependencies file needs to be updated accordingly to fix the failing check

drcgjung:
> @drcgjung if you need another review, please re-request it. looks like dependencies file needs to be updated accordingly to fix the failing check

Hmm, the dependency check failed with a strange error:

`Error: Invalid or corrupt jarfile dash.jar`

does not look like an old DEPs file, but a problem in the workflow?

ndr-brt:
> > @drcgjung if you need another review, please re-request it. looks like dependencies file needs to be updated accordingly to fix the failing check
> 
> Hmm, the dependency check failed with a strange error:
> 
> `Error: Invalid or corrupt jarfile dash.jar`
> 
> does not look like an old DEPs file, but a problem in the workflow?

@drcgjung you just need to update the dependencies with the ones you can find in the console log:
https://github.com/eclipse-edc/Connector/actions/runs/11177205158/job/31246114707?pr=4512

paullatzelsperger (approved):
looks good to me, i'll re-trigger the WF run

drcgjung:
@ndr-brt could you please approve/merge?

PR #4513, opened 2024-09-30
refactor: avoid potential unnecessary store accesses on dataset resolution

### LINKED ISSUE #4499: Optimize Catalog-Endpoint Performance
# Feature Request

The performance of the Catalog-Endpoint degrades as the number of `ContractDefinitions` increases. This appears to be due to an n+1 query issue within the `ContractDefinitionResolverImpl`. Currently, for each `ContractDefinition`, the corresponding access policy is resolved by its ID, resulting in a separate database query for each policy.

This issue seems to persists in the `main` branch. The relevant code can be found here: 
https://github.com/eclipse-edc/Connector/blob/dfa1be6100ed7037aed03a89badb805044fcc74a/core/control-plane/control-plane-contract/src/main/java/org/eclipse/edc/connector/controlplane/contract/offer/ContractDefinitionResolverImpl.java#L72-L89

- Control-Plane
- PolicyDefinitionStore

As the number of `ContractDefinitions` grows, the Catalog-Endpoint experiences significant performance degradation due to the increasing number of database queries.

## Solution Proposal

Implement a database join to retrieve all relevant policies in a single query, reducing the number of queries executed.

ndr-brt:
## What this PR changes/adds

Refactors `ContractDefinitionResolver` and `DatasetResolver` adding a thread local cache to diminish database accesses.

## Why it does that

Avoid potential multiple and unnecessary calls to `PolicyDefinitionStore.findById`. 
In cases where access policy id is the same as contract policy id this will halve database accesses, will also provide advantages where same policies are used for different contract negotiations

## Further notes

- moved contract definition resolver from "contract" to "catalog", because it's only meant to be used in the catalog creation.

## Linked Issue(s)

Closes #4499


core/control-plane/control-plane-catalog/src/main/java/org/eclipse/edc/connector/controlplane/catalog/ContractDefinitionResolverImpl.java
```diff
@@ -0,0 +1,74 @@
+/*
+ *  Copyright (c) 2024 Contributors to the Eclipse Foundation
```
paullatzelsperger, line 2: we only have explicit copyrights everywhere. possibly an old IDE setting :)?

core/control-plane/control-plane-contract/src/main/java/org/eclipse/edc/connector/controlplane/contract/validation/ContractValidationServiceImpl.java
```diff
@@ -40,7 +40,7 @@
 import java.util.Optional;
 
 import static java.lang.String.format;
-import static org.eclipse.edc.connector.controlplane.contract.spi.offer.ContractDefinitionResolver.CATALOGING_SCOPE;
+import static org.eclipse.edc.connector.controlplane.contract.ContractCoreExtension.CATALOG_SCOPE;
```
paullatzelsperger, line 43: this creates a dependency onto an extension class from a service class. We should avoid that, if possible, e.g. by moving the constant into an SPI class, or a `XYZConstants` class

spi/control-plane/catalog-spi/src/main/java/org/eclipse/edc/connector/controlplane/catalog/spi/ResolvedContractDefinitions.java
```diff
@@ -0,0 +1,29 @@
...
+package org.eclipse.edc.connector.controlplane.catalog.spi;
+
+import org.eclipse.edc.connector.controlplane.contract.spi.types.offer.ContractDefinition;
+import org.eclipse.edc.policy.model.Policy;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public record ResolvedContractDefinitions(List<ContractDefinition> contractDefinitions, Map<String, Policy> policies) {
```
paullatzelsperger, line 24: this deserves some javadoc

spi/control-plane/catalog-spi/src/main/java/org/eclipse/edc/connector/controlplane/catalog/spi/ContractDefinitionResolver.java
```diff
@@ -0,0 +1,36 @@
+/*
+ *  Copyright (c) 2024 Contributors to the Eclipse Foundation
```
paullatzelsperger, line 2: we only have explicit copyrights everywhere. possibly an old IDE setting :)?

spi/control-plane/catalog-spi/src/main/java/org/eclipse/edc/connector/controlplane/catalog/spi/ResolvedContractDefinitions.java
```diff
@@ -0,0 +1,29 @@
+/*
+ *  Copyright (c) 2024 Contributors to the Eclipse Foundation
```
paullatzelsperger, line 2: we only have explicit copyrights everywhere. possibly an old IDE setting :)?

core/control-plane/control-plane-contract/src/main/java/org/eclipse/edc/connector/controlplane/contract/validation/ContractValidationServiceImpl.java
ndr-brt: we don't have a convention on where to put `@PolicyScope`s, apparently they can be found both in extensions, service interfaces, and service implementations.

Given that they are (or should be) subject to automatic documentation generation, like for the `@Setting` annotations they seem to fit somewhat good in the extension class, but I don't have a strong opinion here.

Let's eventually discuss offline and tackle those in a dedicated issue/PR maybe

paullatzelsperger (approved):

PR #4514, opened 2024-09-30
refactor: supports multi protocols dsp-http

### LINKED ISSUE #4507: Refactor http `dsp` modules
# Feature Request

We should refactor the current dsp module implementation. 

- `DspRequestHandlerImpl` and `JsonLdRemoteMessageSerializerImpl`  should rely on the root `TypeTransformerRegistry` and then fetch the correct `ContextTransformerRegistry` based on the `dsp-api` prefix + version extracted from `RemoteMessage` and `DspRequest`. 

- For backward compatibility if no version is found the lookup for transformers is `dsp-api` otherwise will be `dsp-api:<version>`

core dsp

_Are there any requirements?_

## Solution Proposal

wolf4ood:
## What this PR changes/adds

Supports multi protocols in DspRequest and DspHttpRemoteMessageDispatcherImpl

- Introduced a new `ProtocolVersion` for `v0.8` default one, bind in the default controllers without a version in the path. 
- Introduced `DspProtocolTypeTransformerRegistry` spi for fetching the right `TypeTransformerRegistry` given a protocol. 
- Introduced `DspProtocolParser` for parsing the protocol format `dataspace-protocol-http:<version>` into a `ProtocolVersion`
- Introduced a `ProtocolRemoteMessage` for holding the protocol and changing it in ingress on `DspRequestHandlerImpl`
- `PostDspHttpRequestFactory` and `GetDspHttpRequestFactory` uses the `DspProtocolParser` for extracting the right protocol version from a `RemoteMessage#protocol`

## Why it does that

supporting multiple dsp protocols

## Further notes

Currently both `v0.8` and `2024/1` have the same transformers registered. In subsequent PRs the `2024/1` should be fixed to point to the right namespace.

The `JsonLd` context refactor has no been applied in this PR. Currently they still use the same `JsonLD` configuration under the `DSP` scope. 

## Linked Issue(s)

Closes #4507 

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -30,7 +30,7 @@ maven/mavencentral/com.fasterxml.jackson.core/jackson-databind/2.18.0, Apache-2.
 maven/mavencentral/com.fasterxml.jackson.dataformat/jackson-dataformat-yaml/2.14.0, Apache-2.0, approved, #5933
 maven/mavencentral/com.fasterxml.jackson.dataformat/jackson-dataformat-yaml/2.16.2, Apache-2.0, approved, #11855
 maven/mavencentral/com.fasterxml.jackson.dataformat/jackson-dataformat-yaml/2.18.0, Apache-2.0, approved, #16370
-maven/mavencentral/com.fasterxml.jackson.datatype/jackson-datatype-jakarta-jsonp/2.18.0, , restricted, clearlydefined
+maven/mavencentral/com.fasterxml.jackson.datatype/jackson-datatype-jakarta-jsonp/2.18.0, Apache-2.0, restricted, clearlydefined
 maven/mavencentral/com.fasterxml.jackson.datatype/jackson-datatype-jsr310/2.14.0, Apache-2.0, approved, #4699
 maven/mavencentral/com.fasterxml.jackson.datatype/jackson-datatype-jsr310/2.16.2, Apache-2.0, approved, #11853
 maven/mavencentral/com.fasterxml.jackson.datatype/jackson-datatype-jsr310/2.18.0, , restricted, clearlydefined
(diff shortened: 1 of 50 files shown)


data-protocols/dsp/dsp-catalog/dsp-catalog-http-api/src/main/java/org/eclipse/edc/protocol/dsp/catalog/http/api/decorator/CatalogPaginationResponseDecorator.java
```diff
@@ -44,19 +44,19 @@ public CatalogPaginationResponseDecorator(String requestUrl, ContinuationTokenSe
...
         }
 
         return responseBuilder;
     }
 
-    private void addLink(String rel, Response.ResponseBuilder responseBuilder, QuerySpec currentQuerySpec, IntBinaryOperator newOffsetOperator) {
+    private void addLink(String rel, String protocol, Response.ResponseBuilder responseBuilder, QuerySpec currentQuerySpec, IntBinaryOperator newOffsetOperator) {
         var newOffset = newOffsetOperator.applyAsInt(currentQuerySpec.getOffset(), currentQuerySpec.getLimit());
-        continuationTokenSerDes.serialize(currentQuerySpec.toBuilder().offset(newOffset).build())
+        continuationTokenSerDes.serialize(currentQuerySpec.toBuilder().offset(newOffset).build(), protocol)
```
ndr-brt, line 59: could the `continuationTokenSerDes` instance already contain the correct transformer registry as field without the need to break the interface? (by instantiate two different instances that will be passed to the two different controllers)

ndr-brt (commented):
great stuff, just a comment, and agree about the e2e test refactoring pointed by the TODO, I also found out that the current design has limits

Issue #4440, opened 2024-08-28
data-plane: restart interrupted data flows 


ndr-brt:
# Feature Request

Currently if a data plane goes down while doing handling long-running data flow, there's no way to recover it back other than creating a new transfer process.

the data-flow should be handled by another runtime in the same clustered environment or by the same dataplane if restarted. 

## Which Areas Would Be Affected?

_e.g., DPF, CI, build, transfer, etc._

## Why Is the Feature Desired?

_Are there any requirements?_

## Solution Proposal

_If possible, provide a (brief!) solution proposal._

Issue #4439, opened 2024-08-28
clustered `data-plane` identification


ndr-brt:
# Feature Request

Currently if a data-plane is deployed in a clustered environment there's no way to identify the runtime that started a transfer to being able to suspend/terminate it correctly.

Specifically, a call to the `suspend`/`terminate` method should be applied on all the runtimes of the same component.

## Which Areas Would Be Affected?

_e.g., DPF, CI, build, transfer, etc._

## Why Is the Feature Desired?

_Are there any requirements?_

## Solution Proposal

_If possible, provide a (brief!) solution proposal._

PR #4522, opened 2024-10-02
docs: DR for clustered data-plane

ndr-brt:
## What this PR changes/adds

DR for implementing clustered data-plane.

## Why it does that

## Further notes

## Linked Issue(s)

Part of: #4439 #4440

### DIFF
--- a/docs/developer/decision-records/2024-10-02-clustered-data-plane/README.md
+++ b/docs/developer/decision-records/2024-10-02-clustered-data-plane/README.md
@@ -26,7 +26,12 @@ For termination (the same logic will be duplicated for suspension), in the `Data
 two new `Processor` registered:
 - one filters by `TERMINATING` state and `runtimeId`: it will stop the data flow and transition it to `TERMINATED`
 - one filters by `TERMINATING` and by `updatedAt` passed by at least 2/3 times `flowLease`: it will transition the data flow to 
-  `TERMINATED` (as cleanup for dangling data flows) 
+  `TERMINATED` (as cleanup for dangling data flows).
+
+Note: once the "termination" message is sent from the control-plane to the data-plane and the ACK received, the control-plane
+will consider the `DataFlow` as terminated, and it will continue evaluating the termination logic on the `TransferProcess` 
+(send protocol message, transition to `TERMINATED`).
+We consider this acceptable because `DataFlow` termination is generally a cleanup operation that shouldn't take too much time.
 
 ### Re-start interrupted data flow
 
--- a/docs/developer/decision-records/2024-10-02-clustered-data-plane/README.md
+++ b/docs/developer/decision-records/2024-10-02-clustered-data-plane/README.md
@@ -1,0 +1,39 @@
+# Clustered data-plane
+
+## Decision
+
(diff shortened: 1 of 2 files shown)


wolf4ood (approved):

docs/developer/decision-records/2024-10-02-clustered-data-plane/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Currently, data-plane cannot run effectively in a clustered environment because:
+- there's no way to identify a specific replica that is running a data flow and "terminate/suspend" it
+- there's no way to re-start a data flow that was interrupted because the replica crashed
+
+## Approach
+
+We will provide this feature through the `DataPlaneStore` persistence layer.
+A `runtimeId` will be added in the `DataFlow`, and it will be set when it gets started with the replica's `runtimeId`.
+There will be a configured duration `T` (that can be in milliseconds, seconds at most. A good name for it needs to be found).
```
jimmarino, line 17: how about flow lease? We may want to use the feature in the future to also check for stuck processes but we don't need that at the moment

docs/developer/decision-records/2024-10-02-clustered-data-plane/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Identify specific replica to suspend/terminate
+
+The now synchronous `suspend`/`terminate` will become asynchronous by putting the `DataFlow` in a `-ING` state like:
```
jimmarino, line 21: These will be propagated back to the control plane, correct?

jimmarino (approved):

docs/developer/decision-records/2024-10-02-clustered-data-plane/README.md
ndr-brt: these are releated to `DataFlow`, because `SUSPENDING` and `TERMINATING` in the control-plane have different meaning.
only when the `DataFlow` would been correctly `SUSPENDED` or `TERMINATED` the control-plane will be advertised and it would transition the transfer from the `SUSPENDING`/`TERMINATING` state to `SUSPENDED`/`TERMINATED`

jimmarino: Yes, I meant the transition to Dataflow terminated, now that the operation is async. There's two ways to approach this:

1. Have the control plane transition to TERMINATING and wait until the DataFlow is TERMINATED before itself transitioning to TERMINATED
2. Have the control plane send a terminate message to the data plane and immediately transition to TERMINATED once the ack is received.

Appraoch 2 is a lot easier and straightforward without much downside. It does place the burden to handle the terminate message correctly or perform some form of cleanup if it fails but that I think is acceptable.

What do you think? Perhaps this should be documented?

ndr-brt: I think 2 will be enough because termination of data flow it's just a cleanup operation, generally speaking it should be quick to complete. I will add some details to the DR

paullatzelsperger (approved):

PR #4535, opened 2024-10-10
refactor: bind policy scope to policy context hierarchy

ndr-brt:
## What this PR changes/adds

NOTE: wait for #4542 to be merged before reviewing

Sorry for the big PR but this is quite an impactful change.
Highlights:
- `PolicyContext` is bound to the string `scope`
- `PolicyContext` hierarchy reflects the scope one
- `PolicyEngine` no more accepts (or better, only methods that been deprecated do) `String scope` but it relies on the `PolicyContext` hierarchy to filter out validators and functions to be applied.

all the other changes are related to that.

note: no breaking changes in the API, as the new interfaces have been set as older ones superinterfaces. Deprecations annotations have been put in place.

Upcoming work:
- adapt `RuleBindingRegistry` to the `PolicyEngine` hierarchy (maybe it could also be inlined in the `PolicyEngine`, as it is only used internally by the latter. 

## Why it does that

_Briefly state why the change was necessary._

## Further notes

_List other areas of code that have changed but are not necessarily linked to the main feature. This could be method
signature changes, package declarations, bugs that were encountered and were fixed inline, etc._

## Linked Issue(s)

Closes #3511 

_Please be sure to take a look at the [contributing guidelines](https://github.com/eclipse-edc/.github/blob/main/CONTRIBUTING.md#submit-a-pull-request) and our [etiquette for pull requests](https://github.com/eclipse-edc/.github/blob/main/contributing/pr_etiquette.md)._

Issue #3511, opened 2023-10-04
policy-engine: typed scope/context


ndr-brt:
# Feature Request

Currently the `PolicyContext` is a free form object that can accept every kind of object, but in fact, the `PolicyContext` structure is strictly dependent from the policy scope, so it could beneficial to have it strictly typed.

## Which Areas Would Be Affected?

policy-engine

## Why Is the Feature Desired?

_Are there any requirements?_

## Solution Proposal

_If possible, provide a (brief!) solution proposal._

PR #4542, opened 2024-10-11
refactor(policy): provide new contexts and function interfaces for policy engine

ndr-brt:
## What this PR changes/adds

Given that #4535 is really huge, I extracted a smaller PR, that takes care of:
- creating new `PolicyContext` implementation
- redefine policy function and policy validator hierarchy 

## Why it does that

help the review process

## Further notes

## Linked Issue(s)

Part of #3511 


core/control-plane/control-plane-aggregate-services/build.gradle.kts
```diff
@@ -17,14 +17,15 @@ plugins {
 }
 
 dependencies {
-    implementation(project(":spi:common:validator-spi"))
-    implementation(project(":spi:control-plane:control-plane-spi"))
+    api(project(":spi:common:boot-spi"))
```
jimmarino, line 20: Why are these exposed as APIs? They are not needed to extend this module.

jimmarino (commented):
One comment about the dependencies

core/control-plane/control-plane-aggregate-services/build.gradle.kts
ndr-brt: must have been some rebase/pr split issue, will remove it

jimmarino: I think a module should only expose another module as an `api` if it is transitively exposed to interact with it, e.g. an SPI interface references a type from the other module.

jimmarino (approved):

wolf4ood (approved):

PR #4557, opened 2024-10-17
refactor: multi-namespace in dsp controllers

### LINKED ISSUE #4510: Refactor `dsp` api modules
# Feature Request

Each Dsp Controller should work with a given DSP message types compatible with the protocol version they implement

_Are there any requirements?_

## Solution Proposal

wolf4ood:
## What this PR changes/adds

Adds support for multi-namespace in dsp controllers by passing the namespace in the controller constructor.

In order to type the available dsp namespaces a class `DspNamespace` has been introduced.

## Why it does that

## Further notes

## Linked Issue(s)

Closes #4510 

### DIFF
--- a/data-protocols/dsp/dsp-spi/src/main/java/org/eclipse/edc/protocol/dsp/spi/type/DspNamespace.java
+++ b/data-protocols/dsp/dsp-spi/src/main/java/org/eclipse/edc/protocol/dsp/spi/type/DspNamespace.java
@@ -14,30 +14,10 @@
 
 package org.eclipse.edc.protocol.dsp.spi.type;
 
-import static org.eclipse.edc.jsonld.spi.Namespaces.DSPACE_SCHEMA;
-
 /**
- * Enum representing different versions of the DSP (Data Space Protocol) namespace.
- * Each version is associated with a specific namespace URI.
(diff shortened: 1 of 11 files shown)


data-protocols/dsp/dsp-spi/src/main/java/org/eclipse/edc/protocol/dsp/spi/type/DspNamespace.java
```diff
@@ -0,0 +1,51 @@
...
+
+package org.eclipse.edc.protocol.dsp.spi.type;
+
+import static org.eclipse.edc.jsonld.spi.Namespaces.DSPACE_SCHEMA;
+
+/**
+ * Enum representing different versions of the DSP (Data Space Protocol) namespace.
+ * Each version is associated with a specific namespace URI.
+ */
+public enum DspNamespace {
```
ndr-brt, line 23: I would model this as a `record`, and then keep the 2 instances as public constants. 
an `enum` is often a blocker for extensibility. I'm not sure we need extensibility here, but I can imagine in the future having a dataspace that wants to use an older version of the protocol while we'll have already deprecated it

ndr-brt (approved):

Issue #4389, opened 2024-08-01
Remove references to `core:common:token-core`


jimmarino:
Currently `:core:common:token-core` is referenced by `dataplane-core`, `auth-delegated`, `jwt-verifiable-credentials`, and `oauth2-core`. This needs to be refactored to  a combination of service and library extensions.

PR #4560, opened 2024-10-18
feat: add JTI validation feature

### LINKED ISSUE #3749: Implement infrastructure to verify `jti` claims
# Feature Request

Related to the refactoring of how tokens are handled in EDC (#3743) we need a way to validate `jti` claims. `jti` claims are supposed to be random strings with a good-as-zero possibility to be repeated within a reasonable amount of time. 

Thus, the recipient of a JWT must record every `jti` claim value until at least time specified by the `exp` claim and reject an incoming JWT if it carries a "known" `jti`. 

After that time, `jti` claims could theoretically be re-used because any prior token carrying the same `jti` have already expired and wouldn't be valid anyway.

identity services, token validation rules

guarding against replay attacks

## Solution Proposal

1. persistent storage for `jti` claims: This could actually be a generic implementation, something like a `TemporaryValueStorage`, which records arbitrary values plus the start and end of their validity periods.
1. both an In-Mem and a Postgres implementation
2. an aggregate service: uses the `TemporaryValueStorage` to CRUD `jti` values
3. a reaper thread running inside the service that cleans out expired entries from the storage

paullatzelsperger:
## What this PR changes/adds

adds a feature to validate the uniqueness of Token-IDs ("jti"). To do that, a `JtiValidationStore` class was implemented (in-mem + SQL) that holds valid JTI strings and is cleaned up periodically by a reaper thread.

Store implementations for in-mem and SQL were added.

## Why it does that

to be able to validate JTI claims

## Further notes
- the `token-core` module should be split up into a `token-core` (containing the extension) and a `token-lib` modules (containing the rules, in-mem store, and other impls etc.). This will come in a subsequent PR (#4389), I didn't want to blow the changeset up even more.
- Token-IDs are recorded in the `EmbeddedSecureTokenService`, but are validated in the `AccessTokenVerifierImpl` of IdentityHub. A related PR will be created shortly.

## Linked Issue(s)

Closes #3749

### DIFF
--- a/extensions/common/store/sql/jti-validation-store-sql/src/test/java/org/eclipse/edc/edr/store/index/sql/SqlJtiValidationStoreExtensionTest.java
+++ b/extensions/common/store/sql/jti-validation-store-sql/src/test/java/org/eclipse/edc/edr/store/index/sql/SqlJtiValidationStoreExtensionTest.java
@@ -1,0 +1,58 @@
+/*
+ *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.edr.store.index.sql;
+
+import org.eclipse.edc.edr.store.index.SqlJtiValidationStore;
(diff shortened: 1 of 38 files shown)


extensions/common/store/sql/jti-validation-store-sql/src/test/java/org/eclipse/edc/edr/store/index/sql/SqlJtiValidationStoreExtensionTest.java
```diff
@@ -0,0 +1,58 @@
+/*
+ *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
```
wolf4ood, line 2:
```suggestion
 *  Copyright (c) 2024 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
```

wolf4ood (approved):

PR #4578, opened 2024-10-23
chore: remove deprecated Dataplane properties

paullatzelsperger:
## What this PR changes/adds

Removes the deprecated `turnCount` and `allowedDestTypes` from the DataPlaneInstance

## Why it does that

deprecated since 0.6 and 0.7 respectively

## Further notes

## Linked Issue(s)

### DIFF
--- a/extensions/control-plane/transfer/transfer-data-plane-signaling/src/test/java/org/eclipse/edc/connector/controlplane/transfer/dataplane/flow/DataPlaneSignalingFlowControllerTest.java
+++ b/extensions/control-plane/transfer/transfer-data-plane-signaling/src/test/java/org/eclipse/edc/connector/controlplane/transfer/dataplane/flow/DataPlaneSignalingFlowControllerTest.java
@@ -66,6 +66,36 @@ public class DataPlaneSignalingFlowControllerTest {
             () -> URI.create("http://localhost"), selectorService, propertiesProvider, dataPlaneClientFactory,
             "random", transferTypeParser);
 
+    @NotNull
+    private DataPlaneInstance.Builder dataPlaneInstanceBuilder() {
+        return DataPlaneInstance.Builder.newInstance().url("http://any");
+    }
+
+    private DataPlaneInstance createDataPlaneInstance() {
+        return dataPlaneInstanceBuilder().build();
+    }
+
+    private DataAddress testDataAddress() {
+        return DataAddress.Builder.newInstance().type("test-type").build();
+    }
+
+    private TransferProcess transferProcess(String destinationType, String transferType) {
+        return TransferProcess.Builder.newInstance()
+                .transferType(transferType)
+                .dataDestination(DataAddress.Builder.newInstance().type(destinationType).build())
+                .build();
+    }
+
+    private TransferProcess.Builder transferProcessBuilder() {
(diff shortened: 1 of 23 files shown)


extensions/control-plane/transfer/transfer-data-plane-signaling/src/test/java/org/eclipse/edc/connector/controlplane/transfer/dataplane/flow/DataPlaneSignalingFlowControllerTest.java
```diff
@@ -66,6 +66,36 @@ public class DataPlaneSignalingFlowControllerTest {
             () -> URI.create("http://localhost"), selectorService, propertiesProvider, dataPlaneClientFactory,
             "random", transferTypeParser);
 
+    @NotNull
+    private DataPlaneInstance.Builder dataPlaneInstanceBuilder() {
```
ndr-brt, line 70: I know that this is likely some automated plugin, but moving private methods above the nested classes that contains the tests causes readability issues.
Automated conventions could be nice, but I would say that in cases like this they could be avoided.

generally speaking, I prefer an approach where "most important stuff" stays at the top, while "least important thing" stays at the bottom.

paullatzelsperger: I agree, but the problem with that while it is possible to sort "nested class before private method", there is no code style rule in IntelliJ (to my knowledge) that lets you specify that by class annotation `@Nested`. 

However, it is possible to arrange "nested classes - private/protected method - static nested classes".

ndr-brt (commented):
another thing that I forgot, the change needs to be reflected on `DataPlaneInstanceSchema` that will make this become a breaking change on management api v3:
https://eclipse-edc.github.io/Connector/openapi/management-api/#/Dataplane%20Selector%20V3/getAllDataPlaneInstancesV3

extensions/control-plane/transfer/transfer-data-plane-signaling/src/test/java/org/eclipse/edc/connector/controlplane/transfer/dataplane/flow/DataPlaneSignalingFlowControllerTest.java
ndr-brt: could it be applied on the other test classes that are involved in the PR?

paullatzelsperger:
> another thing that I forgot, the change needs to be reflected on `DataPlaneInstanceSchema` that will make this become a breaking change on management api v3: https://eclipse-edc.github.io/Connector/openapi/management-api/#/Dataplane%20Selector%20V3/getAllDataPlaneInstancesV3

ok, to do this properly, we'll have to:
- introduce a `v4-alpha` version of the Management API
- not serialize `allowedDestTypes` and `turnCount` through it
- once `v4` is stable, deprecate `v3`
- when `v3` is finally removed, remove the properties from the `DataPlaneInstance`

All in all, this PR should be closed :)

PR #4590, opened 2024-10-29
refactor: optimize service extensions boot process

ndr-brt:
## What this PR changes/adds

Another optimization PR. Main focuses:
- scan `@Provider` annotation only once in the `DependencyGraph`, then: 
    - keep the default providers in the injection point
    - keep the non-default providers in the injection containers
- simplify all the life cycle logic, removing all the `Phase` class hierarchy overhead, so now the lifecycle is easily readable in the `ExtensionLifecycleManager`, plus, less objects are instantiated at startup (before it was 4 for every extension)
- remove the injection container validation, no point in doing that because it's already done by the `DependencyGraph`

The improvements registered by starting an EDC runtime with 78 extensions approx the 15% (mean values out of 10 runs):
- without the refactor it takes approx 2700 millis to startup
- with the refactor it takes approx 2240 millis to startup

## Why it does that

startup optimization

## Further notes

## Linked Issue(s)


jimmarino:
One note let's *not* make the bootstrap process multithreaded (it is single-threaded by design). I've seen this with other runtimes and it has generally resulted in a highly complex process where all operations must be thread-safe for very little (and possibly) slower performance gains.

jimmarino (approved):

PR #4600, opened 2024-11-04
refactor: save one iteration of the extension list in DependencyGraph

ndr-brt:
## What this PR changes/adds

Last PR of the "startup" optimization work (see also #4593 #4590 #4587)

This one ensures that the injection containers are instantiated in the first extensions loop so there's no need to create them in the additional loop that was executed right before method's return.
This saves some other millyseconds and memory, as the `serviceProviders` map is not necessary anymore.

## Why it does that

startup optimization

## Further notes

- removed 2 unused methods in the `ProviderMethodScanner` component

## Linked Issue(s)


paullatzelsperger (approved):
successfully tried out in MVD

PR #4611, opened 2024-11-06
docs: decision record about configuration injection

paullatzelsperger:
## What this PR changes/adds

Adds a decision record that outlines the configuration injection feature

## Linked Issue(s)

### DIFF
--- a/docs/developer/decision-records/2024-11-06-configuration-injection/README.md
+++ b/docs/developer/decision-records/2024-11-06-configuration-injection/README.md
@@ -1,0 +1,139 @@
+# Configuration injection
+
+## Decision
+
+In addition to service injection, the EDC project will support _configuration injection_ ("CI") in future releases.
+
+## Rationale
+
+In an effort to improve the ease-of-use and to lower the barrier of entry for developers we will provide a feature to
+configure extensions with an annotation mechanism. Resolving configuration can result in convoluted and difficult to
+read code.
+
+## Approach
+
+### Requirements
+
+- default values: it should be possible to provide a default value for configuration fields
+- optionality: configuration fields that are not required should be `null` in case there is no config value for them
+- type flexibility: at least `String`, `Integer`, `Long`, `Double` and `Boolean` must be supported
+- annotation-based: the configuration injection is triggered by annotations
+
+Configuration values are resolved during runtime startup, more specifically during the dependency injection phase.
+Technically they are another type of `InjectionPoint`.
+
+### Resolving and injecting plain configuration values
+
+Extension classes can declare fields of type `String`, `Integer`, `Long`, `Double` or `Boolean` and annotate them with
+the `@Setting` annotation:
+
+```java
+public class SomeExtension implements ServiceExtension {
+
+    @Setting(key = "edc.iam.publickey.alias")
+    private String publicKeyAlias;
+}
+```
+
+This would check if a config value `edc.iam.publickey.alias` is present on the `Config` object, and if so, assign its
+value to the field. An injection error would be raised if no config value is found for `edc.iam.publickey.alias`. This
+can be avoided by declaring a default value:
+
+```java
+public class SomeExtension implements ServiceExtension {
+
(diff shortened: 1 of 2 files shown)


wolf4ood (approved):

jimmarino (commented):
When we have errors reported, can we not use exceptions and stack traces? I would prefer a severe message containing the class requiring the injection point, field name, and expected type. That will be easier than reading stacktraces.

paullatzelsperger:
> When we have errors reported, can we not use exceptions and stack traces? I would prefer a severe message containing the class requiring the injection point, field name, and expected type. That will be easier than reading stacktraces.

The runtime won't boot in the case a dependency or a mandatory config setting is not satisfied, so an exception is perfectly fine. In addition, this would impact areas of code outside of the scope of this DR, so that would be a separate discussion anyway.

jimmarino:
> > When we have errors reported, can we not use exceptions and stack traces? I would prefer a severe message containing the class requiring the injection point, field name, and expected type. That will be easier than reading stacktraces.
> 
> The runtime won't boot in the case a dependency or a mandatory config setting is not satisfied, so an exception is perfectly fine, and the information you mentioned is indeed logged. In addition, changing the way dependency errors are reported would impact areas of code outside of the scope of this DR, so that would be a separate discussion anyway. My goal here was to align the error reporting of config injection with that of dependency injection.

That's not what I'm saying. When the missing config error is encountered, it should be reported to a context where it is collated, the loading for that extension is aborted, and loading then continues. After the phase is complete, the context should be checked, and the recorded errors (not exception) reported to the monitor. After the errors are reported, an exception can be thrown (or the runtime terminated).

docs/developer/decision-records/2024-11-06-configuration-injection/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Error reporting
+
+By piggy-backing on the dependency injection mechanism, errors are automatically reported in a collated way. So multiple
+unresolved config values would be reported as one:
+
+```
+Caused by: org.eclipse.edc.boot.system.injection.EdcInjectionException: The following injected fields were not provided or could not be resolved:
+Configuration value "fooBarBaz" of type class java.lang.Long (config key 'foo.bar.baz')
+Configuration object 'config' of type 'class org.eclipse.edc.some.FooBarExtension$FooBarConfig' in class class org.eclipse.edc.some.FooBarExtension
```
ndr-brt, line 133:
```suggestion
Configuration object 'config' of type 'class org.eclipse.edc.some.FooBarExtension$FooBarConfig' in class org.eclipse.edc.some.FooBarExtension
```

ndr-brt (approved):

ndr-brt:
> That's not what I'm saying. When the missing config error is encountered, it should be reported to a context where it is collated, the loading for that extension is aborted, and loading then continues. After the phase is complete, the context should be checked, and the recorded errors (not exception) reported to the monitor. After the errors are reported, an exception can be thrown (or the runtime terminated).

how could loading continues if an extension is aborted? other extensions could be dependent of that one so likely other kind of errors could appear.
As it is done for `@Injection` points, it should fail at the first inconsistency

docs/developer/decision-records/2024-11-06-configuration-injection/README.md
paullatzelsperger: copy paste...

jimmarino:
> > That's not what I'm saying. When the missing config error is encountered, it should be reported to a context where it is collated, the loading for that extension is aborted, and loading then continues. After the phase is complete, the context should be checked, and the recorded errors (not exception) reported to the monitor. After the errors are reported, an exception can be thrown (or the runtime terminated).
> 
> how could loading continues if an extension is aborted? other extensions could be dependent of that one so likely other kind of errors could appear. As it is done for `@Injection` points, it should fail at the first inconsistency

The point of this is to collate all missing configuration points. This may need to perform config injection before service injection to achieve that. During config injection, if there are errors, all operations on the extension in question should cease and the booter moves to the next extension.

jimmarino (approved):

PR #4612, opened 2024-11-08
feat: restart interrupted data flows

ndr-brt:
## What this PR changes/adds

Restarts long running transfers after a data-plane restart

## Why it does that

dataplane should be able to recover the transfers after a reboot

## Further notes

- this only provides the feature in a single data-plane instance environment, clustered environment will be supported in subsequent PRs
- added @Postgresql version of E2E tests for streaming. also because this scenario, that provides for the reboot is only testable with a persistent store and not in memory.
- added a `closeAll` method on `TransferService` that will shut down all the transfers during the shut down phase.
- implemented the `<` operator, I was actually surprised that it didn't already existed. this was necessary, I will provide an issue/PR to implement also the greater than.

the `restartFlows` method will in the future called in a different way as we're reasoning about the startup phases

## Linked Issue(s)

Part of #4440 

### DIFF
--- a/core/common/lib/query-lib/src/test/java/org/eclipse/edc/query/LessThanOperatorPredicateTest.java
+++ b/core/common/lib/query-lib/src/test/java/org/eclipse/edc/query/LessThanOperatorPredicateTest.java
@@ -32,8 +32,14 @@ class LessThanOperatorPredicateTest {
 
     @ParameterizedTest
     @ArgumentsSource(ValidValues.class)
-    void shouldReturnTrue_whenOperandAreNumbers(Object value, Object comparedTo) {
+    void shouldReturnTrue_whenValueLessThanComparedOne(Object value, Object comparedTo) {
         assertThat(predicate.test(value, comparedTo)).isTrue();
+    }
+
+    @ParameterizedTest
+    @ArgumentsSource(InvalidValues.class)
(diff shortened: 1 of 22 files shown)


jimmarino:
Can we get a DR first on the startup processors? I'd like to have this information documented and also have a quick discussion if this should be generalized. A common pattern is to be able to register a general callback during bootstrap for when a runtime enters a ready state.

core/common/lib/query-lib/src/test/java/org/eclipse/edc/query/LessThanOperatorPredicateTest.java
```diff
@@ -0,0 +1,52 @@
...
+    private static class ValidValues implements ArgumentsProvider {
+
+        @Override
+        public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
+            return Stream.of(
+                    arguments(1, 2),
+                    arguments(1, 2L),
+                    arguments(1, 1.01f),
+                    arguments(1, 1.01d),
+                    arguments("a", "b")
```
paullatzelsperger, line 48: nit: maybe add a test case for "b" < "a" and "b" == "b" as well

paullatzelsperger (approved):

wolf4ood (approved):

PR #4616, opened 2024-11-11
docs: DR about state machine startup operation

ndr-brt:
## What this PR changes/adds

Decision about "startup processors" in the state machine component (as requested in https://github.com/eclipse-edc/Connector/pull/4612#issuecomment-2464999629)

## Why it does that

## Further notes

## Linked Issue(s)

### DIFF
--- a/docs/developer/decision-records/2024-11-11-state-machine-startup-processors/README.md
+++ b/docs/developer/decision-records/2024-11-11-state-machine-startup-processors/README.md
@@ -1,0 +1,32 @@
+# State machine startup processors
+
+## Decision
+
+We will add the possibility to register one-time, at-startup processors to the `StateMachine` component.
+
+## Rationale
+
+A state machine could need some bootstrap work to getting ready at startup, as restarting stopped jobs.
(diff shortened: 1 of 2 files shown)


jimmarino (commented):
Why wouldn't we make this functionality more general buy creating a startup hook for all extensions?

ndr-brt:
> Why wouldn't we make this functionality more general buy creating a startup hook for all extensions?

I don't think this should be part of the extension lifecycle, it's strictly state-machine related.
Plus, doing some "data" related operations during startup phase would make the runtime slower to start for no good reason.

jimmarino:
> > Why wouldn't we make this functionality more general buy creating a startup hook for all extensions?
> 
> I don't think this should be part of the extension lifecycle, it's strictly state-machine related. Plus, doing some "data" related operations during startup phase would make the runtime slower to start for no good reasons.

Let's discuss this during the Wednesday weekly. This is a common pattern used for many runtime boot requirements, for example, transaction and other types of recovery.

PR #4621, opened 2024-11-13
feat: adds flag to enable validation on create/update policy definitions

### LINKED ISSUE #4619: Configurable policy definition validation
# Feature Request

Currently we validate the `JsonObject` of the policy definition using the `Validator` pattern when creating and updating the policy definition. The `Validator` only validates the structure of the json input. 

Recently we introduced a set of API in order to validate and get the execution plan of a given policy within a scope.
https://github.com/eclipse-edc/Connector/issues/4422

Would be cool to plug the validator optionally when creating and updating the policy definition for a stricter validation
based on the configured `PolicyEngine`

  

_Are there any requirements?_

## Solution Proposal

wolf4ood:
## What this PR changes/adds

adds a configuration `edc.policy.validation.enabled` for enabling the call to `PolicyEngine#validate` while creating/updating the `PolicyDefinition` 

## Why it does that

improves policy management

## Further notes

Renamed also the failure message from `left operand` to `leftOperand` for consistency 

## Linked Issue(s)

Closes #4619

### DIFF
--- a/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/ControlPlaneServicesExtension.java
+++ b/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/ControlPlaneServicesExtension.java
@@ -75,6 +75,7 @@ import org.eclipse.edc.policy.engine.spi.PolicyEngine;
 import org.eclipse.edc.runtime.metamodel.annotation.Extension;
 import org.eclipse.edc.runtime.metamodel.annotation.Inject;
 import org.eclipse.edc.runtime.metamodel.annotation.Provider;
+import org.eclipse.edc.runtime.metamodel.annotation.Setting;
 import org.eclipse.edc.spi.command.CommandHandlerRegistry;
 import org.eclipse.edc.spi.event.EventRouter;
 import org.eclipse.edc.spi.iam.IdentityService;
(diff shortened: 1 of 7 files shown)


jimmarino (approved):

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/ControlPlaneServicesExtension.java
```diff
@@ -99,6 +100,9 @@ public class ControlPlaneServicesExtension implements ServiceExtension {
 
     public static final String NAME = "Control Plane Services";
 
+    @Setting(key = "edc.policy.validation.enabled", description = "If true enable the policy validation when creating and updating policy definitions", defaultValue = "false")
```
ndr-brt, line 103:
```suggestion
    @Setting(key = "edc.policy.validation.enabled", description = "If true enables the policy validation when creating and updating policy definitions", defaultValue = "false")
```

I'm also wondering if we want to set this disabled by default. I'd see more value in having it enabled by defaut

ndr-brt (approved):

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/ControlPlaneServicesExtension.java
wolf4ood: For now i think we have to explicitly enable it, first because it's still part of the experimental period i think
and second most important if it's enabled by default it's a breaking change :)

ndr-brt: ok, but having an invalid policy in the system couldn't be considered as a malfunction?

ndr-brt: ok, so, let's merge! :man_dancing:

PR #4633, opened 2024-11-18
feat: use new Configuration Injection mechanism everywhere

### LINKED ISSUE #4610: Feature: Configuration Injection
# Feature Request

Implement a feature that enables _configuration injection_. By annotating certain field values it should be possible to automatically resolve and set config values, for example:

```java
public class SomeExtension extends ServiceExtension {

    @Setting(key = "edc.some.config.value", required = true, defaultValue = "barbaz")
    private String someConfigValue
}
```

thus, `someConfigValue` gets assigned the value of `edc.some.config.value`, or - lacking that - the default value `"barbaz"`. 

In addition, it should be able to group config values together in a _configuration object_, which is basically a POJO:

```java
public class SomeExtension extends ServiceExtension {

    @Configuration
    private SomeConfigObject configObject;

}

public record SomeConfigObject(@Setting(key="...") String stringValue, 
                               @Setting(key="...") Double doubleValue){

}
```

dependency injection, ultimately the entire code base

ease of use

## Solution Proposal

enhance the `InjectionPoint` implementations, move some responsibility into the `FieldInjectionPoint` to allow changing the behaviour in other impls.

## Subtasks

- [x] Decision-Record #4611
- [ ] Implementation of the configuration injection (including runtime metamodel and autodoc) 
  - https://github.com/eclipse-edc/Runtime-Metamodel/pull/75
  - https://github.com/eclipse-edc/GradlePlugins/pull/289
  - https://github.com/eclipse-edc/Connector/pull/4617
  - https://github.com/eclipse-edc/Connector/pull/4620
  - https://github.com/eclipse-edc/Connector/pull/4628
- [ ] Adapting the code base to utilize this feature
  - [ ] https://github.com/eclipse-edc/Connector/pull/4633
  - [ ] https://github.com/eclipse-edc/IdentityHub/pull/491
  - [ ] https://github.com/eclipse-edc/FederatedCatalog/pull/265
- [ ] Update the EDC documentation

paullatzelsperger:
## What this PR changes/adds

This PR upgrades the entire code base to use the new Configuration Injection feature that was developed recently.

Note that this does **not** include:
- runtime id and component id, as they are not read by an extension but by the `ServiceExtensionContext`
- any property with a `context` attribute, as that is not yet implemented

## Why it does that

reduce code verbosity

## Further notes

- in tests, the `ObjectFactory` has to be used to construct the extension, because at the time the extension is injected into the test method, it is already fully constructed (i.e. all config values are already assigned) and that means, that mocking config values _in_ the test method is not possible, because it's too late. Directly injecting the extension into test methods only works when the config is set directly in the `@BeforeEach` method.
- sorry for the big change set

## Linked Issue(s)

Closes #4610

### DIFF
--- a/core/common/boot/src/main/java/org/eclipse/edc/boot/BootServicesExtension.java
+++ b/core/common/boot/src/main/java/org/eclipse/edc/boot/BootServicesExtension.java
@@ -37,13 +37,13 @@ public class BootServicesExtension implements ServiceExtension {
 
     public static final String NAME = "Boot Services";
 
-    @Setting(value = "Configures the participant id this runtime is operating on behalf of")
+    @Setting(description = "Configures the participant id this runtime is operating on behalf of")
     public static final String PARTICIPANT_ID = "edc.participant.id";
 
-    @Setting(value = "Configures the runtime id. This should be fully or partly randomized, and need not be stable across restarts. It is recommended to leave this value blank.", defaultValue = "<random UUID>")
+    @Setting(description = "Configures the runtime id. This should be fully or partly randomized, and need not be stable across restarts. It is recommended to leave this value blank.", defaultValue = "<random UUID>")
     public static final String RUNTIME_ID = "edc.runtime.id";
 
-    @Setting(value = "Configures this component's ID. This should be a unique, stable and deterministic identifier.", defaultValue = "<random UUID>")
+    @Setting(description = "Configures this component's ID. This should be a unique, stable and deterministic identifier.", defaultValue = "<random UUID>")
     public static final String COMPONENT_ID = "edc.component.id";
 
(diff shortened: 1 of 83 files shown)


ndr-brt (approved):
just some nits, is not easy to go deep in this :)

about the `DependencyInjectionExtension`, why not have the `Configuration` object automatically injected as mock objects so in every test the object can be controlled accordingly?

paullatzelsperger:
> just some nits, is not easy to go deep in this :)

yeah, i know, unfortunately there isn't a good way to segment this PR :(

> about the `DependencyInjectionExtension`, why not have the `Configuration` object automatically injected as mock objects so in every test the object can be controlled accordingly?

PR #4634, opened 2024-11-19
docs: decision record about transformers versioning scheme

wolf4ood:
## What this PR changes/adds

Decision Record about alternative versioning scheme


jimmarino (approved):

paullatzelsperger (approved):

ndr-brt (approved):
I don't remember why I put that extension, if that can be removed, let's do it :rocket:

PR #4646, opened 2024-11-26
fix: not required properties in DcpDefaultServicesExtension

wolf4ood:
## What this PR changes/adds

The following properties config in `DcpDefaultServicesExtension` are marked required:

- `edc.iam.sts.publickey.id`
- `edc.iam.sts.privatekey.alias`

They are only required when the embedded STS is used. 

## Why it does that

It fixes the case where the `SecureTokenService` is not default (provided by another extension)

but the above properties are still required

## Further notes

## Linked Issue(s)

### DIFF
--- a/extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/DcpDefaultServicesExtension.java
+++ b/extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/DcpDefaultServicesExtension.java
@@ -87,7 +87,7 @@ public class DcpDefaultServicesExtension implements ServiceExtension {
 
         if (context.getSetting(OAUTH_TOKENURL_PROPERTY, null) != null) {
             context.getMonitor().warning("The property '%s' was configured, but no remote SecureTokenService was found on the classpath. ".formatted(OAUTH_TOKENURL_PROPERTY) +
-                    "This could be an indicator of a configuration problem.");
+                    "This could be an indication of a configuration problem.");
         }
 
         return new EmbeddedSecureTokenService(new JwtGenerationService(externalSigner), () -> privateKeyAlias, () -> publicKeyId, clock, TimeUnit.MINUTES.toSeconds(stsTokenExpirationMin), jtiValidationStore);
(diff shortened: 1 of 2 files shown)


paullatzelsperger (commented):
couldn't we leave the "new" injection mechanism in place, but declare it as `isRequired=false`, and then throw an exception in the default provider method, if the value is null?

paullatzelsperger:
> We could but it's actually required if you use the embedded sts

yes, exactly, so in the provider method for the embedded STS (`createDefaultTokenService`) we could check if the values are null and log a warning or throw an exception. the provider method won't get invoked if the embedded STS isn't used

wolf4ood:
I was thinking more on the autodoc that we would print that's not required, while it is

paullatzelsperger:
> I was thinking more on the autodoc that we would print that's not required, while it is

yes, but if we make it required, it'll come up as "required" in autodoc, even if it isn't required, so same problem. I guess we could implement a "delegated requirement", which would be the cleanest solution.

for now, I'd keep using the config injection (`required=false`), and explain in the description that it is indeed required if the default service is used, etc.

extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/DcpDefaultServicesExtension.java
```diff
@@ -75,13 +75,21 @@ public class DcpDefaultServicesExtension implements ServiceExtension {
...
+
+        if (!errors.isEmpty()) {
+            var msg = String.join(", ", errors);
+            throw new EdcException("The following errors were found in the configuration of the Embedded STS: [%s]".formatted(msg));
+        }
 
         if (context.getSetting(OAUTH_TOKENURL_PROPERTY, null) != null) {
             context.getMonitor().warning("The property '%s' was configured, but no remote SecureTokenService was found on the classpath. ".formatted(OAUTH_TOKENURL_PROPERTY) +
-                                         "This could be an indicator of a configuration problem.");
+                    "This could be an indicator of a configuration problem.");
```
paullatzelsperger, line 90:
```suggestion
                    "This could be an indication of a configuration problem.");
```

paullatzelsperger (approved):

Issue #4591, opened 2024-10-30
Fatal error while resume PULL transfer process


AndrYurk:
# Bug Report

## Describe the Bug

Not possible to Resume PULL transfer process from provider side due to fatal error on consumer side request.

### Expected Behavior

Resume called from provider should start transfer process without exceptions.

### Observed Behavior
Currently system throw an exception on consumer side while trying to change state from "SUSPENDED" to "STARTED".

```
[PROVIDER] DEBUG 2024-10-29T16:08:13.635279121 Resumption requested for TransferProcess with ID dd7d56ca-38e9-4203-afc8-b826592b9a5e
[CONSUMER] DEBUG 2024-10-29T16:08:13.850276352 [ExecutionManager] Run pre-execution task
[CONSUMER] DEBUG 2024-10-29T16:08:13.850403681 [ExecutionManager] No WorkItems found, skipping execution
[PROVIDER] DEBUG 2024-10-29T16:08:14.242948008 TransferProcess: ID dd7d56ca-38e9-4203-afc8-b826592b9a5e. Resume data flow
[PROVIDER] DEBUG 2024-10-29T16:08:14.246843283 [DataPlaneManagerImpl] DataFlow dd7d56ca-38e9-4203-afc8-b826592b9a5e is now in state STARTED
[PROVIDER] DEBUG 2024-10-29T16:08:14.246929808 TransferProcess: ID dd7d56ca-38e9-4203-afc8-b826592b9a5e. send transfer start to http://localhost:3326/protocol
[CONSUMER] DEBUG 2024-10-29T16:08:14.27007983 DSP: Incoming TransferStartMessage for class org.eclipse.edc.connector.controlplane.transfer.spi.types.TransferProcess process: 96942366-01d6-4660-a39d-01fea334106e
[CONSUMER] DEBUG 2024-10-29T16:08:14.274445504 DSP: Service call failed: Cannot process TransferStartMessage because transfer cannot be started
[PROVIDER] DEBUG 2024-10-29T16:08:14.490688961 TransferProcess: ID dd7d56ca-38e9-4203-afc8-b826592b9a5e. send transfer start to http://localhost:3326/protocol
[...]

PR #4647, opened 2024-11-26
fix: Fatal error while resume PULL transfer from provider

AndrYurk:
## What this PR changes/adds

Adds a ```SUSPENDED``` state to the list of allowed states to be consumer started.

## Why it does that

When provider resume the transfer it initiate transitionStarted for the consumer from ```SUSPENDED``` state, that is currently not allowed, as a result consumer return an error for the provider and transfer move to ```TERMINATED``` state. 

## Further notes

To make possible create e2e test by provider commands, method awaitTransferToBeInState from TransferEndToEndTestBase need to be unified.

## Linked Issue(s)

Closes [#4591](https://github.com/eclipse-edc/Connector/issues/4591)


ndr-brt (approved):
LGTM

PR #4678, opened 2024-12-16
chore(DataPlanePublicAPI): Error message handling

### LINKED ISSUE #4670: DataplanePublicApiControler completable future failure messages are returned as contatenated string
# Bug Report

## Describe the Bug

In the `DataplanePublicApiV2Controler`, if the `CompletableFuture` handled by the `processRequest` method has a result with multiple failure messages, these messages will all be concatenated into a single string, even though the `TransferErrorResponse` handles a list of errors.

### Expected Behavior

Since the end result to the client is a list of errors, it would be expected that response includes an actual list of errors and not a concatenated string of errors.

### Observed Behavior

image

Here a Details list should be a single error, but all are concatenated.

kkotowiczz:
## What this PR changes/adds
Changed error handling method to be returned directly as list, instead of being concatenated first and then wrapped to the list

## Why it does that

In order to make an error response being more readable

## Further notes
N/A
## Linked Issue(s)
Closes #4670


jimmarino (changes requested):
Please provide unit tests

extensions/data-plane/data-plane-public-api-v2/src/test/java/org/eclipse/edc/connector/dataplane/api/controller/DataPlanePublicApiV2ControllerTest.java
```diff
@@ -117,7 +117,24 @@ void should_returnInternalServerError_if_transferFails() {
...
+                .thenReturn(completedFuture(StreamResult.error(errorMsg)));
+
+        baseRequest()
+                .header(AUTHORIZATION, token)
+                .when()
+                .post("/any")
+                .then()
+                .statusCode(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode())
+                .contentType(JSON)
+                .body("errors", is(List.class));
```
jimmarino, line 137: This test does not properly verify the change. Compare it to how the original test verifies the error messages.

jimmarino (changes requested):
This PR does not properly test the changes. Please fix this.

jimmarino:
In the future, please:

1. Open an issue
2. After your issue has been approved, please create a DRAFT PR to avoid repetitive reviews.

Thx

extensions/data-plane/data-plane-public-api-v2/src/test/java/org/eclipse/edc/connector/dataplane/api/controller/DataPlanePublicApiV2ControllerTest.java
```diff
@@ -117,7 +120,33 @@ void should_returnInternalServerError_if_transferFails() {
...
+                .then()
+                .statusCode(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode())
+                .contentType(JSON)
+                .body("errors", isA(List.class))
+                .extract()
+                .jsonPath();
+        
+        var errors = response.getList("errors", String.class);
+        assertEquals(firstErrorMsg, errors.get(0));
+        assertEquals(secondErrorMsg, errors.get(1));
```
ndr-brt, line 149: you can use restassured to test this, like:
```
.body("errors[0]", is(firstErrorMsg)
```

ndr-brt (approved):
LGTM

PR #4680, opened 2024-12-17
feat: Add resource parameter to the OAuth2 token request to follow RFC-8707

### LINKED ISSUE #4668: Support of resource parameter when a connector request an access token from an IDP
# Feature Request

IAM/Oauth2

A `resource` parameter, as defined by the [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707), allows a client (EDC) to request an access token to its IDP by specify the `aud` claim that should be added to the token.Without this parameter the IDP have no mean to restrict the usage of the token to a specific EDC. In a Dataspace where multiples organisations manages their connectors the missing of this feature can allow a organisation to use an access token to target another EDC that it was unintended for.

## Solution Proposal

Update the `org.eclipse.edc.iam.oauth2.identity.Oauth2ServiceImpl` to add this parameter with a value corresponding to the target EDC URL.

scandinave:
## What this PR changes/adds

This PR makes the OAuth2 implementation ask for a token with a resource from param to follow the RFC-8707. 

## Why it does that

This allows an IDP to set the correct `aud` claim in the access token given to the connector.

## Linked Issue(s)

Closes #4668

### DIFF
--- a/extensions/common/iam/oauth2/oauth2-core/src/main/java/org/eclipse/edc/iam/oauth2/identity/Oauth2ServiceImpl.java
+++ b/extensions/common/iam/oauth2/oauth2-core/src/main/java/org/eclipse/edc/iam/oauth2/identity/Oauth2ServiceImpl.java
@@ -19,7 +19,7 @@ package org.eclipse.edc.iam.oauth2.identity;
 
 import org.eclipse.edc.iam.oauth2.spi.client.Oauth2Client;
 import org.eclipse.edc.iam.oauth2.spi.client.Oauth2CredentialsRequest;
-import org.eclipse.edc.iam.oauth2.spi.client.PrivateKeyOauth2CredentialsRequest;
+import org.eclipse.edc.iam.oauth2.spi.client.PrivateKeyOauth2CredentialsRequest.Builder;
 import org.eclipse.edc.jwt.spi.JwtRegisteredClaimNames;
 import org.eclipse.edc.keys.spi.PublicKeyResolver;
 import org.eclipse.edc.spi.iam.ClaimToken;
@@ -54,6 +54,7 @@ public class Oauth2ServiceImpl implements IdentityService {
     private final TokenValidationService tokenValidationService;
     private final PublicKeyResolver publicKeyResolver;
     private final TokenValidationRulesRegistry tokenValidationRuleRegistry;
+    private final boolean tokenResourceEnabled;
 
     /**
      * Creates a new instance of the OAuth2 Service
@@ -63,10 +64,11 @@ public class Oauth2ServiceImpl implements IdentityService {
      * @param client                 client for Oauth2 server
      * @param jwtDecoratorRegistry   Registry containing the decorator for build the JWT
      * @param tokenValidationService Service used for token validation
+     * @param tokenResourceEnabled   Add support for generating access token request with resource parameter
      */
     public Oauth2ServiceImpl(String tokenUrl, TokenGenerationService tokenGenerationService, Supplier<String> privateKeyIdSupplier,
                              Oauth2Client client, TokenDecoratorRegistry jwtDecoratorRegistry, TokenValidationRulesRegistry tokenValidationRuleRegistry, TokenValidationService tokenValidationService,
-                             PublicKeyResolver publicKeyResolver) {
+                             PublicKeyResolver publicKeyResolver, boolean tokenResourceEnabled) {
         this.tokenUrl = tokenUrl;
         this.privateKeySupplier = privateKeyIdSupplier;
         this.client = client;
@@ -75,6 +77,7 @@ public class Oauth2ServiceImpl implements IdentityService {
         this.tokenGenerationService = tokenGenerationService;
         this.tokenValidationService = tokenValidationService;
         this.publicKeyResolver = publicKeyResolver;
+        this.tokenResourceEnabled = tokenResourceEnabled;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


scandinave:
I've updated tests to validate the new parameter.

jimmarino (changes requested):
This can change existing behavior in some cases. If the resource is not set as a param, it should not be set on the client OAuth request. 

Also, this should be an opt-in configuration parameter that is off by default (please document with `@Setting`). This behavior also needs to be tested.

jimmarino:
> I was thinking that the resource will always be set because there is always a counterPartyAddress right? Connector validates the audience of the access token when receiving request. If no resource is set, the audience will not match.
> 
> But I can definitely make a configurable option to enable this on demands for people that don't care about that.
> 
> I will rework my PR.

Yes, please do to preserve the existing behavior.

scandinave:
I've had a test in the Daps extension that highlight the current situation. It is called Rfc8707IntegrationTest. With this PR, if we have the DAPS extension that tells the IDP to set a specific audience and we also enabled the new resource parameter feature, the audience will have the value sent via the resource parameter. 

What behavior do you expect for this ? : 

* Always make the resource parameter do its job
* Make the Daps override the resource parameter
* Make the Daps throw a warning/error telling why it can't due its job.

scandinave:
Can you reopen this PR please ?

extensions/common/iam/oauth2/oauth2-daps/src/test/java/org/eclipse/edc/iam/oauth2/daps/Rfc8707IntegrationTest.java
```diff
@@ -0,0 +1,104 @@
+/*
```
jimmarino, line 1: This header needs to be changed to your corporate sponsor that aligns with the CLA

extensions/common/iam/oauth2/oauth2-daps/src/test/java/org/eclipse/edc/iam/oauth2/daps/Rfc8707IntegrationTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.nio.file.Path;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.eclipse.edc.junit.testfixtures.TestUtils.findBuildRoot;
+
+@ExtendWith(EdcExtension.class)
+@ComponentTest
+public class Rfc8707IntegrationTest {
```
jimmarino, line 40: Please don't use an integration test for this as they slow down the build. Unit tests should cover this.

jimmarino (changes requested):
Comments inline

extensions/common/iam/oauth2/oauth2-daps/src/test/java/org/eclipse/edc/iam/oauth2/daps/Rfc8707IntegrationTest.java
scandinave: I have made this test because I found that the `DapsIntegrationTest` test fail if in used with the feature flag of this PR set to true. This test is to demonstrate the concomitance used of the Daps extension with the new feature but with a correct configuration. 
I don't know why someone would use both at the same time so if you think, it is not needed, I can remove it. The new feature is already covert by a updated unit test included in this PR.

extensions/common/iam/oauth2/oauth2-daps/src/test/java/org/eclipse/edc/iam/oauth2/daps/Rfc8707IntegrationTest.java
scandinave: In progress on our side

extensions/common/iam/oauth2/oauth2-daps/src/test/java/org/eclipse/edc/iam/oauth2/daps/Rfc8707IntegrationTest.java
jimmarino: Yes, please delete it. DAPs will be removed at some point as well as it is not a production-grade identity provider.

scandinave:
Unit test are in failure on the main branch. See #4699. Waiting for the correction.

scandinave:
@jimmarino Do I need to wait for a review by the second reviewer, or it is ok for you to merge this PR ?

extensions/common/iam/oauth2/oauth2-core/src/main/java/org/eclipse/edc/iam/oauth2/identity/Oauth2ServiceImpl.java
```diff
@@ -98,12 +102,16 @@ private Result<String> generateClientAssertion() {
 
     @NotNull
     private Oauth2CredentialsRequest createRequest(TokenParameters parameters, String assertion) {
-        return PrivateKeyOauth2CredentialsRequest.Builder.newInstance()
+        PrivateKeyOauth2CredentialsRequest.Builder<?> builder = Builder.newInstance()
```
jimmarino, line 105: Please use var

jimmarino (changes requested):
One minor change and it will be good to merge.

extensions/common/iam/oauth2/oauth2-core/src/main/java/org/eclipse/edc/iam/oauth2/identity/Oauth2ServiceImpl.java
scandinave: Done. Thanks for your support.

jimmarino (approved):

PR #4685, opened 2024-12-19
refactor(test): improving `EmbeddedRuntime` exception handling

ndr-brt:
## What this PR changes/adds

Avoids unnecessary exception wrapping in the `EmbeddedRuntime` `boot` method

## Why it does that

improve test failure message

## Further notes

- changed the way the properties are provided to the runtime, legacy way was to set them tosystem properties, I changed this behavior by registering a `ConfigurationExtension` that will load them into a `Config` object directly, more EDC-y :)

## Linked Issue(s)

### DIFF
--- a/core/common/junit/src/main/java/org/eclipse/edc/junit/extensions/EmbeddedRuntime.java
+++ b/core/common/junit/src/main/java/org/eclipse/edc/junit/extensions/EmbeddedRuntime.java
@@ -18,16 +18,17 @@ import org.eclipse.edc.boot.system.runtime.BaseRuntime;
 import org.eclipse.edc.spi.EdcException;
 import org.eclipse.edc.spi.monitor.ConsoleMonitor;
 import org.eclipse.edc.spi.monitor.Monitor;
+import org.eclipse.edc.spi.system.ConfigurationExtension;
 import org.eclipse.edc.spi.system.ServiceExtensionContext;
 import org.eclipse.edc.spi.system.SystemExtension;
 import org.eclipse.edc.spi.system.configuration.Config;
+import org.eclipse.edc.spi.system.configuration.ConfigFactory;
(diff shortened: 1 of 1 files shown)


core/common/junit/src/main/java/org/eclipse/edc/junit/extensions/EmbeddedRuntime.java
```diff
@@ -70,54 +71,51 @@ private EmbeddedRuntime(MultiSourceServiceLocator serviceLocator, String name, M
...
+        var runtimeThrowable = new AtomicReference<Throwable>();
+        var latch = new CountDownLatch(1);
 
-                    Thread.currentThread().setContextClassLoader(classLoader);
+        runtimeThread = executorService.submit(() -> {
+            try {
+                var classLoader = URLClassLoader.newInstance(classPathEntries);
 
-                    super.boot(false);
+                Thread.currentThread().setContextClassLoader(classLoader);
```
jimmarino, line 86: I just saw this now. Technically this should reset the previous classloader for the TCCL in a finally block (on error) or when the runtime shuts down. Probably for another PR.

ndr-brt: It shouldn't be necessary as this code runs in a separated thread, so this class assignment is done only for the specific thread

ndr-brt: it's currently working like this:
https://github.com/eclipse-edc/Connector/blob/c530bcf29a86e5b1ed1218eda5ec2c75bc4bc838/core/common/junit/src/main/java/org/eclipse/edc/junit/extensions/EmbeddedRuntime.java#L87

PR #4687, opened 2024-12-19
fix: transfer type resolution on dp self registration

### LINKED ISSUE #4686: Transfer Type is wrongly resolved during dataplane self registration
# Bug Report

## Describe the Bug

With the bi-directional transfer feature, a set of supported `responseTypes` was introduced in the public endpoint generation service. These supported responses types are concatened to the overall supported transferTypes of the dataplane.

Since the feature was initially proposed with the transfer types to be `HttpData-PULL/Websocket`, but then changed to `HttpData-PULL-Websocket`, some reminiscences of the first proposal where not removed.

The self registration extensions tests didn't caught this because no response types where added to the tests and "publicEndpointGeneratorService.supportedResponseTypes()" returns an empty set, not a null one.

### Expected Behavior

When there are supported response types, the dataplane self registration extension should advertise to the control plane the supported transfer types in the correct format (HttpData-PULL-Websocket)

rafaelmag110:
## What this PR changes/adds

The responseType is now concatenated to the transferType as expected.

## Why it does that

It was not being correctly concatenated.

## Further notes

Also fixed some code docs lying around that had "HttpData-PULL/Websocket"

## Linked Issue(s)

Closes #4686

### DIFF
--- a/core/control-plane/control-plane-transfer/src/main/java/org/eclipse/edc/connector/controlplane/transfer/flow/TransferTypeParserImpl.java
+++ b/core/control-plane/control-plane-transfer/src/main/java/org/eclipse/edc/connector/controlplane/transfer/flow/TransferTypeParserImpl.java
@@ -25,7 +25,7 @@ public class TransferTypeParserImpl implements TransferTypeParser {
 
     /**
      * Parses a compose transfer type string into a {@link TransferType}:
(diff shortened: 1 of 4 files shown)


rafaelmag110:
@paullatzelsperger as reviewer as well since he implemented the bi-directional channel feature.

paullatzelsperger (approved):

Issue #4592, opened 2024-10-30
Termination of pull transfer process from consumer side not success in real but stated as "TERMINATED".


AndrYurk:
# Bug Report

## Describe the Bug

During termination PULL transfer process from consumer, system does not call method terminate() from DataFlowManager on provider side.
As a result process not terminated in real but transfer process had state "TERMINATED", and then it is not possible to initiate terminating process again not from consumer neither form provider side.

### Expected Behavior

Terminating from consumer should invoke org.eclipse.edc.connector.controlplane.transfer.flow.DataFlowManagerImpl#terminate on provider side.

### Observed Behavior

Terminating from consumer does not invoke org.eclipse.edc.connector.controlplane.transfer.flow.DataFlowManagerImpl#terminate on provider side.

## Steps to Reproduce
End to end test for tractusx-edc/edc-tests/edc-controlplane/transfer-tests/TransferPullEndToEndTest
```
@Test
        void transferData_withTerminateByConsumer() {
            var assetId = "api-asset-1";

            var requestDefinition = request().withMethod("GET").withPath(MOCK_BACKEND_PATH);

            Map<String, Object> dataAddress = Map.of(
                    "baseUrl", privateBackendUrl,
                    "type", "HttpData",
                    "contentType", "application/json"
            );

            PROVIDER.createAsset(assetId, Map.of(), dataAddress);

            var accessPolicyId = PROVIDER.createPolicyDefinition(createAccessPolicy(CONSUMER.getBpn()));
            var contractPolicyId = PROVIDER.createPolicyDefinition(createContractPolicy(CONSUMER.getBpn()));
            PROVIDER.createContractDefinition(assetId, "def-1", accessPolicyId, contractPolicyId);
            var transferProcessId = CONSUMER.requestAssetFrom(assetId, PROVIDER).withTransferType("HttpData-PULL").execute();

            CONSUMER.waitForTransferProcess(transferProcessId, TransferProcessStates.STARTED);

            // wait until EDC is available on the consumer side
            server.when(requestDefinition).respond(response().withStatusCode(200).withBody("test response"));

            var edr = CONSUMER.edrs().waitForEdr(transferProcessId);

            // consumer can fetch data with a valid token
            var data = CONSUMER.data().pullData(edr, Map.of());
            assertThat(data).isNotNull().isEqualTo("test response");

            server.verify(requestDefinition, VerificationTimes.exactly(1));

[...]

PR #4688, opened 2024-12-20
fix: Termination of pull transfer process from consumer side not success

AndrYurk:
## What this PR changes/adds
Start termination process from data flow manager for provider when termination initiated by consumer

## Why it does that
During the termination of PULL transfer process from consumer, system does not call method terminate() from DataFlowManager on provider side.
As a result process not terminated in real but transfer process had state "TERMINATED", and then it is not possible to initiate terminating process again not from consumer neither form provider side.

## Further notes
After provider moved to TERMINATED state, it also should move to the DEPROVISIONING automatically

## Linked Issue(s)

Closes [#4592](https://github.com/eclipse-edc/Connector/issues/4592)

### DIFF
--- a/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/transferprocess/TransferProcessProtocolServiceImpl.java
+++ b/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/transferprocess/TransferProcessProtocolServiceImpl.java
@@ -212,13 +212,13 @@ public class TransferProcessProtocolServiceImpl implements TransferProcessProtoc
 
     @NotNull
     private ServiceResult<TransferProcess> suspendedAction(TransferSuspensionMessage message, TransferProcess transferProcess) {
-        if (transferProcess.canBeTerminated()) {
-            if (transferProcess.getType() == PROVIDER) {
-                var suspension = dataFlowManager.suspend(transferProcess);
-                if (suspension.failed()) {
-                    return ServiceResult.conflict("Cannot suspend transfer process %s: %s".formatted(transferProcess.getId(), suspension.getFailureDetail()));
-                }
+        if (transferProcess.getType() == PROVIDER) {
+            var suspension = dataFlowManager.suspend(transferProcess);
+            if (suspension.failed()) {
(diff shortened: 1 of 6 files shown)


jimmarino (commented):
There are a lot of tests changed in a critical area of the codebase. Can you explain what those changes are and also, why `@Nested` is being used?

AndrYurk:
@jimmarino, I used @Nested to combine all the tests related to the termination process (similar to what was done for Suspended, StartedResumed, and Started).

Actually, only one new test, `provider_shouldTerminateDataFlowAndTransitionToDeprovisioning`, was added to check the termination of the data flow and the transition to the "DEPROVISIONING" state for the provider transfer process.
The other three tests were left unchanged, they were simply moved and renamed.

jimmarino:
> @jimmarino, I used `@Nested` to combine all the tests related to the termination process (similar to what was done for Suspended, StartedResumed, and Started).
> 
> Actually, only one new test, `provider_shouldTerminateDataFlowAndTransitionToDeprovisioning`, was added to check the termination of the data flow and the transition to the "DEPROVISIONING" state for the provider transfer process. The other three tests were left unchanged, they were simply moved and renamed.

Please *do not* do that, as it makes it difficult to review this PR. A PR should be limited and focused on the change, particularly in a critical part of the codebase. Please refocus this PR only on the change and tests to verify it, maintaining the test structure that is currently in place.

AndrYurk:
@jimmarino structure of tests was rollbacked

jimmarino (changes requested):
I think we need to step back and go through what the expected behavior is when termination does not happen cleanly. The current PR will leave the connector in an inconsistent state if the data plane (or 'DataFlowManager`) returns a failure. For example, by leaving provisioned resources in place and not updating the transfer process. 

Ideally, this would be handled by continuing the cleanup on the controller side and sending a data plane failure event to be reconciled by a listener. However, I'd prefer to discuss and design this before running into an implementation.

I'm going to make the associated issue as triage so the committers can discuss this at our next meeting and come up with a plan for how to proceed.

AndrYurk:
@jimmarino thank you, I look forward to the decision.
I have one question about quality gate, currently one of the End-To-End-Tests `pullFromHttp_httpProvision()` is failing, but I did not touch it at all. Looks like MockServer try to use Address that already in use. Could run test be initiated again manually?

Issue #4710, opened 2025-01-09
flaky testcontainers tests


ndr-brt:
# Feature Request

Looks like we have some flaky tests, caused by the way we are using testcontainers.
An analysis on the issue has been made in the tractus-x edc context:

> every test that involves testcontainers in fact could cause issues, because testcontainers itself spins up a connector with an exposed port, and this happens always after the `getFreePort` is called, here's the suspect:
> 
> ```
> 301c3c46fe23   testcontainers/ryuk:0.11.0   "/bin/ryuk"              29 seconds ago   Up 28 seconds      0.0.0.0:49158->8080/tcp, :::49158->8080/tcp                           testcontainers-ryuk-5b664308-1f43-4381-962d-ed6531f5c6ef
> ```
> 
> the port that it uses is generated randomly after the `getFreePort` is called to define the connector configuration but before the connectors are actually started, like:
> 
> * EDC test runtime defines ports to be used by choosing not used ones
> * testcontainers spins up a container that could use one of the ports already defined by the EDC test runtime
> * EDC test runtime spins up the runtimes, that could break because one of the port has been used by testcontainers

(ref. https://github.com/eclipse-tractusx/tractusx-edc/issues/1027#issuecomment-2579548558)

## Which Areas Would Be Affected?

end to end tests

## Why Is the Feature Desired?

avoid test flakyness

## Solution Proposal

* spin testcontainers before defining ports to be used (so make them not statically generated as it is now)
* force the `ryuk` port to be generated by `getFreePort` (if that's possible)

PR #4712, opened 2025-01-10
ci: fix flaky testcontainers tests

ndr-brt:
## What this PR changes/adds

Refactor EndToEnd tests to avoid flaky test executions when testcontainers is used.
Concepts:
- `EmbeddedRuntime` configurations are passed lazily through `configurationProvider`s
- The configuration values in `Participant` subclasses are generated lazily (using the provided `LazySupplier`) during the EDC runtime startup so they will take into account eventual testcontainers ports used 

## Why it does that

avoid test flakyness by design

## Further notes

- there still are two potential pityfalls in `Participant`: `managementEndpoint` and `protocolEndpoint` that are currently instantiated statically while they should follow the same pattern. It will be fixed in a subsequent PR to avoid clogging this one
- I will also provide some guidance in the documentation that will be useful also for downstream projects

## Linked Issue(s)

Part of #4710 

### DIFF
--- a/core/common/junit/src/main/java/org/eclipse/edc/junit/extensions/EmbeddedRuntime.java
+++ b/core/common/junit/src/main/java/org/eclipse/edc/junit/extensions/EmbeddedRuntime.java
@@ -27,7 +27,9 @@ import org.jetbrains.annotations.NotNull;
 
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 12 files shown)


rafaelmag110:
Could we have the getFreePort lazily supplied instead? Does that make sense? Would that be simpler? I'm trying to get my head around this.

ndr-brt:
> Could we have the `getFreePort` lazily supplying ports instead? Does that make sense? Would that be simpler? I'm trying to get my head around this.

it looks a more impactful change, as only in this repository alone the `getFreePort` it's used more than 100 times, plus, to be honest, laziness is a crazy horse to deal with, it's better to keep its scope as narrow as possible (in this case, to the `Participant` class)

wolf4ood (approved):

Issue #4711, opened 2025-01-10
Multiprotocol DSP path


wolf4ood:
# Feature Request

Currently for out going messages a connector takes the `counterPartyAddress` and append the path based on the chosen dsp protocol version which is configured in the connector. This might not be the case when communicating with a different dsp implementation which might expose protocol version under different paths. 

Since EDC now expose the management api for fetching supported DSP version from a counter party we should let the users decide the base path of the dsp version (counterPartyAddress + dsp version path) directly in input, without automatically infer that.

Meanwhile for the `callbackAddress` which currently comes from `ProtocolWebhook` should automatically infer and append ther dsp version path based on the connector supported version in the `ProtocolVersionRegistry`

To keep the current behavior we might introduce a configuration flag which should be disable by default.

## Which Areas Would Be Affected?

dsp

## Why Is the Feature Desired?

interoperability

PR #4730, opened 2025-01-16
feat: add support for multiple protocol webooks

### LINKED ISSUE #4722: Multiple protocol webhooks
# Feature Request

Currently when doing a protocol communication the `callbackAddress` is taken from the `ProtocolWebhook` injectable. This means that at runtime only one protocol is supported. In order to better support multiple protocols and versions in the same runtime a `ProtocolWebhook` should be configurable for each protocol/version

protocol, state-machine

_Are there any requirements?_

## Solution Proposal

wolf4ood:
## What this PR changes/adds

add support for multiple protocol webooks [DR](https://github.com/eclipse-edc/Connector/pull/4747)

Prerequisite for #4711 

 
## Why it does that

extensibility, multi-protocol versions

## Further notes

## Linked Issue(s)

Closes #4722 

### DIFF
--- a/core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/protocol/ProtocolWebhookRegistryImpl.java
+++ b/core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/protocol/ProtocolWebhookRegistryImpl.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2025 Metaform Systems, Inc.
+ *  Copyright (c) 2025 Cofinity-X
  *
  *  This program and the accompanying materials are made available under the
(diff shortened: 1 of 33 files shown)


core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/protocol/ProtocolWebhookRegistryImpl.java
```diff
@@ -0,0 +1,36 @@
+/*
+ *  Copyright (c) 2025 Metaform Systems, Inc.
```
paullatzelsperger, line 2: is that correct?

paullatzelsperger (approved):

PR #4735, opened 2025-01-17
refactor: hard requirement to switch to the new way to configure api authentication

### LINKED ISSUE #4729: Mandatory switch from old api authentication
# Feature Request

Currently we have fallback (deprecated) authentication set in:

`DelegatedAuthenticationExtension`
https://github.com/eclipse-edc/Connector/blob/011edf8a125b9d7346d87b73e98c0d5e69d88af8/extensions/common/auth/auth-delegated/src/main/java/org/eclipse/edc/api/auth/delegated/DelegatedAuthenticationExtension.java#L91-L103

`TokenBasedAuthenticationExtension`
https://github.com/eclipse-edc/Connector/blob/011edf8a125b9d7346d87b73e98c0d5e69d88af8/extensions/common/auth/auth-tokenbased/src/main/java/org/eclipse/edc/api/auth/token/TokenBasedAuthenticationExtension.java#L74-L81

In this one still the old one has been set, however I think we should migrate this to use the `ApiAuthenticationProviderRegistry`
`StsAccountsApiExtension`
https://github.com/eclipse-edc/Connector/blob/011edf8a125b9d7346d87b73e98c0d5e69d88af8/extensions/common/iam/identity-trust/identity-trust-sts/identity-trust-sts-accounts-api/src/main/java/org/eclipse/edc/api/iam/identitytrust/sts/accounts/StsAccountsApiExtension.java#L61-L64

We need to move on and cleanup the old stuff without potentially causing security breaches into adopters runtimes.

api authentication

cleanup

## Solution Proposal

If the "fallback" path is followed (using the old way to configure api authentication), we just throw an exception and stop the runtime with a detail and force adopters to migrate to the new way to configure authentication.

ndr-brt:
## What this PR changes/adds

Stops runtime executions if old and deprecated auth settings are used with a message that suggest how to configure authentication with the new method (see `auth-configuration` extension)

## Why it does that

cleanup

## Further notes

## Linked Issue(s)

Closes #4729 


wolf4ood (approved):
:rocket:

PR #4736, opened 2025-01-17
docs: decision record about key management

paullatzelsperger:
## What this PR changes/adds

adds a decision record outlining planned changes in key management (remote signing, automatic key rotation)

## Why it does that

improved security, centralized key management

## Further notes

## Linked Issue(s)

### DIFF
--- a/docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
+++ b/docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
@@ -64,10 +64,10 @@ To accomodate `Vault` implementations, that do not support either remote signing
 public interface Vault {
   default VaultCapabilities getVaultCapabilities() {
     return VaultCapabilities.Builder.newInstance()
-            .supportsSigning(true)
-            .supportsVerifying(true)
-            .supportsAutomaticRotation(true)
-            .supportsManualRotation(true)
+            .supportsSigning(false)
+            .supportsVerifying(false)
+            .supportsAutomaticRotation(false)
+            .supportsManualRotation(false)
             .build();
   }
 // ...
--- a/docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
+++ b/docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
@@ -36,19 +36,19 @@ VerifiablePresentations that contain many large VerifiableCredentials!_
   }
   ```
 - the default implementation performs signing and verifying exactly as implemented currently, pulling keys out of the
-  vault and performing locally.
-- for each vault technology, there would be an implementation
-- implement custom `JWSSigner` and `JWSVerifier` (from Nimbus) named `RemoteJwsSigner` and `RemoteJwsVerifier`
-  respectively, that delegate signing and verifying to the `SigningService`.
+  vault and performing the signing locally.
+- For each vault technology, there would be a `SigningService` implementation
+- Custom implementations of `JWSSigner` and `JWSVerifier` (from Nimbus) named `RemoteJwsSigner` and `RemoteJwsVerifier`
+  respectively, will delegate signing and verifying to the `SigningService`.
   These objects are provided by an extension.
-- configuration values: `edc.signing.remote.enable` to enable/disable remote signing
+- A configuration value `edc.signing.remote.enable` will be added to enable/disable remote signing
 
 ### Implementation Key Rotation
 
-- add `rotate(String keyId, Map<String, Object> keyProperties)` method to the `SigningService` to specify the algorithm
+- Add `rotate(String keyId, Map<String, Object> keyProperties)` method to the `SigningService` to specify the algorithm
   and cryptographic properties as well as the TTL (time-to-live)
 - IdentityHub: `KeyPairServiceImpl` calls `SigningService#rotate()` when rotating keys
-- configuration values:
+- Configuration values:
     - `edc.signing.keys.rotation.enable`: enable/disable automatic rotation
     - `edc.signing.keys.rotation.algorithm`: the cryptographic algorithm
     - `edc.signing.keys.rotation.ttl`: time-to-live (in days) for keys, until they get rotated automatically
--- a/docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
+++ b/docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
@@ -9,13 +9,15 @@ We will extend the capabilities of our private key management in the following a
 
 ## Rationale
 
-Production-grade enterprise deployments of the Connector and other components will require 
+Production-grade enterprise deployments of the Connector and other components will require
 centralized key management capabilities, where key pairs are automatically rotated to defend against key attrition. This
 is typically implemented by HSMs such as Hashicorp Vault.
 
 In a similar fashion EDC will provide a feature to "remotely sign" payloads. Instead of transmitting the private key
-from the vault to EDC over the network, and signing the payload locally, which carries some risk of exposure, leakage, mishandling etc.,
-EDC will transmit the payload to the vault, have the payload signed/encrypted there and transmit the signature or encrypted payload back.
+from the vault to EDC over the network, and signing the payload locally, which carries some risk of exposure, leakage,
+mishandling etc.,
+EDC will transmit the payload to the vault, have the payload signed/encrypted there and transmit the signature or
+encrypted payload back.
 
 With this, the private key never leaves the vault, thus avoiding any risk of key exposure or leakage.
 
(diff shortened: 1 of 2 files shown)


docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+> _Note that there may be a penalty on latency, especially when transmitting very large payloads, such as
+VerifiablePresentations that contain many large VerifiableCredentials!_
+
+## Approach
+
+### Implementation Remote signing
+
+- add `byte[] sign(String keyId, byte[] data)` method to `Vault`
+- add `boolean verify(String keyId, byte[] signature)` method to `Vault`
```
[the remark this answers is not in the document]
paullatzelsperger: Coupling the signing process to the Vault isn't necessarily a bad thing, because it reflects reality: most Vaults (Azure, AWS, HCV) support facilities to sign, verify and rotate, in addition to be secure Key-Value stores.

So to me that coupling, and introducing a `DigitalSignatureService` is not mutually exclusive: the `DigitalSignatureService` would simply be the "client code" that calls methods on the `Vault`. To do that, the vault would still have to expose `sign`, `verify` and `rotate` methods. 

This would mean, there could be two `DigitalSignatureServices`: 
- a default one, that pulls secrets and signs/verifies locally
- a `VaultSignatureService`, that delegates those calls to the `Vault`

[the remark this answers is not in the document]
ndr-brt: > [edit]: it would be the `DigitalSignatureService` implementation that uses `JWSSigner` and `JWSVerifier` objects from Nimbus. IF we went with separate services _per Vault_, this would introduce a dependency onto Nimbus in each Vault.

it's the other way around: the "nimbus way" to sign will just need to be implemented once in a "DigitalSignatureService" implementation (registered by default and eventually overridable), instead to need to be provided for every `Vault` implementation (because the `sign`, `verify` and `rotate` will need separate implementations for different `Vault` implementation, but, at the end, they will be the same).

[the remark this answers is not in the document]
paullatzelsperger: Nimbus is mainly relevant for JWTs, and for that, it is sufficient to provide custom JWSSigner and JWSVerifier objects that we can plug in. We already have a `TokenGenerationService`, that can be customized with a `JWSSigner`, all we need is to do the same for the `TokenValidationService`.

In cases, where non-JWT content is signed, we could use such a DigitalSignatureService`. I don't think we have such a case ATM though.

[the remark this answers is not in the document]
ndr-brt: I was more referring to changing the `Vault` interface, that doesn't look the best approach, because signature (also if only JWT) is something that "relies" on vault to sign or it can be done by a vault service, but also by a separate service.

An example is how AWS manages this: Secret Manager keeps secrets, while the signature as a service is made by a different product called KMS.

Keeping the interfaces separated will help flexibility and avoid hard coupling between vault and a signature service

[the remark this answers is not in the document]
paullatzelsperger: To sum up: 

If we keep this separate, there would have to be a `AwsKmsSignatureService`, a `AzureVaultSignatureService`, a `HashicorpVaultSignatureService`, a `GcpSecretsManagerSignatureService`, etc.

Combining this, the `JWSSigner` and `JWSVerifier` would delegate to the `SignatureService` (instead of the Vault).

The default implementation of the SignatureService would simply take a vault, pull the keys out and sign/verify locally.

do i have this correct so far?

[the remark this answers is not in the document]
ndr-brt: Yes, looks good :+1:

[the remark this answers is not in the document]
paullatzelsperger: wanna give it another look?

docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+- add new interface `SigningService` that declares two methods:
+  ```java
+  public interface SigningService {
+    byte[] sign(String keyId, byte[] data);  
+    boolean verify(String keyId, byte[] signature);
+  }
+  ```
+- the default implementation performs signing and verifying exactly as implemented currently, pulling keys out of the
+  vault and performing locally.
```
jimmarino, line 39:
```suggestion
  vault and performing the signing locally.
```

docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
```diff
@@ -0,0 +1,66 @@
...
+- for each vault technology, there would be an implementation
```
jimmarino, line 40:
```suggestion
- For each vault technology, there would be a `SigningService` implementation
```

docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
```diff
@@ -0,0 +1,66 @@
...
+- implement custom `JWSSigner` and `JWSVerifier` (from Nimbus) named `RemoteJwsSigner` and `RemoteJwsVerifier`
```
jimmarino, line 41:
```suggestion
- Custom implementations of `JWSSigner` and `JWSVerifier` (from Nimbus) named `RemoteJwsSigner` and `RemoteJwsVerifier`
```

docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
```diff
@@ -0,0 +1,66 @@
...
+  respectively, that delegate signing and verifying to the `SigningService`.
```
jimmarino, line 42:
```suggestion
  respectively, will delegate signing and verifying to the `SigningService`.
```

docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
```diff
@@ -0,0 +1,66 @@
...
+  These objects are provided by an extension.
+- configuration values: `edc.signing.remote.enable` to enable/disable remote signing
```
jimmarino, line 44:
```suggestion
- A configuration value `edc.signing.remote.enable` will be added to enable/disable remote signing
```

docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Implementation Key Rotation
+
+- add `rotate(String keyId, Map<String, Object> keyProperties)` method to the `SigningService` to specify the algorithm
```
jimmarino, line 48:
```suggestion
- Add `rotate(String keyId, Map<String, Object> keyProperties)` method to the `SigningService` to specify the algorithm
```

docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
```diff
@@ -0,0 +1,66 @@
...
+  and cryptographic properties as well as the TTL (time-to-live)
+- IdentityHub: `KeyPairServiceImpl` calls `SigningService#rotate()` when rotating keys
+- configuration values:
```
jimmarino, line 51:
```suggestion
- Configuration values:
```

jimmarino (approved):
I added wording suggestions for clarity.

wolf4ood (approved):

ndr-brt (approved):

docs/developer/decision-records/2025-01-17-key-management-improvement/README.md
```diff
@@ -0,0 +1,66 @@
...
+    boolean verify(String keyId, byte[] signature);
+  }
+  ```
+- the default implementation performs signing and verifying exactly as implemented currently, pulling keys out of the
+  vault and performing the signing locally.
+- For each vault technology, there would be a `SigningService` implementation
+- Custom implementations of `JWSSigner` and `JWSVerifier` (from Nimbus) named `RemoteJwsSigner` and `RemoteJwsVerifier`
+  respectively, will delegate signing and verifying to the `SigningService`.
+  These objects are provided by an extension.
+- A configuration value `edc.signing.remote.enable` will be added to enable/disable remote signing
```
rafaelmag110, line 44: Small nit:
Isn't this configuration redundant? If you add a remote signing service to your runtime, you'd expect remote signing to be done.
I could imagine a case where the remote signing service is added, only for the edc operator to find out later it wasn't working because it wasn't activated...

PR #4737, opened 2025-01-17
fix: add deterministic ObjectMapper for JSON-LD

### LINKED ISSUE #4734: JSON-LD ObjectMapper resolution
# Bug Report

## Describe the Bug

Currently we do fetch most of the time the `ObjectMapper`  for JSON-LD processing in the `initialize` phase of extensions

by injecting the `TypeManager` and then doing

```
var jsonLdMapper = typeManager.getMapper(JSON_LD);
```

This can lead to an issue when the `JSON_LD` object mapper is not registered yet in the `TypeManager`.

I've encounter this while working on multi protocol webhooks [PR](https://github.com/eclipse-edc/Connector/pull/4730)

where I removed the `ProtocolWebhook` injectable replacing it by a `ProtocolWebhookRegistry`. This led to a different dependency injection timeline which causes some extension to fetch the `JSON_LD` ObjectMapper before it was configured by the `JsonLdExtension`

## Possible Implementation

Move in the prepare phase or make it lazy the fetch of the required ObjectMapper

Additional but for future issues, we should make the `Registry` patter safe, ensuring that values cannot be read from a `Registry` during
the initialize phase.

wolf4ood:
## What this PR changes/adds

add deterministic ObjectMapper for JSON-LD

## Why it does that

## Further notes

## Linked Issue(s)

Closes #4734 


paullatzelsperger (approved):

wolf4ood:
On a second though this reduce the impact of the changes, but it will limit the extensibility of the `TypeManager` on the long run. I will go back to my original solution with having a lazy fetch of the ObjectMapper

wolf4ood:
I've re-target the PR to not change the usage of `TypeManager` but instead having the lazy fetch of the `ObjectMapper`

paullatzelsperger (approved):

core/common/lib/json-ld-lib/src/main/java/org/eclipse/edc/jsonld/util/JacksonJsonLd.java
```diff
@@ -36,6 +36,7 @@ public void setupModule(SetupContext context) {
             }
         };
         mapper.registerModule(module);
+        mapper.enable(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY);
```
ndr-brt, line 39: should this be part of a separate PR for the sake of clarity? It could also deserve a test

ndr-brt (approved):

core/common/lib/json-ld-lib/src/main/java/org/eclipse/edc/jsonld/util/JacksonJsonLd.java
wolf4ood: This was already present in the code base, I just move it here centrally. 
A test make sense :+1:

jimmarino (approved):

PR #4766, opened 2025-01-28
fix: make provider transfer termination async

ndr-brt:
## What this PR changes/adds

Apply transfer termination in asynchronous manner.

## Why it does that

as stated in the [related DR](https://github.com/eclipse-edc/Connector/tree/1eb74c3768d314b5e0119d74af31e7cac7c607ad/docs/developer/decision-records/2025-01-28-async-protocol-message-processing)

## Further notes

- to distinguish when a termination has been requested by one party or the other another status `TERMINATING_REQUESTED` has been added (best name that came to mind :sweat_smile: ). For this status there will be no message dispatch, because the message was already sent from the counter party. If this proposal goes through I will update the documentation accordingly.
- this PR does not focus on other async operations (like the suspension), they will be tackled by separated PRs

## Linked Issue(s)

Part of #4592 


core/control-plane/control-plane-transfer/src/main/java/org/eclipse/edc/connector/controlplane/transfer/process/TransferProcessManagerImpl.java
```diff
@@ -177,6 +180,7 @@ protected StateMachineManager.Builder configureStateMachineManager(StateMachineM
                 .processor(processConsumerTransfersInState(RESUMING, this::processConsumerResuming))
                 .processor(processTransfersInState(COMPLETING, this::processCompleting))
                 .processor(processTransfersInState(TERMINATING, this::processTerminating))
+                .processor(processTransfersInState(TERMINATING_REQUESTED, this::processTerminating))
```
ronjaquensel, line 183: If the state `TERMINATING_REQUESTED` is handled by `processTerminating`, wouldn't that mean that a `TransferTerminationMessage` is sent again after one has already been sent by the counter-party? Afaik, only one party should send this message, right?

ronjaquensel (dismissed):

core/control-plane/control-plane-transfer/src/main/java/org/eclipse/edc/connector/controlplane/transfer/process/TransferProcessManagerImpl.java
ndr-brt: yes, that's why there's a condition into the `sendTransferTerminationMessage`.
I thought to bring the condition to the upper lever, but it would had complicated the code much more, because of the nested `RetryProcess` used. I can take a look if it could refactored into a better shape

ronjaquensel: You're right, I completely missed that. While it reads a bit intransparent at first, if the alternative is refactoring it all in a more compley way, this would be fine with me.

ronjaquensel (approved):

jimmarino (approved):
Let's see if the condition mentioned in the comments can be cleaned up in a subsequent PR

PR #4784, opened 2025-02-03
fix: Add missing audience validation when accessing management api with oauth

### LINKED ISSUE #4768: API management is missing token Audience validation
Hi, 

I've seen that management API does not validate the token audience. The AudienceRule is not applied in the `DelegatedAuthenticationExtension`. I think it must be done like it is in the `Oauth2ServiceExtension` otherwise the Management API will authorize tokens emitted by the correct Authorization Server but intended for a different resources serveur protected with the authorization server

Do I miss something?

_Originally posted by @scandinave in https://github.com/eclipse-edc/Connector/discussions/4684_

scandinave:
## What this PR changes/adds

This PR add the audience Validation for the API management context

## Why it does that

Not validating audience is a security risk.

## Further notes

I have added some tests to validate the modification in the extension. But I could not find a way to change the value of an injected configuration field annotated by `@Settings`. So I can't unit test that when I set an audience, everything works as expected.

I have validated the code manually though. 

@paullatzelsperger 

## Linked Issue(s)

Closes #4768 

### DIFF
--- a/extensions/common/auth/auth-delegated/src/main/java/org/eclipse/edc/api/auth/delegated/DelegatedAuthenticationExtension.java
+++ b/extensions/common/auth/auth-delegated/src/main/java/org/eclipse/edc/api/auth/delegated/DelegatedAuthenticationExtension.java
@@ -9,6 +9,7 @@
  *
  *  Contributors:
  *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *       Dawex Systems - Add audience validation
  *
  */
 
@@ -27,6 +28,7 @@ import org.eclipse.edc.spi.result.Result;
 import org.eclipse.edc.spi.system.ServiceExtension;
 import org.eclipse.edc.spi.system.ServiceExtensionContext;
 import org.eclipse.edc.spi.system.configuration.Config;
+import org.eclipse.edc.token.rules.AudienceValidationRule;
 import org.eclipse.edc.token.rules.ExpirationIssuedAtValidationRule;
 import org.eclipse.edc.token.rules.NotBeforeValidationRule;
 import org.eclipse.edc.token.spi.TokenValidationRulesRegistry;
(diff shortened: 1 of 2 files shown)


scandinave:
Not sure about the use of the context and key value of the  `@Setting` annotation for the audience configuration. Is it ok ?

extensions/common/auth/auth-delegated/src/test/java/org/eclipse/edc/api/auth/delegated/DelegatedAuthenticationExtensionTest.java
```diff
@@ -57,4 +73,34 @@ public void delegatedProvider(DelegatedAuthenticationExtension extension) {
         verify(config).getLong(AUTH_CACHE_VALIDITY_MS, DEFAULT_CACHE_TIME_TO_LIVE);
 
     }
+
+    @Test
+    void initializeWithAudience(ServiceExtensionContext context, ObjectFactory factory) {
```
paullatzelsperger, line 78: instead of using the objectfactory, you can directly inject the `DelegatedAuthExtension` into the test method, c.f. the previous test method

extensions/common/auth/auth-delegated/src/main/java/org/eclipse/edc/api/auth/delegated/DelegatedAuthenticationExtension.java
```diff
@@ -96,6 +102,15 @@ public void initialize(ServiceExtensionContext context) {
             throw new EdcException(message);
         }
 
+        if (audience == null) {
+            context.getMonitor().warning("No audience configured for delegated authentication, defaulting to the participantId");
```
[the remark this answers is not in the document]
scandinave: maybe I can also replace the previous context.monitor by the one form line 95.

extensions/common/auth/auth-delegated/src/test/java/org/eclipse/edc/api/auth/delegated/DelegatedAuthenticationExtensionTest.java
scandinave: This is what I have done at first but if I do that the audience value is already injected with a null value before I can set it in the test.

paullatzelsperger: ah, that's right. sorry, overlooked that.

paullatzelsperger (approved):

PR #4796, opened 2025-02-05
feat: add `toBuilder()` methods

paullatzelsperger:
## What this PR changes/adds

adds `toBuilder()` methods to several VC model classes

## Why it does that

to enable easy copying, and setting values. this is needed when updating credentials, e.g. status list credentials in IH

## Further notes

## Linked Issue(s)

### DIFF
--- a/spi/common/verifiable-credentials-spi/src/main/java/org/eclipse/edc/iam/verifiablecredentials/spi/model/VerifiableCredential.java
+++ b/spi/common/verifiable-credentials-spi/src/main/java/org/eclipse/edc/iam/verifiablecredentials/spi/model/VerifiableCredential.java
@@ -119,6 +119,10 @@ public class VerifiableCredential {
         return dataModelVersion;
     }
 
+    public <T extends VerifiableCredential, B extends Builder<T, B>> Builder<T, B> toBuilder() {
+        return new Builder(this);
+    }
+
     public static class Builder<T extends VerifiableCredential, B extends Builder<T, B>> {
         protected final T instance;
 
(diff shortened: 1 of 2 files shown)


spi/common/verifiable-credentials-spi/src/main/java/org/eclipse/edc/iam/verifiablecredentials/spi/model/VerifiableCredential.java
```diff
@@ -119,6 +119,10 @@ public DataModelVersion getDataModelVersion() {
         return dataModelVersion;
     }
 
+    public <T extends VerifiableCredential, B extends Builder<T, B>> Builder<T, B> toBuilder() {
+        return new Builder(this);
```
ndr-brt, line 123: we must be aware with this approach (that differs from the current `toBuilder` implementations we have e.g. in `Policy` or `Asset` instances, that the changes applied on the builder will affect the original object instance.
Maybe we should decide on a standard approach about builders, because the different approaches used currently could cause confusion.

paullatzelsperger: excellent point.
as far as i can see, both `Asset` and `Policy` _affect_ the original instance (i.e. modify some related objects), whereas the `toBuilders` implemented here directly modify the original instance. subtle but an important difference.

to me, the implementation here is more correct, because it directly operates on the original instance. 
as an addition, we could introduce a `copy()` method that creates a carbon-copy. WDYT?

ndr-brt: maybe we could state a clear decision and change all the builders appropriately and to document it, I'm not sure if there are other builders that behave as `Asset` and `Policy`, but I guess so

ndr-brt (approved):
btw, this PR is good to go :)

spi/common/verifiable-credentials-spi/src/main/java/org/eclipse/edc/iam/verifiablecredentials/spi/model/VerifiableCredential.java
paullatzelsperger: i'll open an issue about this

PR #4814, opened 2025-02-11
feat: implement JsonArrayTranslator

### LINKED ISSUE #4813: Support JSON Arrays via `JsonFieldTranslator`
# Feature Request

In our current `JsonFieldTranslator`, we don't really support JSON Arrays. For example, imagine the following JSON structure, that is serialized as `TestObject.jsonStruct`:
```json
{
  "someKey": "someValue",
  "someNumber": 42,
  "someObject": {
    "nestedArray": ["nestedValue1", "nestedValue2"]
  }
}
```
Querying for `jsonStruct.someObject.nestedArray contains "nestedValue1"` would yield the following SQL statement:

```sql
SELECT * FROM testobjects WHERE (jsonstruct -> 'someObject' -> 'nestedArray')::jsonb ?? 'nestedValue1'
```

This would work as intended. However, when we imagine a class `TestObject` that contains a `jsonArray`, this won't work anymore:

```java
[
  "value1", "value2"
]
```

because a `Criterion` `jsonArray contains "value1"` would result in the follwoing SQL statement:

```sql
SELECT * FROM testobjects WHERE (jsonArray -> 'jsonArray')::jsonb ?? 'value1'
```

which would be incorrect.

Thus, a dedicated `JsonArrayTranslator` should be implemented, that can handle JSON arrays.

`sql-lib`

support JSON arrays in SQL queries.

## Solution Proposal

implement a JsonArrayTranslator, with the following requirements and constraints:
- can only handle the `"contains"` operator, no `"endWith"`, `"startsWith"` or similar.
- can only handle scalar right-operands, no lists. That means, checking if a list contains another list is not supported.
- the `JsonArrayTranslator` only has a no-arg constructor, no `columnName` is passed, as that would conflict with the left-operand

## Further comments

maybe we could re-use the `JsonArrayTranslator` or parts of it within the `JsonFieldTranslator`, but investigating this now seemed not worth the effort TBH.

paullatzelsperger:
## What this PR changes/adds

implements the `JsonArrayTranslator`, to enable querying top-level JSON arrays in SQL stores.

## Further notes

- this feature is an addition to the current `JsonFieldTranslator` and should only be used for top-level JSON arrays

## Linked Issue(s)

Closes #4813 

### DIFF
--- a/core/common/lib/sql-lib/src/main/java/org/eclipse/edc/sql/translation/JsonArrayTranslator.java
+++ b/core/common/lib/sql-lib/src/main/java/org/eclipse/edc/sql/translation/JsonArrayTranslator.java
@@ -1,0 +1,58 @@
+/*
+ *  Copyright (c) 2025 Cofinity-X
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Cofinity-X - initial API and implementation
+ *
+ */
+
+package org.eclipse.edc.sql.translation;
+
+import org.eclipse.edc.spi.query.Criterion;
+import org.eclipse.edc.util.reflection.PathItem;
+
+import java.util.Collection;
+import java.util.List;
+
+import static org.eclipse.edc.sql.translation.FieldTranslator.toParameters;
+import static org.eclipse.edc.sql.translation.FieldTranslator.toValuePlaceholder;
+
+/**
+ * This is a specialized translator, that targets object properties that are JSON array, for example {@code someObject.jsonArray},
+ * and a criterion of "jsonArray contains foobar".
+ * <p>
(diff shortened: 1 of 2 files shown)


core/common/lib/sql-lib/src/main/java/org/eclipse/edc/sql/translation/JsonArrayTranslator.java
```diff
@@ -0,0 +1,58 @@
...
+import static org.eclipse.edc.sql.translation.FieldTranslator.toParameters;
+import static org.eclipse.edc.sql.translation.FieldTranslator.toValuePlaceholder;
+
+/**
+ * This is a specialized translator, that targets object properties that are JSON array, for example {@code someObject.jsonArray},
+ * and a criterion of "jsonArray contains foobar".
+ * <p>
+ * This is necessary, because with a conventional {@link JsonFieldTranslator} we would get {@code (jsonArray -> 'jsonArray')::jsonb ?? ?},
+ * because it always assumes one indirection from columnName to object name.
+ * <p>
```
ndr-brt, line 32: nit
```suggestion
 * </p>
```

ndr-brt (approved):

PR #4818, opened 2025-02-12
fix: support VP without VCs in IdentityAndTrustService

### LINKED ISSUE #4807: `IdentityAndTrustService` does not support presentations with no VC
# Bug Report

## Describe the Bug

Current implementation of `IdentityAndTrustService` return verification failure if the returned VP contains no VC. There can be use-cases wherein no VC is required to request access to a resource (e.g. public dataset). 

`IdentityAndTrustService` should not block the access is there is no VC. Instead this should be done by the subsequent Policy Engine invokation, which might be configured to fail if there is no verified claims. 

### Expected Behavior

`IdentityAndTrustService` should return empty ParticipantAgent when there is no VC in the VP, instead of failing.

### Observed Behavior

## Context Information

- OS: [e.g. iOS, Windows]
- ...

## Detailed Description

## Possible Implementation

bscholtes1A:
## What this PR changes/adds

Support VPs without VCs in IdentityAndTrustService.

## Why it does that

## Further notes

## Linked Issue(s)

Closes #4807 

### DIFF
--- a/extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/VerifiableCredentialValidationServiceImpl.java
+++ b/extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/VerifiableCredentialValidationServiceImpl.java
@@ -36,6 +36,7 @@ import java.util.Collection;
 import java.util.List;
 
 import static org.eclipse.edc.spi.result.Result.failure;
+import static org.eclipse.edc.spi.result.Result.success;
 
 public class VerifiableCredentialValidationServiceImpl implements VerifiableCredentialValidationService {
     private final PresentationVerifier presentationVerifier;
@@ -60,7 +61,7 @@ public class VerifiableCredentialValidationServiceImpl implements VerifiableCred
             var presentationIssuer = verifiablePresentation.presentation().getHolder();
             return presentationVerifier.verifyPresentation(verifiablePresentation)
                     .compose(u -> validateVerifiableCredentials(credentials, presentationIssuer, additionalRules));
-        }).reduce(Result.success(), Result::merge);
+        }).reduce(success(), Result::merge);
     }
 
     @NotNull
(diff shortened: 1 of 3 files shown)


extensions/common/iam/identity-trust/identity-trust-service/src/main/java/org/eclipse/edc/iam/identitytrust/service/IdentityAndTrustService.java
```diff
@@ -133,6 +133,15 @@ public Result<ClaimToken> verifyJwtToken(TokenRepresentation tokenRepresentation
         var claimToken = claimTokenResult.getContent();
         var accessToken = claimToken.getStringClaim(PRESENTATION_TOKEN_CLAIM);
         var issuer = claimToken.getStringClaim(ISSUER);
+        var scopeResult = extractScope(accessToken);
```
paullatzelsperger, line 136: I'm afraid we can't do this. The shape of the `accessToken` is not defined in the spec, so it could be any string. It might not contain a `scope` claim, in fact, it might not even be a JWT. From the [spec](https://eclipse-dataspace-dcp.github.io/decentralized-claims-protocol/HEAD/#verifiable-presentation-access-token):

> The format of the token is implementation-specific and therefore should be treated as an opaque string by the Verifier.

extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/VerifiableCredentialValidationServiceImpl.java
```diff
@@ -75,11 +73,11 @@ private Result<Void> validateVerifiableCredentials(List<VerifiableCredential> cr
                 new HasValidSubjectSchema(mapper)));
 
         filters.addAll(additionalRules);
-        var results = credentials
+        return credentials
```
paullatzelsperger, line 76: better to simply add at the beginning of the method
```java
if(credentials.isEmpty()){
        return success();
}
```

this will be much more explicit, easier to read and avoids needless instantiation of classes

paullatzelsperger (changes requested):
We cannot introspect the `accessToken` on the Verifier side, because its shape is undefined.

paullatzelsperger (approved):

PR #4822, opened 2025-02-13
feat: make HashiCorp vault authentication extensible

### LINKED ISSUE #4751: Refactor HAshicorpHealthService
# Feature Request

The `HAshicorpHealthCheckService` currently does too many things: sending a heartbeat to HCV and refreshing tokens.

This needs to be refactored.

hashicorp vault extension

separation of concerns, cleanliness of the code base

## Solution Proposal

### LINKED ISSUE #4810: Make HashiCorp vault authentication extensible
# Feature Request

As outlined in [this DR](https://github.com/eclipse-edc/Connector/tree/main/docs/developer/decision-records/2025-02-06-hashicorp-vault-authentication-refactor), we want to refactor the `vault-hashicorp` module.

`vault-hashicorp`

To support the usage of different HCV authentication methods.

me

## Solution Proposal

as outlined in the DR

ronjaquensel:
## What this PR changes/adds

Refactors the `vault-hashicorp` module as outlined in [this DR](https://github.com/eclipse-edc/Connector/tree/main/docs/developer/decision-records/2025-02-06-hashicorp-vault-authentication-refactor).

## Why it does that

To make vault authentication extensible and thus enable using different authentication methods.

## Further notes

In addition to what's already mentioned in the DR:
- created `HashicorpVaultTokenRenewService`, which now contains the token look-up & renew methods previously located in the health service. I did not add them to the `HashicorpVaultTokenRenewTask` directly, to keep a separation between the methods for renewing the token and the periodic task triggering the token renewal.
- created `vault-hashicorp-spi`, which as of now contains only the `HashicorpVaultTokenProvider` interface.

me

## Linked Issue(s)

Closes #4810
Closes #4751


extensions/common/vault/hashicorp/vault-hashicorp-spi/build.gradle.kts
ndr-brt, line 1: I would put this in the `spi` folder, at the moment all the spis reside there (apart from some `dsp` ones, that I think they should be put there as well)

extensions/common/vault/hashicorp/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultTokenRenewService.java
```diff
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2022 Mercedes-Benz Tech Innovation GmbH
+ *  Copyright (c) 2025 Cofinity-X
```
ndr-brt, line 2: copyright

extensions/common/vault/hashicorp/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultTokenRenewService.java
```diff
@@ -174,48 +156,31 @@ public Result<Long> renewToken() {
...
                 .url(requestUri)
-                .headers(headers)
+                .headers(getHeaders(tokenProvider.vaultToken()))
                 .post(createRequestBody(requestBody))
                 .build();
     }
-
-    private Headers getHeaders() {
+    
+    private Headers getHeaders(String token) {
```
ndr-brt, line 178: given that `tokenProvider` is a class field, maybe it could be used internally in this method instead of having the token passed every time

extensions/common/vault/hashicorp/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultTokenRenewService.java
ronjaquensel: I think GitHub got confused which file was old and which was new after I changed the module structure (since all files where moved then) 😄 reverted that part now and it fits again

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/auth/HashicorpVaultAuthenticationExtension.java
```diff
@@ -14,23 +14,22 @@
...
 import org.eclipse.edc.vault.hashicorp.spi.auth.HashicorpVaultTokenProvider;
 
 @Extension(value = HashicorpVaultAuthenticationExtension.NAME)
 public class HashicorpVaultAuthenticationExtension implements ServiceExtension {
     
     public static final String NAME = "Hashicorp Vault Authentication";
     
-    @Configuration
-    private HashicorpVaultSettings config;
+    @Setting(description = "The token used to access the Hashicorp Vault", key = "edc.vault.hashicorp.token")
```
ndr-brt, line 28: detail, I think this way the setting by default it is required, that's ok in a case where we are using the default token provider, but in case there's another extension registered that overrides it, the setting should not be required.
I would put `required = false` on this and then do a check in the `tokenProvider()` method (that will be called only if no other implementation has been registered)

ronjaquensel: good point 👍🏻

extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/auth/HashicorpVaultAuthenticationExtension.java
```diff
@@ -20,16 +20,19 @@
...
     
     public static final String NAME = "Hashicorp Vault Authentication";
     
-    @Setting(description = "The token used to access the Hashicorp Vault", key = "edc.vault.hashicorp.token")
+    @Setting(description = "The token used to access the Hashicorp Vault. Only required, if default token authentication is used.", key = "edc.vault.hashicorp.token", required = false)
     private String token;
     
     @Provider(isDefault = true)
     public HashicorpVaultTokenProvider tokenProvider() {
+        requireNonNull(token, "Using default TokenProvider: Vault token is required.");
```
ndr-brt, line 35: reporting also the setting key could help out operator to fix the error in an easier way

ndr-brt (approved):

PR #4833, opened 2025-02-21
refactor: incorrect paramaters order in dataplane logs regarding DataFlow Id

### LINKED ISSUE #4804: fix: Incorrect parameters' order in dataplane logs regarding DataFlow Id
# Bug Report

## Describe the Bug

[In this PR](https://github.com/eclipse-edc/Connector/pull/4743) adding the DataFlow ID to the DataPlane logs, the order of the parameters added is not correct and needs to be updated.

### Expected Behavior

The log is should be printed as follows.

`Transferring from AmazonS3 to AmazonS3 for flow id: c9a1d5e9-b4a7-4a31-a508-c27a7ca61419.`

### Observed Behavior

The log is disconnected
`Transferring from c9a1d5e9-b4a7-4a31-a508-c27a7ca61419 to AmazonS3 for flow id: AmazonS3.`

> AwsS3 used as an example here, but issue applies to any type of TP.

Same is seen for `Unknown data source` and `Unknown data sink type` logs.

## Steps to Reproduce

Perform any TP and can be confirmed in the logs

## Possible Implementation

This is a very small change, only needs to update the (three) logs on [this](https://github.com/eclipse-edc/Connector/blob/main/core/data-plane/data-plane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/pipeline/PipelineServiceImpl.java) class.

bmg13:
## What this PR changes/adds

Fixes the order of the parameters in three logs in the DataPlane to ensure consistent reading of DataSource, DataDestination and DataFlowId fields in a Transfer Process.

## Why it does that

To keep logic order in log reading.

## Linked Issue(s)

Closes #4804

### DIFF
--- a/core/data-plane/data-plane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/pipeline/PipelineServiceImpl.java
+++ b/core/data-plane/data-plane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/pipeline/PipelineServiceImpl.java
@@ -108,7 +108,7 @@ public class PipelineServiceImpl implements PipelineService {
         var source = sourceFactory.createSource(request);
         sources.put(request.getProcessId(), source);
         monitor.debug(() -> format("Transferring from %s to %s for flow id: %s.",
-                request.getProcessId(), request.getSourceDataAddress().getType(), request.getDestinationDataAddress().getType()));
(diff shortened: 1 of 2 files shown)


jimmarino:
Tests?

bmg13:
> Tests?

Since this was a simple log order change, assumed no need of tests, but would avoid happening again, will add, thank you!

jimmarino (approved):

PR #4842, opened 2025-02-25
Fixed Action building logic in JsonObjectFromPolicyTransformer

### LINKED ISSUE #4838: ODRL policy output error
# Bug Report

## Describe the Bug

When outputting negotiated `Agreement`s (and policies in general), wrapped ODRL `action` entities (children of ODRL `duty` entities contained in ODRL `permission` entities) are wrapped twice, which is an error and not ODRL spec compliant.

### Expected Behavior

Valid ODRL serializations (JSON-LD) serialize `action` entities like this:

Image

### Observed Behavior

This is what the EDC serialization produces:

Image

## Steps to Reproduce

1. Negotiate some policy with aforementioned elements.
2. Call `curl -s -H "content-type: application/json" "http://localhost:29193/management/v3/contractagreements/$AGREEMENT_ID" | jq` or similar.

## Possible Implementation

The error is caused by an implementation flaw in `org/eclipse/edc/connector/controlplane/transform/odrl/from/JsonObjectFromPolicyTransformer.java`.
My PR with the fix is going to follow soon.

milux:
## What this PR changes/adds

Fixes #4838.

## Why it does that

The `action` is wrapped into another `action` under certain circumstances, which, according to our understanding of the ODRL IM, seems to be an error.

## Linked Issue(s)

Closes #4838

### DIFF
--- a/core/control-plane/control-plane-transform/src/main/java/org/eclipse/edc/connector/controlplane/transform/odrl/from/JsonObjectFromPolicyTransformer.java
+++ b/core/control-plane/control-plane-transform/src/main/java/org/eclipse/edc/connector/controlplane/transform/odrl/from/JsonObjectFromPolicyTransformer.java
@@ -264,16 +264,14 @@ public class JsonObjectFromPolicyTransformer extends AbstractJsonLdTransformer<P
             if (action == null) {
                 return actionBuilder.build();
             }
+            actionBuilder.add(ID, action.getType());
             if (action.getIncludedIn() != null || action.getConstraint() != null) {
-                actionBuilder.add(ODRL_ACTION_ATTRIBUTE, jsonFactory.createObjectBuilder().add(ID, action.getType()));
                 if (action.getIncludedIn() != null) {
                     actionBuilder.add(ODRL_INCLUDED_IN_ATTRIBUTE, action.getIncludedIn());
                 }
                 if (action.getConstraint() != null) {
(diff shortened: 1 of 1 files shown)


jimmarino:
Please see my comments in the issue on how this should be fixed (i.e. it must correspond to the DSP schema),

paullatzelsperger:
@milux please also respect our process. issues must be triaged before working on them, and you *will need a sponsor*.
in addition, there are no tests on this, which is a definite death sentence for any PR.

milux:
> @milux please also respect our process. issues must be triaged before working on them, and you _will need a sponsor_. in addition, there are no tests on this, which is a definite death sentence for any PR.
> 
> for reference: https://eclipse-edc.github.io/documentation/for-contributors/guidelines/pr-etiquette/

@paullatzelsperger I would have been happy to adapting any tests covering that code. Since I essentially only removed 2 lines, I can hardly see how this would negatively affect coverage, though.

Since the corresponding issue was already closed, that doesn't matter anymore, though.

paullatzelsperger:
> I would have been happy to adapting any tests covering that code. Since I essentially only removed 2 lines, I can hardly see how this would negatively affect coverage, though.

It may be a moot point, but still: a change like this _always_ warrants a test. If you can't see that, then that is a problem.

paullatzelsperger:
Writing tests is simply good practice and not up for debate. While our code base might be lacking in that regard on occasion, it cannot serve as reasoning to Not to strive to do better. Quite the opposite.

PR #4908, opened 2025-04-09
feat: allow multiple validators per type

### LINKED ISSUE #4892: Custom validators replace built-in validators
# Bug Report

## Describe the Bug

When registering custom validators as outlined in the [documentation](https://eclipse-edc.github.io/documentation/for-adopters/extensions/#validation), the custom validator actually replaces the built-in one.

### Expected Behavior

The custom validator is invoked _in addition to_ the built-in one.

### Observed Behavior

depending on the extension ordering, the custom validator either gets overwritten by the built-in one, or it overwrites the built-in one. 

Either way, the behaviour is non-deterministic and cannot be controlled.

## Detailed Description

In the particular case of the `JsonObjectValidatorImpl`, registering new validators replaces existing ones:

```java
@Override
public void register(String type, Validator<JsonObject> validator) {
    this.validators.put(type, validator);
}
```
Obviously, registering two validators for the same `type` (e.g. `EDC_ASSET_TYPE`) creates a conflict.

Therefor it depends on whether the built-in validator (e.g. in `AssetApiExtension`) or a custom validator is registered first. More specifically, assuming registration happens in the extension's `initialize()` method, it depends on extension initialization order.

## Possible Implementation

- either change the `validators` field of the `JsonObjectValidatorRegistryImpl` class to a `Map<String, Collection<Validator<JsonObject>>`, and change the registration semantics accordingly
- make the behaviour configurable, e.g. via a boolean flag, to specify whether other validators should be replaced or added to. in that case there may even be use for an ordering, similar to the `@Priority` annotation of `ContainerRequestFilter`s

paullatzelsperger:
## What this PR changes/adds

this PR changes the `JsonObjectValidatorRegistryImpl` such that it now contains multiple validators per type and it merges all the results of the individual validations.

## Why it does that

to easily add custom validators to built-in ones

## Further notes

## Linked Issue(s)

Closes #4892 

### DIFF
--- a/core/common/runtime-core/src/main/java/org/eclipse/edc/runtime/core/validator/JsonObjectValidatorRegistryImpl.java
+++ b/core/common/runtime-core/src/main/java/org/eclipse/edc/runtime/core/validator/JsonObjectValidatorRegistryImpl.java
@@ -20,20 +20,24 @@ import org.eclipse.edc.validator.spi.ValidationResult;
 import org.eclipse.edc.validator.spi.Validator;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 public class JsonObjectValidatorRegistryImpl implements JsonObjectValidatorRegistry {
 
-    private final Map<String, Validator<JsonObject>> validators = new HashMap<>();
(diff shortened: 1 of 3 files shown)


ndr-brt (approved):
nice and tidy :ok_hand:

PR #4938, opened 2025-04-22
chore: remove authentication from DSP versions endpoint

### LINKED ISSUE #4918: DSP version API should not require an authentication
# Bug Report

Currently the implementation for the dsp versions expect to receive a token and validate that with the identity service. That endpoint should be public available for discovery

[Ref](https://eclipse-dataspace-protocol-base.github.io/DataspaceProtocol/2025-1-RC1/#exposure-of-dataspace-protocol-versions)  

_A clear and concise description of the bug._

### Expected Behavior

_A clear and concise description of what you expected to happen._

### Observed Behavior

## Context Information

- OS: [e.g. iOS, Windows]
- ...

## Detailed Description

## Possible Implementation

ronjaquensel:
## What this PR changes/adds

Removes the authentication check from the DSP versions endpoint.

Sets a "dummy" token/string for the `GetDspRequest` in the `DspVersionApiController`, as the `DspRequestHandler` returns `unauthorized` if the token is null. Alternatively, the null check in `DspRequestHandler` could be extended to only return unauthorized if the token is null and the request's result type is **not** `ProtocolVersions`, but the former seemed like the cleaner solution. Let me know what you think.

## Why it does that

In `2025/1`, the version endpoint must be unauthenticated.

## Further notes

I did not make any changes to the dispatchers, i.e. the authorization header will be set also on outgoing `ProtocolVersionRequestMessages`.  I think this makes sense for now, as older versions would still have the versions endpoint authenticated. Could be refactored at a later point to not set the authorization header for version requests.

me

## Linked Issue(s)

Closes #4918


core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/protocol/VersionProtocolServiceImpl.java
```diff
@@ -9,32 +9,27 @@
...
 
-    public VersionProtocolServiceImpl(ProtocolVersionRegistry registry, ProtocolTokenValidator tokenValidator) {
+    public VersionProtocolServiceImpl(ProtocolVersionRegistry registry) {
         this.registry = registry;
-        this.tokenValidator = tokenValidator;
     }
 
     @Override
-    public ServiceResult<ProtocolVersions> getAll(TokenRepresentation tokenRepresentation) {
-        return tokenValidator.verify(tokenRepresentation, RequestVersionPolicyContext::new)
```
ndr-brt, line 37: is this `RequestVersionPolicyContext` not used anymore right? :axe:

ndr-brt (approved):

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/protocol/VersionProtocolServiceImpl.java
ronjaquensel: good point! other than registration of scopes/policy contexts doesn't seem to be used anywhere else. Since thematically it fits to this PR, I'll do so right away 🪓

PR #4939, opened 2025-04-22
fix: add null destinationAddress check in DataPlaneInstance

### LINKED ISSUE #4933: NullPointerException thrown when DataPlaneSignalingFlowController.provision called with with null destination
# Bug Report

## Describe the Bug

A NullPointerException is thrown by `DataPlaneSignalingFlowController` in the provisioning phase if the `dataDestination` is null in the `TransferProcess`.

### Expected Behavior

Null `dataDestination` are valid cases, for pull data transfers and thus they should not cause a NPE.

### Observed Behavior

A NPE is thrown.

## Context Information

- OS: [e.g. iOS, Windows]
- ...

## Detailed Description

## Possible Implementation

bscholtes1A:
## What this PR changes/adds

Add null check on `destination` address in `DataPlaneInstance`.

## Why it does that

Prevent NPE to be thrown in case `destination` is null which is a valid use-case for pull transfers. 

## Linked Issue(s)

Closes #4933 


spi/data-plane-selector/data-plane-selector-spi/src/test/java/org/eclipse/edc/connector/dataplane/selector/spi/instance/DataPlaneInstanceTest.java
```diff
@@ -86,6 +87,22 @@ void verifyCanHandle_withTransferType() throws MalformedURLException {
...
+        var provisionType = "provisionType";
+        var inst = DataPlaneInstance.Builder.newInstance()
+                .id("test-id")
+                .url(new URL("http://localhost:8234/some/path"))
+                .allowedSourceType("srcType")
+                .allowedTransferType("destType")
+                .destinationProvisionTypes(Set.of("provisionType"))
+                .build();
+
+        assertThat(inst.canProvisionDestination(null)).isTrue();
```
ndr-brt, line 101: if no destination is provided then there's no need to provision, this should return false.

ndr-brt (approved):

PR #4940, opened 2025-04-22
docs: contract agreement policy scope DR

### LINKED ISSUE #4919: Additional `contract.agreement` policy scope
# Feature Request

Currently the contract agreement always happens automatically, the only way obtain "manual approved" flow would be through the implementation of `PendingGuard`s in the negotiation state machine.
This looks not the best option as the policy engine is made for exactly this kind of things.

The proposal is to add a new `contract.agreement` scope with a relative engine evaluation right before the automatic agreement. If this policy doesn't evaluate, the negotiation will be put in a `pending` state, waiting for manual approval (or rejection).

A written proposal in the form of a DR will come up.

_Are there any requirements?_

## Solution Proposal

ndr-brt:
## What this PR changes/adds

Provides DR for introducing `contract.agreement` policy scope.

## Why it does that

## Further notes

## Linked Issue(s)

Part of #4919

### DIFF
--- a/docs/developer/decision-records/2025-04-25-agreement-policy-scope/README.md
+++ b/docs/developer/decision-records/2025-04-25-agreement-policy-scope/README.md
@@ -1,0 +1,73 @@
+# Contract Agreement policy scope
+
+## Decision
+
+We will add a `contract.agreement` policy scope, that will be evaluated by the provider before the automatic agreement.
+
+## Rationale
+
+The negotiation flow is at the moment an automatic one: a negotiation can happen if the `contract.negotiation` policy 
+scope evaluation succeeds, or it gets terminated if the evaluation fails.
+
+To implement "manual negotiation approval", the `PendingGuard` service can be used, but this doesn't permit to
+
+## Approach
+
+There will be a new policy scope/context to permit the new `PolicyEngine` evaluation, the class will look like:
+
+```java
+public class ContractAgreementPolicyContext extends PolicyContextImpl {
+
+    @PolicyScope
+    public static final String AGREEMENT_SCOPE = "contract.agreement";
+
+    @Override
+    public String scope() {
+        return AGREEMENT_SCOPE;
+    }
+}
+```
+
+An additional evaluation of the `PolicyEngine` will be added in the `ProviderContractNegotiationManager` in the
+`processRequested` method, so that every negotiation that gets `REQUESTED` can be pass through policy functions:
+```java
+    @WithSpan
+    private boolean processRequested(ContractNegotiation negotiation) {
+        var evaluation = policyEngine.evaluate(negotiation.getLastContractOffer().getPolicy(), new ContractAgreementPolicyContext());
+        if (evaluation.succeeded()) {
+            transitionToAgreeing(negotiation);
+        } else {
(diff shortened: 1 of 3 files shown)


docs/developer/decision-records/2025-04-25-agreement-policy-scope/README.md
```diff
@@ -0,0 +1,77 @@
...
+## Decision
+
+We will add a `contract.agreement` policy scope, that will be evaluated by the provider before the automatic agreement.
+
+## Rationale
+
+The negotiation flow is at the moment an automatic one: a negotiation can happen if the `contract.negotiation` policy 
+scope evaluation succeeds, or it gets terminated if the evaluation fails.
+
+To implement "manual negotiation approval", the `PendingGuard` service can be used, but this doesn't permit to
```
ronjaquensel, line 12: looks like the last part of the sentence is missing 😄

docs/developer/decision-records/2025-04-25-agreement-policy-scope/README.md
```diff
@@ -0,0 +1,77 @@
...
+{
+  "@context": "http://www.w3.org/ns/odrl.jsonld",
+  "@type": "Offer",
+  "permission": [{
+    "action": "use",
+    "duty": [{
+      "action": "reviewPolicy"
+    }]
+  }]
+}
```
ronjaquensel, line 69: I like the idea of having an out-of-the-box available policy function for this functionality, but when putting this as a duty into the policy, wouldn't that technically (in ODRL terms) mean that the consumer has to review the policy (since when initiating a negotiation, the provider is set as the `assigner` and the consumer is set as the `assignee`)?

docs/developer/decision-records/2025-04-25-agreement-policy-scope/README.md
ndr-brt: whops :sweat_smile:

docs/developer/decision-records/2025-04-25-agreement-policy-scope/README.md
ndr-brt: Uhm... I'm definitely no expert on this, but your comment makes total sense.
what could be an alternative?

ronjaquensel: maybe basing the decision whether to set the negotiation to `pending` on the presence of specific attributes on the asset or policy? ID (or private properties directly) of asset or policy could be passed in the policy context for this scope. The provider would then need a way to configure for which properties the negotiation should be set to pending, e.g. via a configuration or an API to have it more dynamic. The policy function would check whether any of the configured properties is present on the policy/asset and return false if it is. The provider could then simply flag assets/policies accordingly.

ndr-brt: tbh the odrl model is too vague and looks lke the same result can be achieved with different rule actions, so I just removed this part, no default policy function will be there, it will be up to adopters to implement their own

ronjaquensel (approved):

jimmarino (approved):

Issue #4793, opened 2025-02-05
Move provisioning phase in the data-plane


ndr-brt:
# Feature Request

Currently the provisioning phase is handled by the control-plane, but it would be better to have the data-plane to manage it.
DR will come with further explaination

## Which Areas Would Be Affected?

_e.g., DPF, CI, build, transfer, etc._

## Why Is the Feature Desired?

_Are there any requirements?_

## Who will sponsor this feature?

_Please @-mention the committer that will sponsor your feature_.

## Solution Proposal

_If possible, provide a (brief!) solution proposal._

PR #4981, opened 2025-05-09
refactor(provisioning): manage single resource provision results

ndr-brt:
## What this PR changes/adds

Manage failures in provision/deprovisioning in data plane

## Why it does that

moving provision process in data plane

## Further notes

- renamed `ProvisionResourceDefinition` in `ProvisionResource` and transforming it in a stateful entity

## Linked Issue(s)

Part of #4793 


ndr-brt:
> @ndr-brt Let's discuss this a bit. Is the change to a stateful entity intended to have each provisioned entity transitioned independently via a Provisioner and then having the data flow state machine monitor the states to transition? Or are you intending to create individual state machines for the provisioning using a fan-out-->fan-in where the data flow state machine somehow joins for the provisioning state machines? The latter approach would introduce a lot of complexities that we don't need.

The provisioner would just apply the provisioning, but the actual state change is made by the `DataPlaneManager`, because `ProvisionerResource` is a child entity of the `DataFlow` aggregate root, no needs nor plans to have a dedicated state machine, that would be definitely unnecessary complexity.

> Maybe a DR and discussion are needed to go over some of this? Also, we should make sure we catch and change all method names if *Definition is dropped.

I think this is a design decision that doesn't diverge from the root DR about moving provisioning in the data plane. 
The "definition" term makes sense in the current design because it never gets fully correlated with the provisioned and the deprovisioned, but superficially in the cryptic `ResponseHandler` implementations. 
While the provision resource is not just a definition, but it is an entity that gets created, provisioned and then deprovisioned.

To summarize:
**Before**

Container Entity (`TransferProcess`):
- `ResourceManifest`
  - `List<ResourceDefinition>`
- `ProvisionedResourceSet`
  - `List<ProvisionedResource>`
- `List<DeprovisionedResource>`

anemic domain, manual and not straightforward correlation, poor encapsulation

**After**
Container Entity (`DataFlow`):
- `List<ProvisionResource>`
  - every resource contains identifier and other attributes like `DataAddress` and free-form properties plus:
    - `ProvisionedResource`
    - `DeprovisionedResource`

more descriptive domain, straightforward correlation, better encapsulation (`ProvisionResource` "knows")

paullatzelsperger (approved):
lgtm

ronjaquensel (approved):

PR #4992, opened 2025-05-16
fix: use numeric JWT claims instead of ISO-8601

### LINKED ISSUE #4991: JWTs use wrong data type for temporal claims
# Bug Report

## Describe the Bug

Temporal claims in JSON Web Tokens are expected to be numeric date values (="epoch seconds"), but are in fact ISO-8601 strings.

### Expected Behavior

Temporal claims (`nbf`, `exp`, `iat`) should be numeric values.

### Observed Behavior

In many places they are String values

## Detailed Description

[RFC 7519 Section 4.1.4](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4) defines `exp`, `nbf` and `iat` to be opțional claims that must contain a `NumericDate` value, which is the number of seconds since the Unix Epoch.

In many places, e.g. in the `IdentityAndTrustService`, we set those claims to be of type string: 
https://github.com/eclipse-edc/Connector/blob/1c3a71136c71c3823da08a70b6c798e71bdcba8d/extensions/common/iam/identity-trust/identity-trust-service/src/main/java/org/eclipse/edc/iam/identitytrust/service/IdentityAndTrustService.java#L138-L142

## Possible Implementation

use the correct representation, i.e.: 
```java
//bad:
claims.put("exp", Instant.now().toString());

//correct: 
claims.put("exp", Instant.now().getEpochSecond());
```

## Further notes

this will have a rippling effect through the code base, because many method arguments must be changed from `Map<String, String>` to `Map<String, Object>` and it will affect downstream repositories as well. 

**Also, this is a breaking change!**

paullatzelsperger:
## What this PR changes/adds

this PR changes the use of ISO-8601 strings as temporal JWT claims to be numeric date values.

## Why it does that

RFC 7519 requires them to be `NumericDate` values, i.e. Epoch seconds

## Further notes

The culprit is the Nimbus library, which fails to parse JSON Web tokens where the `nbf`, `exp` and `iat` claims are not numeric.

## Linked Issue(s)

Closes #4991 

### DIFF
--- a/core/common/token-core/src/test/java/org/eclipse/edc/jwt/rules/ExpirationIssuedAtValidationRuleTest.java
+++ b/core/common/token-core/src/test/java/org/eclipse/edc/jwt/rules/ExpirationIssuedAtValidationRuleTest.java
@@ -19,7 +19,6 @@ import org.eclipse.edc.token.rules.ExpirationIssuedAtValidationRule;
 import org.eclipse.edc.token.spi.TokenValidationRule;
 import org.junit.jupiter.api.Test;
 
-import java.sql.Date;
 import java.time.Clock;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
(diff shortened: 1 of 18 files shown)


ndr-brt (approved):
All my tests have been positive, let's get this merged

PR #5007, opened 2025-05-22
docs: participant identity for dsp version

ndr-brt:
## What this PR changes/adds

Provides DR for refactoring to support different identifiers for different DSP versions.

## Why it does that

_Briefly state why the change was necessary._

## Further notes

_List other areas of code that have changed but are not necessarily linked to the main feature. This could be method
signature changes, package declarations, bugs that were encountered and were fixed inline, etc._

## Who will sponsor this feature?

_Please @-mention the committer that will sponsor your feature_.

## Linked Issue(s)

Part of #5006

_Please be sure to take a look at the [contributing guidelines](https://github.com/eclipse-edc/.github/blob/main/CONTRIBUTING.md#submit-a-pull-request) and our [etiquette for pull requests](https://github.com/eclipse-edc/.github/blob/main/contributing/pr_etiquette.md)._

PR #5016, opened 2025-05-26
refactor: pull protocol token validation up in the create request handler

ndr-brt:
## What this PR changes/adds

Prototypes preparatory refactoring to reach the concepts explained here: https://github.com/eclipse-edc/Connector/pull/5007

The idea is to pull the token validation out of the services to the request handler. 
In this PR only the "create" endpoints have been involved ("get" and "update" would eventually come later).

Given that to validate the input token there's the need for a policy evaluation, I added another step in the request handler that permits to create a `ProtocolRequestContext` object, that necessarily offers a `Policy` and a `PolicyContext` provider through interface methods, plus eventual additional data needed for different entities (e.g. `ValidatableContractOffer` for `ContractNegotiation` and `ContractAgreement` for `TransferProcess`.).

With that context the token can be validated, and the service call will get the `ParticipantAgent` directly, already validated.
The capabilities to create these contexts have been left in the service layer, because that's where the knowledge resides.

This is a simple explanatory diagram that shows how the logic in the `DspRequestHandler` will become  
```mermaid
graph TD;
  A[Protocol endpoint call]
  B[Message deserialization]
  C[Context provision]
  D[Token validation]
  E[Service call]
  A-->B;
  B-->C;
  C-->D;
  D-->E;
```

This PR is still a draft to validate the idea. Still left some documentation TODO that will be eventually be resolved if the proposal will be approved.

## Why it does that

## Further notes

## Linked Issue(s)

Part of #5007 

### DIFF
--- a/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/catalog/CatalogProtocolServiceImpl.java
+++ b/core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/catalog/CatalogProtocolServiceImpl.java
@@ -83,6 +83,7 @@ public class CatalogProtocolServiceImpl implements CatalogProtocolService {
     }
 
     @Override
+    @NotNull
     public ServiceResult<RequestCatalogContext> provideRequestContext(CatalogRequestMessage message) {
         return ServiceResult.success(new RequestCatalogContext());
     }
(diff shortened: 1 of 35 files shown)


ndr-brt:
> > To "run the token through the IdentityService" we need scopes, and scopes are extracted by the policy engine, so the two operations cannot be separated
> 
> yes, in the current implementation that is true. At present, we use the scopes for two things:
> 
> 1. to send them in the PresentationQueryMessage
> 2. to check that every scope string has a corresponding credential in the PresentationResponseMessage.
> 
> At this time, "Running through the IdentityService" means performing authentication, i.e. establishing the token sender's identity, and the first half of authorization, i.e. obtaining the token sender's claims (through DCP presentation flow). The second half of authorization, i.e. comparing the policy rule against the claims, happens elsewhere.
> 
> For authentication we don't need context or scopes - we simply check that the token was signed by the sender and that a few basic claims are correct etc. But for authorization we _do_ need context, b/c we need to determine which operation to authorize and under what circumstances (i.e. policy). So for that, we need the scopes in order to make the Presentation Query.
> 
> The point I was trying to make is if we can separate those two operations out, we could have a generalized authentication functionality by moving authentication code to the DSP controller, and leave authorization in the service, after context is established.
> 
> Whether we add one method to the `IdentityService`, or we split up the service into two would have to be investigated in a spike.
> 
> This would obviously be a more impactful (and breaking) change to the `IdentityService` with the added value being:
> 
> * a generalized and re-usable `authn` code piece
> * fail fast: not having to engage the policy engine if `authn` fails -> saves some compute cycles and time.
> * more clearly delineated concerns

I think this refactor makes sense and it's not that impactful: we'd just need to add `authenticate` and `authorize` methods in the `IdentityService` interface and split the `verifyJwtToken` in 2. We can keep then a deprecated default implementation for it that calls `authenticate` and authorize` one before the other to avoid breaking changes.
I already did a little POC, will add it on my list.

ronjaquensel (approved):

PR #5018, opened 2025-05-29
fix: fixes last two CN dsp-tck tests

### LINKED ISSUE #4978: Tck contract negotiation negative test scenarios
# Bug Report

## Describe the Bug

Currently on the integration of the dsp tck we allows failure for those tests:

- [ ] `CN:03-01`
- [x] `CN:03-02`
- [x] `CN:03-03`
- [x] `CN:03-04`
- [ ] `CN_C:01-02`
- [x] `CN_C:03-01`
- [x] `CN_C:03-02`
- [x] `CN_C:03-03`
- [x] `CN_C:03-04`
- [x] `CN_C:03-05`
- [x] `CN_C:03-06`

Those tests invalid state transition/messages. Some tests fails because we throw an `IllegalArgumentException` which results in an HTTP Status 500 instead of a `400` as expected. Other tests fails because the tck expects a failure while the edc returns 200, since the EDC
do not fails when receiving messages on terminal state (finilized or terminated ), but simply skip all messages if the current state is a terminal one

### Expected Behavior

dsp compliance with `CN:03-*` and `CN_C:03-*`  scenarios

### Observed Behavior

## Context Information

- OS: [e.g. iOS, Windows]
- ...

## Detailed Description

## Possible Implementation

wolf4ood:
## What this PR changes/adds

Fixes the remaining two failing tests:

- `CN:03-01`
- `CN_C:01-02`

For `CN:03-01` additional checks on the protocol layer where added in order to return `conflict` when  the message cannot be processed. Previously messages were ignored when on final states which means that the response of the
connector was 200, which was not what `CN:03-01` test was expecting since the tests try to deliver a terminated message while the state is finalized.

For `CN_C:01-02` the fix was if when processing the requesting state in the consumer side there are more than 1 contract offer, then it's considered a counter offer from the consumer and it uses a process bound request endpoint.

## Why it does that

dsp compliance

## Further notes

## Linked Issue(s)

Closes #4978 

### DIFF
--- a/core/control-plane/control-plane-contract/src/main/java/org/eclipse/edc/connector/controlplane/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
+++ b/core/control-plane/control-plane-contract/src/main/java/org/eclipse/edc/connector/controlplane/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
@@ -116,10 +116,15 @@ public class ConsumerContractNegotiationManagerImpl extends AbstractContractNego
         var callbackAddress = protocolWebhookRegistry.resolve(negotiation.getProtocol());
 
         if (callbackAddress != null) {
+            var type = ContractRequestMessage.Type.INITIAL;
+            // if there are multiple offers, it's a counter-offer
+            if (negotiation.getContractOffers().size() > 1) {
+                type = ContractRequestMessage.Type.COUNTER_OFFER;
+            }
             var messageBuilder = ContractRequestMessage.Builder.newInstance()
                     .contractOffer(negotiation.getLastContractOffer())
                     .callbackAddress(callbackAddress.url())
-                    .type(ContractRequestMessage.Type.INITIAL);
+                    .type(type);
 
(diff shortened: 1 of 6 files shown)


core/control-plane/control-plane-contract/src/main/java/org/eclipse/edc/connector/controlplane/contract/negotiation/ConsumerContractNegotiationManagerImpl.java
```diff
@@ -116,10 +116,15 @@ private boolean processRequesting(ContractNegotiation negotiation) {
         var callbackAddress = protocolWebhookRegistry.resolve(negotiation.getProtocol());
 
         if (callbackAddress != null) {
+            var type = ContractRequestMessage.Type.INITIAL;
+            // if there are multiple offers, it's a counter-offer
```
ndr-brt, line 120: I think the code is already self-explaining, no need for the comment

ndr-brt (approved):

PR #5020, opened 2025-05-30
fix: allowing sorting Contract Definitions by creation date

### LINKED ISSUE #4921: Fix allowing sorting Contract Definitions by creation date
# Bug Report

## Describe the Bug

Enable retrieving Contract Definitions sorted by their creation date.

### Expected Behavior

We should be able to sort Contract Definitions by their creation date, similar to what is seen with other entities (like assets and policies).

### Observed Behavior

Currently is not possible to sort Contract Definitions based on their creation time.

## Steps to Reproduce

Perform the GET Contract Definitions request

```/v3/contractdefinitions/request```

With a sorting order defined in the body, like

```
{
 "@context": {},
 "@type": "QuerySpecDto",
 "https://w3id.org/edc/v0.0.1/ns/offset": 0,
 "https://w3id.org/edc/v0.0.1/ns/limit": 50,
 "sortField": "createdAt",
 "sortOrder": "DESC"
}
```

It will return a 5XX.

## Context Information

Keep in mind that workaround to use the createdAt value from response is not possible at the meantime since this value is not in the response, as explained [here](https://github.com/eclipse-edc/Connector/issues/4895).

## Possible Implementation

add the creation date mapping in [here](https://github.com/eclipse-edc/Connector/blob/c0e67c6bc46ea41cbf2ac1971304a155af12d2f4/extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/controlplane/store/sql/contractdefinition/schema/postgres/ContractDefinitionMapping.java#L27) like so

```
add("createdAt", statements.getCreatedAtColumn());
```

similar to what is seen in other Entities, like [Assets](https://github.com/eclipse-edc/Connector/blob/c0e67c6bc46ea41cbf2ac1971304a155af12d2f4/extensions/control-plane/store/sql/asset-index-sql/src/main/java/org/eclipse/edc/connector/controlplane/store/sql/assetindex/schema/postgres/AssetMapping.java#L37).

## Further notes

This issue was created as a bug since this may not be intended behavior. If is intended, we could propose changing the 5XX http status response to a 4XX.

## Detailed Description

The following are additional elements to show that storing in the db is working as expected and may help explain the proposed fix as well.

Image

Image

### PR DESCRIPTION
**Pull request by bmg13**
## What this PR changes/adds

Include fix to enable retrieving Contract Definitions sorted by their creation date. Simply updated the mapping for this entity.

## Why it does that

We should be able to sort Contract Definitions by their creation date, similar to what is seen with other entities (like assets and policies).

## Linked Issue(s)

Closes #4921

### CODE REVIEW

File: extensions/control-plane/store/sql/contract-definition-store-sql/src/main/java/org/eclipse/edc/connector/controlplane/store/sql/contractdefinition/schema/postgres/ContractDefinitionMapping.java [hunk 1]
```diff
@@ -26,6 +26,7 @@
 public class ContractDefinitionMapping extends TranslationMapping {
     public ContractDefinitionMapping(ContractDefinitionStatements statements) {
         add("id", statements.getIdColumn());
+        add("createdAt", statements.getCreatedAtColumn());
```
paullatzelsperger, line 29: there should be a test for this.

bmg13: Updated/Adde tests regarding ContractDefinition's createdAt, but no UT related with that specific block, since similar mappings do not seem to have them. Would you propose to create a ContractDefinitionMappingTest?

rafaelmag110: The tests you provided do not represent the behavior that you described as failing in the attached issue. Please write a test that checks if the result of a contract definition api request is correctly ordered by creation date, when the request has a query spec contains such an ordering criteria.

bmg13: [Added suggested test](https://github.com/eclipse-edc/Connector/pull/5020/files#diff-1a7f691e4113c74f4d4c1ac3eb2e6bb7ea43a6efe5a4ea6f8db14add0b710f8bR15). Keep in mind that the `createdAt` field is not returned, so I enforced the result extraction as a list to keep the order and validate it with the return ids.

system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/ContractDefinitionApiEndToEndTest.java
```diff
@@ -117,6 +122,47 @@ void queryPolicyDefinitionWithSimplePrivateProperties(ManagementEndToEndTestCont
...
 
+        @Test
+        void queryContractDefinitions_sortByCreatedDate(ManagementEndToEndTestContext context, ContractDefinitionStore store) throws JsonProcessingException {
+            var id1 = UUID.randomUUID().toString();
+            var id2 = UUID.randomUUID().toString();
+            var id3 = UUID.randomUUID().toString();
+            var ids = List.of(id1, id2, id3);
+            for (String id : ids) {
+                store.save(createContractDefinition(id).build());
+            }
```
ndr-brt:
```suggestion
            Stream.of(id1, id2, id3).forEach(id -> store.save(createContractDefinition(id).build()));
```

ndr-brt (approved):

system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/ContractDefinitionApiEndToEndTest.java
bmg13: Added as suggested in [here](https://github.com/eclipse-edc/Connector/pull/5020/commits/c957151641957e582f4062b08a6bc982c9c2f68d).
Updated to nanoTime since test was failing in mem due to two entries having the same creation time in milli.

system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/ContractDefinitionApiEndToEndTest.java
```diff
@@ -117,6 +123,44 @@ void queryPolicyDefinitionWithSimplePrivateProperties(ManagementEndToEndTestCont
                     .body("size()", is(0));
         }
 
+        @Test
+        void queryContractDefinitions_sortByCreatedDate(ManagementEndToEndTestContext context, ContractDefinitionStore store) throws JsonProcessingException {
+            var id1 = UUID.randomUUID().toString();
+            var id2 = UUID.randomUUID().toString();
+            var id3 = UUID.randomUUID().toString();
+            Stream.of(id1, id2, id3).forEach(id -> store.save(createContractDefinition(id).createdAt(System.nanoTime()).build()));
```
jimmarino: Please don't do this. Start with a time and then increment using an offset so it is deterministic.

bmg13: followed suggestion, changed in [here](https://github.com/eclipse-edc/Connector/pull/5020/commits/c3f9e1c1af293a64c0baf500468a85894078da20).

system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/ContractDefinitionApiEndToEndTest.java
```diff
@@ -128,7 +129,9 @@ void queryContractDefinitions_sortByCreatedDate(ManagementEndToEndTestContext co
             var id1 = UUID.randomUUID().toString();
             var id2 = UUID.randomUUID().toString();
             var id3 = UUID.randomUUID().toString();
-            Stream.of(id1, id2, id3).forEach(id -> store.save(createContractDefinition(id).createdAt(System.nanoTime()).build()));
+            var createdAtTime = 1000L;
+            var increment = new AtomicInteger(0);
+            Stream.of(id1, id2, id3).forEach(id -> store.save(createContractDefinition(id).createdAt(createdAtTime + increment.getAndIncrement()).build()));
```
jimmarino: Why not use an AtomicLong and have it as one operation?

bmg13: You're right, simplified in [here](https://github.com/eclipse-edc/Connector/pull/5020/commits/6456ff3bb6f83601510048370d2714072ec27f8c).

spi/control-plane/contract-spi/src/testFixtures/java/org/eclipse/edc/connector/controlplane/contract/spi/testfixtures/offer/store/ContractDefinitionStoreTestBase.java
```diff
@@ -393,6 +393,16 @@ void verifySorting() {
             assertThat(getContractDefinitionStore().findAll(QuerySpec.Builder.newInstance().sortField("id").sortOrder(SortOrder.DESC).build())).hasSize(10).isSortedAccordingTo((c1, c2) -> c2.getId().compareTo(c1.getId()));
         }
 
+        @Test
+        void verifySortingCreatedAt() {
+            IntStream.range(0, 10).mapToObj(i -> createContractDefinition("id" + i)).forEach(getContractDefinitionStore()::save);
+
+            var querySpec = QuerySpec.Builder.newInstance().sortField("createdAt");
+
+            assertThat(getContractDefinitionStore().findAll(querySpec.sortOrder(SortOrder.ASC).build())).isSortedAccordingTo(Comparator.comparing(ContractDefinition::getCreatedAt));
```
jimmarino: Please format this so the method calls appear on subsequent lines - it makes it easier to read

system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/ContractDefinitionApiEndToEndTest.java
```diff
@@ -117,6 +124,45 @@ void queryPolicyDefinitionWithSimplePrivateProperties(ManagementEndToEndTestCont
                     .body("size()", is(0));
         }
 
+        @Test
+        void queryContractDefinitions_sortByCreatedDate(ManagementEndToEndTestContext context, ContractDefinitionStore store) throws JsonProcessingException {
```
jimmarino, line 126: Why are you doing this in an end-to-end test? This should be verified by individual unit tests and an integration test for Postgres. End-to-end tests should only verify complete codepaths and should not be used to verify individual functionality.

jimmarino (changes requested):

system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/ContractDefinitionApiEndToEndTest.java
bmg13: My reasoning is due to the raised issue was not having the ordered response and with this test that can be confirmed, being a `PostgresqlIntegrationTest`.
Similar to what found in [here](https://github.com/eclipse-edc/Connector/blob/52695d04658deee841a391c757e59c5567296bd6/system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/TransferProcessApiEndToEndTest.java#L234) for `TransferProcess`s.
If the end result was just the existence of a field, it would not make much sense, but having the order on the final response I believe it adds value.
What if I update this test to be `queryContractDefinitions_withQuerySpec` since it is broader and is not currently covered?

spi/control-plane/contract-spi/src/testFixtures/java/org/eclipse/edc/connector/controlplane/contract/spi/testfixtures/offer/store/ContractDefinitionStoreTestBase.java
bmg13: Done in [here](https://github.com/eclipse-edc/Connector/pull/5020/commits/52d6903a82f7618356110f8ba22bfc942b03bc25).

ndr-brt:
@jimmarino could you review it again? I'd be ready to merge this

system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/ContractDefinitionApiEndToEndTest.java
jimmarino: I see `verifySortingCreatedAt` covers the case for all store implementations. I don't understand the comment:

>What if I update this test to be queryContractDefinitions_withQuerySpec since it is broader and is not currently covered?

Is this test intended to verify that ordering is preserved end-to-end, or does it do something else?

system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/ContractDefinitionApiEndToEndTest.java
```diff
@@ -117,6 +124,45 @@ void queryPolicyDefinitionWithSimplePrivateProperties(ManagementEndToEndTestCont
...
+            var id1 = UUID.randomUUID().toString();
+            var id2 = UUID.randomUUID().toString();
+            var id3 = UUID.randomUUID().toString();
+            var createdAtTime = new AtomicLong(1000L);
+            Stream.of(id1, id2, id3).forEach(id -> store.save(createContractDefinition(id).createdAt(createdAtTime.getAndIncrement()).build()));
+
+            var content = """
```
jimmarino: Why is a JSON string created and parsed into a query object like this. The query object should be created in code directly.

jimmarino (changes requested):

system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/ContractDefinitionApiEndToEndTest.java
bmg13: This test is for the ordering only.

system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/ContractDefinitionApiEndToEndTest.java
bmg13: this was based on existing testing for timestamp sorting, like [this](https://github.com/eclipse-edc/Connector/blob/52695d04658deee841a391c757e59c5567296bd6/system-tests/management-api/management-api-test-runner/src/test/java/org/eclipse/edc/test/e2e/managementapi/TransferProcessApiEndToEndTest.java#L234) one

PR #5038, opened 2025-06-12
fix: use proper wildcard for HTTP accept header

### LINKED ISSUE #5037: Invalid Accept header when using default wildcard in `BaseRevocationListService.downloadStatusListCredential`
# Bug Report

## Describe the Bug

In `BaseRevocationListService.downloadStatusListCredential`, the `Accept` header is built from `acceptedContentTypes`. If this list is not configured, it defaults to `TrustedIssuerRegistry.WILDCARD`, which is currently set to `"*"`. However, `"*"` is not a valid value for an HTTP `Accept` header — the correct value should probably be `"*/*"` for a wildcard.

## Steps to Reproduce

1. Do not explicitly set `acceptedContentTypes` for revocation service.
2. Trigger `downloadStatusListCredential(...)`.
3. Inspect the outgoing HTTP request.
4. Observe `Accept: *` in the headers.

## Possible Implementation
`TrustedIssuerRegistry.WILDCARD` probably has a different meaning as it is also used in other context, so I think it should not be used here. It should probably use a separate constant like `HttpWildcard = "*/*"` for a HTTP context.

paullatzelsperger:
## What this PR changes/adds

use `"*/*"` wildcard instead of the wrong `"*"` one

## Why it does that

fix wildcard

## Further notes

## Linked Issue(s)

Closes #5037


paullatzelsperger:
FYI: @thomasrutger

jimmarino (approved):

PR #5040, opened 2025-06-13
feat: support DSP correct state in deprovisioning

### LINKED ISSUE #5032: Deprovision related state considered completed
# Feature Request

All the three states related to deprovisioning

- `DEPROVISIONING`,
- `DEPROVISIONING_REQUESTED`,
- `DEPROVISIONED`

Are considered `COMPLETED` in the EDC dsp protocol implementation. 
This makes the following DSP tck tests fail

- `TP:01-01`
- `TP:01-03`
- `TP:01-05`

since the TCK will check for the terminated state while the EDC returns completed

_Are there any requirements?_

## Solution Proposal

wolf4ood:
## What this PR changes/adds

Fixes dsp-tck tests:

- `TP:01-01`
- `TP:01-03`
- `TP:01-05`

We will base the decision if a TP is `TERMINATED` or `COMPLETED` only by checking the `errorDetail` field as a temporary workaround. This will not require any changes for users.

We can  revisit this behavior once https://github.com/eclipse-edc/Connector/issues/4793 is in.

## Why it does that

dsp-compliance

## Further notes

## Linked Issue(s)

Closes #5032 

### DIFF
--- a/spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/controlplane/transfer/spi/types/TransferProcess.java
+++ b/spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/controlplane/transfer/spi/types/TransferProcess.java
@@ -308,14 +308,17 @@ public class TransferProcess extends StatefulEntity<TransferProcess> {
     }
 
     public void transitionCompleting() {
+        this.errorDetail = null;
         transition(COMPLETING, state -> canBeCompleted());
     }
 
     public void transitionCompletingRequested() {
+        this.errorDetail = null;
         transition(COMPLETING_REQUESTED, state -> canBeCompleted());
     }
 
     public void transitionCompleted() {
+        this.errorDetail = null;
         transition(COMPLETED, COMPLETED, COMPLETING, COMPLETING_REQUESTED, STARTED);
     }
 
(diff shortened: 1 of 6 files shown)


spi/control-plane/transfer-spi/src/main/java/org/eclipse/edc/connector/controlplane/transfer/spi/types/TransferProcess.java
```diff
@@ -143,6 +143,8 @@ public class TransferProcess extends StatefulEntity<TransferProcess> {
     private String transferType;
     private String dataPlaneId;
 
+    private Integer previousState;
```
ronjaquensel: just a nit: for other state-related fields we use `int` instead of `Integer`, maybe worth to have this aligned

ndr-brt (commented):
Woulnd't be a more "holistic" approach to just keep memory of all the state changes? That will permit other logic based on the entity history to be evaluated in a better way (see the `<action>WasRequestedByTheCounterParty`), because a `previousState` field that's only set in particular occasions could be misunderstood easily.

wolf4ood:
We could discuss about naming, but I don't think memory will fit the purpose. Currently we have a the deprovisining states that are considered completed for DSP, which is wrong. For us deprovisioning is also considered a final state and we don't know how to map the deprovisioning in the correct DSP state unless we know which is the origination or the cause of the deprovisioning eg the transfer as either terminated or completed.

ndr-brt:
> We could discuss about naming, but I don't think memory will fit the purpose. Currently we have a the deprovisining states that are considered completed for DSP, which is wrong. For us deprovisioning is also considered a final state and we don't know how to map the deprovisioning in the correct DSP state unless we know which is the origination or the cause of the deprovisioning eg the transfer as either terminated or completed.

My proposal wasn't just to rename the field, but to not to have a simple field that stores only (eventually) the previous state, instead to keep a record of the state changes, from which we can understand if the transfer has been `COMPLETED` or `TERMINATED`, and it could also be used for other logic conditionals.
It would have been to catch the opportunity, once we have to migrate the database, to achieve more at one go.

in any case, theoretically provisioning/deprovisioning states should go away once #4793 is completed (I'm still working on it... slowly!)

jimmarino:
> Then if they are going we could think about leaving those DSP failing tests as allowed for the time being if we do not want to dirty the DB just for this
> 
> For the history of changes it would be ok for me but only in memory will not solve this use case and probably others. and probably would be a not so smol change :)

We need to pass the tests. I think we should put the simple fix in for now as we are under pressure to complete the TCK in the next couple of days. We can work on a loner-term solution later.

wolf4ood:
Here's the changeset with just the workaround

https://github.com/eclipse-edc/Connector/compare/main...wolf4ood:DataSpaceConnector:feat/5032_complete_dsp_tck_integration_workaround

wolf4ood:
I've applied the changeset to this PR. Now it does not require a migration and It only check the `errorDetail` field when serializing for DSP

ronjaquensel (approved):

PR #5118, opened 2025-07-14
fix: fixes DataAddress updates on InMemoryAssetIndex

### LINKED ISSUE #5117: Updates on an asset DataAddress fail for InMemoryAssetIndex
# Bug Report

## Describe the Bug

Given a connector installation running an in memory  database, during an asset update its data address is not properly updated.

The bug is hard to detect since a `findById() ` after an update returns the correct address. However there is also a map for data addresses used by the resolveForAsset(), and that is not considered during an updated, leading subsequent transfer processes to start with the not updated address.

## Context Information

Problem exists on main.

rafaelmag110:
## What this PR changes/adds

Fixes data address updates on the `InMemoryAssetIndex`

## Further notes

Took the liberty to adapt some existing update tests as those were not actually updating the reference - they would pass even if you removed the `assetIndex.update()` instruction.

@wolf4ood? Should be a quick one 🙏 

## Linked Issue(s)

Closes #5117


wolf4ood (approved):
LGTM thanks!

Issue #5006, opened 2025-05-22
Participant Identifier for different protocol versions


ndr-brt:
# Feature Request

In Catena-X we're about to go into a identifier change, to manage that in an interoperable way there's the need to have the connector being able to do that.

This will be a propaedeutic activity for the "multiple dataspaces support"

DR will come up

## Which Areas Would Be Affected?

_e.g., DPF, CI, build, transfer, etc._

## Why Is the Feature Desired?

_Are there any requirements?_

## Who will sponsor this feature?

_Please @-mention the committer that will sponsor your feature_.

## Solution Proposal

_If possible, provide a (brief!) solution proposal._

PR #5141, opened 2025-07-25
docs: DR for multi-identifier support

ronjaquensel:
## What this PR changes/adds

Adds a decision record for supporting different participant identifiers for different dataspace profile contexts

## Why it does that
Part of the dataspace profile context concept defined in [this DR](https://github.com/eclipse-edc/Connector/tree/main/docs/developer/decision-records/2025-05-28-dataspace-profile-context) to enable a connector to participate in multiple dataspaces.

me

## Linked Issue(s)

Part of #5006 

### DIFF
--- a/docs/developer/decision-records/2025-07-25-multiple-participant-identifiers/README.md
+++ b/docs/developer/decision-records/2025-07-25-multiple-participant-identifiers/README.md
@@ -20,9 +20,9 @@ to be considered for both the resolution of the own participant ID and the extra
 ### Resolution of own participant ID
 
 Currently, the participant ID is determined through one setting and will be used all throughout the connector. We will
-add a participant ID to the `DataspaceProfileContext`, which allows assigning a specific identifier to a specific
-context. The `DataspaceProfileContextRegistry` will provide a new method for getting the participant ID for a given
-protocol:
+add a new field `participantId` to the `DataspaceProfileContext`, which allows assigning a specific identifier to a
+specific context. The `DataspaceProfileContextRegistry` will provide a new method for getting the participant ID for a
+given protocol:
 
 ```java
 @NotNull
@@ -58,7 +58,8 @@ extraction of any other participant attributes, which will still be handled by t
 The extracted ID will then be passed to the `ParticipantAgentService` as a parameter.
 
 As ID extraction should depend on the `DataspaceProfileContext`, a function for extracting the participant ID from a
-`ClaimToken` will be added to the `DataspaceProfileContext`. For this function, we add a new interface:
+`ClaimToken` needs to be associated with a `DataspaceProfileContext`. For this function, we add a new interface as
+shown below. A new field `idExtractionFunction` will be added to the `DataspaceProfileContext` of this type.
 
 ```java
 @FunctionalInterface


docs/developer/decision-records/2025-07-25-multiple-participant-identifiers/README.md
```diff
@@ -0,0 +1,112 @@
...
+The previously used participant ID will remain the default and will be set in the `DataspaceProfileContextRegistry` as
+such. Thus, if no dedicated participant ID is defined for the given protocol, the method will return the default ID:
+
+```java
+public String getParticipantId(String protocol) {
+    return profiles().stream()
+            .filter(it -> it.name().equals(protocol))
+            .map(DataspaceProfileContext::participantId)
+            .findAny()
+            .orElse(defaultParticipantId);
```
ndr-brt: The current behavior is that in case of an unknown protocol/profile it returns null, so caller it's able to tell if the protocol/profile is a supported one or not. 
I think it should work that way to avoid hard-to-detect errors when there's a typo in the request.

docs/developer/decision-records/2025-07-25-multiple-participant-identifiers/README.md
```diff
@@ -0,0 +1,112 @@
...
+defined through configuration.
+
+The extraction will be pulled up into the `ProtocolTokenValidatorImpl` to separate extraction of the ID from the
+extraction of any other participant attributes, which will still be handled by the `ParticipantAgentServiceExtensions`.
+The extracted ID will then be passed to the `ParticipantAgentService` as a parameter.
+
+As ID extraction should depend on the `DataspaceProfileContext`, an additional method will be added to the
+`DataspaceProfileContextRegistry` for registering a new profile. In addition to the profile itself, this method will
+also accept a function for extracting the participant ID from a `ClaimToken`, an instance of
+`Function<ClaimToken, String>`:
```
ndr-brt: wouldn't a dedicated interface give better readability and control?

docs/developer/decision-records/2025-07-25-multiple-participant-identifiers/README.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+All registered profiles that were created without passing a function for ID extraction will use the default function.
+This will be supplied during creation of the `DataspaceProfileContextRegistryImpl` and contain the current default
+behaviour of looking for a configurable claim in the `ClaimToken`. As the default logic may be different depending on
+the `iam` extension used (e.g. for DCP the credential subject ID from received VCs is used), the default function must
+be overridable. Therefore, a third method will be added to the `DataspaceProfileContextRegistry`: 
+
+```java
+void setDefaultIdExtractionFunction(Function<ClaimToken, String> extractionFunction);
```
ndr-brt: I think the default function should only be functional for the "out of the box" profile, useful for testing and samples, but for a production dataspace I foresee a custom profile definition in any case so the adopters will be in control of the profile and they wouldn't be surprised when the default EDC implementation change for whatever reason.

TL;DR: I don't think we need this set method

docs/developer/decision-records/2025-07-25-multiple-participant-identifiers/README.md
ronjaquensel: Makes sense, I'll adapt this.

docs/developer/decision-records/2025-07-25-multiple-participant-identifiers/README.md
ronjaquensel: Probably yes, I'll add an interface for this function

docs/developer/decision-records/2025-07-25-multiple-participant-identifiers/README.md
ronjaquensel: I agree that for non-default profiles this would not be needed, as I would also expect adopters to explicitely define this when registering a profile. The reason I included this is because the default extraction looks different when using e.g. `iam-mock` vs. using DCP. So I intended this method to be used in the DCP extensions to set a new default behaviour.  But maybe this could also be handled another way as to not confuse adopters to use this method, let me give this some thought

ndr-brt: you're right, different identity services have different ways to extract the identifier, and it's correct that there's at least one default way to do so... should this be described by a `DefaultIdExtractionFunction` that will be registered by either iam-mock and dcp implementations and it will be put on the default dataspace profile.

ronjaquensel: agree, that sounds like a good option. Updated the DR accordingly. I also pulled the function for id extraction into the `DataspaceProfileContext` now instead of having these 2 separate, if a function needs to be defined explicitely for every profile anyway

docs/developer/decision-records/2025-07-25-multiple-participant-identifiers/README.md
```diff
@@ -0,0 +1,114 @@
...
+@ExtensionPoint
+public interface ParticipantIdExtractionFunction extends Function<ClaimToken, String> { }
+```
+
+Accordingly, we add another method to the `DataspaceProfileContextRegistry` for obtaining the ID extraction function
+for a given profile context:
+
+```java
+ParticipantIdExtractionFunction getIdExtractionFunction(String protocol);
+```
```
wolf4ood, line 75: This is for read, shouldn't also have a way to register `ParticipantIdExtractionFunction` for a specific profile?

ronjaquensel: Andrea mentioned under his last comment that the default function should only be functional for default profiles, which I think makes sense. Therefore I changed the DR a bit to include the function into the `DataspaceProfileContext`, thus no new registration method would be needed. Wdyt?

Nevertheless, I could probably make it a bit more explicit that the function would be added to the `DataspaceProfileContext`, as this is easily missable right now I think 😄

wolf4ood: Make sense

ronjaquensel: made it a bit more explicit in the text so it's clear that this will be added to the `DataspaceProfileContext` 🙂

ndr-brt (approved):

PR #5154, opened 2025-08-01
feat: add participant id extraction function to dataspace profile context

### LINKED ISSUE #5006: Participant Identifier for different protocol versions

ronjaquensel:
## What this PR changes/adds

Adds a function for extracting the counter-party id from a `ClaimToken` to the `DataspaceProfileContext`. Pulls the extraction of the participant id from the `ParticipantAgentService` up to the `ProtocolTokenValidator` to ensure the extracted id is not overridden by a `ParticipantAgentServiceExtension`. 

Previously, the id was extracted by either using a `ParticipantAgentServiceExtension` or, as a fallback solution, by looking for a configurable claim in the `ClaimToken`. This behaviour has been moved to implementations of the new interface `DefaultParticipantIdExtractionFunction`:
- provided by iam mock extension: using the configurable claim
- provided by DCP extension: extracting the credential subject id from a list of VCs

This is the second part of the implementation outlined in [this DR](https://github.com/eclipse-edc/Connector/tree/main/docs/developer/decision-records/2025-07-25-multiple-participant-identifiers).

## Why it does that

to support different dataspaces potentially using different participant identifiers

## Further notes

I introduced 2 new `RemoteMessage` implementations for requesting a negotiation or a transfer process by id respectively. These are required, as we need to pass protocol information for every DSP request. For now, the protocol is passed to the respective service as a string (since this was already done for the `CatalogProtocolService`), the service then builds the `RemoteMessage` and passes it to the token validator.

As now `RemoteMessages` are available for every request that is handled with a `GetDspRequest`, this can also be refactored to take a `RemoteMessage` instead of string. But this would likely make more sense in its own PR. Once this PR is merged, I will create a follow-up issue for that.

me

## Linked Issue(s)

Closes #5006

### DIFF
--- a/spi/common/participant-spi/src/main/java/org/eclipse/edc/participant/spi/ParticipantAgent.java
+++ b/spi/common/participant-spi/src/main/java/org/eclipse/edc/participant/spi/ParticipantAgent.java
@@ -53,10 +53,7 @@ public class ParticipantAgent {
 
     @Deprecated(since = "0.14.0")
     public ParticipantAgent(Map<String, Object> claims, Map<String, String> attributes) {
-        this.claims = Map.copyOf(claims);
-        this.attributes = Map.copyOf(attributes);
-        
-        this.id = attributes.get(PARTICIPANT_IDENTITY);
+        this(attributes.get(PARTICIPANT_IDENTITY), claims, attributes);
     }
 
     /**


core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/agent/ParticipantAgentServiceImpl.java
```diff
@@ -23,35 +24,24 @@
...
         extensions.stream().map(extension -> extension.attributesFor(token)).forEach(attributes::putAll);
-        if (!attributes.containsKey(PARTICIPANT_IDENTITY)) {
-            var claim = token.getClaim(identityClaimKey);
-            if (claim != null) {
-                attributes.put(PARTICIPANT_IDENTITY, claim.toString());
-            }
-        }
+
+        attributes.put(PARTICIPANT_IDENTITY, participantId);
         return new ParticipantAgent(token.getClaims(), attributes);
```
ndr-brt: could we add `id` as a field of `ParticipantAgent` at this point?

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/protocol/ProtocolTokenValidatorImpl.java
```diff
@@ -38,19 +40,21 @@ public class ProtocolTokenValidatorImpl implements ProtocolTokenValidator {
...
         this.identityService = identityService;
         this.monitor = monitor;
         this.policyEngine = policyEngine;
         this.agentService = agentService;
+        this.dataspaceProfileContextRegistry = dataspaceProfileContextRegistry;
     }
 
     @Override
-    public ServiceResult<ParticipantAgent> verify(TokenRepresentation tokenRepresentation, RequestPolicyContext.Provider policyContextProvider, Policy policy, RemoteMessage message) {
+    public ServiceResult<ParticipantAgent> verify(TokenRepresentation tokenRepresentation, RequestPolicyContext.Provider policyContextProvider, Policy policy, RemoteMessage message, String protocol) {
```
ndr-brt: Would provide a `RemoteMessage` implementation for the GET endpoints avoid this overload?

core/common/connector-core/src/main/java/org/eclipse/edc/connector/core/agent/ParticipantAgentServiceImpl.java
ronjaquensel: good point. I'll add an `id` field, add a new constructor that also takes an id and deprecate the old constructor. Once that deprecated constructor is removed in one of the next versions, we can then also enforce a participant id.

core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/protocol/ProtocolTokenValidatorImpl.java
ronjaquensel: yes, introducing `RemoteMessages` for the 3 scenarios would avoid the need to overload. I'll look into it

spi/common/participant-spi/src/main/java/org/eclipse/edc/participant/spi/ParticipantAgent.java
```diff
@@ -41,12 +41,22 @@ public class ParticipantAgent {
...
+        this.claims = Map.copyOf(claims);
+        this.attributes = Map.copyOf(attributes);
+    }
 
+    @Deprecated(since = "0.14.0")
     public ParticipantAgent(Map<String, Object> claims, Map<String, String> attributes) {
         this.claims = Map.copyOf(claims);
         this.attributes = Map.copyOf(attributes);
+        
+        this.id = attributes.get(PARTICIPANT_IDENTITY);
```
ndr-brt: nit: `this` constructor could be called

ronjaquensel: good point, that is indeed more elegant. Updated it

PR #5181, opened 2025-08-21
fix: add index for foreign key references

### LINKED ISSUE #5179: Index for lease_id references
Dear community, 
we are experiencing issues with unsuccessful processing of contract negotiations and transfer processes. Related error messages within DSP message processing show "EdcException: Entity is currently leased!". The analyses showed that this comes from a race condition where the counterparty connector's async DSP message is received even before the internal update to object itself (contract negotiation or transfer process) is fully performed.
Looking at the runtime of the single statements we saw that there are spikes in the removal of the lease from the edc_lease table. This is because the lease entries are referenced by other tables and those other tables do not have indexes for the foreign keys.

For primary keys Postgresql adding implicitly indexes, but for foreign keys this does not happen automatically. Therefore, I suggest to create indexes for all foreign keys referencing to the lease ID of the lease table.

```sql
CREATE UNIQUE INDEX IF NOT EXISTS contract_negotiation_lease_id_uindex
    ON edc_contract_negotiation (lease_id);

CREATE UNIQUE INDEX IF NOT EXISTS transfer_process_lease_id_uindex
    ON edc_transfer_process (lease_id);

CREATE UNIQUE INDEX IF NOT EXISTS policy_monitor_lease_id_uindex
    ON edc_policy_monitor (lease_id);

CREATE UNIQUE INDEX IF NOT EXISTS data_plane_lease_id_uindex
    ON edc_data_plane (lease_id);
```

Performance evaluation shows that there is a significant acceleration by putting those index.

Runtime of a DELETE of one lease from the edc_lease table.
Before adding the indexes: 922ms
Image

After adding the indexes: 0,5ms
Image

May I suggest a PR for adding those indexes to the schema SQL files? Additionally, I would suggest the same in the migration extension of the tractus-x repository.

lholthof:
## What this PR changes/adds

Add index for foreign key references which Postgresql does not create automatically. Relevant for SQL store extensions having foreign keys, e.g. to edc_lease table.

## Why it does that

Performance analysis showed that queries slow down without these indexes having about 50k entities in the tables.

## Further notes

-

@ndr-brt 

## Linked Issue(s)

Closes #5179


wolf4ood (approved):
LGTM

Thanks @lholthof

PR #5298, opened 2025-10-28
feat: enable participant context in the DCP identity service

### LINKED ISSUE #5295: Dcp Identity service: support for multiple participant context
# Feature Request

Currently the dcp (ex iatp) modules needs to  be configured for a single participant. In order to use those modules in a multi participant scenario we should move to `ParticipantContextConfig` when fetching
the did and the sts configurations

_Are there any requirements?_

## Solution Proposal

wolf4ood:
## What this PR changes/adds

enable participant context in the DCP identity service:

- The setting `edc.iam.issuer.id` is not loaded at boot time but derived from `ParticipantContextConfig`  when needed at runtime.
- The `TokenValidationAction` have been changed to receive the `participantContextId` in input
- The `AudienceValidationRule` has been moved into `SelfIssueIdTokenValidationAction` as runtime check using the above mechanism.
- The `VerifiableCredentialValidationService` and  `PresentationVerifier` have been changed to receive in input the `audience` when checking the VP. 

## Why it does that

## Further notes

Breaking change only at SPI layer

## Linked Issue(s)

Closes #5295 


extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IdentityAndTrustExtension.java
```diff
@@ -94,10 +94,10 @@ public class IdentityAndTrustExtension implements ServiceExtension {
     public static final String DCP_CLIENT_CONTEXT = "dcp-client";
     public static final String JSON_2020_SIGNATURE_SUITE = "JsonWebSignature2020";
     public static final long DEFAULT_CLEANUP_PERIOD_SECONDS = 60;
+    @Setting(description = "DID of the participant")
+    private static final String ISSUER_ID_KEY = "edc.iam.issuer.id";
```
paullatzelsperger, line 98: i'm wondering if `issuer` is the correct term, considering we now have an IssuerService and that might get a bit confusing....

Maybe we could deprecate this, and add something like `edc.iam.connector.did` or something?

wolf4ood: Yeah it's not super clear, also it could make sense to use the `edc.participant.id` to have a single setting for the id

wolf4ood: In case i would defer this in a separated PR wdyt?

paullatzelsperger: yep sure

PR #5301, opened 2025-10-29
fix(dataplanecore): Restore the final status of the lease renewal co…

### LINKED ISSUE #5294: The state override issue occurs during the lease renewal process of DataPlaneManagerImpl
# Bug Report

## Describe the Bug

Status coverage for lease renewal

### Expected Behavior

There is a state-overwriting issue in the updateFlowLease method of the DataPlaneManagerImpl class under certain extreme circumstances. When the processReceived method is executed asynchronously, if the asynchronous task is completed or fails, and there is concurrent execution with the updateFlowLease method, the completion or failure status of the asynchronous task may be overwritten as "STARTED".

### Observed Behavior

Renewal of lease in concurrent scenarios covers the actual task result status

## Context Information

- Used version EDC v0.14.1
- OS: Macos

## Detailed Description

## Possible Implementation

Image

Modify to the following implementation
```java
    private boolean updateFlowLease(DataFlow dataFlow) {
        dataFlow.setModified();
        store.save(dataFlow);
        return true;
    }
```

cycle2zhou:
…verage task

When the asynchronous execution result of a task and the lease renewal occur concurrently, it may cause the status of the task execution result to be reset to "STARTED" due to the lease renewal. By fixing the lease renewal process to only update the "updatedAt" field, we ensure that the final state is not overwritten.

## What this PR changes/adds

_Briefly describe WHAT your pr changes, which features it adds/modifies._

## Why it does that

## Further notes

## Linked Issue(s)

Closes #5294


paullatzelsperger (changes requested):
please add tests

core/data-plane/data-plane-core/src/test/java/org/eclipse/edc/connector/dataplane/framework/manager/DataPlaneManagerImplTest.java
```diff
@@ -1169,19 +1170,20 @@ class UpdateFlowLease {
...
             await().untilAsserted(() -> {
                 var captor = ArgumentCaptor.forClass(DataFlow.class);
                 verify(store).save(captor.capture());
                 var storedDataFlow = captor.getValue();
-                assertThat(storedDataFlow.getState()).isEqualTo(STARTED.code());
+                assertThat(storedDataFlow.getState()).isEqualTo(RECEIVED.code());
                 assertThat(storedDataFlow.getRuntimeId()).isEqualTo(runtimeId);
-                assertThat(storedDataFlow.getStateCount()).isEqualTo(1);
+                assertThat(storedDataFlow.getStateCount()).isEqualTo(0);
+                assertTrue(storedDataFlow.getUpdatedAt() > updatedAt);
```
paullatzelsperger: use assertJ assertions please: `assertThat(....).isGreaterThan(...)`

paullatzelsperger (changes requested):

core/data-plane/data-plane-core/src/test/java/org/eclipse/edc/connector/dataplane/framework/manager/DataPlaneManagerImplTest.java
```diff
@@ -1169,19 +1169,20 @@ class UpdateFlowLease {
 
         @Test
         void shouldUpdateFlow_whenFlowStartedAfterFlowLease() {
-            var dataFlow = dataFlowBuilder().state(RECEIVED.code()).build();
+            var dataFlow = dataFlowBuilder().runtimeId(runtimeId).state(RECEIVED.code()).build();
```
ndr-brt: please explicitly set also `stateCount = 0` so the assertion will verify that the value remains the same

ndr-brt: let's use the `STARTED` state, as only started data flow should have the flow lease updated.

core/data-plane/data-plane-core/src/test/java/org/eclipse/edc/connector/dataplane/framework/manager/DataPlaneManagerImplTest.java
```diff
@@ -1169,19 +1169,20 @@ class UpdateFlowLease {
...
             when(store.nextNotLeased(anyInt(), startedFlowOwnedByThisRuntime()))
                     .thenReturn(List.of(dataFlow)).thenReturn(emptyList());
 
             manager.start();
-
+            var updatedAt = dataFlow.getUpdatedAt();
```
ndr-brt: this is quite fragile, because the instance of `DataFlow` in input is the same of the one captured by the captor, so this work only because the variable is instanciated before the `storedDataFlow` one.

I'd propose this:
```java
 var clock = Clock.systemDefaultZone();
            var updatedAt = clock.millis();
            var dataFlow = dataFlowBuilder().runtimeId(runtimeId).state(RECEIVED.code()).updatedAt(updatedAt).clock(clock).build();

...

// assertion
assertThat(storedDataFlow.getUpdatedAt()).isGreaterThan(updatedAt);
```

core/data-plane/data-plane-core/src/test/java/org/eclipse/edc/connector/dataplane/framework/manager/DataPlaneManagerImplTest.java
```diff
@@ -1169,19 +1170,21 @@ class UpdateFlowLease {
...
         void shouldUpdateFlow_whenFlowStartedAfterFlowLease() {
-            var dataFlow = dataFlowBuilder().state(RECEIVED.code()).build();
+            var clock = Clock.systemDefaultZone();
+            var updatedAt = clock.millis();
+            var dataFlow = dataFlowBuilder().runtimeId(runtimeId).state(STARTED.code()).stateCount(0).updatedAt(updatedAt).clock(clock).build();
             when(store.nextNotLeased(anyInt(), startedFlowOwnedByThisRuntime()))
                     .thenReturn(List.of(dataFlow)).thenReturn(emptyList());
 
             manager.start();
-
```
ndr-brt, line 1177: one last nit, please leave an empty line between the "act" phase and the "assert" one

ndr-brt (approved):

PR #5352, opened 2025-11-18
feat: add multi-tenant vault SPI

paullatzelsperger:
## What this PR changes/adds

adds a `ParticipantVault`, that takes the `participantContextId` to store/resolve/delete secrets

## Why it does that

separation of secrets in the HSM

## Further notes

a default implementation `DefaultParticipantVaultImpl` was also provided, that just delegates to a conventional `Vault` instance.

In another PR I will remove usages of the old (and now deprecated) pieces of code that still use the legacy vault.

## Linked Issue(s)

Closes #5251

### DIFF
--- a/core/common/lib/keys-lib/src/main/java/org/eclipse/edc/keys/VaultPrivateKeyResolver.java
+++ b/core/common/lib/keys-lib/src/main/java/org/eclipse/edc/keys/VaultPrivateKeyResolver.java
@@ -18,7 +18,6 @@ import org.eclipse.edc.keys.spi.KeyParserRegistry;
 import org.eclipse.edc.spi.monitor.Monitor;
 import org.eclipse.edc.spi.result.Result;
 import org.eclipse.edc.spi.security.ParticipantVault;
-import org.eclipse.edc.spi.security.Vault;
 import org.eclipse.edc.spi.system.configuration.Config;
 import org.jetbrains.annotations.NotNull;
 
@@ -33,24 +32,17 @@ import static java.util.Optional.ofNullable;
 public class VaultPrivateKeyResolver extends AbstractPrivateKeyResolver {
 
     private final ParticipantVault participantVault;
-    private final Vault fallbackVault;
 
-    public VaultPrivateKeyResolver(KeyParserRegistry registry, ParticipantVault participantVault, Monitor monitor, Config config, Vault fallbackVault) {
+    public VaultPrivateKeyResolver(KeyParserRegistry registry, ParticipantVault participantVault, Monitor monitor, Config config) {
         super(registry, config, monitor);
         this.participantVault = participantVault;
-        this.fallbackVault = fallbackVault;
     }
 
     @NotNull
     @Override
     protected Result<String> resolveInternal(String participantContextId, String keyId) {
 
-        String key;
-        if (participantContextId == null) {
-            key = fallbackVault.resolveSecret(keyId);
-        } else {
-            key = participantVault.resolveSecret(participantContextId, keyId);
-        }
+        var key = participantVault.resolveSecret(participantContextId, keyId);
 
         return ofNullable(key)
                 .map(Result::success)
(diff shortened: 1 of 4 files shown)


core/common/boot/src/main/java/org/eclipse/edc/boot/VaultCompatibilityExtension.java
```diff
@@ -0,0 +1,44 @@
+/*
+ *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
```
ndr-brt: nit: copyright

core/common/lib/keys-lib/src/main/java/org/eclipse/edc/keys/VaultPrivateKeyResolver.java
```diff
@@ -31,18 +32,28 @@
...
-                .map(Result::success)
-                .orElseGet(() -> Result.failure("Private key with ID '%s' not found in Vault".formatted(keyId)));
+    protected Result<String> resolveInternal(String participantContextId, String keyId) {
+
+        String key;
+        if (participantContextId == null) {
+            key = fallbackVault.resolveSecret(keyId);
+        } else {
+            key = participantVault.resolveSecret(participantContextId, keyId);
+        }
```
ndr-brt: is the fallback vault necessary here? I feel like the participant vault should already manage this case (null participant context id) by itself

core/control-plane/control-plane-transfer-manager/src/main/java/org/eclipse/edc/connector/controlplane/transfer/process/TransferProcessManagerImpl.java
```diff
@@ -146,20 +146,20 @@ public StatusResult<TransferProcess> initiateConsumerRequest(ParticipantContext
...
+                              .contractId(transferRequest.getContractId())
+                              .protocol(transferRequest.getProtocol())
+                              .type(CONSUMER)
+                              .clock(clock)
+                              .transferType(transferRequest.getTransferType())
+                              .privateProperties(transferRequest.getPrivateProperties())
+                              .callbackAddresses(transferRequest.getCallbackAddresses())
+                              .traceContext(telemetry.getCurrentTraceContext())
+                              .participantContextId(participantContext.getParticipantContextId())
+                              .build();
```
ndr-brt: I just noticed this here but an huge part of the changelog are this changes in the spacing, could we stick with the typical style? that'll make the PR easier to read, then we can decide to change the checkstyle setting in a separate PR eventually

paullatzelsperger: yeah, no idea how that got in... i also did not change any of my IntelliJ Settings :D i can revert those whitespace changes

core/common/lib/keys-lib/src/main/java/org/eclipse/edc/keys/VaultPrivateKeyResolver.java
paullatzelsperger: unfortunately, yes. The code flows in through some of the dataplane-related code, that's why I kept the deprecated method around. once that is chopped off, we can remove all the fallback code and the deprecated methods

core/control-plane/control-plane-transfer-manager/src/main/java/org/eclipse/edc/connector/controlplane/transfer/process/TransferProcessManagerImpl.java
paullatzelsperger: it was an -- apparently new and silently activated -- extension in IntelliJ that changes indents... 🤦‍♂️

core/common/lib/keys-lib/src/main/java/org/eclipse/edc/keys/VaultPrivateKeyResolver.java
paullatzelsperger: edit: technically, we could get rid of it, but that would put the requirement on all `ParticipantVault` implementations, to be able to deal with nullable `vaultPartition` arguments, and I didn't like that

ndr-brt: But this could be well detailed into the javadoc: `participantContextId` could be null and the implementation needs to deal with that, so all the callers won't need to check it out (and inject an additional fallback vault)

paullatzelsperger: yeah, I guess you're right... however, in practice there likely would be a "default partition", rather than a fallback vault. Such a "default partition" might even require elevated access rights... different discussion though :) will update 👍

PR #5356, opened 2025-11-20
feat: allow encrypted properties on ParticipantContextConfig

### LINKED ISSUE #5354: Support sensitive config values in `ParticipantContextConfig`
# Feature Request

On occasion, the `ParticipantContextConfig` object may need to hold sensitive configuration values, such as access credentials to vaults.

In those situations the configuration value must be persisted in encrypted form, to avoid secret leakage.

`ParticipantContextConfig` and related classes

some configuration values are sensitive, but cannot be held in the vault, for example, Vault access credentials.

## Solution Proposal

add method(s): `getStringSensitive(String key)` to obtain the encrypted value out of the database, decrypt it and return the decrypted value.

One possible solution would be to maintain a global (EDC-wide) encryption key and use that to salt/encrypt the value.

wolf4ood:
## What this PR changes/adds

Enable encryption of `ParticipantContextConfiguration` `privateEntries`:

- Introduced an `EncryptionService` which providers encrypt and decrypt methods
- By default the runtime inject a no-op one,
- Plugged `EncryptionService` in `ParticipantContextConfigServiceImpl` that encrypt on save then properties in the new field `privateEntries`
- Plugged `EncryptionService` in `ParticipantContextConfigImpl` on `getStringSensitive` which calls the decryption of the property 
- Implements an extension that proviedes AES/GCM 

## Why it does that

## Further notes

## Linked Issue(s)

Closes #5354 

### DIFF
--- a/extensions/common/encryption/aes-encryption/README.md
+++ b/extensions/common/encryption/aes-encryption/README.md
@@ -1,1 +1,0 @@
-# File System Configuration
--- a/extensions/common/encryption/aes-encryption/src/main/java/org/eclipse/edc/encryption/aes/AesEncryptionService.java
+++ b/extensions/common/encryption/aes-encryption/src/main/java/org/eclipse/edc/encryption/aes/AesEncryptionService.java
@@ -50,6 +50,12 @@ public class AesEncryptionService implements EncryptionService {
 
     }
 
+    @Override
+    public Result<String> decrypt(String cipherText) {
+        return getKey(aesKeyAlias)
+                .compose(key -> doDecrypt(cipherText, key));
+    }
+
     private Result<String> doEncrypt(String plainText, SecretKey key) {
         try {
             var iv = generateIv(IV_SIZE_BYTES);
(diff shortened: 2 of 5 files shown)


extensions/common/encryption/aes-encryption/src/main/java/org/eclipse/edc/encryption/aes/AesEncryptionService.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Override
+    public Result<String> encrypt(String plainText) {
+        return getKey(aesKeyAlias)
+                .compose(key -> doEncrypt(plainText, key));
+
+    }
+
+    private Result<String> doEncrypt(String plainText, SecretKey key) {
```
paullatzelsperger, line 59: nit: method ordering: public should go above private ones

extensions/common/encryption/aes-encryption/README.md
```diff
@@ -0,0 +1 @@
+# File System Configuration
```
paullatzelsperger: remove

spi/common/participant-context-config-spi/src/main/java/org/eclipse/edc/participantcontext/spi/config/ParticipantContextConfig.java
```diff
@@ -100,4 +100,14 @@ public interface ParticipantContextConfig {
...
+
+    /**
+     * Returns the sensitive String representation of the value
+     *
+     * @param participantContextId the participant context identifier
+     * @param key                  of the setting
+     * @return a String representation of the setting
+     * @throws EdcException if no setting is found
+     */
+    String getStringSensitive(String participantContextId, String key);
```
paullatzelsperger: micro nit: would call this `getSensitiveString()`

PR #5408, opened 2025-12-12
feat: support for JWT status list credential

### LINKED ISSUE #5398: Add JWT parsing when evaluating status list credential response
# Feature Request

When requesting a status list credential from an issuer, the connector expects the response to be JSON in the `BaseRevocationListService`. As per specification. the status list credential may be returned as either JSON or as part of a JWT depending on the `Accept` header in the request. Thus, the connector should support both options as well. We will therefore introcuce JWT parsing in the `BaseRevocationListService` depending on the value of the `Accept` header.

DCP

Specification support, alignment with IdentityHub's default behaviour

me

## Solution Proposal

In  the `downloadStatusListCredential` of the `BaseRevocationListService`, evaluate the `Accept` header before parsing the response. Depending on its value, either parse the JSON (as is done right now), or first parse the JWT and extract the `vc` field, before parsing it to JSON.

ronjaquensel:
## What this PR changes/adds

Adds support for requesting and parsing the status list credential as JWT in the `BaseRevocationListService`.

## Why it does that

specification support, alignment with identity hub's default behaviour

me

## Linked Issue(s)

Closes #5398

### DIFF
--- a/extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/revocation/BaseRevocationListService.java
+++ b/extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/revocation/BaseRevocationListService.java
@@ -9,6 +9,7 @@
  *
  *  Contributors:
  *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *       Fraunhofer-Gesellschaft zur Förderung der angewandten Forschung e.V. - JWT parsing for status list credential
  *
  */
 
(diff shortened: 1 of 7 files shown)


extensions/common/iam/verifiable-credentials/src/test/java/org/eclipse/edc/iam/verifiablecredentials/revocation/bitstring/BitstringStatusListRevocationServiceTest.java
```diff
@@ -317,4 +333,30 @@ void getStatusPurpose_multipleCredentialStatus() {
                     .isEqualTo("revocation, suspension");
         }
     }
+
+    @Test
+    void downloadStatusListCredential_asJwt_callTokenValidationService() throws Exception {
```
ndr-brt: the "token validation fails" case is missing, same for status list 2021.

extensions/common/iam/verifiable-credentials/src/main/java/org/eclipse/edc/iam/verifiablecredentials/revocation/BaseRevocationListService.java
```diff
@@ -160,21 +172,36 @@ protected Result<C> getCredential(String credentialUrl) {
...
                               .url(credentialUrl)
-                              .header("Accept", String.join(",", acceptedContentTypes))
+                              .header("Accept", acceptHeader)
                               .get()
                               .build();
         try (var response = httpClient.execute(request)) {
             if (response.isSuccessful()) {
                 if (response.body() == null) {
                     throw new EdcException("Response body is null for URL: " + credentialUrl);
                 }
```
ndr-brt: nit: from the latest major okhttp version, response body became non-null by default, so no need to check it for nullability

extensions/common/iam/verifiable-credentials/src/test/java/org/eclipse/edc/iam/verifiablecredentials/revocation/bitstring/BitstringStatusListRevocationServiceTest.java
ronjaquensel: good point!

PR #5423, opened 2025-12-18
feat: introduce encryption algo registry

### LINKED ISSUE #5422: Refactor: EncryptionService
# Feature Request

Currently the `EncryptionService` is an injectable service, with only one implementation of `aes` algorithm.

We should make more flexible introducing an `EncryptionAlgorithmRegistry` where users can register encryption strategy. 

This will also solve a potential cyclic dependencies when using the hashicorp vault and aes implemetation.

_Are there any requirements?_

## Solution Proposal

wolf4ood:
## What this PR changes/adds

introduce encryption algo registry

## Why it does that

## Further notes

## Linked Issue(s)

Closes #5422 

### DIFF
--- a/core/common/participant-context-config-core/src/main/java/org/eclipse/edc/participantcontext/config/ParticipantContextConfigServicesExtension.java
+++ b/core/common/participant-context-config-core/src/main/java/org/eclipse/edc/participantcontext/config/ParticipantContextConfigServicesExtension.java
@@ -36,7 +36,7 @@ public class ParticipantContextConfigServicesExtension implements ServiceExtensi
 
     @Setting(
             description = "The encryption algorithm used for encrypting and decrypting sensitive config.",
-            key = "edc.participants.config.encryption",
+            key = "edc.participants.config.encryption.algorithm",
             defaultValue = "aes"
     )
     private String encryptionAlgorithm;


core/common/participant-context-config-core/src/main/java/org/eclipse/edc/participantcontext/config/ParticipantContextConfigServicesExtension.java
```diff
@@ -14,41 +14,53 @@
...
+import static org.eclipse.edc.participantcontext.config.ParticipantContextConfigServicesExtension.NAME;
 
 @Extension(NAME)
 public class ParticipantContextConfigServicesExtension implements ServiceExtension {
 
     public static final String NAME = "Participant Context Config Services Extension";
 
+    @Setting(
+            description = "The encryption algorithm used for encrypting and decrypting sensitive config.",
+            key = "edc.participants.config.encryption",
```
paullatzelsperger: nit: maybe something like `edc.participants.config.encryption.algorithm` would be more intuitive?

extensions/common/encryption/aes-encryption/src/main/java/org/eclipse/edc/encryption/aes/AesEncryptionExtension.java
```diff
@@ -33,15 +34,27 @@ public class AesEncryptionExtension implements ServiceExtension {
...
+
+    @Inject
+    private EncryptionAlgorithmRegistry registry;
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+        if (aesKeyAlias == null || aesKeyAlias.isBlank()) {
+            monitor.info("AES encryption key alias not set; AES encryption algorithm will not be registered");
+        } else {
+            registry.register("aes", new AesEncryptionAlgorithm(vault, aesKeyAlias));
```
paullatzelsperger, line 56: not sure we can store the encryption key in the vault. What happens if we want to store per-participant vault access credentials? that would create a chicken-and-egg-problem...

wolf4ood: this is the global key in the global vault, which allows to encrypt per participant `privateEntries`

paullatzelsperger: ok, that's fine then.

PR #5436, opened 2026-01-08
fix: policy evaluation NPE

### LINKED ISSUE #5426: Policy evaluation NPE
# Bug Report

Hey everyone

## Describe the Bug

If a policy constraint's left operand is bound to a scope, but no evaluation function for that scope is registered, a set of default cases applies depending on the constraint operator ([ref](https://github.com/eclipse-edc/Connector/blob/main/core/common/lib/policy-evaluator-lib/src/main/java/org/eclipse/edc/policy/evaluator/PolicyEvaluator.java#L176-L194)). As of right now, only the operators `EQ`, `IN` and `NEQ` are being handled. If the policy constraint contains any other operator, the function returns null, which propagates a Null Pointer Exception (NPE).

### Expected Behavior

If a constraint is being evaluated but no evaluation function for it was found, the evaluation should fail.

Note that the `policydefinitions` API tries to enforce this by searching for an evaluation function when a `PolicyDefinition` is being registered, but one can be created for a different scope. For example, a policy constraint left operand can be bound to the `CATALOG_SCOPE`, but it's function bound to the `TRANSFER_SCOPE`.

### Observed Behavior

NPE propagated on Policy evaluation.

For example, when requesting a Catalog, the NPE is propagated on the Provider Connector and the Consumer Connector receives a 500.

## Steps to Reproduce

1. Create an extension that registers the following:
```java
// Action needs to be bound so that the permission in not filtered out (ScopeFilter, L76-L78)
ruleBindingRegistry.bind("http://www.w3.org/ns/odrl/2/use", CATALOG_SCOPE);
// Bind a left operand to CATALOG_SCOPE
ruleBindingRegistry.bind("https://w3id.org/edc/v0.0.1/ns/leftOp", CATALOG_SCOPE);
// Bind an evaluation function to another scope (e.g. TRANSFER_SCOPE). Needed to be able to create the PolicyDefinition
policyEngine.registerScope(TRANSFER_SCOPE, TransferProcessPolicyContext.class);
policyEngine.registerFunction(
                TransferProcessPolicyContext.class,
                Permission.class,
                "https://w3id.org/edc/v0.0.1/ns/leftOp",
                (operator, rightValue, rule, ctx) -> true);
```
2. Start a Connector pair, Consumer and Provider
3. Create any dummy `Asset`
4. Create a `PolicyDefinition` as follows:
```json
{
	"@context": {
		"@vocab": "https://w3id.org/edc/v0.0.1/ns/"
	},
	"@id": "test-policy",
	"policy": {
		"@context": "http://www.w3.org/ns/odrl.jsonld",
		"@type": "Set",
		"permission": [
			{
				"action": "odrl:use", // use bound action
				"constraint": {
					"leftOperand": "leftOp", // use bound operand
					"operator": "isAnyOf", // use an unhandled operator (e.g. isAnyOf)
					"rightOperand": "rightOp"
				}
			}
		]
	}
}
```
5. Create a `ContractDefinition` with the created policy definition as the access policy and usage policy
6. Send a Catalog Request
7. Verify receives 500 error from Provider

## Context Information

- Tested in version 0.14.0, but I believe the Policy Engine / Evaluator has not changed since

## Detailed Description

500 error:

Image

NPE logs:

```text
SEVERE 2025-12-22T17:42:03.084785254 JerseyExtension: Unexpected exception caught

java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because "result" is null

	at org.eclipse.edc.policy.evaluator.PolicyEvaluator.visitRule(PolicyEvaluator.java:240)

	at org.eclipse.edc.policy.evaluator.PolicyEvaluator.visitPermission(PolicyEvaluator.java:99)

	at org.eclipse.edc.policy.evaluator.PolicyEvaluator.visitPermission(PolicyEvaluator.java:48)

	at org.eclipse.edc.policy.model.Permission.accept(Permission.java:42)

	at org.eclipse.edc.policy.evaluator.PolicyEvaluator.lambda$visitPolicy$0(PolicyEvaluator.java:75)

	at java.base/java.util.ArrayList.forEach(Unknown Source)

	at org.eclipse.edc.policy.evaluator.PolicyEvaluator.visitPolicy(PolicyEvaluator.java:75)

	at org.eclipse.edc.policy.evaluator.PolicyEvaluator.visitPolicy(PolicyEvaluator.java:48)

	at org.eclipse.edc.policy.model.Policy.accept(Policy.java:94)

	at org.eclipse.edc.policy.evaluator.PolicyEvaluator.evaluate(PolicyEvaluator.java:67)

	at org.eclipse.edc.policy.engine.PolicyEngineImpl.evaluate(PolicyEngineImpl.java:140)

	at org.eclipse.edc.connector.controlplane.catalog.ContractDefinitionResolverImpl.lambda$resolveFor$2(ContractDefinitionResolverImpl.java:62)

	at java.base/java.util.Optional.map(Unknown Source)

	at org.eclipse.edc.connector.controlplane.catalog.ContractDefinitionResolverImpl.lambda$resolveFor$3(ContractDefinitionResolverImpl.java:62)

	at java.base/java.util.stream.ReferencePipeline$2$1.accept(Unknown Source)

	at java.base/java.util.stream.SliceOps$1$1.accept(Unknown Source)

	at java.base/java.util.stream.ReferencePipeline$2$1.accept(Unknown Source)

	at java.base/java.util.concurrent.ConcurrentHashMap$ValueSpliterator.tryAdvance(Unknown Source)

	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(Unknown Source)

	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(Unknown Source)

	at java.base/java.util.stream.AbstractPipeline.copyInto(Unknown Source)

	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source)

	at java.base/java.util.stream.AbstractPipeline.evaluate(Unknown Source)

	at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(Unknown Source)

	at java.base/java.util.stream.ReferencePipeline.toArray(Unknown Source)

	at java.base/java.util.stream.ReferencePipeline.toArray(Unknown Source)

	at java.base/java.util.stream.ReferencePipeline.toList(Unknown Source)

	at org.eclipse.edc.connector.controlplane.catalog.ContractDefinitionResolverImpl.resolveFor(ContractDefinitionResolverImpl.java:67)

	at org.eclipse.edc.connector.controlplane.catalog.DatasetResolverImpl.query(DatasetResolverImpl.java:67)

	at org.eclipse.edc.connector.controlplane.services.catalog.CatalogProtocolServiceImpl.lambda$getCatalog$0(CatalogProtocolServiceImpl.java:62)

	at org.eclipse.edc.spi.result.AbstractResult.map(AbstractResult.java:145)

	at org.eclipse.edc.connector.controlplane.services.catalog.CatalogProtocolServiceImpl.lambda$getCatalog$1(CatalogProtocolServiceImpl.java:61)

	at org.eclipse.edc.transaction.spi.NoopTransactionContext.execute(NoopTransactionContext.java:34)

	at org.eclipse.edc.connector.controlplane.services.catalog.CatalogProtocolServiceImpl.getCatalog(CatalogProtocolServiceImpl.java:60)

	at org.eclipse.edc.protocol.dsp.http.message.DspRequestHandlerImpl.createResource(DspRequestHandlerImpl.java:132)

	at org.eclipse.edc.protocol.dsp.catalog.http.api.controller.BaseDspCatalogApiController.requestCatalog(BaseDspCatalogApiController.java:84)
```

## Possible Implementation

I believe that the default cases that execute when no evaluation function is found were added in the very early stages of the Policy Engine, but now are not needed. A possible implementation would be to return false if an evaluation function was not found, and never return null.

correiaafonso12:
## What this PR changes/adds

Adjusts the `PolicyEvaluator` to remove policy literal evaluations and to return `false` if an evaluation function was not found.

## Why it does that

Avoid possible NPE.

## Further notes

If no evaluation function was found an entry is added to the `ruleProblems`. Note that this doesn't mean that the error is logged. The caller function will have access to all "problems" and may handle it as needed.

@ndr-brt 

## Linked Issue(s)

Closes #5426

### DIFF
--- a/core/common/lib/policy-evaluator-lib/src/main/java/org/eclipse/edc/policy/evaluator/PolicyEvaluator.java
+++ b/core/common/lib/policy-evaluator-lib/src/main/java/org/eclipse/edc/policy/evaluator/PolicyEvaluator.java
@@ -156,27 +156,35 @@ public class PolicyEvaluator implements Policy.Visitor<Boolean>, Rule.Visitor<Bo
 
     @Override
     public Boolean visitAtomicConstraint(AtomicConstraint constraint) {
-        var rightValue = constraint.getRightExpression().accept(this);
         var leftRawValue = constraint.getLeftExpression().accept(this);
-        if (leftRawValue instanceof String) {
-            AtomicConstraintFunction<Object, Rule, Boolean> function;
-            if (ruleContext instanceof Permission) {
-                function = getEvaluationFunction((String) leftRawValue, permissionFunctions, dynamicPermissionFunctions);
-            } else if (ruleContext instanceof Prohibition) {
-                function = getEvaluationFunction((String) leftRawValue, prohibitionFunctions, dynamicProhibitionFunctions);
-            } else {
-                function = getEvaluationFunction((String) leftRawValue, dutyFunctions, dynamicDutyFunctions);
-            }
-            if (function != null) {
-                return function.evaluate(constraint.getOperator(), rightValue, ruleContext);
-            }
+        if (!(leftRawValue instanceof String)) {
+            ruleProblems.add(RuleProblem.Builder.newInstance()
+                    .rule(ruleContext)
+                    .description(ruleContext.toString())
+                    .constraintProblem(new ConstraintProblem("Left operand value is not a String", constraint))
+                    .build());
+            return false;
+        }
+
+        AtomicConstraintFunction<Object, Rule, Boolean> function;
+        if (ruleContext instanceof Permission) {
+            function = getEvaluationFunction((String) leftRawValue, permissionFunctions, dynamicPermissionFunctions);
+        } else if (ruleContext instanceof Prohibition) {
+            function = getEvaluationFunction((String) leftRawValue, prohibitionFunctions, dynamicProhibitionFunctions);
+        } else {
+            function = getEvaluationFunction((String) leftRawValue, dutyFunctions, dynamicDutyFunctions);
+        }
+        if (function == null) {
             ruleProblems.add(RuleProblem.Builder.newInstance()
                     .rule(ruleContext)
                     .description(ruleContext.toString())
                     .constraintProblem(new ConstraintProblem("No evaluation function found", constraint))
                     .build());
+            return false;
         }
-        return false;
+
+        var rightValue = constraint.getRightExpression().accept(this);
+        return function.evaluate(constraint.getOperator(), rightValue, ruleContext);
     }
 
     @Override
--- a/core/common/lib/policy-evaluator-lib/src/test/java/org/eclipse/edc/policy/evaluator/PolicyEvaluatorTest.java
+++ b/core/common/lib/policy-evaluator-lib/src/test/java/org/eclipse/edc/policy/evaluator/PolicyEvaluatorTest.java
@@ -48,11 +48,23 @@ import static org.mockito.Mockito.when;
 class PolicyEvaluatorTest {
 
     @Test
+    void verifyEvaluationFails_whenLeftOperandIsNotString() {
+        var constraint = AtomicConstraint.Builder.newInstance()
+                .leftExpression(new LiteralExpression(new Object()))
+                .operator(Operator.EQ)
+                .rightExpression(new LiteralExpression("foo"))
+                .build();
         var permission = Permission.Builder.newInstance().constraint(constraint).build();
         var policy = Policy.Builder.newInstance().permission(permission).build();
 
         var evaluator = PolicyEvaluator.Builder.newInstance().build();
+        var result = evaluator.evaluate(policy);
 
+        assertThat(result.valid()).isFalse();
+        assertThat(result.getProblems())
+                .flatExtracting(RuleProblem::getConstraintProblem)
+                .extracting(ConstraintProblem::getDescription)
+                .contains("Left operand value is not a String");
     }
 
     @ParameterizedTest(name = "{displayName} {0}")


core/common/lib/policy-evaluator-lib/src/test/java/org/eclipse/edc/policy/evaluator/PolicyEvaluatorTest.java
```diff
@@ -39,33 +39,30 @@
...
         var constraint = createLiteralAtomicConstraint("foo", "foo");
-
-        var duty = Duty.Builder.newInstance().constraint(constraint).build();
-        var policy = Policy.Builder.newInstance().duty(duty).build();
+        var permission = Permission.Builder.newInstance().constraint(constraint).build();
+        var policy = Policy.Builder.newInstance().permission(permission).build();
 
         var evaluator = PolicyEvaluator.Builder.newInstance().build();
-        assertTrue(evaluator.evaluate(policy).valid());
+        assertFalse(evaluator.evaluate(policy).valid());
```
ndr-brt: could you please add assertion also for the rule problem, to avoid false negatives.

core/common/lib/policy-evaluator-lib/src/test/java/org/eclipse/edc/policy/evaluator/PolicyEvaluatorTest.java
```diff
@@ -39,33 +39,30 @@
...
     }
 
     @Test
-    void verifyProhibitionNotEqualEval() {
-        var constraint = createLiteralAtomicConstraint("baz", "bar");
-
+    void verifyProhibitionEvaluationFails_whenNoFunctionWasDefined() {
```
ndr-brt: nit: these three tests could be described with a parameterized test

core/common/lib/policy-evaluator-lib/src/main/java/org/eclipse/edc/policy/evaluator/PolicyEvaluator.java
```diff
@@ -171,28 +170,13 @@ public Boolean visitAtomicConstraint(AtomicConstraint constraint) {
...
                 return function.evaluate(constraint.getOperator(), rightValue, ruleContext);
             }
+            ruleProblems.add(RuleProblem.Builder.newInstance()
+                    .rule(ruleContext)
+                    .description(ruleContext.toString())
+                    .constraintProblem(new ConstraintProblem("No evaluation function found", constraint))
+                    .build());
         }
-
-        // TODO handle expression eval errors
```
ndr-brt, line 176: that's a whole lot of dead code, nice to 🪓 it!
could you please add a rule problem in this case so in the case of left raw being not a string the caller will be notified accordingly

core/common/lib/policy-evaluator-lib/src/test/java/org/eclipse/edc/policy/evaluator/PolicyEvaluatorTest.java
correiaafonso12: Implemented at [b1e711f](https://github.com/eclipse-edc/Connector/pull/5436/commits/b1e711f1299e1ffccd1a72323cab8b8ffe8e0215)

core/common/lib/policy-evaluator-lib/src/test/java/org/eclipse/edc/policy/evaluator/PolicyEvaluatorTest.java
correiaafonso12: Implemented at [b1e711f](https://github.com/eclipse-edc/Connector/pull/5436/commits/b1e711f1299e1ffccd1a72323cab8b8ffe8e0215)

core/common/lib/policy-evaluator-lib/src/main/java/org/eclipse/edc/policy/evaluator/PolicyEvaluator.java
correiaafonso12: Implemented at [1086e80](https://github.com/eclipse-edc/Connector/pull/5436/commits/1086e8029010cb37d0fd03f3359da78c4c7cacff), and also added a unit test.
The unit test only evaluates a policy with `Permission`, and not with the other rules, as it shouldn't have any impact. Hope that's ok.

Issue #5323, opened 2025-11-07
Dataplane Signaling Protocol support


ndr-brt:
# Feature Request

The [Dataplane Signaling Specification](https://github.com/eclipse-dataplane-signaling/dataplane-signaling) is on its way, and we will support it fully, to do that we'll need to apply some changes

## Solution Proposal

- [x] Add additional fields on `Asset`
- [x] Add additional fields on `TransferRequest` and `TransferProcess`
- [x] Dataplane registration
- [x] Data plane availability check
- [x] Dataplane Signaling Protocol DataFlowController
- [x] Dataplane Signaling Protocol Client
- [x] refactor: inline DataFlowManager
- [x] Data Flow suspension
- [x] Async Prepare
- [x] Async Start
- [x] Cleanup DataAddress from store when TP reaches final states
- [x] Fail transfer when consumer data-plane not found
- [x] In NotifyPrepared command handler make call .prepared and not .provisioned
- [x] remove data destination from management v4
- [x] discovery through transfer type and labels
- [x] Authentication between CP and DP

PR #5495, opened 2026-02-09
feat: cleanup TransferProcess's DataAddress on final state

ndr-brt:
## What this PR changes/adds

Removes DataAddress entries from vault when TransferProcess reaches final states (TERMINATED, COMPLETED)

## Why it does that

secrets cleanup.

## Further notes

## Linked Issue(s)

Part of #5323

### DIFF
--- a/core/control-plane/control-plane-transfer/src/main/java/org/eclipse/edc/connector/controlplane/transfer/dataaddress/VaultDataAddressStore.java
+++ b/core/control-plane/control-plane-transfer/src/main/java/org/eclipse/edc/connector/controlplane/transfer/dataaddress/VaultDataAddressStore.java
@@ -81,7 +81,7 @@ public class VaultDataAddressStore implements DataAddressStore {
     }
 
     @Override
-    public StoreResult<DataAddress> remove(TransferProcess transferProcess) {
+    public StoreResult<Void> remove(TransferProcess transferProcess) {
         var dataAddressAlias = transferProcess.getDataAddressAlias();
         if (dataAddressAlias == null) {
             return StoreResult.success();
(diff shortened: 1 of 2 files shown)


core/control-plane/control-plane-transfer/src/main/java/org/eclipse/edc/connector/controlplane/transfer/dataaddress/VaultDataAddressStore.java
```diff
@@ -80,6 +80,22 @@ public StoreResult<DataAddress> resolve(TransferProcess transferProcess) {
...
+    public StoreResult<DataAddress> remove(TransferProcess transferProcess) {
+        var dataAddressAlias = transferProcess.getDataAddressAlias();
+        if (dataAddressAlias == null) {
+            return StoreResult.success();
+        }
+
+        var result = vault.deleteSecret(transferProcess.getParticipantContextId(), dataAddressAlias);
+        if (result.succeeded()) {
+            transferProcess.setDataAddressAlias(null);
+            return StoreResult.success();
```
ronjaquensel, line 93: According to the interface method's javadoc, shouldn't the removed data address be returned in case of successful deletion? Though not sure if this is required, so I would also be fine with just removing that statement from the javadoc in `DataAddressStore`.

ronjaquensel (approved):

core/control-plane/control-plane-transfer/src/main/java/org/eclipse/edc/connector/controlplane/transfer/dataaddress/VaultDataAddressStore.java
ndr-brt: you're right, I'll fix the javadoc: no reason to return the data address

PR #5517, opened 2026-02-18
fix: make dataDestination accessible for TP that relies on legacy data plane specs

### LINKED ISSUE #5516: TransferProcess DataAddress is not accessible anymore for PUSH transfers
# Feature Request

At the moment the DataAddress is correctly stored in the vault instead of being plain in the database because it contains secrets to get the data.

The only issue is that when the transfer is a PUSH one and the address is provisioned by the data-plane (e.g. using random strings in the destination), the control-plane could not be able to get the address, as the vault entry gets removed as soon as it reaches the `COMPLETED` state.

The proposal would be to store it in the database as long as the registered `FlowController` is the legacy one.
With the upcoming data-plane signaling the DataAddress will be managed by the data plane in its entirety so the consumer will need to ask the data-plane for that information.

_Are there any requirements?_

## Solution Proposal

ndr-brt:
## What this PR changes/adds

Adds missing storage of `DataAddress` on Transfer initialization on consumer side.
Avoid deleting `dataDestination` from TransferProcess when the data plane protocol used is the legacy one: this will avoid breaking the management contract for which the data destination is readable from the management API.
With the new data plane signaling specs the data destination/data address will be totally managed by the data-plane so it could be eventually requested there, but the control-plane will just store it securely to being able to pass it to the counter party but nothing more..

To permit this I introduced a `DataPlaneProtocolInUse` service class that permits the `VaultDataAddress` to understand when the DataAddress needs to be stored in the vault (new protocol) or in the vault and the database (legacy protocol).
It's already deprecated so it will be deleted together with the legacy protocol.

## Why it does that

## Further notes

## Linked Issue(s)

Closes #5516 

### DIFF
--- a/core/control-plane/control-plane-transfer/src/main/java/org/eclipse/edc/connector/controlplane/transfer/TransferProcessDefaultServicesExtension.java
+++ b/core/control-plane/control-plane-transfer/src/main/java/org/eclipse/edc/connector/controlplane/transfer/TransferProcessDefaultServicesExtension.java
@@ -21,6 +21,7 @@ import org.eclipse.edc.connector.controlplane.transfer.spi.TransferProcessPendin
 import org.eclipse.edc.connector.controlplane.transfer.spi.flow.TransferTypeParser;
 import org.eclipse.edc.connector.controlplane.transfer.spi.observe.TransferProcessObservable;
 import org.eclipse.edc.connector.controlplane.transfer.spi.types.DataAddressStore;
+import org.eclipse.edc.connector.controlplane.transfer.spi.types.DataPlaneProtocolInUse;
 import org.eclipse.edc.jsonld.spi.JsonLd;
 import org.eclipse.edc.runtime.metamodel.annotation.Extension;
 import org.eclipse.edc.runtime.metamodel.annotation.Inject;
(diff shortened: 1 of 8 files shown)


wolf4ood (approved):

core/control-plane/control-plane-transfer/src/main/java/org/eclipse/edc/connector/controlplane/transfer/TransferProcessDefaultServicesExtension.java
```diff
@@ -63,7 +66,11 @@ public TransferTypeParser transferTypeParser() {
 
     @Provider
     public DataAddressStore dataAddressStore() {
-        return new VaultDataAddressStore(vault, typeTransformerRegistry, jsonLd);
+        return new VaultDataAddressStore(vault, typeTransformerRegistry, jsonLd, dataPlaneProtocolInUse);
     }
 
+    @Provider
+    public DataPlaneProtocolInUse dataPlaneProtocolInUse() {
+        return new DataPlaneProtocolInUse();
```
ronjaquensel: shouldn't this return the same instance used by the `DataAddressStore`? Right now, 2 separate instances are created

ndr-brt: indeed! it was a typo by going back and forth trying to decide which extension should provide it :)
thanks

ronjaquensel (approved):

Issue #5462, opened 2026-01-21
Add `context` to `VerifiableCredential` class


arnoweiss:
# Feature Request

The `VerifiableCredential` POJO is currently missing a means to differentiate its semantics if the `type` property is equivalent. It should contain that.

## Which Areas Would Be Affected?

controlplane, transformers, validation functions

## Why Is the Feature Desired?

In a single-issuer, multi-dataspace scenario, verifiers currently can't distinguish credentials of a type.
https://github.com/eclipse-tractusx/tractusx-edc/issues/2379
Previously discussed in https://github.com/eclipse-edc/Connector/discussions/5446

## Who will sponsor this feature?

@ndr-brt 

## Solution Proposal

expand the class, let the validation functions decide whether to json-ld-expand or not.

PR #5518, opened 2026-02-18
docs: decision record about dynamic JSON-LD context inference

wolf4ood:
## What this PR changes/adds

decision record about dynamic JSON-LD context inference

## Why it does that

## Further notes

## Linked Issue(s)

### DIFF
--- a/docs/developer/decision-records/2026-02-18-dynamic-jsonld-context-inference/README.md
+++ b/docs/developer/decision-records/2026-02-18-dynamic-jsonld-context-inference/README.md
@@ -1,0 +1,86 @@
+# Dynamic JSON-LD Context Inference
+
+## Decision
+
+We will implement a mechanism to dynamically infer JSON-LD contexts based on the structure and content of the data being
+processed.
+
+## Rationale
+
+Currently, EDC relies on predefined JSON-LD contexts to interpret and process data. Users can register
+custom JSON-LD contexts to use when compacting JSON-LD data depending on the scope where the compaction is performed.
+We have different scopes for DSP. DCP and Management API but once the context is registered
+it's always used for all compaction operations within that scope. This has served well for connectors operating in a
+single dataspace, but as we expand to support multiple dataspaces/participans in the same runtime, having static
+JSON-LD context for single scopes becomes a limitation.
+
+For example let's say we have two dataspaces X and Y, each with its own set of JSON-LD contexts for custom policy terms
+definitions or custom asset terms. Currently, the only way to support this use case is to inject a `@context` of all the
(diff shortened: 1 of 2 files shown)


wolf4ood:
Storing the `@context` in entities could be another approach, but that's a ton of work and it will make the entities/store to be json-ld aware. Having this in the json-ld service is the easiest way. We could also achieve this without additional configuration and inspecting the configured JSON-LD context and do the reverse mapping. It might be tricky but doable. 

For the ParticipantContext not sure we need that for every participant. That would be more
a `DataspaceContext` configured with JSON-LD context and having an association between `ParticipantContext` and `DataspaceContext` for choosing which contexts to use in each dataspace context when compacting. But that means that we have to store for each entity which dataspace  context they belong to

ndr-brt:
> Yeah that's why the offered assets and policy is a CD choices so we cannot assign profile or something to policies. That's why having a dynamic JSON-LD context help here. Based on the "found" namespaces it construct a JSON-LD context which is consistent.
> 
> The semantic profile can be easily inferred on input when creating the asset or policies by the users using the right `@context` the issue is when giving that entity back. Which in this DR is automatically inferred by the entities namespaces found in properties.
> 
> The other approach would be storing that along with the entity or having a "semantic profile" in the management API to associate manually

Maybe it's simple as it is: given that an ingested asset/policy can only be interpreted correctly with a specific context, it's necessary for the control-plane to store these contexts as well along with the entity, there's nothing wrong with it, as the entities will need that context in egestion phase.
Not by accident, there's an issue that reports this exact need for another entity: https://github.com/eclipse-edc/Connector/issues/5462

wolf4ood:
That's probably the cleaner solution, but it will require major refactoring, on entities, store,
and in compaction/expansion strategy where we do that in the interceptor and controllers
only receive/return expanded objects. I could do a spike on this.

wolf4ood:
I did a little spike on this. Storing the `@context` is doable but it's a bit work:

We'd have to change all the entities that can be customized  by storing the `@context` at least

- Asset
- Policy Definition
- Contract Negotiation
- Contract Agreement

We'd have to change how we manage JSON-LD on those controller. I think we cannot use a generic interceptor that expand and compact but rather going back to managing within the single controller

On ingress we'd have to extract the `@context`, run the validation, expand it and transforming to Java entities associating the extracted context.

On Egress we'd have to load the entity/entities transform to JsonObject and then compact by calling a variant of `JsonLd#compact` which supports additional context likely in a List of Object/String to be passed in stored in the entity.

It's a major refactor but we can do it in step and only for new controllers v5 for example.

Let's discuss if we want to go with this solution

PR #5696, opened 2026-04-23
fix: add @context to QuerySpec, Criterion, IdResponse OpenAPI schemas

### LINKED ISSUE #5690: QuerySpec OpenAPI definition is missing the @context property
# Bug Report

## Describe the Bug

I'm coming from https://github.com/eclipse-tractusx/tractusx-edc/issues/2754 since I noticed that the OpenAPI definition comes from this project.
I'm generating a REST Client from the OpenAPI definition. While then searching for Assets with a limit of 10 results in 50 elements returned.

### Expected Behavior

The requested number of elements

### Observed Behavior

The search always returns 50 elements

## Steps to Reproduce

1. Generate a client from the OpenAPI definition
2. Make a search on the `/v3/assets/request` endpoint with `limit` and `offset` (e.g. offset = 0, limit = 10)
3. Endpoint always returns 50 elements

## Context Information

- Used version: 0.10.3

## Detailed Description

When sending the request with the generated client the body has the following content

```json
{
  "limit": 10,
  "offset": 0
}
```

Sending this manually via e.g. Postman the same behaviour can be observed (50 elements returned).

Changing the body to the following results in the correct number of requested elements

```json
{
  "@context": {
  },
  "limit": 10,
  "offset": 0
}
```

Looking at the QuerySpec definition the @context is only present in the example and not in the properties which are used when generating the client.

## Possible Implementation

The QuerySpec definition would need the following additional property

```diff yml
QuerySpec:
  type: object
  properties:
+   '@context':
+     type: object
    '@type':
      type: string
      example: https://w3id.org/edc/v0.0.1/ns/QuerySpec
    filterExpression:
      type: array
      items:
        $ref: "#/components/schemas/Criterion"
    limit:
      type: integer
      format: int32
    offset:
      type: integer
      format: int32
    sortField:
      type: string
    sortOrder:
      type: string
      enum:
        - ASC
        - DESC
  example:
    '@context':
      '@vocab': https://w3id.org/edc/v0.0.1/ns/
    '@type': QuerySpec
    offset: 5
    limit: 10
    sortOrder: DESC
    sortField: fieldName
    filterExpression: [ ]
```

mvanhorn:
Closes #5690.

## Why

`QuerySpecSchema` in `core/common/lib/api-lib/src/main/java/org/eclipse/edc/api/model/ApiCoreSchema.java` declares the fields used by springdoc to generate the OpenAPI definition for `QuerySpec`. The example string for that record includes `@context`, but the record fields themselves don't — so OpenAPI generators don't emit `@context` as a property and clients leave it out of the request body.

The reporter's symptom is consistent with that: posting a body without `@context` falls through the JSON-LD context expansion in the management API and the `limit` / `offset` never reach the database layer. The endpoint then returns the configured default of 50 rows regardless of what the caller asked for. Adding `"@context": {}` to the body fixes the response, which confirms the diagnosis the issue points at.

## Change

Adds an `Object context` field to the `QuerySpecSchema` record annotated with `@Schema(name = CONTEXT)`, where `CONTEXT` is the existing `JsonLdKeywords.CONTEXT` constant (`"@context"`). This mirrors the existing pattern for `TYPE` on the same record.

I left it non-required (no `requiredMode = REQUIRED`) — the issue's proposed YAML diff lists it under `properties` only, not under `required`, so generators describe it as a property without forcing every existing client to send it.

`CriterionSchema` has the same example-vs-properties asymmetry but the issue scopes the fix to `QuerySpec`, so this PR does too. Happy to add the same field to `CriterionSchema` in a follow-up if maintainers want them aligned.

## Verification

The generated OpenAPI YAML is built from this Java source via springdoc, so I cannot regenerate it without the full build toolchain locally. The change is structurally identical to the `@Schema(name = TYPE) String type` field directly above it, which produces the `@type` property in the published OpenAPI.

### DIFF
--- a/core/common/lib/api-lib/src/main/java/org/eclipse/edc/api/model/ApiCoreSchema.java
+++ b/core/common/lib/api-lib/src/main/java/org/eclipse/edc/api/model/ApiCoreSchema.java
@@ -21,6 +21,7 @@ import org.eclipse.edc.spi.types.domain.DataAddress;
 import java.util.List;
 
 import static io.swagger.v3.oas.annotations.media.Schema.RequiredMode.REQUIRED;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.CONTEXT;
 import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.ID;
 import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.TYPE;
 import static org.eclipse.edc.spi.query.Criterion.CRITERION_TYPE;
@@ -30,6 +31,8 @@ public interface ApiCoreSchema {
 
     @Schema(name = "Criterion", example = CriterionSchema.CRITERION_EXAMPLE)
     record CriterionSchema(
+            @Schema(name = CONTEXT)
+            Object context,
             @Schema(name = TYPE, example = CRITERION_TYPE)
             String type,
             @Schema(requiredMode = REQUIRED)
@@ -52,6 +55,8 @@ public interface ApiCoreSchema {
 
     @Schema(name = "QuerySpec", example = QuerySpecSchema.QUERY_SPEC_EXAMPLE)
     record QuerySpecSchema(
+            @Schema(name = CONTEXT)
+            Object context,
             @Schema(name = TYPE, example = EDC_QUERY_SPEC_TYPE)
             String type,
             int offset,
@@ -75,6 +80,8 @@ public interface ApiCoreSchema {
 
     @Schema(name = "IdResponse", example = IdResponseSchema.ID_RESPONSE_EXAMPLE)
     record IdResponseSchema(
+            @Schema(name = CONTEXT)
+            Object context,
             @Schema(name = ID)
             String id,
             long createdAt
--- a/extensions/common/api/control-api-configuration/src/main/resources/control-api-version.json
+++ b/extensions/common/api/control-api-configuration/src/main/resources/control-api-version.json
@@ -2,7 +2,7 @@
   {
     "version": "2.1.4",
     "urlPath": "/v1",
-    "lastUpdated": "2026-04-09T12:00:01Z",
+    "lastUpdated": "2026-04-25T09:00:00Z",
     "maturity": "stable"
   }
 ]
--- a/extensions/common/api/management-api-configuration/src/main/resources/management-api-version.json
+++ b/extensions/common/api/management-api-configuration/src/main/resources/management-api-version.json
@@ -14,7 +14,7 @@
   {
     "version": "5.0.0-beta",
     "urlPath": "/v5beta",
-    "lastUpdated": "2026-04-24T09:00:00Z",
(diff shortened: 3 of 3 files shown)


paullatzelsperger:
@mvanhorn you need to agree to the ECA before we can accept your PR

theobisproject:
I also noticed that the `Criterion` and `IdResponseSchema` definition have the `@context` in the OpenAPI examples but are also missing it in the properties like in the `QuerySpec`

mvanhorn:
Thanks @theobisproject - pushed 9bd449f which adds `@Schema(name = CONTEXT) Object context` to both `CriterionSchema` and `IdResponseSchema`, mirroring the `QuerySpecSchema` fix. @paullatzelsperger I'll sign the ECA; seems the check is still blocking so let me know if there's anything else I should do on that front.

ndr-brt:
> Thanks @theobisproject - pushed [9bd449f](https://github.com/eclipse-edc/Connector/commit/9bd449f3db2ffe2bc9d46feb800a54293f1a6da0) which adds `@Schema(name = CONTEXT) Object context` to both `CriterionSchema` and `IdResponseSchema`, mirroring the `QuerySpecSchema` fix. @paullatzelsperger I'll sign the ECA; seems the check is still blocking so let me know if there's anything else I should do on that front.

the ECA needs to be signed for all the commits, so you either squash them into one signed by you, otherwise also @theobisproject needs to sign it.

Please also fix the PR name to follow [our guidelines](https://eclipse-edc.github.io/documentation/for-contributors/guidelines/pr-etiquette/), and you need to modify the `management-api-version.json` and `control-api-version.json`  files in order to make the related checks pass, as this is a change in the API documentation.

mvanhorn:
Squashed into a single signed commit (01e1aad) and rebased onto current main:

- Title renamed: `fix: add @context to QuerySpec, Criterion, IdResponse OpenAPI schemas`
- `control-api-version.json` and `management-api-version.json` (v5beta) lastUpdated bumped to today
- The single commit is signed off by me, so the per-commit ECA check should be satisfied without @theobisproject needing to sign

ECA on my Eclipse Foundation account is the remaining blocker.

ndr-brt (approved):
once the ECA is signed, we can merge this

mvanhorn:
ECA cleared after amending the commit author from the GitHub no-reply email to my real one (2b0f498). Signed-off-by preserved. Ready to merge whenever you have a window.

mvanhorn:
@ndr-brt @paullatzelsperger - the Tck-Tests failure is the known-flaky `DspTckDockerTest$InMemoryTest` `TP:03-02` timeout. Same test passes in this run for the `PostgresTest` and `DspTckEmbeddedTest` modes, and main has been failing the InMemory variant intermittently this week (4 of 10 recent main runs). The diff here is +9/-2 across three OpenAPI schema files and cannot affect the transfer-process state machine timing the test exercises. Re-run or merge as you see fit.

ndr-brt:
thanks!

PR #5709, opened 2026-04-28
feat(dps): implement suspend/resume cases

### LINKED ISSUE #5707: DSP: manage suspension and resumption
# Feature Request

As described in the specs:
https://github.com/eclipse-dataplane-signaling/dataplane-signaling/blob/658908b546a67a1967bb2a433e0211a55d181821/docs/signaling.md#data-flow-suspensionresumption

_Are there any requirements?_

## Solution Proposal

ndr-brt:
## What this PR changes/adds

Implements the suspension/resumption cases as detailed [in the specs](https://github.com/eclipse-dataplane-signaling/dataplane-signaling/blob/658908b546a67a1967bb2a433e0211a55d181821/docs/signaling.md#data-flow-suspensionresumption).

**BREAKING CHANGE**: a new `data_address_owner BOOLEAN DEFAULT false` column needs to be added to the `edc_transfer_process` table

## Why it does that

DPS

## Further notes
- the "resume" related status now are: `RESUMING`, `RESUMING_REQUESTED`, `RESUMED`, we need 3 of them because there's a distinct way to handle resumption based on the fact that the transfer process "owns" the data address or not. the ownership is assigned based on the direction of the transfer: is owner if "PULL and PROVIDER" or "PUSH and CONSUMER". The ownership is assigned when the "prepare" or "start" DPS message returns: if the data address is returned there, the TP is the data address owner. This was the only way to assure this without giving the CP the knowledge about the data flow direction.
- added a `TransferProcessResumed` event, to permit the DSP TCK to work correctly (in the case that the suspension/resumption is triggered by a "non owner of the data address"

## Linked Issue(s)

Closes #5707 


core/control-plane/control-plane-aggregate-services/src/main/java/org/eclipse/edc/connector/controlplane/services/transferprocess/TransferProcessProtocolServiceImpl.java
```diff
@@ -183,22 +183,46 @@ private ServiceResult<TransferProcess> requestedAction(ParticipantContext partic
...
     @NotNull
     private ServiceResult<TransferProcess> startedAction(TransferStartMessage message, TransferProcess transferProcess) {
+        if (transferProcess.currentStateIsOneOf(SUSPENDED)) {
+            transferProcess.protocolMessageReceived(message.getId());
+            transferProcess.transitionResuming();
+            return update(transferProcess)
+                    .onSuccess(i -> observable.invokeForEach(l -> l.resuming(transferProcess)))
+                    .flatMap(ServiceResult::from)
+                    .map(it -> transferProcess);
+        }
```
ronjaquensel, line 198: Maybe I'm missing something here, but shouldn't the data address also be stored in this case, if present in the message? In the cases `provider-push - consumer suspend/start` and `consumer-pull - provider suspend/start` the counter-party would immediately send the data address in the `TransferStartMessage` without this control plane having prior knowledge about the resumption happening, i.e. the process would be in the `SUSPENDED` state, right?

ndr-brt: good catch, yes, in case the receiver of the message is not a "data address owner" it should store the message that is in the message.

ronjaquensel (approved):

PR #5747, opened 2026-05-15
feat: add profile field

### LINKED ISSUE #5733: Introduce profile field in V5 API
# Feature Request

Since we are starting to rollout the dataspace profile management, and we basically map the profile to the current `protocol` that we use in the management APIs, it would be nice if in the api v5 we support the `profile` field and deprecate/remove the `protocol` one.

_Are there any requirements?_

## Solution Proposal

wolf4ood:
## What this PR changes/adds

 Since we are sharing the schema v4 with v4 api and v5, i've added profile field in schema v4. the `protocol` will continue to work on both versions. on the *Request object is `protocol` or `profile` are mutually exclusive required. On V5 we should advertise that `protocol` is being replaced by `profile`. The underlying storage and field in pojo is not changed.
 
 All the v5 tests are now switched to `profile` field.

## Why it does that

## Further notes

## Linked Issue(s)

Closes #5733 

### DIFF
--- a/spi/control-plane/catalog-spi/src/main/java/org/eclipse/edc/connector/controlplane/catalog/spi/CatalogRequest.java
+++ b/spi/control-plane/catalog-spi/src/main/java/org/eclipse/edc/connector/controlplane/catalog/spi/CatalogRequest.java
@@ -96,6 +96,7 @@ public class CatalogRequest {
             return this;
         }
 
+        @Deprecated(since = "management-api:v5")
         public Builder protocol(String protocol) {
             instance.protocol = protocol;
             return this;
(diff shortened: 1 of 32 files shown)


spi/control-plane/catalog-spi/src/main/java/org/eclipse/edc/connector/controlplane/catalog/spi/CatalogRequest.java
```diff
@@ -27,7 +27,9 @@ public class CatalogRequest {
 
     public static final String CATALOG_REQUEST_TYPE_TERM = "CatalogRequest";
     public static final String CATALOG_REQUEST_TYPE = EDC_NAMESPACE + CATALOG_REQUEST_TYPE_TERM;
+    @Deprecated(since = "management-api:v5")
```
ndr-brt: If the idea is to remove it in version 5, let's put deprecated since `management-api:v4`

ndr-brt (approved):

PR #5858, opened 2026-07-01
fix: update PostgreSQL type casts in JsonFieldTranslator and add tests

### LINKED ISSUE #5843: invalid PostgreSQL type casts in JsonFieldTranslator
# Bug Report

## Describe the Bug
When applying a filter expression on a Catalog for a numeric type property (double, integer). The control plane fails with exception error. 

`org.eclipse.edc.spi.persistence.EdcPersistenceException: org.postgresql.util.PSQLException: ERROR: type "double" does not exist
     at org.eclipse.edc.sql.SqlQueryExecutor.query(SqlQueryExecutor.java:97)
     at org.eclipse.edc.connector.controlplane.store.sql.assetindex.SqlAssetIndex.lambda$queryAssets$0(SqlAssetIndex.java:62)
     at org.eclipse.edc.connector.controlplane.catalog.DatasetResolverImpl.query(DatasetResolverImpl.java:74)
     at org.eclipse.edc.connector.controlplane.services.catalog.CatalogProtocolServiceImpl.lambda$getCatalog$0(CatalogProtocolServiceImpl.java:62)`

### Expected Behavior

Type casting to double should be working fine.

## Steps to Reproduce

apply this filter expression on a property of a numeric type ex: 

```json
{
     "@context": { "@vocab": "https://w3id.org/edc/v0.0.1/ns/" },
     "@type": "CatalogRequest",
     "counterPartyAddress": "<provider-dsp-address>",
     "counterPartyId": "<provider-id>",
     "protocol": "dataspace-protocol-http",
     "querySpec": {
       "filterExpression": [
         { "operandLeft": "'https://w3id.org/edc/v0.0.1/ns/score'", "operator": "=", "operandRight": 10.0 }
       ]
     }
   }
```

## Context Information

After checking this file [JsonFieldTranslator.java](https://github.com/eclipse-edc/Connector/blob/baab7fd146b6ba1d761e5df004f4e56d1da93cf0/core/common/lib/sql-lib/src/main/java/org/eclipse/edc/sql/translation/JsonFieldTranslator.java#L65) seems to use some casting syntax that is not supported by postgres. 

```java
   private String createStatementForType(Class<?> type, String statement) {
       if (type.equals(Boolean.class)) return format("(%s)::boolean", statement); 
       if (type.equals(Integer.class)) return format("(%s)::integer", statement); 
       if (type.equals(Double.class))  return format("(%s)::double", statement);  
       if (type.equals(Float.class))   return format("(%s)::float", statement);   
       if (type.equals(Long.class))    return format("(%s)::long", statement);   
       if (type.equals(Short.class))   return format("(%s)::short", statement); 
       return statement;
   }

```

PostgreSQL has no types named double, long, or short. The correct names are double precision (float8), bigint (int8), and smallint (int2).

even though integer doesn't produce an exception error, the numeric values are stored in the DB as a double and thus integer can't be successfully compared.

## Possible Implementation

Map the Java numeric types to valid PostgreSQL type names:
```java
   if (type.equals(Double.class)) return format("(%s)::double precision", statement); 
   if (type.equals(Long.class))   return format("(%s)::bigint", statement);       
   if (type.equals(Short.class))  return format("(%s)::smallint", statement);    
```

mokhairymahmoud:
## What this PR changes/adds

This PR fixes the Postgres cast type error in [JsonFieldTranslator.java](https://github.com/eclipse-edc/Connector/compare/main...mokhairymahmoud:Connector:fix/PostgreSQL-type-casts?expand=1#diff-569a6a3fe3110fe6a92d41b9009b7e760bea4a243a848476124bd233b5c69183), as Postgres doesn't have "double" syntax but rather "double percision" similarly for the float, short and long.  

## Why it does that

To allow using numeric filter expression in the catalog and everywhere.

## Linked Issue(s)

Closes #5843 

### DIFF
--- a/core/common/lib/core-lib/src/main/java/org/eclipse/edc/sql/translation/JsonFieldTranslator.java
+++ b/core/common/lib/core-lib/src/main/java/org/eclipse/edc/sql/translation/JsonFieldTranslator.java
@@ -72,19 +72,19 @@ public class JsonFieldTranslator implements FieldTranslator {
         }
 
         if (type.equals(Double.class)) {
-            return format("(%s)::double", statement);
+            return format("(%s)::double precision", statement);
         }
 
         if (type.equals(Float.class)) {
-            return format("(%s)::float", statement);
+            return format("(%s)::real", statement);
         }
 
         if (type.equals(Long.class)) {
-            return format("(%s)::long", statement);
+            return format("(%s)::bigint", statement);
         }
 
         if (type.equals(Short.class)) {
-            return format("(%s)::short", statement);
+            return format("(%s)::smallint", statement);
         }
 
         return statement;
--- a/core/common/lib/core-lib/src/test/java/org/eclipse/edc/sql/translation/JsonFieldTranslatorTest.java
+++ b/core/common/lib/core-lib/src/test/java/org/eclipse/edc/sql/translation/JsonFieldTranslatorTest.java
@@ -104,7 +104,7 @@ class JsonFieldTranslatorTest {
 
             var result = translator.toWhereClause(PathItem.parse("field"), criterion, operator);
 
-            assertThat(result.sql()).isEqualTo("(column_name ->> 'field')::double = ?");
+            assertThat(result.sql()).isEqualTo("(column_name ->> 'field')::double precision = ?");
             assertThat(result.parameters()).containsExactly(100.0);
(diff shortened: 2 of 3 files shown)


ndr-brt (approved):
thanks

PR #5921, opened 2026-07-24
docs: add ADR for map-based type transformer registry

mokhairymahmoud:
## What this PR changes/adds

Adds a decision record to change typeTransformer registry from an arraylist to a hashmap.

## Why it does that

It allows a reliable replacement of the default transformer registered in the transformer registry.

### DIFF
--- a/docs/developer/decision-records/2026-07-24-transformers-registry/README.md
+++ b/docs/developer/decision-records/2026-07-24-transformers-registry/README.md
@@ -1,0 +1,58 @@
+# Change the type-transformer registry from a list to a map
+
+## Decision
+
+`TypeTransformerRegistryImpl` will store registered transformers in a map rather than a list.
+
+The registry will use a nested map, keyed first by input type and then by output type:
+
+```java
+Map<Class<?>, Map<Class<?>, TypeTransformer<?, ?>>> transformers
+```
+A registration is identified by the pair formed by `TypeTransformer#getInputType()` and
+`TypeTransformer#getOutputType()`. Registering a transformer with an existing pair replaces the
+previous transformer for that pair. The last registration therefore wins.
+
+Registries created with `forContext` remain isolated from their parent registries. A context
+registry first resolves its own transformer and falls back to its parent only when it has no match,
+so a context-specific registration can override a default registration without changing the
+parent registry.
+
+## Rationale
+
+The registry currently stores transformers in an `ArrayList` and returns a compatible transformer
+during lookup using `findAny()`.
+
+Registering a transformer for a pair already handled by a default transformer does not reliably
+replace the default; the default remains eligible for selection. The selected transformer depends
+on the registry's encounter order, which is an implicit and unreliable override mechanism.
+
+The map-based implementation ensures that the last registration for a given input/output pair is
+the one returned by `transformerFor`. To override a default transformer, an extension must
+register its transformer for the same pair after the default transformer has been registered.
+
+## Approach
+
+1. Replace the list in `TypeTransformerRegistryImpl` with a map keyed by registered input and
+	output types.
+2. Make `register` replace the entry for an already registered type pair.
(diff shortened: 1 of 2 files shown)


docs/developer/decision-records/2026-07-24-transformers-registry/README.md
```diff
@@ -0,0 +1,58 @@
...
+The map-based implementation ensures that the last registration for a given input/output pair is
+the one returned by `transformerFor`. To override a default transformer, an extension must
+register its transformer for the same pair after the default transformer has been registered.
+
+## Approach
+
+1. Replace the list in `TypeTransformerRegistryImpl` with a map keyed by registered input and
+	output types.
+2. Make `register` replace the entry for an already registered type pair.
+3. Update `transformerFor` to retain compatible-supertype matching and parent-context fallback.
```
ndr-brt, line 39: It would be nice to hear a little more about what will be the modification to "retain compatible-supertype matching"

docs/developer/decision-records/2026-07-24-transformers-registry/README.md
```diff
@@ -0,0 +1,58 @@
...
+4. Add tests for same-registry replacement, context override, parent fallback, and supertype
+   matching, including ambiguous compatible types.
+
+The public `TypeTransformerRegistry` API remains unchanged.
+
+### Consequences
+
+- A later registration for the same input/output pair intentionally replaces the earlier one.
+- Extensions can override defaults by registering in the applicable registry context.
+- The most specific compatible input type wins. Equally specific unrelated types fail explicitly.
```
ndr-brt: could you please elaborate more on the second part of this point?

docs/developer/decision-records/2026-07-24-transformers-registry/README.md
mokhairymahmoud: I added a sample code for how I think it can be done. Just to mention that this is in case we need to support supertype matching, if we decide to drop it completely and depend on exact inputType exact matching, it would be much easier.  what do you think ?

docs/developer/decision-records/2026-07-24-transformers-registry/README.md
mokhairymahmoud: I added an example to clarify what I mean. Again, this will only appear if we decide to support superType matching, if we decide to drop it and depend on exact matching, this will not be needed.

mokhairymahmoud: @paullatzelsperger what do you think here ?

docs/developer/decision-records/2026-07-24-transformers-registry/README.md
mokhairymahmoud: @paullatzelsperger what do you think here ?

docs/developer/decision-records/2026-07-24-transformers-registry/README.md
```diff
@@ -0,0 +1,83 @@
...
+	For example, `List -> Output` is selected over `Collection -> Output` for an `ArrayList` input.
+	Conversely, an input implementing both `Serializable` and `Comparable` classes causes an `EdcException`
+	when transformers are registered for both types and the same output.
+
+
+### Target modules and mapping
+
+| Module | Change |
+| --- | --- |
+| `core:common:lib:jsonld-lib` | Replace storage and lookup in `TypeTransformerRegistryImpl`. |
```
ndr-brt, line 81: I just noticed that the `TypeTransformerRegistryImpl` is in the `jsonld-lib`.
not needed to be touched in this DR, but I find it curious :)

ndr-brt (approved):

wolf4ood (approved):

mokhairymahmoud:
can someone merge the PR please because I don't have the permission to merge PRs @ndr-brt @wolf4ood

PR #5929, opened 2026-07-29
refactor: change TypeTransformerRegistry to use a Map instead of Arraylist

mokhairymahmoud:
## What this PR changes/adds

This PR is to change the current TypeTransformerRegistry from an ArrayList to be a Map. This guarantees the existence of only one unique transformer type in the registry and make the transformer selection more deterministic. The PR is the implementation of this [DR](https://github.com/mokhairymahmoud/Connector/tree/main/docs/developer/decision-records/2026-07-24-transformers-registry) 

## Why it does that

The current way of tackling the typeTransformerRegistry is based on ArrayList with a selection mechanism based on findAny() which makes it indeterministic to select a specific transformer specially if there are two of the same type. 

## Linked Issue(s)

Related to #5921 

_Please be sure to take a look at the [contributing guidelines](https://eclipse-edc.github.io/documentation/for-contributors/guidelines/#submitting-a-pull-request) and our [etiquette for pull requests](https://eclipse-edc.github.io/documentation/for-contributors/guidelines/pr-etiquette/)._

### DIFF
--- a/core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
+++ b/core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
@@ -25,21 +25,16 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.BinaryOperator;
 
 import static java.lang.String.format;
 
 public class TypeTransformerRegistryImpl implements TypeTransformerRegistry {
-    private static final Monitor NOOP_MONITOR = new Monitor() {
-    };
 
     private final Map<Class<?>, Map<Class<?>, TypeTransformer<?, ?>>> transformers = new HashMap<>();
     private final Map<String, TypeTransformerRegistry> contextRegistries = new HashMap<>();
     private final Monitor monitor;
     private TypeTransformerRegistry parent;
-
-    public TypeTransformerRegistryImpl() {
-        this(NOOP_MONITOR);
-    }
 
     public TypeTransformerRegistryImpl(Monitor monitor) {
         this.monitor = monitor;
@@ -47,14 +42,14 @@ public class TypeTransformerRegistryImpl implements TypeTransformerRegistry {
 
     private TypeTransformerRegistryImpl(TypeTransformerRegistryImpl parent) {
         this.parent = parent;
-        monitor = parent.monitor;
+        this.monitor = parent.monitor;
     }
 
     @Override
     public void register(TypeTransformer<?, ?> transformer) {
-        var registeredTransformer = transformers.computeIfAbsent(transformer.getInputType(), key -> new HashMap<>())
+        var overriddenTransformer = transformers.computeIfAbsent(transformer.getInputType(), key -> new HashMap<>())
                 .put(transformer.getOutputType(), transformer);
-        if (registeredTransformer != null) {
+        if (overriddenTransformer != null) {
             monitor.warning(format("Overriding transformer registered for %s -> %s", transformer.getInputType(), transformer.getOutputType()));
         }
     }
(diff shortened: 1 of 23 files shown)


core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
```diff
@@ -31,7 +29,7 @@
 
 public class TypeTransformerRegistryImpl implements TypeTransformerRegistry {
     private final Map<String, Class<?>> aliases = new HashMap<>();
```
ndr-brt: it's been a while that I see this aliases unused field, since you're here, could you please delete it?

core/common/lib/jsonld-lib/src/test/java/org/eclipse/edc/transform/TestTypeTransformer.java
```diff
@@ -0,0 +1,43 @@
+/*
+ *  Copyright (c) 2022 - 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
```
ndr-brt: nit: copyright

core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
```diff
@@ -44,7 +42,8 @@ private TypeTransformerRegistryImpl(TypeTransformerRegistry parent) {
 
     @Override
     public void register(TypeTransformer<?, ?> transformer) {
-        this.transformers.add(transformer);
+        transformers.computeIfAbsent(transformer.getInputType(), key -> new HashMap<>())
```
ndr-brt: I think it could worth to add a warning log in case of transformer override

core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
```diff
@@ -54,14 +53,33 @@ public void register(TypeTransformer<?, ?> transformer) {
...
+                .toList();
+
+        if (mostSpecificInputTypes.size() > 1) {
+            throw new EdcException(format("Ambiguous transformers registered for %s -> %s", input.getClass(), outputType));
+        }
+
+        return mostSpecificInputTypes.stream()
+                .findFirst()
+            .map(inputType -> transformers.get(inputType).get(outputType));
+    }
```
ndr-brt: this logic is not wrong but I think it could be written in a more readable way - and slightly more performant, by avoiding multiple iterations over `inputTypes`, using streams `reduce` function:

```java
    private Optional<TypeTransformer<?, ?>> findTransformer(Object input, Class<?> outputType) {
        return transformers.entrySet().stream()
                .filter(entry -> entry.getKey().isInstance(input))
                .filter(entry -> entry.getValue().containsKey(outputType))
                .map(Map.Entry::getKey)
                .map(Optional::of)
                .reduce(Optional.empty(), (current, candidate) -> {
                    if (current.isEmpty()) {
                        return candidate;
                    }

                    if (candidate.get().isAssignableFrom(current.get())) {
                        return current;
                    }

                    if (current.get().isAssignableFrom(candidate.get())) {
                        return candidate;
                    }

                    throw new EdcException(format("Ambiguous transformers registered for %s -> %s", input.getClass(), outputType));
                })
                .map(transformers::get)
                .map(it -> it.get(outputType));
    }
```

so the "most specific" logic emerges in the reduce function itself:
- set the first item as current
- is the candidate a supertype of current? ignore it
- is the candidate a subtype of current? set it as current
- neither of the two? it means that it's a type that's on a different hierarchy: throw exception

the reduce function could also be extracted and called `findMostSpecificInputType`

core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
mokhairymahmoud: done

core/common/lib/jsonld-lib/src/test/java/org/eclipse/edc/transform/TestTypeTransformer.java
mokhairymahmoud: changed it

core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
mokhairymahmoud: Already did, thanks.

core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
mokhairymahmoud: Your implementation is cleaner so I applied it. thanks

core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
```diff
@@ -15,36 +15,48 @@
...
 import java.util.Objects;
 import java.util.Optional;
 
 import static java.lang.String.format;
 
 public class TypeTransformerRegistryImpl implements TypeTransformerRegistry {
-    private final Map<String, Class<?>> aliases = new HashMap<>();
-    private final List<TypeTransformer<?, ?>> transformers = new ArrayList<>();
+    private static final Monitor NOOP_MONITOR = new Monitor() {
+    };
```
ndr-brt: what's this? If this class has a collaborator please ensure that every instantiation passes the needed collaborators.
Plus, the added `Monitor` interaction would deserve an unit test

core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
```diff
@@ -15,36 +15,48 @@
...
-    private TypeTransformerRegistryImpl(TypeTransformerRegistry parent) {
+    private TypeTransformerRegistryImpl(TypeTransformerRegistryImpl parent) {
         this.parent = parent;
+        monitor = parent.monitor;
     }
 
     @Override
     public void register(TypeTransformer<?, ?> transformer) {
-        this.transformers.add(transformer);
+        var registeredTransformer = transformers.computeIfAbsent(transformer.getInputType(), key -> new HashMap<>())
```
ndr-brt:
```suggestion
        var overriddenTransformer = transformers.computeIfAbsent(transformer.getInputType(), key -> new HashMap<>())
```

core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
```diff
@@ -54,14 +66,40 @@ public void register(TypeTransformer<?, ?> transformer) {
...
-        return transformers.stream()
-                .filter(t -> t.getInputType().isInstance(input) && t.getOutputType().equals(outputType))
-                .findAny()
+        return findTransformer(input, outputType)
                 .map(it -> (TypeTransformer<INPUT, OUTPUT>) it)
                 .or(() -> Optional.ofNullable(parent).map(p -> p.transformerFor(input, outputType)))
                 .orElseThrow(() -> new EdcException(format("No Transformer registered that can handle %s -> %s", input.getClass(), outputType)));
     }
 
+    private Optional<TypeTransformer<?, ?>> findTransformer(Object input, Class<?> outputType) {
```
ndr-brt: This additional method doesn't add much more than indirection, let's either:
- inline this into `transformerFor`
- inline `findMostSpecificInputType` into this.

in my past comment I suggested to extract only the function passed to the `reduce` method into a separate one, that would move away the complexity under a meaningful name, that could be like:

```
BinaryOperator<Class<?>> selectMostSpecificType(Class<?> inputType, Class<?> outputType) {
    return (current, candidate) -> {
        if (current.isEmpty()) {
                        return candidate;
                    }

                    if (candidate.get().isAssignableFrom(current.get())) {
                        return current;
                    }

                    if (current.get().isAssignableFrom(candidate.get())) {
                        return candidate;
                    }
                    throw new EdcException(format("Ambiguous transformers registered for %s -> %s", inputType, outputType));
    }
}
```

ndr-brt (changes requested):

core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
mokhairymahmoud: done

core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
mokhairymahmoud: fixed

core/common/lib/jsonld-lib/src/main/java/org/eclipse/edc/transform/TypeTransformerRegistryImpl.java
mokhairymahmoud: I changed it as requested, there is a unit test for the warning from the monitor. let me know if you need any other changes. Thanks @ndr-brt

ndr-brt (approved):
looks good!

